PR #3, opened 2016-07-04
Fix combine-plants recursion bug

kane-t:
(Finally) fixed a recursion bug in the original script that prevented it from recursing into sub-containers.

### DIFF
--- a/combine-plants.lua
+++ b/combine-plants.lua
@@ -16,7 +16,7 @@ if args.max then max = tonumber(args.max) end
 
 local stockpile = nil
 if args.stockpile then stockpile = df.building.find(tonumber(args.stockpile)) end
-    
+
 local container = nil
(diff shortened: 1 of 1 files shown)


combine-plants.lua
```diff
@@ -16,32 +16,38 @@ if args.max then max = tonumber(args.max) end
 
 local stockpile = nil
 if args.stockpile then stockpile = df.building.find(tonumber(args.stockpile)) end
-
+
```
lethosor: Should get rid of this (and on new lines 47 and 78)
(We should probably get travis set up in this repo eventually.)

kane-t: Oh, you mean the extra tabs?  Sure thing, I'll do it right now.  Mind if I ask why?

PR #4, opened 2016-07-27
syndrome-trigger updated to work with syndrome classes

AtomicChicken:
New arg added: **-synclass**
Allows a target SYN_CLASS to be specified instead of a syndrome name. This allows modders to specify an effect for a bunch of syndromes at once, hence saving time. It can also be used with syndromes like the generated vampire and werebeast curses, which lack names but have syndrome classes.

### DIFF
--- a/modtools/syndrome-trigger.lua
+++ b/modtools/syndrome-trigger.lua
@@ -80,7 +80,8 @@ validArgs = validArgs or utils.invert({
  'clear',
  'help',
  'command',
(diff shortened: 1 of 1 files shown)


lethosor:
Thanks! Not sure why travis decided to run there. There was a whitespace issue that I fixed manually (there was a tab on line 126), so keep that in mind in the future. Ideally travis would have caught that if it were working, but it isn't yet.

PR #6, opened 2016-09-15
Proposed fixes for ban-cooking.rb

rndmvar:
1) All ban-able items must be cook-able to be banned.
2) Checks to ensure that kitchen.exc_types[index] exists before attempting to write to it.
 -- Ran into issues with the original script attempting to ban non-cookable plant items, which lead to mismatches in the length of the kitchen.exc_types versus the others.
3) Change the index variable in the ban_cooking function from a key lookup to a numerical index value
4) Added subtype parameter to allow handling of tree fruits.
5) Added printing of the item that has been banned.
6) Seeds banning bans the seeds from being cooked now.
 -- Required that banning not be for seed producers that can't be brewed into alcohol for logistic/time reasons
7) Fixed mill and thread banning, as they referenced local variable 'm', but didn't set it.
 -- I don't know if there are any cook-able mill or thread plants...
8) Added fruit banning, for those that wish only to ban cooking of wine producing plants and plant growths
9) Added a pipe delimited print out via the 'show' argument for displaying in the console all items currently banned from cooking.
10) Added lots of comments, because I like comments.

All of this was tested on Dwarf Fortress 0.43.03
This is my first time coding in Ruby, as my typical go to is Python.  So the code may be ugly, as I stayed away from messing with classes/functions for this.

### DIFF
--- a/ban-cooking.rb
+++ b/ban-cooking.rb
@@ -6,114 +6,395 @@ ban-cooking
 A more convenient way to ban cooking various categories of foods than the
 kitchen interface.  Usage:  ``ban-cooking <type>``.  Valid types are ``booze``,
 ``honey``, ``tallow``, ``oil``, ``seeds`` (non-tree plants with seeds),
-``brew``, ``mill``, ``thread``, and ``milk``.
+``brew``, ``fruit``, ``mill``, ``thread``, and ``milk``.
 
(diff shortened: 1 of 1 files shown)


lethosor:
Done. Sorry for the delay!

PR #8, opened 2016-11-22
Add "change-build-menu" and "if-entity".

milochristiansen:
change-build-menu provides a simple API for adding or removing buildings from DF's build sidebar. if-entity runs a command if the current entity matches the given ID.

### DIFF
--- a/modtools/change-build-menu.lua
+++ b/modtools/change-build-menu.lua
@@ -1,0 +1,560 @@
+-- Edit the build mode sidebar menus.
+-- 
+-- Based on my script for the Rubble addon "Libs/Change Build List".
+
+--[[
+Copyright 2016 Milo Christiansen
+
+This software is provided 'as-is', without any express or implied warranty. In
+no event will the authors be held liable for any damages arising from the use of
+this software.
+
+Permission is granted to anyone to use this software for any purpose, including
+commercial applications, and to alter it and redistribute it freely, subject to
+the following restrictions:
+
+1. The origin of this software must not be misrepresented; you must not claim
+that you wrote the original software. If you use this software in a product, an
+acknowledgment in the product documentation would be appreciated but is not
+required.
+
+2. Altered source versions must be plainly marked as such, and must not be
+misrepresented as being the original software.
+
+3. This notice may not be removed or altered from any source distribution.
+]]
+
+--@ module = true
+--@ enable = true
+
+-- From here until I say otherwise the APIs are generic building manipulation functions from "Libs/Buildings"
+
+local utils = require "utils"
+
+local wshop_type_to_id = {
+	[df.workshop_type.Carpenters] = "CARPENTERS",
+	[df.workshop_type.Farmers] = "FARMERS",
+	[df.workshop_type.Masons] = "MASONS",
+	[df.workshop_type.Craftsdwarfs] = "CRAFTSDWARFS",
+	[df.workshop_type.Jewelers] = "JEWELERS",
+	[df.workshop_type.MetalsmithsForge] = "METALSMITHSFORGE",
+	[df.workshop_type.MagmaForge] = "MAGMAFORGE",
+	[df.workshop_type.Bowyers] = "BOWYERS",
+	[df.workshop_type.Mechanics] = "MECHANICS",
+	[df.workshop_type.Siege] = "SIEGE",
+	[df.workshop_type.Butchers] = "BUTCHERS",
+	[df.workshop_type.Leatherworks] = "LEATHERWORKS",
+	[df.workshop_type.Tanners] = "TANNERS",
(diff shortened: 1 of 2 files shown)


lethosor:
Yep, these look like modtools scripts. Thanks for organizing them!

The one thing I noticed is that they use tabs for indentation. That should probably be changed to spaces to be consistent with other DFHack code. (Travis would've complained about it, but it's been disabled - I really ought to fix and re-enable that.)

milochristiansen:
I always use tabs and forgot that DFHack uses spaces. Is it a big deal? I know some scripts use four spaces, and others use 1 (?!?)

lethosor:
Yeah, we should standardize on something (all of the C++ code uses 4 spaces, as far as I know, but some scripts vary). The only rule I'm aware of for scripts is "no tabs". It's not a lot of effort to change on our end, though, so as long as this script works, I'd be fine with merging this and changing the indentation afterward. 

I do recognize that having the version here use spaces would make it a little harder to merge future changes from your end, though (and I don't want to force you to change your indentation style). Would it be possible to stick with spaces in your dfhack/scripts fork? It looks to me like you're basically copying changes from wherever you keep your Rubble code, so hopefully changing whitespace when you do that wouldn't interfere with your Rubble repo.

milochristiansen:
OK, I'll make the changes, its not a big deal.

I don't see me making many changes to them in the future, the main reason I want to get them into DFHack now is that they are "finished" and stable...

lethosor:
Thanks! Do let us know if you make any changes later, though.

modtools/change-build-menu.lua
```diff
@@ -0,0 +1,560 @@
...
+    function RevertBuildingChangesAdv(typ, subtyp, custom, category)
+        These two functions revert changes made by "ChangeBuilding" and
+        "ChangeBuildingAdv". Like those two functions there are two versions,
+        a simple one that takes a string ID and one that takes a numeric ID
+        triplet.
+]==]
+end
+
+if not dfhack.isWorldLoaded() then
+    dfhack.color(RED)
```
lethosor: Did you avoid dfhack.printerr() here intentionally? I was about to suggest it, but it looks like you used it later in this file.

milochristiansen: Yes, but I forget why...

It had something to do with the exact way I wanted the output formatted, but for the life of me I can't remember why I changed it now...

lethosor: I can't see where RED is defined (it's not built-in), so maybe this isn't actually changing the color. Also, if it were COLOR_RED, it could be a darker red or non-bold red, depending on the terminal. Either way, I think printerr() might be better for consistency, unless there's a good reason for this.

milochristiansen: Looking at it I can only assume I was sleep deprived at the time, since printerr would do exactly what I had intended... Like I said, I can't remember what I was thinking (was I thinking?). I'll change it.

lethosor:
(Reopened this and #9 to see if Travis complained about anything else.)

milochristiansen:
What is the expected format for documentation?

milochristiansen:
Oh, and \r\n newlines are a fluke, I have my editor set to use \n always. I must have copied an old file instead of creating a new one when I started writing those scripts...

modtools/change-build-menu.lua
```diff
@@ -0,0 +1,558 @@
...
+            KILN
+    
+    change-build-menu remove <ID> <CATEGORY>
+        Remove the workshop or furnace with the ID <ID> from <CATEGORY>.
+        
+        <CATEGORY> and <ID> may have the same values as for the "add" option.
+    
+    change-build-menu revert <ID> <CATEGORY>
+        Revert an earlier remove or add operation. It is NOT safe to "remove"
+        and "add"ed building or vice versa, use this option to reverse any
```
dscorbett: `and` → `an`.

milochristiansen: Oops...

Actually I was reviewing the offline master copy this morning (backporting changes I had made to the DFHack copy) and noticed that myself. Thank you for reminding me though, or I would have forgotten to change it in the online copy.

milochristiansen:
I did not fix the newlines or documentation. The newlines because it would be practically impossible on this computer (the linux mint default editor does not support that operation and neither does the github web UI), the documentation because I have no idea what the supported format is...

milochristiansen:
Closer, but it seems I have errors in the doc formatting...

Looks like I'll have to read up on yet another doc format.

lethosor:
(Fixed in f0080ac)

milochristiansen:
I had a "hopefully fixed" version ready to go (created after much reading of the ReST docs), but your fixes are much better, so it looks like this is all done...

PR #9, opened 2016-11-22
Add season-palette

milochristiansen:
season-palette is a simple script that allows automatic switching of the color palette with the changing seasons. It also provides a Lua function that allows loading arbitrary color palettes.

### DIFF
--- a/season-palette.lua
+++ b/season-palette.lua
@@ -28,92 +28,92 @@ misrepresented as being the original software.
 --@ module = true
 --@ enable = true
 
-local mapingsA = {
-	["BLACK"] = 0,
-	["BLUE"] = 1,
-	["GREEN"] = 2,
-	["CYAN"] = 3,
-	["RED"] = 4,
-	["MAGENTA"] = 5,
-	["BROWN"] = 6,
-	["LGRAY"] = 7,
-	["DGRAY"] = 8,
-	["LBLUE"] = 9,
-	["LGREEN"] = 10,
-	["LCYAN"] = 11,
-	["LRED"] = 12,
-	["LMAGENTA"] = 13,
-	["YELLOW"] = 14,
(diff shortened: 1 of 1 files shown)


season-palette.lua
```diff
@@ -0,0 +1,221 @@
...
+misrepresented as being the original software.
+
+3. This notice may not be removed or altered from any source distribution.
+]]
+
+--@ module = true
+--@ enable = true
+
+local mapingsA = {
+	["BLACK"] = 0,
```
lethosor: Not a huge issue, but `BLACK = 0` could be cleaner here (or maybe using utils.invert, although I think that starts counting at 1 instead of 0, so it would require more effort elsewhere).

lethosor:
Nice! I remember seeing this on the forums. +1 for the enable/disable support.

season-palette.lua
milochristiansen: I'll fix my "mapings" :)

I use `["string"]` as a convention when the key is a string and not an identifier (which is arguably the case here). Do you really think I should change it?

Also I meant to add this to modtools instead of the main list, should I cancel the request and redo it?

lethosor: Omitting the brackets results in the keys being strings (except you're essentially restricted to valid identifiers as keys, which isn't an issue here). I suppose it doesn't really matter, though.

This is fine in the main folder because it can be used directly by end-users. `modtools` is only for mod-oriented utilities that users shouldn't be running directly (well, except for testing purposes).

PeridexisErrant:
Please add some documentation to the script too - you can see examples in the repo, or read [the docs on docs](https://dfhack.readthedocs.io/en/stable/Contributing.html#documentation-standards).

lethosor:
It's there already, just not in the expected format. I can take care of that.

PR #14, opened 2017-04-24
fixed a few issues

Amostubal:
occasionally the original point of creation has no map data (its below the bottom floor of hell, and therefore does not exist).
figured that since the point it tries to create the unit is a series of simulated actions that moves through the arena spawn unit menus, that before opening the menu, you could just move the cursor there, if location was set, it worked.  

also added an argument to set the unit to civ_id and group_id of the the fort, to reduce on the verbose of most commands (nearly every use I have for create-unit is set to the civ and group of the fort...).  

moved the civ_id change closer to the point of actually creating the unit,  thought about moving it into the createNemesis function... but didn't.

corrected error involving out of bounds vector, on old forts in the setting of the age(o.l. 523/n.l. 540).  It was just a bad way to look up the historical figure, used the appropriate call.

the entire thing no longer needs teleport to operate.

### DIFF
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -25,7 +25,6 @@ local usage = [====[
 modtools/create-unit
 ====================
 Creates a unit.  Usage::
-
     -race raceName
         specify the race of the unit to be created
         examples:
@@ -38,6 +37,8 @@ Creates a unit.  Usage::
             FEMALE
     -domesticate
         if the unit can't learn or can't speak, make it a friendly animal
+    -setUnitToFort
+        Sets the groupId and civId to the local fort.
     -civId id
         Make the created unit a member of the specified civ
         (or none if id = -1).  If id is \\LOCAL, make it a member of the
@@ -76,7 +77,7 @@ end
 
 local utils=require 'utils'
 
-function createUnit(race_id, caste_id)
+function createUnit(race_id, caste_id, location)
   local view_x = df.global.window_x
   local view_y = df.global.window_y
   local view_z = df.global.window_z
@@ -105,6 +106,13 @@ function createUnit(race_id, caste_id)
   local old_gametype = df.global.gametype
   df.global.gametype = df.game_type.DWARF_ARENA
   gui.simulateInput(dwarfmodeScreen, 'D_LOOK_ARENA_CREATURE')
+
+-- move cursor to location instead of moving unit later, corrects issue of missing mapdata when moving the created unit.
+	if location then
+		df.global.cursor.x = tonumber(location[1])
+		df.global.cursor.y = tonumber(location[2])
+		df.global.cursor.z = tonumber(location[3])
+	end
 
   local spawnScreen = dfhack.gui.getCurViewscreen()
   if dfhack.world.isArena() then
@@ -199,13 +207,11 @@ function createFigure(trgunit,he,he_group)
   -- set values that seem related to state and do event
   --change_state(hf, dfg.ui.site_id, region_pos)
 
-
   --lets skip skills for now
   --local skills = df.historical_figure_info.T_skills:new() -- skills snap shot
   -- ...
   -- note that innate skills are automaticaly set by DF
   hf.info.skills = {new=true}
-
 
   he.histfig_ids:insert('#', hf.id)
   he.hist_figures:insert('#', hf)
@@ -275,8 +281,8 @@ function createNemesis(trgunit,civ_id,group_id)
 end
 
 --createNemesis(u, u.civ_id,group)
-function createUnitInCiv(race_id, caste_id, civ_id, group_id)
-  local uid = createUnit(race_id, caste_id)
+function createUnitInCiv(race_id, caste_id, location, civ_id, group_id)
+  local uid = createUnit(race_id, caste_id, location)
   local unit = df.unit.find(uid)
   if ( civ_id ) then
     createNemesis(unit, civ_id, group_id)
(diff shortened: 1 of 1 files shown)


modtools/create-unit.lua
```diff
@@ -25,7 +25,6 @@ local usage = [====[
 modtools/create-unit
 ====================
 Creates a unit.  Usage::
-
```
lethosor: I'm not sure why you're removing extra newlines. They help improve readability sometimes, and this one in particular is necessary for the docs to build.

Amostubal: ? @ lethosor is that an automated message or can you actually see what I'm doing as I do it?

lethosor: I can see individual commits at https://github.com/DFHack/scripts/pull/14/commits, but that's it.

Amostubal: yeah ignore those if you want those lines.   I was having an issue where travis kept saying I had some funky indentation.... I can't find anything else in the file... was going to move a few lines and see if travis saw it differently.

lethosor: Oh, it might complain about lines that are just whitespace (e.g. a bunch of spaces on their own line). Anyway, I can take care of whatever Travis is complaining about, so don't feel like you have to make a bunch of commits to fix that.

Amostubal: @ lethosor yeah I didn't realize it pushed a new commit, I thought it was just editing the original commit.

lethosor: No worries, I can combine those into a single commit too.

Amostubal:
@lethosor I found some other things that I didn't realize about this script.

my current version is more robust and is functional, but I haven't uploaded it to github yet.  want to make sure am I putting in pull requests to the right place on this file?

lethosor:
Sure, I'll check. The most recent commit I'm seeing here is still from 11 days ago.

Amostubal:
@lethosor Yeah I just hadn't pushed it to my change yet: so all the changes according to what I can remember:
1. corrected the issue where the unit is created at a bad location (mapdata missing/below hell), by adding the ability to send location to the creation and setting the (k) look location before calling the creation from arena menus.

2. added this ability to call location to the end of all the args in all the function calls, so that they don't break any older scripts that call from outside, that haven't been updated (should at least be receiving a null location, so the script wont try and move the cursor before calling), and therefore they can update when they want or continue to do their own teleports, etc,

3.  fixed a historical_figure call that breaks the script if the unit index goes to high, basically replaced a direct index with the correct find function.

4.  added a line of text where units not given new names by the script still has its arena name (i.e. "dog -123456789").

5. removed the teleport at the end, its no longer needed.

6.  a basic -setUnitToCivAndFort to get around the \\LOCAL issues that plagued my original issues with the script.

concerns: 
line 294: not sure if using 0, is the best way to do that, its just the first way that accomplished the job I was wanting it to do, and it didn't break or fail afterwards.

line 508-511: I moved the setting of unit.civ_id at some point, not sure why; it was down around line 599 after nickname setting.  Not sure if this has any adverse affects, I couldn't find any, but not sure why it was so far down the script to begin with.

Well that's it... I'm packing it into the next version of masterwork as is right now.

lethosor:
I'm curious - what do these commands print for you?
```
devel/print-args \LOCAL
devel/print-args \\LOCAL
devel/print-args \\\LOCAL
devel/print-args \\\\LOCAL
```

Amostubal:
@lethosor directly into the console they print the \ perfectly....

I think my issue was from reaction-trigger.

\\LOCAL would work in the console, but the moment I went to the onLoad.init
modtools/reaction-trigger -reactionName BUY_GOBLIN_1 -command [ modtools/create-unit -race HUMAN -caste MALE -civId \\LOCAL -groupId \\LOCAL -location [ \\LOCATION ] -name PLAINS -age 20 ]

would fail. I'd had to change the \\LOCAL to \\\\LOCAL to get it to work.

Amostubal:
And the added in newlines corrected the travisCL complaint.... I'm happier than ever, with it now.

modtools/create-unit.lua
```diff
@@ -579,11 +601,8 @@ if args.nick and type(args.nick) == 'string' then
   dfhack.units.setNickname(df.unit.find(unitId), args.nick)
 end
 
-if civ_id then
-  local u = df.unit.find(unitId)
-  u.civ_id = civ_id
-end
 
+--[[ depracated by amostubal.  no longer necessary as we move the cursor in the creation of the unit.  no longer needs to utilize teleport...
 if args.location then
```
lethosor: Passing a location argument still works, right? If so, can you just remove this section?

Amostubal: yes we can, I really try not to delete whole lines sections, that much.   The location is just passed along to the actual function that creates the unit.

modtools/create-unit.lua
```diff
@@ -566,6 +587,7 @@ if args.name then
 else
   local unit = df.unit.find(unitId)
   unit.name.has_name = false
+  unit.name.first_name = "" -- remove the first name altogether. gets rid of units having number names.
```
lethosor: I'm not sure I like this change - it results in units with blank names in arena mode, and even without this change in fortress mode, the number names aren't visible in the unit or job lists.
I'm thinking of just changing this back, but is there something I'm missing here?

Amostubal: Hold on I went out of town after loading 1.25, this is from my phone... okay the issue was that when you view them either with v or k I think on the map it shows up, additionally i thought it was in the unit list... I can understand the arena mode thing... We could add a verify of arena mode and ignore that line?

I won't be able to test any code until at least Monday evening,  it's not a dire fix more of a 'visual immersion fix.  It was primarily being used her for summoned/purchased pets, a user side fix is easy enough for that.

Another note: I'm not sure but github appears to have waited 3 days to tell me you wrote here? Lol

lethosor: Oh, sorry about that. I'm not sure what's up with the notifications - I managed to comment in a way that I don't usually use, and my comment didn't show up on my end for a few minutes, so maybe that's related.

I'll look into this again as soon as I can (in the middle of r1 right now), and see about adding separate handling for arena mode if necessary.

Amostubal: Awesome I'll look at updating rd hack after the release this week in the MW repository and run tests on it this week... I've declared a break in that release schedule anyways.

PR #16, opened 2017-04-27
Add unforbid script to unforbid all items.

rrauenza:
Started it with a mandatory 'all' param in case other options are wanted in the future.


lethosor:
Looks good!

PR #17, opened 2017-05-10
fix to small issue in full-heal.lua

Amostubal:
there was a .job_type missing in line 151 causing it to throw errors.

### DIFF
--- a/full-heal.lua
+++ b/full-heal.lua
@@ -148,7 +148,7 @@ if unit then
 
     if unit.job.current_job and unit.job.current_job.job_type == df.job_type.Rest then
         --print("Wake from rest -> clean self...")
-        unit.job.current_job = df.job_type.CleanSelf
+        unit.job.current_job.job_type = df.job_type.CleanSelf
     end
 end
 


lethosor:
Does this actually work (the job-switching part)?

Amostubal:
It was interesting as he just canceled the clean injury claiming to rest injury, but instead of actually switching to rest, he ran to hospital then went to work... I dropped a floor on him.

lethosor:
I mean, I was more concerned about it breaking completely or crashing. What you described sounds better than that, and better than resting indefinitely, so it's fine with me.

lethosor:
The build failure looks like my fault (I assumed that `probe` would be a valid link target, as practically every other plugin is, so I didn't bother to actually double-check that the docs built).

Amostubal:
I've only tested it twice... on the same dwarf, on the same save... so I may drop a few more floors on some dwarfs and see if it does crash... if you want some more testing.

yeah I assumed it had to be another issue for travis, as I had only added the job_type into a line.

Amostubal:
@lethosor well the only thing that did happen after dropping several more floors, I had one guy that continued to sit on his bed, listed as job clean self and refused to move, I had to manually switch him back to rest and let him sit a minute and he got up...

PR #19, opened 2017-06-03
incorporated Z change into digfort

Amostubal:
simple check for a line to be #< or #> edit then added a case for it.  It now will go up or down z based on the symbol.


Amostubal:
reference issue: https://github.com/DFHack/dfhack/issues/878

Amostubal:
#874

Amostubal:
#https://github.com/DFHack/dfhack/issues/878

PR #21, opened 2017-06-27
Created intro-loader.lua

ChrisThirtle:
I made a little script that can change the introduction text after the embark screen, using some simple file I/O, and a bit of finagling.
It requires that those wishing to implement the script compress the text files according to the Zlib Compressed Data Format Specification (RFC 1950), and notes an available program (Andux's WTFtools) that will do so for them; this is due to Toady's implementation of the data files. If there were any usability improvements to this script, it would be implementing this compression directly.

Once the files are prepared, the script can be set up to load files with entity-specific intros for mods that implement new races, or it can be set to load up random intro files. With both settings enabled, the script can load up a randomly chosen file from a set of entity-specific ones.

I've probably got some documentation or formatting to fix, but as far as I can tell it is at least functional. The program _should_ work cross-platform, though since I am dealing with file IO here, it is entirely possible that I have messed something up in the sections I had to do compatibility work with.

### DIFF
--- a/modtools/intro-loader.lua
+++ b/modtools/intro-loader.lua
@@ -3,7 +3,7 @@
 
 local help = [====[
 intro-loader.lua
-===============
+================
 
 Alters the fort mode intro text before embarking
 
@@ -92,16 +92,6 @@ if args.folder then
     inputfolder = dfhack.getDFPath() ..  args.folder .. "/"
 end
 
---Build current index of files in directory
-do
-    local index = io.open(dfhack.getDFPath() .. "/data/announcement/index.txt","w")
-	local files = dfhack.filesystem.listdir(inputfolder)
-    for foo,filename in pairs(files) do
-		index:write(filename .. "\n")
-	end
-    io.close(index)
-end
-
 rngesus = dfhack.random.new()
 
 
@@ -116,11 +106,13 @@ local function introloader(folder,dorand,doent,verbose)
     
     if args.rand then
         --Determine number of usable files
-		index = io.open(dfhack.getDFPath() .. "/data/announcement/index.txt","r")
-		local indextext = index:read("*all")
+		local index = " "
+		for foo,filename in pairs(dfhack.filesystem.listdir(folder)) do
+			index = index .. "\n" .. filename
+		end
+		
         local filecount = 0
-		print(indextext)
-        for foo in string.gmatch(indextext,inputfile .. "_%d+") do
+        for foo in string.gmatch(index,inputfile .. "_%d+") do
             filecount = filecount + 1
         end
         
--- a/modtools/intro-loader.lua
+++ b/modtools/intro-loader.lua
@@ -22,7 +22,7 @@ Usage:
 -help
     print this text
 -v
-    Activate verbose mode
+    Activate verbose mode, kinda spammy
 -folder
     Designate a folder to get the input intro files from relative to
     the DF directory
@@ -67,7 +67,7 @@ local utils = require 'utils'
 
 validArgs = validArgs or utils.invert({
     'help',
-    'v'
+    'v',
     'now',
     'folder',
     'file',
@@ -82,63 +82,64 @@ if args.help then
     return
 end
 
-local inputfolder = dfhack.getDFPath() .. "\\data\\announcement\\"
-if args.folder then
-    inputfolder = dfhack.getDFPath() ..  args.folder .. "\\"
-end
-
 local prefix = "INTRO"
 if args.file then
     prefix = args.file
 end
 
-if args.rand then
-    math.randomseed(os.time())
+local inputfolder = dfhack.getDFPath() .. "/data/announcement/"
+if args.folder then
+    inputfolder = dfhack.getDFPath() ..  args.folder .. "/"
 end
-
 
 --Build current index of files in directory
 do
-    local OS = dfhack.getOSType()
-    if OS == 'windows' then
-        os.execute("dir \"" .. inputfolder .. "\" /L /B >index.txt")
-    elseif OS == 'linux' or OS == 'darwin' then --apple's real name is darwin, apparently
-        os.execute("ls \"" .. inputfolder .. "\" -1 -A >index.txt")
-    end
-    
-    local list = io.open(dfhack.getDFPath() .. "\\data\\announcement\\index.txt","r")
-    indextext = list:read("*all")
-    io.close(list)
+    local index = io.open(dfhack.getDFPath() .. "/data/announcement/index.txt","w")
+	local files = dfhack.filesystem.listdir(inputfolder)
+    for foo,filename in pairs(files) do
+		index:write(filename .. "\n")
+	end
+    io.close(index)
 end
+
+rngesus = dfhack.random.new()
 
 
 --Code main body, this is where the magic happens
 local function introloader(folder,dorand,doent,verbose)
-    local entity = "_" .. df.historical_entity.find(df.global.ui.civ_id).entity_raw.code
-    if ~doent then entity = "" end
-    local inputfile = folder .. "\\" .. prefix .. entity
+    local entity = ""
+	if args.entity then
+		entity = "_" .. df.historical_entity.find(df.global.ui.civ_id).entity_raw.code
+	end
     
-    if dorand then
+    local inputfile = prefix .. entity
+    
+    if args.rand then
         --Determine number of usable files
+		index = io.open(dfhack.getDFPath() .. "/data/announcement/index.txt","r")
+		local indextext = index:read("*all")
         local filecount = 0
-        while string.gmatch(indextext,inputfile .. "_%d+") do
-            filecount++
+		print(indextext)
+        for foo in string.gmatch(indextext,inputfile .. "_%d+") do
+            filecount = filecount + 1
         end
         
         --Random decision
         if filecount > 0 then
-            local rand = math.random(1,filecount)
+            local rand = rngesus:random(filecount)+1
             inputfile = inputfile .. "_" .. rand
         end
     end
+	
+	inputfile = folder .. inputfile
     
-    if verbose then print("Loading intro text") end
+    if verbose then print("Loading intro text from " .. inputfile) end
     local input = assert(io.open(inputfile, "rb"),"No matching intro file, leaving as-is")
     if input == false then return end
-    local output = assert(io.open(dfhack.getDFPath() .. "\\data\\announcement\\fortressintro", "wb"),"Could not create or open data/announcement/fortressintro, somehow.")
+    local output = assert(io.open(dfhack.getDFPath() .. "/data/announcement/fortressintro", "wb"),"Could not create or open data/announcement/fortressintro, somehow.")
     if output == false then input:close() return end
     local t = input:read("*all")
-    t = string.gsub(t, "\r\n", "\n")
+    t = string.gsub(t, "\r\n", "\n") --You can never be too sure with these things; Windows is absolutely obsessive with its \r characters
     output:write(t)
     if verbose then print("Civilization intro text written") end
     output:close()
@@ -149,12 +150,12 @@ end
 
 --Flow control for the code; runs process code after embark selection has been made, to ensure proper operation.
 dfhack.onStateChange.loadNewFortressIntro = function(code)
-if(dorand|doent)
-    if args.v then print("Intro loader ready") end
-    if (code == SC_VIEWSCREEN_CHANGED and dfhack.gui.getCurFocus() == 'setupdwarfgame') or (args.now) then
-            dfhack.with_suspend(introloader,inputfolder,args.rand,args.entity, args.v)
-    end
-else
-    dfhack.printerr("Intro loader has invalid parameters, aborting")
-    end
+	if(args.rand or args.entity) then
+		if args.v then print("Intro loader ready") end
+		if (code == SC_VIEWSCREEN_CHANGED and dfhack.gui.getCurFocus() == 'setupdwarfgame') or (args.now) then
+			dfhack.with_suspend(introloader,inputfolder,args.rand,args.entity, args.v)
+		end
+	else
+		dfhack.printerr("Intro loader has invalid parameters, aborting")
+	end
 end
--- a/modtools/intro-loader.lua
+++ b/modtools/intro-loader.lua
@@ -7,9 +7,15 @@ intro-loader.lua
 
 Alters the fort mode intro text before embarking
 
-This script should be called in "dfhack.init" for proper operation. Attempting to call the script at any other point in execution will have unpredictable results.
+This script should be called in "dfhack.init" for proper operation. Attempting
+ to call the script at any other point in execution will have unpredictable
+ results.
+
 This script should not be called from the terminal outside of debug usage.
-The script requires either the -entity, -rand, or both parameters be included in the call to operate, and will indicate failure without
+
+The script requires either the -entity, -rand, or both parameters be included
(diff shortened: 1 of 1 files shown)


modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+end
+
+if args.rand then
+    math.randomseed(os.time())
+end
+
+
+--Build current index of files in directory
+do
+    local OS = dfhack.getOSType()
```
lethosor: Use dfhack.filesystem.listdir() instead.

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+    if OS == 'windows' then
+        os.execute("dir \"" .. inputfolder .. "\" /L /B >index.txt")
+    elseif OS == 'linux' or OS == 'darwin' then --apple's real name is darwin, apparently
+        os.execute("ls \"" .. inputfolder .. "\" -1 -A >index.txt")
+    end
+    
+    local list = io.open(dfhack.getDFPath() .. "\\data\\announcement\\index.txt","r")
```
lethosor: Backslashes aren't portable - use forward slashes. (Although this wouldn't be necessary using listdir().)

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+end
+
+
+
+--Flow control for the code; runs process code after embark selection has been made, to ensure proper operation.
+dfhack.onStateChange.loadNewFortressIntro = function(code)
+if(dorand|doent)
+    if args.v then print("Intro loader ready") end
+    if (code == SC_VIEWSCREEN_CHANGED and dfhack.gui.getCurFocus() == 'setupdwarfgame') or (args.now) then
+            dfhack.with_suspend(introloader,inputfolder,args.rand,args.entity, args.v)
```
lethosor: Is with_suspend really necessary here? No other scripts use it, and it's not necessary for ordinary scripts as far as I know.

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+    if verbose then print("Civilization intro text written") end
+    output:close()
+    input:close()
+end
+
+
+
+--Flow control for the code; runs process code after embark selection has been made, to ensure proper operation.
+dfhack.onStateChange.loadNewFortressIntro = function(code)
+if(dorand|doent)
```
lethosor: This will break if either one is nil, and I don't see anywhere that they're being set that would be accessible to this function. Using `or` would be better, but this condition would still always be false.

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+    inputfolder = dfhack.getDFPath() ..  args.folder .. "\\"
+end
+
+local prefix = "INTRO"
+if args.file then
+    prefix = args.file
+end
+
+if args.rand then
+    math.randomseed(os.time())
```
lethosor: I wasn't aware that this was necessary, but it does seem to be. I'd say this would be a better thing to put in the DFHack core than in individual scripts, but it's probably fine to leave it for now. You could also use DFHack's [random number generation API](https://dfhack.readthedocs.io/en/latest/docs/Lua%20API.html#random-number-generation) if you truly care about randomness.

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+    local list = io.open(dfhack.getDFPath() .. "\\data\\announcement\\index.txt","r")
+    indextext = list:read("*all")
+    io.close(list)
+end
+
+
+--Code main body, this is where the magic happens
+local function introloader(folder,dorand,doent,verbose)
+    local entity = "_" .. df.historical_entity.find(df.global.ui.civ_id).entity_raw.code
+    if ~doent then entity = "" end
```
lethosor: This won't work - if `doent` is a non-number (boolean, nil, etc.), this will raise an error. Otherwise, `~0 == -1` and `~1 == -2`, so this will always be true.
(Use `not` instead.)

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+    local inputfile = folder .. "\\" .. prefix .. entity
```
lethosor: Change to "/"

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+    
+    if dorand then
+        --Determine number of usable files
+        local filecount = 0
+        while string.gmatch(indextext,inputfile .. "_%d+") do
+            filecount++
```
lethosor: `++` isn't a Lua operator.

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+        if filecount > 0 then
+            local rand = math.random(1,filecount)
+            inputfile = inputfile .. "_" .. rand
+        end
+    end
+    
+    if verbose then print("Loading intro text") end
+    local input = assert(io.open(inputfile, "rb"),"No matching intro file, leaving as-is")
+    if input == false then return end
+    local output = assert(io.open(dfhack.getDFPath() .. "\\data\\announcement\\fortressintro", "wb"),"Could not create or open data/announcement/fortressintro, somehow.")
```
lethosor: Backslashes

modtools/intro-loader.lua
lethosor: Also, even if you're expecting both of these to be numbers, and both are 0, Lua treats anything besides false or nil as true in conditionals:
```
> if 0|0 then print('a') end
a
```

modtools/intro-loader.lua
```diff
@@ -0,0 +1,160 @@
...
+      - The files must not have file extensions
+]====]
+
+
+local utils = require 'utils'
+
+
+validArgs = validArgs or utils.invert({
+    'help',
+    'v'
```
lethosor: Missing a comma after this line

lethosor:
(Some of the Travis warnings are because this repo's build script hasn't been updated to use Lua 5.3 yet - sorry about that!)

KohaiKhaos:
I'll run over everything, I certainly missed a bunch of stuff because I ended up making far too many changes since I last had enough to test with. I definitely did not have enough simple *material* put together to test everything.
I'll make changes and run through some thorough testing before I submit again. I especially appreciate being pointed at the dfhack.filesystem.listdir() function though, as I spent so long looking over the DFhack documentation I think I went blind to finding the bits I actually needed.
If you can't tell, this is kind of my first go with making scripts like this, and Lua and exactly how it works are kind of off for me. I'll get back with this in a little while.

modtools/intro-loader.lua
KohaiKhaos: Okay, I am an absolutely silly person for using a bitwise operator there. I need to think more on the little bits, clearly, so it's time for me to run some thorough testing of every little tidbit.

lethosor:
You can just make changes to your repo and they'll be reflected here (although keep in mind that since you're using the master branch in your repo, *anything* you add there will be reflected here).
Anyway, thanks for the PR! Feel free to ask about anything else Lua-related.

modtools/intro-loader.lua
KohaiKhaos: Yes, because of the file IO being just a bit too slow. I was doing it several times to be sure, and made sure this was specifically the cause, and this was with an earlier, faster version, but essentially...
You move from the map screen to the embark prep screen, which starts the critical section that is the function. If you _immediately_ press enter, you will cause DF to experience an error due to this function being in the middle of writing to fortressintro while DF wants to read from it.
Pausing execution for the incredibly short time here ensures that you _can't_ move to fast, and the normal runtime for this is incredibly short. I almost didn't even notice this error in my earlier testbuilds, but when I started running through my test embarks a bit faster, it became quite obvious that the file I/O was dragging on just a bit too long for DF to deal with.

Additionally the function cannot be effectively run any earlier in DF's execution because your entity hasn't been set until you are leaving the map screen. If I _only_ implemented the random choice loader, it'd be fine running earlier, but the entity-specific code definitely has to be run here, and it's definitely possible for someone with a decent computer to pass through fast enough to cause an error. I can go get you a screencap of what happens if I purposefully set off the error, if you'd like.

modtools/intro-loader.lua
KohaiKhaos: I was having some issues really reading through processArgs(), honestly. Can you possibly explain what its returns really are?
The arguments being passed to the function in those variables should be getting derived from the work processArgs() is doing; the only reason they're named that there is because I'm not quite used to lua and its very global variables and their scope yet; I'm _pretty_ sure I don't even need to pass them into the function like I did.
Also, I clearly need to stop using bitwise operators.

expwnent: Basically if they type "myScript -argName1 argval1 -argName2 -argName3 argVal3" and you say args = processArgs(blah blah) then args.argName1 is "argval1", args.argName2 is non-null, and args.argName3 is "argVal3". If I remember correctly.

modtools/intro-loader.lua
KohaiKhaos: I mean, is that not what the DFhack API is really doing, except maybe calling srand() in C?
I just find it weird to *not* make a call to a seeding function when I intend to deal with random numbers, and it's just always convenient to use OS time for it since that's always changing.
Especially with a function like this where lua's normal behavior of starting every function with a fixed seed would result in you never getting a different result for your intros.

modtools/intro-loader.lua
lethosor: It should return a table with the names and values of the arguments passed to the script. I'm not sure if there's a good example, but I'll see if I can find one later.

modtools/intro-loader.lua
expwnent: In general it would be better to use the API to get your own independent random number generator so that it'll be easier to debug if other scripts use randomness, but it's probably fine.

modtools/intro-loader.lua
KohaiKhaos: And empty values are regarded as "True"s, and I assume ones that aren't put in the call are returned as "Nil"s then?

KohaiKhaos:
And now I get settled in for some nice hard testing of every way this can go.

modtools/intro-loader.lua
expwnent: What do you mean by empty values? Do you mean `args[nil]` or `args.someArgumentThatWasNotSpecified`?

modtools/intro-loader.lua
lethosor: DFHack's RNG API is more complicated than that. It was implemented for C++ tools that needed more complicated features, like 3dveins, then exposed to Lua.

KohaiKhaos:
I got my first run on pure random working. Entity comes next.
I would like to point out that I've gone through like, 5-6 runs just ensuring the random operation; I'm pretty sure the documentation on rng:random([limit]) is wrong; it says the range is 0 to [limit], but so far my tests seem to indicate a range from 0 to [limit]-1.
Actually I'm kinda sorta 100% sure of it.
Open up the lua interpreter in DFhack with: lua dfhack.interpreter()
In there, go with
rng = dfhack.random.new()
for i=1,100000 do if rng:random(2)>1 then print("I did it") end end

It won't print a single time.

lethosor:
Yes, that's what the docs say:
> * rng:random([limit])
> Returns a random integer. If limit is specified, the value is in the range [0, limit); otherwise it uses the whole 32-bit unsigned integer range.

`[0, limit)` means that the range includes 0 but excludes limit.

KohaiKhaos:
Ack, curse these parentheses. It's been way way way too long since I had to remember that they meant silly things like that.

modtools/intro-loader.lua
```diff
@@ -82,63 +82,64 @@ if args.help then
...
-    
-    local list = io.open(dfhack.getDFPath() .. "\\data\\announcement\\index.txt","r")
-    indextext = list:read("*all")
-    io.close(list)
+    local index = io.open(dfhack.getDFPath() .. "/data/announcement/index.txt","w")
+	local files = dfhack.filesystem.listdir(inputfolder)
+    for foo,filename in pairs(files) do
+		index:write(filename .. "\n")
+	end
+    io.close(index)
```
lethosor: You could just use the result of listdir() directly (it's a Lua table), instead of writing it to a file and reading from the file again. It's probably faster too.

lethosor:
Yeah, I suppose the RNG docs could be clarified a bit.

modtools/intro-loader.lua
KohaiKhaos: Oh my brain derped on that, since I had to do it the hard way earlier and that needed to be way separated from everything else just to make the document sensible. Notably, the filenames are actually in the second 'column' from listdir, oddly enough. That was confounding until I actually looked into index.txt and saw it was just full of numbers.

KohaiKhaos: local index = " "
		for foo,filename in pairs(dfhack.filesystem.listdir(folder)) do
			index = index .. "\n" .. filename
		end

Inside the function should probably work

modtools/intro-loader.lua
lethosor: That's interesting. I was under the impression that all code in onStateChange hooks was run with DF suspended. This *seems* to confirm that, although it's possible that DF starts processing the screen (calling its logic() method or something) before DFHack's state change event can trigger.
```lua
dfhack.onStateChange.test=function() print('sleeping') os.execute('sleep 1') print('done') end
```
(This will pause DF for a second every time you switch screens, and I haven't seen it display the new screen before that time is up yet.)

modtools/intro-loader.lua
lethosor: Empty values are empty strings, but they still evaluate to `true`. Here's the result of playing around with it a bit in the Lua interpreter:
```
[lua]# utils = require 'utils'
[lua]# ~utils.processArgs({'-arg1','value1'})
table: 0x1171565d0
arg1                          = value1
[lua]# ~utils.processArgs({'-arg1','value1','-arg2'})
table: 0x116f943f0
arg1                          = value1
arg2                          = 
[lua]# ~utils.processArgs({'-arg1','value1','-arg2'}).arg1
value1
[lua]# ~utils.processArgs({'-arg1','value1','-arg2'}).arg2

[lua]# ~utils.processArgs({'-arg1','value1','-arg2'}).arg3
nil
[lua]# 
```

modtools/intro-loader.lua
lethosor: To be fair, I had to double-check this. The bitwise operators are new in Lua 5.3, so I wasn't sure what was possible to do with them.

modtools/intro-loader.lua
lethosor: > Notably, the filenames are actually in the second 'column' from listdir, oddly enough.

Oh, that's because pairs() returns pairs of (index, value). You'll get the same thing with this:
```
[lua]# for k, v in pairs({'a','b','c'}) do print(k, v) end
1    a
2    b
3    c
```

KohaiKhaos:
Entity works with this last push too, that one was a simple look over since I had it working in a much earlier build that used identical logic; it's the addition of randomness that's new.

lukebemish:
Is this feature still being developed/worked on? I think it would be quite useful, and was wondering what it's status is.

lethosor:
It has been a while since I looked at it. From what I remember, it looks like the main question still is about how to clean up event handling, e.g. so `with_suspend()` (which is a no-op) can be removed. It looks like the fork has been deleted, so making changes within this PR is a bit difficult.

myk002:
If nobody wants to pick this up, could we close the PR? The code will still be available in the PR if someone wants to dig it up later.

PR #27, opened 2017-07-30
Update skill-change.lua

Atkana:
Add in support for increasing/decreasing levels as a result of adding/removing experience to beyond the level's range, to the best of my mathematical ability/limited knowledge on how exp works.
Also adding in an arg for the printing of the changes, since IMO a modtool shouldn't do that automatically.

Sorry if I've screwed something up, I barely understand how to use github :P

### DIFF
--- a/modtools/skill-change.lua
+++ b/modtools/skill-change.lua
@@ -1,7 +1,6 @@
 -- Sets or modifies a skill of a unit
 --author expwnent
 --based on skillChange.lua by Putnam
---TODO: update skill level once experience increases/decreases
 --TODO: skill rust?
 local help = [====[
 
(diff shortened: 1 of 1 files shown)


maxthyme:
Oh hell are we supposed to be pushing to the 1130 branch? Also, isn't this already a part of gm-unit?

lethosor:
No. That branch shouldn't even exist in this repo.

Atkana:
> Oh hell are we supposed to be pushing to the 1130 branch? Also, isn't this already a part of gm-unit?

gm-unit can be used to edit a unit's skill rating via gui, but can't be used for exp or via scripts.
Also sorry about the wrong branch thing, about the only instruction I understood from the contributing guidelines was the bit that explicitly said to not branch from master and I just assumed dfhack-1130 was the develop branch. Is it all sorted or is there anything I need to do?

lethosor:
The "develop branch only" part only applies to the main DFHack repo. Sorry for the confusion there. The dfhack-1130 branch was from #23, for https://github.com/DFHack/dfhack/issues/1130.

PR #30, opened 2017-08-20
Update siren.lua

OwnageIsMagic:
`flags` [now bitfield](https://github.com/DFHack/df-structures/blob/master/df.units.xml#L1697) and [docs](https://dfhack.readthedocs.io/en/stable/docs/Lua%20API.html#bitfield-references) says that I can simply assign a lua table for that
Will it call default constructor or something similar? 
Or I should set every field in it?

> offtopic:
> xml says there only 4 fields but `pairs` iterate up to 31
> can I safely use last bits in bitfiels for my own purposes?

### DIFF
--- a/siren.lua
+++ b/siren.lua
@@ -51,7 +51,7 @@ function add_thought(unit, emotion, thought)
     thought=thought,
     subthought=0,
     severity=0,
-    flags=0,
+    flags={},
     unk7=0,
     year=df.global.cur_year,
     year_tick=df.global.cur_year_tick})


lethosor:
Thanks! I think flags probably got changed to an actual bitfield in the XML at some point, which broke this.

You can probably use flags for your own purposes, until DF starts using them (probably unlikely, but be careful).

OwnageIsMagic:
Does dfhack have any policies on setting `unk*` values?
looks like empty table just create `<uint32_t>0` as value, is it right?

lethosor:
0 should be right. Not sure what you were asking about `unk` fields. You can use them, although suggesting names for them is better.

PR #32, opened 2017-10-17
export-dt-ini: Add new offsets

cvuchener:
Future versions of DT will need this.


cvuchener:
I am adding more offsets here, since this is still open.

lethosor:
Thanks! I'm merging this in for r3, but feel free to add more in the future if you need to.

PR #37, opened 2017-12-10
Couple updates, couple new scripts.

maxthyme:
Cleaned out some of the other stuff and fixed a few bugs I know of, kept the travis fixes so it should pass I think.


maxthyme:
ARGH, freaking new firefox profile pointed me at the wrong folder, hang on, had to sort out what happened there.

worldgenedit.lua
```diff
@@ -0,0 +1,165 @@
...
+--[====[
+
+worldgenedit
+============
+Perhaps of dubious utility in general, being able to alter an active worldgen is handy.
+
+]====]
+local gui = require 'gui'
+local widgets = require 'gui.widgets'
+local wv = dfhack.gui.getCurViewscreen()
```
lethosor: Not totally sure what "wv" stands for

worldgenedit.lua
```diff
@@ -0,0 +1,165 @@
...
+local wd = df.global.world.world_data
+local wg = df.global.world.worldgen.worldgen_parms
+local sites = df.global.world.world_data.sites
+local pent, dent, eent, hent, gent, stn, stc, pop, eny
```
lethosor: A lot of these are hard to understand. Longer, more descriptive variable names would be better. I'd recommend some plugin (or editor) that can autocomplete names for you, if you don't have that ability already, so that typing them out is less annoying.

articrap.lua
```diff
@@ -0,0 +1,48 @@
+--A half-assed way to get artifact quality gear without the actual artifact or artifake.
```
lethosor: Not a great script name

assumecontrol.lua
```diff
@@ -0,0 +1,64 @@
+--Assume Direct Control of the unit you're viewing. This Can Hurt You.
```
lethosor: A bunch of newer scripts use hyphens to separate words in their names, like `assume-control`. I know they're not all consistent, but for this (and some others), it might be more readable if you do that.

breakdance.lua
```diff
@@ -0,0 +1,13 @@
+--Breaks up a stuck dance activity.
```
lethosor: Maybe `stop-dance` for clarity.

names.lua
```diff
@@ -1,68 +1,118 @@
...
- return
+    return
 end
-
-if not df.viewscreen_layer_choose_language_namest:is_instance(dfhack.gui.getCurViewscreen()) then
-    choices = df.new(df.viewscreen_setupadventurest) ; choices.subscreen = 3 ; gui = require 'gui' ; gui.simulateInput(choices, 'A_RANDOM_NAME') ; gui.simulateInput(choices, 'A_CUST_NAME')
+namescr = defclass(namescr, gui.Screen)
+namescr.focus_path = 'names'
+function namescr:init()
+    if not dfhack.gui.getViewscreenByType(df.viewscreen_layer_choose_language_namest)==true then
```
lethosor: Instead of `if not x == true`, use `if not x`. What you have could be misinterpreted as `if not (x == true)`, but it's actually `if (not x) == true`. 

As far as I can tell, you're checking to see if the top viewscreen is *not* a `viewscreen_layer_choose_language_namest`. I'm not sure why you didn't use `is_instance` for this, but it would be more consistent. `getViewscreenByType` is useful if you want to get a viewscreen that's *somewhere* in the viewscreen stack by passing a second parameter.

names.lua
```diff
@@ -1,68 +1,118 @@
...
-:-unit:     if viewing a unit
-:-first [Somename | "Some Names like This":
-            if a first name is desired, leave blank to clear current first name
 ]====]
+
+local gui = require 'gui'
+local dlg = require 'gui.dialogs'
+local widgets = require 'gui.widgets'
 local utils = require 'utils'
+local vw
```
lethosor: You don't really have to do this. Global variables in scripts won't leak outside of scripts. Also, these should probably be members of `namescr` instead of globals anyway, and you can replace `vw` with `self._native.parent` in namescr functions (I think), which is the only place you're using `vw` anyway.

updatefinish.lua
```diff
@@ -0,0 +1,144 @@
+--Alter the date which the updating region process finishes at.
```
lethosor: Am I correct in assuming that this requires you to run it when the screen is advancing (so you only have a short window when you can run it, and have to wait for the world to load first)? If so, it might be nice to allow it to be run *before* the world update screen appears, and intercept that screen when it does show up. A state change handler would probably work for that.

Also, this is another example of documentation in different places - the first line here won't show up on https://dfhack.readthedocs.io/.

crithit.lua
```diff
@@ -0,0 +1,22 @@
+--Turns the previously readied attack into a super powerful critical hit.
```
lethosor: Maybe `critical-hit`

articrap.lua
maxthyme: Hmmm, artifake for this and artimake for the artifake script might make more sense?

assumecontrol.lua
maxthyme: Don't like hyphens, and I don't think camelCase fits the styling does it?

breakdance.lua
maxthyme: Not even sure if this one is necessary actually, probably just pull it out entirely since it was mostly a workaround for the bug that caused hung dances which I think Toady fixed.

names.lua
maxthyme: I could swear I did this before moving the vw definition into the namescr setup actually.

names.lua
maxthyme: Weirdly I used the is_instance earlier and derped about the if not x logic.

names.lua
```diff
@@ -1,68 +1,118 @@
...
+            uid = vw.parent.unit.id
+            trg = df.unit.find(uid)
+        elseif df.global.ui_advmode.menu==1 then
+            local t_look=df.global.ui_look_list.items[df.global.ui_look_cursor]
+            if t_look.type==2 then
+                trg=t_look.unit
+            end
+        end
+        choices = df.viewscreen_setupadventurest:new()
+        choices.page = 7
```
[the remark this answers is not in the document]
maxthyme: It changed a few times but I can fix it since we have a defined name for it now.

updatefinish.lua
maxthyme: Not sure exactly how to make it do that but I'll look into it, outside of super quick updating worlds it gives you plenty of time.

worldgenedit.lua
maxthyme: Was probably a typo supposed to be vw.

questport.lua
```diff
@@ -0,0 +1,44 @@
...
+    df.global.ui_advmode.unk_1=qx
+    df.global.ui_advmode.unk_2=qy
+        if df.global.ui_advmode.menu==0 then
+            gui.simulateInput(qp.child,'LEAVESCREEN')
+            df.global.ui_advmode.menu=26
+            df.global.ui_advmode.travel_not_moved=true
+            gui.simulateInput(qp,'CURSOR_DOWN')
+            dfhack.timeout(15,'frames',function()
+            gui.simulateInput(qp,'A_TRAVEL_LOG') end)
+        elseif df.global.ui_advmode.menu==26 then
```
lethosor: Should use the enum for this - `df.ui_advmode_menu.Travel`?

names.lua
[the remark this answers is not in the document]
maxthyme: Hell did I not say that I'm ok with you pushing to my fork?

I've been really loopy and sick for a while.

lethosor:
I forgot I could push to your fork - that might make it easier for me to resolve whatever conflicts are left for you.

crithit.lua
lethosor: Also, I can't get this one to work. I'm trying it from multiple places in the `A` menu in adventure mode, but that might not be right.

lethosor:
Merged so far:
### Updates:
- launch
- names
### New scripts:
- break-dance (renamed)
- cannibalism
- fillneeds
- firestarter
- flashstep
- ghostly
- questport

lethosor:
@maxthyme ok, I updated your branch by removing all the files I modified on DFHack's end from your branch, then merging DFHack's branch into yours. That managed to avoid most of the conflicts; I still had to tell it to use the existing files instead of the ones I deleted in a couple cases, but it seems to have worked.

Anyway, [here](https://github.com/DFHack/scripts/commits?author=lethosor&since=2018-02-03T05:00:00Z&until=2018-02-04T05:00:00Z) is a list of the changes I made when I merged in those 9, if you're interested. It's in reverse chronological order, so start at the bottom. Let me know if you have questions.

I was also able to name some `df.fire` fields thanks to `firestarter` (I don't know exactly what they do, but they're some kind of temperatures), so thanks for that.

myk002:
What is the status of this PR? It looks like the last update was several years ago. Can we close the PR?

PR #41, opened 2018-01-03
Bugfix reaction-product-trigger

rndmvar:
The external function call getWorkerAndBuilding fails in Adventure mode when crafting outside of a building.  Requiring that unit.job.current_job not be 'nil' before calling the external function works around this issue.  These checks are carried out in 'modtools/reaction-trigger.lua' before calling the internal function.  So, changing reaction-product-trigger.lua, to follow the methodology present in reaction-trigger.lua, probably makes the most sense.

### DIFF
--- a/modtools/reaction-product-trigger.lua
+++ b/modtools/reaction-product-trigger.lua
@@ -79,7 +79,11 @@ end
 
 local function afterProduce(reaction,reaction_product,unit,input_items,input_reagents,output_items)
  --printall(unit.job.current_job)
- local _,buildingId = dfhack.script_environment('modtools/reaction-trigger').getWorkerAndBuilding(unit.job.current_job)
+ local _,buildingId = -1,-1
+ -- unit.job.current_job might be nil for Adventure mode
+ if unit.job.current_job then
+  _,buildingId = dfhack.script_environment('modtools/reaction-trigger').getWorkerAndBuilding(unit.job.current_job)
+ end
  for _,hook in ipairs(productHooks[reaction.code] or {}) do
   local command = hook.command
   local processed = processArgs(command, reaction, reaction_product, unit, input_items, input_reagents, output_items, buildingId)


modtools/reaction-product-trigger.lua
```diff
@@ -79,7 +79,11 @@ end
 
 local function afterProduce(reaction,reaction_product,unit,input_items,input_reagents,output_items)
  --printall(unit.job.current_job)
- local _,buildingId = dfhack.script_environment('modtools/reaction-trigger').getWorkerAndBuilding(unit.job.current_job)
+ local _,buildingId = -1,-1
```
lethosor: What happens below if `buildingId` remains -1 now?

modtools/reaction-product-trigger.lua
```diff
@@ -79,7 +79,11 @@ end
...
+ -- unit.job.current_job might be nil for Adventure mode
+ if unit.job.current_job then
+  _,buildingId = dfhack.script_environment('modtools/reaction-trigger').getWorkerAndBuilding(unit.job.current_job)
+ end
  for _,hook in ipairs(productHooks[reaction.code] or {}) do
   local command = hook.command
   local processed = processArgs(command, reaction, reaction_product, unit, input_items, input_reagents, output_items, buildingId)
```
lethosor: @rndmvar specifically, I'm not sure that this line or the line after will work well if buildingId is -1 - it might pass through a building ID of -1 to callbacks that aren't expecting it. Would it be better to bail out in that case, similar to https://github.com/DFHack/scripts/blob/b330e696e6cb6a607d8cc77679bbe0fa5e51598c/modtools/reaction-trigger.lua#L134-L140 ? I'm not very familiar with modtools scripts.

myk002:
Closing old PRs. Please reopen if this PR is still active.

PR #43, opened 2018-01-20
view-unit-reports: View combat reports for currently selected unit

danamlund:
Script to solve the problem of having to manually find combat reports in 'r' for specific unit.

Current unit is unit near cursor in 'v', selected in 'u' list,
unit/corpse/splatter at cursor in 'k'. And newest unit with race when
'k' at race-specific blood spatter.

If this is considered useful enough it can be added to dfhack.init-example in the following scopes (I would prefer it as a single keybinding since 4 lines is too much clutter)
```
# show battle report for selected unit
keybinding add Ctrl-Shift-R@dwarfmode/ViewUnits view-unit-reports
keybinding add Ctrl-Shift-R@unitlist view-unit-reports
keybinding add Ctrl-Shift-R@dwarfmode/LookAround view-unit-reports
keybinding add Ctrl-Shift-R@textviewer view-unit-reports
```

### DIFF
--- a/view-unit-reports.lua
+++ b/view-unit-reports.lua
@@ -47,15 +47,6 @@ local function find_last_unit_with_combat_log_and_race(race_id)
    return output
 end
 
-local function find_unit(unit_id)
-   for _, unit in pairs(df.global.world.units.all) do
-      if unit.id == unit_id then
-         return unit
-      end
-   end
-   return nil
-end
-
 local function current_selected_unit()
    -- 'v' view, 'k' at unit, unit list view
    local unit = dfhack.gui.getSelectedUnit(true)
@@ -66,7 +57,7 @@ local function current_selected_unit()
    -- 'k' over a corpse
    local item = dfhack.gui.getSelectedItem(true)
    if df.item_corpsest:is_instance(item) then
-      return find_unit(item.unit_id)
+      return df.unit.find(item.unit_id)
    end
 
    -- 'k' over some spatter (e.g. blood)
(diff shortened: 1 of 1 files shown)


view-unit-reports.lua
```diff
@@ -0,0 +1,95 @@
...
+   local output = nil
+   for _, unit in pairs(df.global.world.units.all) do
+      if unit.race == race_id and #get_combat_logs(unit) >= 1 then
+         output = unit
+      end
+   end
+   return output
+end
+
+local function find_unit(unit_id)
```
lethosor: Replace this with df.unit.find(id) - it's faster and will behave the same way.

view-unit-reports.lua
```diff
@@ -0,0 +1,95 @@
...
+      return unit
+   end
+
+   -- 'k' over a corpse
+   local item = dfhack.gui.getSelectedItem(true)
+   if df.item_corpsest:is_instance(item) then
+      return find_unit(item.unit_id)
+   end
+
+   -- 'k' over some spatter (e.g. blood)
```
lethosor: Nice features here!

danamlund:
Fixed help formatting and rebased into single commit. to pass travis check.

view-unit-reports.lua
danamlund: Good to know

view-unit-reports.lua
danamlund: Glad to hear, thanks.

lethosor:
Added your keybinding to dfhack.init-example too

danamlund:
That is great, thank you.

lethosor:
Oh, I also just noticed the keybindings listed here. You can actually specify multiple contexts separated with `|`, like this:
```
keybinding add Ctrl-Shift-R@dwarfmode/ViewUnits|unitlist|dwarfmode/LookAround|textviewer view-unit-reports
```
But since getSelectedUnit() works in many more places than that, it might be better to not specify a context anyway.

PR #44, opened 2018-01-25
fix gui/unit-info-viewer and refactor view-unit-reports

danamlund:
gui/unit-info-viewer have nice find-current-unit functionality. Improved it by
- not using deprecated api enum
- fixing some cases
- removing cases that are now handled by dfhack.gui.getSelectedUnit
- find unit from blood spatter of historical figure
- making it a module

view-unit-reports is refactored to use the find-current-unit functionality from gui/unit-info-viewer

### DIFF
--- a/gui/unit-info-viewer.lua
+++ b/gui/unit-info-viewer.lua
@@ -78,9 +78,6 @@ function getUnit_byVS(silent)    -- by view screen mode
         elseif v.mode == 2 then -- select trainer
             u = v.trainer_unit[v.trainer_cursor]
         end
-    elseif df.viewscreen_layer_workshop_profilest:is_instance(v) then
-        -- context: @layer_workshop_profile/Unit     -- (q)uery workshop -> (P)rofile -- df.global.ui.main.mode == 17
-        u = v.workers[v.layer_objects[0].cursor]
     elseif df.viewscreen_layer_overall_healthst:is_instance(v) then
         -- context @layer_overall_health/Units -- z -> health
         u = v.unit[v.layer_objects[0].cursor]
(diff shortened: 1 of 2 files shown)


danamlund:
gui/unit-info-viewer now also finds current user in (r)eports.
And when viewing a combat log for a unit, it finds on the other unit mentioned in the current combat log (if available).

Roses mentioned this case on the forums.

lethosor:
Most of the find-unit stuff should probably be in library/modules/Gui.cpp. Maybe the only exception is the unit-from-spatter feature, which could probably be in library/lua/gui.lua or something - importing the "gui/view-unit-info" script for that doesn't make a lot of sense in general.

I understand that these files are in the dfhack repo, not this one, so I'm willing to take on that refactoring and merge this as-is. I've also been promising some people a r1 at some point, so just to gauge what you think - are these important features that should be merged ASAP, or mostly cleanup? (I do appreciate the whitespace cleanup too, by the way.)

danamlund:
These are just cleanup, they can wait.
I will convert these changes to Gui.cpp and gui.lua and make a new pull request on dfhack.

One question: getting other unit from a combat log requires searching through all reports for all units. Is that extra processing time acceptable to have in Gui::getAnyUnit ?

danamlund:
Since all changes in this pull request are being implemented in Gui.cpp instead I will close it.

After Gui.cpp changes are committed we can clean up all the extra find-unit logic from gui/unit-info-viewer.lua. But that is just code cleanup, the functionality will not change.

danamlund:
Pull request for implementation in Gui.cpp: https://github.com/DFHack/dfhack/pull/1218

PR #46, opened 2018-01-28
Added adv_fix_sleepers

ArrowThunder:
Added a (hopefully buildable version) of adv_fix_sleepers, a script to fix bugged sleepers.


adv_fix_sleepers.lua
```diff
@@ -0,0 +1,63 @@
...
+
+]====]
+
+--========================
+-- Author: ArrowThunder on bay12 & reddit
+-- Version: 1.0
+--=======================
+
+-- gets the army controller based on the army controller id
+local function get_army_controller(id)
```
lethosor: Fairly certain this whole thing is equivalent to df.army_controller.find(id)

lethosor:
I'd probably rename this to adv-fix-sleepers for consistency (and it'd avoid having to press shift on US keyboards, maybe others).

ArrowThunder:
Used DFHack Lua API's .find() function instead of my homebrew one. Retested the script, just to be safe, and it still works. Renamed the script to match the conventions set by other scripts (dashes, not underscores).

adv-fix-sleepers.lua
```diff
@@ -0,0 +1,46 @@
+--Fixes all local bugged sleepers in adventure mode.
+--[====[
+
+adv-fix-sleepers
+================
+Fixes :bug:'6798'. This bug is characterized by sleeping units who refuse to
```
lethosor: Links need to use backticks, not single quotes

PR #47, opened 2018-02-16
export-dt-ini: name offset update

cvuchener:
Use name offsets from language_name instead of unit, since it is no longer at the
beginning of the structure (because unit became a class type).

Sorry, I forgot to send you that change. It is required from DT 39.2 and incompatible with older versions.


devel/export-dt-ini.lua
```diff
@@ -289,9 +291,7 @@ address('tissue_name',df.tissue_template,'tissue_name_singular')
 address('tissue_flags',df.tissue_template,'flags')
 
 header('dwarf_offsets')
-address('first_name',df.unit,'name','first_name')
-address('nick_name',df.unit,'name','nickname')
```
lethosor: This wasn't taking the vtable pointer into account? Weird.

cvuchener: DT was using these offsets in other case than unit.name (e.g. identity). It used to work because it was at the beginning so it was the same as language_name offsets. Because of the added vtable pointer, it is not any more, and it required breaking changes in DT.

PR #48, opened 2018-02-25
export identity information

robertjanetzko:
extends exportlegends.lua to export information about identities uses by historical figures

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -264,6 +264,22 @@ function export_more_legends_xml()
     end
     file:write("</historical_figures>\n")
 
+    file:write("<identities>\n")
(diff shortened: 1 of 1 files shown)


JapaMala:
Those Anon values should never be used. Make a PR for df-structures first, to identify them, then use them in a script.

robertjanetzko:
added pull request to df-structures: https://github.com/DFHack/df-structures/pull/244

lethosor:
That was merged. Can you update the field name here (or I can push to your branch, I think, but I assume you have a better way of testing this than I do).

lethosor:
Whoops, sorry for missing the update to this. Looks good to me.

PR #49, opened 2018-03-27
item-trigger update: specify inventory mode

AtomicChicken:
Previously, item-trigger's inventory update checks only differentiated between equipping and unequipping items. This means that something like
modtools/item-trigger -itemType ITEM_HELM_CAP -onEquip -command [ foo ]
would trigger foo both in cases where a cap was being worn, and in cases where it was simply being held in the unit’s hand.

With this update, the user will now be able to specify the inventory mode alongside onEquip and onUnequip to allow for greater selectivity. Taking the above example, if we wanted to trigger foo only in cases where a cap was being worn, we could specify this via
modtools/item-trigger -itemType ITEM_HELM_CAP -onEquip 2 -command [ foo ]
or
modtools/item-trigger -itemType ITEM_HELM_CAP -onEquip [ 2 ] -command [ foo ]
where ‘2’ is the mode type corresponding to ‘Worn’.

Multiple inventory modes can be specified. For example, 
modtools/item-trigger -itemType RING -onUnequip [ 0 1 2 ] -command [ foo ]
would trigger foo when a ring is unequipped in cases where it was being hauled, held or worn.

A list of inventory mode values has been included in the -help printout.
Specifying mode remains optional, preserving backwards compatibility.

### DIFF
--- a/modtools/item-trigger.lua
+++ b/modtools/item-trigger.lua
@@ -24,10 +24,26 @@ Arguments::
             RING
     -onStrike
         trigger the command on appropriate weapon strikes
-    -onEquip
+    -onEquip mode
         trigger the command when someone equips an appropriate item
-    -onUnequip
+        Optionally, the equipment mode can be specified
+        Possible values for mode:
+            0 = Hauled
+            1 = Weapon
+            2 = Worn
+            3 = Piercing
+            4 = Flask
+            5 = Wrapped around
+            6 = Stuck in
+            7 = In Mouth
+            8 = Pet
+            9 = Sewn into
+            10 = Strapped
+        multiple values can be specified simultaneously
+        example: -onEquip [ 1 2 0 ]
+    -onUnequip mode
         trigger the command when someone unequips an appropriate item
+        see above note regarding 'mode' values
     -material mat
         trigger the commmand on items with the given material
         examples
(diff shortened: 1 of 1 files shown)


modtools/item-trigger.lua
```diff
@@ -120,61 +136,85 @@ function getitemType(item)
...
+ for _,command in ipairs(triggerArg) do
+  if command[""..table.mode..""] then
+   local reqModeType = command[""..table.mode..""]
+   local modeType = tonumber(table.modeType)
+   if #reqModeType == 1 then
+    if (tonumber(reqModeType) or tonumber(reqModeType[1])) == modeType then
+     utils.fillTable(command,table)
+     processTrigger(command)
+     utils.unfillTable(command,table)
+	end
```
lethosor: tab -> 4 spaces

modtools/item-trigger.lua
```diff
@@ -24,10 +24,26 @@ Arguments::
             RING
     -onStrike
         trigger the command on appropriate weapon strikes
-    -onEquip
+    -onEquip mode
         trigger the command when someone equips an appropriate item
-    -onUnequip
+        Optionally, the equipment mode can be specified
+        Possible values for mode:
+            0 = Hauled
```
lethosor: Is this a DF enum? If so, you should use human-readable names instead of making people look up numbers every time they want to want to read/write a command.

modtools/item-trigger.lua
```diff
@@ -120,61 +136,85 @@ function getitemType(item)
...
- local mode = (isEquip and 'onEquip') or (not isEquip and 'onUnequip')
-
+function equipHandler(unit, item, mode, modeType)
  local table = {}
  table.mode = mode
+ table.modeType = modeType
  table.item = df.item.find(item)
  table.unit = df.unit.find(unit)
- if table.item and table.unit then -- they must both be not nil or errors will occur after this point with instant reactions.
+ if table.item and table.unit then--they must both be not nil or errors will occur after this point with instant reactions.
```
lethosor: Spaces around comments make them more readable. No need to remove them.

modtools/item-trigger.lua
lethosor: On second glance, it looks like you just documented this. Not a big deal, then, but it would be nice to be able to use enum values (if this is in fact an enum).

AtomicChicken: Noted. I suppose it'd cause far less hassle to everyone involved.

modtools/item-trigger.lua
AtomicChicken: My bad, not sure why I did that.

modtools/item-trigger.lua
AtomicChicken: Huh, now how did that get in there?

PR #50, opened 2018-03-28
item-trigger: support multiple conditions

AtomicChicken:
A significant rewrite allowing the user to specify any combination of itemType, material and contaminant.
Previously, only one of the above could be used at a time, which greatly restricted the script's flexibility and limited its usefulness to modders.

Also fixed the outdated(?) creature/plant material token format demonstrated in the -help printout, and added a few more examples.

### DIFF
--- a/modtools/item-trigger.lua
+++ b/modtools/item-trigger.lua
@@ -78,13 +78,13 @@ Arguments::
 local eventful = require 'plugins.eventful'
 local utils = require 'utils'
 
-triggers = triggers or {}
+itemTriggers = itemTriggers or {}
 eventful.enableEvent(eventful.eventType.UNIT_ATTACK,1) -- this event type is cheap, so checking every tick is fine
 eventful.enableEvent(eventful.eventType.INVENTORY_CHANGE,5) -- this is expensive, but you might still want to set it lower
 eventful.enableEvent(eventful.eventType.UNLOAD,1)
 
 eventful.onUnload.itemTrigger = function()
- triggers = {}
+ itemTriggers = {}
 end
 
 function processTrigger(command)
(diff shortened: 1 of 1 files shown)


modtools/item-trigger.lua
```diff
@@ -146,57 +143,105 @@ function compareInvModes(reqMode,itemMode)
...
 
-function checkMode(table,triggerArg)
- for _,command in ipairs(triggerArg) do
-  if command[""..table.mode..""] then
-   local reqModeType = command[""..table.mode..""]
-   local modeType = tonumber(table.modeType)
+function checkMode(triggerArgs,table)
+ local mode = table.mode
+ for _,argArray in ipairs(triggerArgs) do
+  if argArray[""..mode..""] then
```
lethosor: `tostring(mode)` or `""..mode` should work

modtools/item-trigger.lua
```diff
@@ -48,14 +49,15 @@ Arguments::
         trigger the commmand on items with the given material
         examples
             INORGANIC:IRON
-            CREATURE_MAT:DWARF:BRAIN
-            PLANT_MAT:MUSHROOM_HELMET_PLUMP:DRINK
+            CREATURE:DWARF:BRAIN
```
lethosor: Not familiar with this script... does this break backwards compatibility?

AtomicChicken: I suppose so? The issue is that dfhack.matinfo.decode(item):getToken() returns strings in the format displayed above, which is incongruous with the format previously documented in this script's -help printout. Specifying something like "-material CREATURE_MAT:DWARF:BRAIN" shouldn't have worked at all in recent versions (tested it as far back as 0.43.05 to check), and the fact that this has gone unreported for so long indicates that the script probably hasn't seen much use in this regard.

modtools/item-trigger.lua
AtomicChicken: Changed.

modtools/item-trigger.lua
lethosor: Oh, I missed that part in the PR description, and thought you had actually changed the format here. I really don't know much about the modtools scripts, so thanks for putting in the time to work on them.

PR #52, opened 2018-05-07
create-unit: bug fixes and update

AtomicChicken:
- creatures of the appropriate age are now spawned as babies or children where applicable
- fix: civ_id is now properly assigned to historical_figure, resolving several hostility issues (spawned pets are no longer attacked by fortress military!)
- fix: unnamed creatures are no longer spawned with a string of numbers as a first name


modtools/create-unit.lua
```diff
@@ -338,13 +340,6 @@ function domesticate(uid, group_id)
     u.animal.population.population_idx = -1
     u.animal.population.depth = -1
 
-    u.counters.soldier_mood_countdown = -1
```
lethosor: Which change is this related to? This section was last changed in 96b118b2e7fe8d91b59e0cc1c7951bf27a5e6698, which led to the merge conflict here, but I can't imagine the first two fields here (at least) corresponding to the bugs you fixed.

lethosor: And for reference, the rest of this section dates back to the [initial commit](https://github.com/DFHack/scripts/blame/038499117c6391431f2e2d57d73856254029c391/modtools/create-unit.lua#L341) of this script (only the field names were changed recently)

AtomicChicken: To clarify, I did not remove this section, but simply relocated it from the domesticate() function to line 552. This is because, as far as I can tell, these changes should apply to all units, not just tamed creatures.

lethosor: Ah, thanks

lethosor:
Just so you know if you try to push more changes - I used the "resolve conflicts" button here, which made a new commit, so be sure to pull that.

PR #55, opened 2018-05-16
Added bodyswap.lua

AtomicChicken:
A revived version of the long defunct "adv-bodyswap" with some added functionality.

This script allows the player to swap their control over to any unit present on the map in adventure mode. Targets are specified either via their unit id or through UI selection prior to running the script (as detailed in the script usage documentation). Unlike previous iterations, units lacking nemesis or historical figure data (such as wild animals) are valid and safe targets, as the script creates this data when it is missing using functions from modtools/create-unit.

### DIFF
--- a/bodyswap.lua
+++ b/bodyswap.lua
@@ -1,0 +1,255 @@
+-- Shifts player control over to another unit in adventure mode.
+-- author: Atomic Chicken
+-- based on "assumecontrol.lua" by maxthyme, as well as the defunct advtools plugin "adv-bodyswap"
+-- nemesis and historical figure creation uses slightly modified functions from modtools/create-unit
+
+--[====[
+
+bodyswap
+========
+
+Shifts player control over to another unit (including wild animals) in adventure mode.
+Usage:
+
+:bodyswap:
+    Swaps into the unit selected in the UI (for example, when viewing the unit's status screen or description).
+
+:bodyswap -unit [unitId]:
+    Swaps into the unit corresponding to the specified id.
+
+]====]
+
+local utils = require 'utils'
+validArgs = validArgs or utils.invert({
+'unit',
+'help'
+})
+local args = utils.processArgs({...}, validArgs)
+
+local usage = [====[
+
+bodyswap
+========
+This script allows the player to gain control over a new unit in adventurer mode
+whilst simultaneously loosing control over their current character.
+
+To specify the target unit, simply select it in the user interface,
+such as by opening the unit's status screen or viewing its description
+and enter "bodyswap" in the DFHack console.
+
+Alternatively, the target unit can be specified by its unit id as shown below.
+
+Arguments::
+
+    -unit id
+        replace "id" with the unit id of your target
+        example: 
+            bodyswap -unit 42
+
+]====]
+
+if args.help then
+ print(usage)
+ return
+end
+
+if not dfhack.world.isAdventureMode() then
+  qerror("This script can only be used in adventure mode!")
+end
+
+function setNewAdvNemFlags(nem)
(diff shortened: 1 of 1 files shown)


bodyswap.lua
```diff
@@ -0,0 +1,255 @@
...
+'help'
+})
+local args = utils.processArgs({...}, validArgs)
+
+local usage = [====[
+
+bodyswap
+========
+This script allows the player to gain control over a new unit in adventurer mode
+whilst simultaneously loosing control over their current character.
```
lethosor: losing

bodyswap.lua
```diff
@@ -0,0 +1,255 @@
...
+]====]
+
+local utils = require 'utils'
+validArgs = validArgs or utils.invert({
+'unit',
+'help'
+})
+local args = utils.processArgs({...}, validArgs)
+
+local usage = [====[
```
lethosor: I don't think two versions of the documentation is ideal. Sphinx should pick up anything between `[====[` and `]====]`, even if it's a string instead of a comment (see repeat, spawnunit, and lots of modtools scripts for examples).

bodyswap.lua
```diff
@@ -0,0 +1,255 @@
...
+  if newUnit == oldUnit then
+    return
+  end
+
+  local activeUnits = df.global.world.units.active
+  local oldUnitIndex
+  if activeUnits[0] == oldUnit then
+    oldUnitIndex = 0
+  else
+    for i,u in pairs(activeUnits) do
```
lethosor: I *think* you could probably use `utils.binsearch_in_vector(activeUnits, oldUnitIndex, field)` here (but haven't tried).

lethosor: Yeah, binsearch_in_vector is the C++ name. `utils.binsearch(activeUnits, oldUnit.id, 'id')` (yeah, I screwed up my earlier example) returns `unit, true, index` if it was found. `select(3, utils.binsearch...)` would give you the index if the unit was found, but the index of another unit otherwise, so something like this might be better:
```lua
_, found, oldUnitIndex = utils.binsearch(activeUnits, oldUnit.id, 'id')
if not found then qerror('old unit not found')
-- find newUnitIndex, swap, etc.
```

AtomicChicken: Thanks for the example!
Unless I'm messing something up, it seems that utils.binsearch isn't working reliably; it sometimes returns failure values (found == false, etc) consistently for certain cases and I have no idea why.

I'm running this in-game (in a well-populated town in adventure mode) for testing purposes. It should help illustrate my point: 
```lua
utils = require 'utils'
local activeUnits = df.global.world.units.active
for _,unit in ipairs(activeUnits) do
  print(utils.binsearch(activeUnits,unit.id,'id'))
end
```

lethosor: Hm, is it not sorted by id? I thought that was how df.unit.find() worked... can you try also checking `df.unit.find(unit.id)==unit`? Not sure if I have an adventure save handy.

lethosor: I haven't been able to reproduce this in adventure mode with ~100 units nearby

AtomicChicken: Oh, I missed the previous reply. That's really quite strange. Mind you, failed cases have been a minority thus far; did you look through the entire output?

lethosor: The second column was all `true`s. I'll see if I can find a bigger city or something.

lethosor:
```
$ python $HOME/dfhack/travis/lint.py
./bodyswap.lua: Contains DOS-style newlines: entire file
The command "python $HOME/dfhack/travis/lint.py" exited with 1.
```
\r\n needs to be replaced with \n. Done on my end.

PR #57, opened 2018-05-19
stamper.lua - gui designation manipulator 

grubsteak:
this tool allows you to manipulate designations by transforming them through reflections, rotations, and inverting them. you can also use these designations as brushes to erase other designations or cancel building designations

please leave my author's note in if it's not a big deal

I've tested this over and over and I think I've squashed all the bugs except for the hacky fixes at the bottom

also it probably doesn't follow the dfhack formatting convention so I'll wait for travis to tell me what I did wrong. (I'm new to github)

there is some janky shit at the bottom and if you know a better way to do that please implement it


grubsteak:
EAT IT TRAVIS

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+        dc:key('CUSTOM_B'):string(": Blink Clipboard",self.blink and COLOR_WHITE or COLOR_GREY) dc:newline(1)
+        dc:newline()
+    elseif self.state=='mark' then
+        if self.buffer==nil then
+            dc:string("Select two corners.")
+        end
+        dc:newline():newline(1)
+        dc:key('CUSTOM_P'):string(": Cull Selections",self.cull and COLOR_WHITE or COLOR_GREY)
+    elseif self.state=='convert' then
+        dc:key('CUSTOM_D'):string(": Mine",COLOR_GREY) dc:newline(2)
```
lethosor: `dc:key_string('CUSTOM_D', 'Mine') is equivalent, and you should chain the `:newline()` instead of putting two statements on the same line, which is harder to read.

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+        dc:key('CUSTOM_H'):string(": Channel",COLOR_GREY) dc:newline(2)
+        dc:key('CUSTOM_U'):string(": Up Stair",COLOR_GREY) dc:newline(2)
+        dc:key('CUSTOM_J'):string(": Up Stair",COLOR_GREY) dc:newline(2)
+        dc:key('CUSTOM_I'):string(": U/D Stair",COLOR_GREY) dc:newline(2)
+        dc:key('CUSTOM_R'):string(": Up Ramp",COLOR_GREY) dc:newline(2)
+         dc:newline(1)
```
lethosor: Indentation doesn't match

grubsteak:
how about `gui/paintbrush`?

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+        dc:string("To undesignate use the erase command",COLOR_WHITE)
+    end
+    dc:newline(2)
+    dc:newline(1)
```
lethosor: You don't really need the 2 there - dc:newline():newline(1) would be equivalent.
Maybe newline() needs a second `dy` parameter to skip multiple lines, since that seems to be used fairly often...

lethosor:
Again, not entirely obvious what it does from the name - we have existing brush-like tools for liquids and tiletypes (no GUI for the latter, though). If you're concerned about typing, use a hotkey like gui/liquids (Alt-L).

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+    end
+    if keys.LEAVESCREEN then
+            self:dismiss()
+    elseif self:propagateMoveKeys(keys) then
+        return
+    end
+end
+
+--------------------------WARNING: JANKY SHIT BELOW--------------------------
+for i=1, 4 do
```
lethosor: I think what gui/teleport does is something you could try, and that could be moved inside the CopyUI class too.

https://github.com/DFHack/scripts/blob/0b6e4a56392e125c01e24694ddd880bbc652451d/gui/teleport.lua#L47-L63

Forcing ui.main.mode to `df.ui_sidebar_mode.LookAround` might be more appropriate for your case. One benefit of that is that it will make DF show the sidebar *temporarily* if it was hidden before - your way overrides user preferences.

~~You probably also don't need to copy/restore the sidebar mode like that script - that was an old feature for supporting the `k` (LookAround) menu too, but I don't think it works any more.~~ Looks like you're already doing this.

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+            self:transformBuffer(function(x,y,xlen,ylen,tile) if tile.dig>0 then tile.dig=2 end  return x,y end)
+            self.state='paste'
+        elseif keys.CUSTOM_R then
+            self:transformBuffer(function(x,y,xlen,ylen,tile) if tile.dig>0 then tile.dig=4 end  return x,y end)
+            self.state='paste'
+        end
+        
+    end
+    if keys.LEAVESCREEN then
+            self:dismiss()
```
lethosor: Indentation

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+         dc:newline(1)
+        dc:string("To undesignate use the erase command",COLOR_WHITE)
+    end
+    dc:newline(2)
+    dc:newline(1)
+    dc:key('LEAVESCREEN'):string(": Back")
+    
+end
+local firstInput=true
+local firstPos=copyall(df.global.cursor)
```
lethosor: These seem like things that should be member variables

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+    else
+        return 0
+    end
+end
+
+local function paintMapTile(dc, vp, cursor, pos, ...)
+    if not same_xyz(cursor, pos) then
+        local stile = vp:tileToScreen(pos)
+        if stile.z == 0 then
+            dc:map(true):seek(stile.x,stile.y):char(...):map(false)
```
lethosor: Good job with this! Did you test it with TWBT, or just steal it from scripts like gui/siege-engine? I assume it works with TWBT since this is here, but it would be nice to know.

grubsteak:
I apologize for the lack of clarity. It lets you lift designations and manipulate them by mirroring them, rotating them, and inverting them.  Sort of like a brush tool for an image manipulator.

will any of these be okay?
1. `gui/stamp` (or `gui/stamper`)
2. `gui/brush`

copy.lua
```diff
@@ -0,0 +1,417 @@
...
+
+]====]
+
+--[[
+ Author's Note:
+ 
+ shoutout to twitch.tv/keupo
+ created by flavorstreet
+ 
+ pipicaca.
```
lethosor: This is fine - I'm less enthusiastic about people adding their names to every script they touch, since that's what Git history is for, but background information like this can be useful to have. (Not sure what significance `pipicaca` has here, but not an issue.)

copy.lua
grubsteak: I "adapted" (stole) this from a script called guide-path

copy.lua
grubsteak: this was to fix a bug where the cursor would jump to the side I should probably put it in the :init() function

copy.lua
grubsteak: I did not test it with TWBT.

copy.lua
lethosor: Yeah, that's what I mean.

copy.lua
lethosor: Yeah, that's one of the other scripts that got changed along with gui/siege-engine when we added TWBT support to the drawing API (I thought of gui/siege-engine because it was the one that inspired the change). As long as you use this every time you draw to the map area, it should work fine.

copy.lua
grubsteak: actually it isn't need I just forgot to take it out.

lethosor:
I didn't see either of your naming comments earlier, sorry. I guess stamper is better; still not too clear from the name what it does, but it is more targeted towards this script than others, so all right.

grubsteak:
I commited the changes you requested.

ivantumanov:
@lethosor just pinging for an update on this PR - I've seen some people asking about its status on stream / in comments

lethosor:
Looks good now. I'll test it soon - not sure when, but it's on the list for r2.

PR #60, opened 2018-06-03
add-recipe

brndd:
You know how sometimes you can't craft high boots? This is because civilizations pick the different types of armour they can craft randomly during worldgen, and this is done per slot/type, so sometimes you end up with civilizations that are satisfied with low boots and never go higher. Normally there's no fix to this other than editing entity_default.txt in the raws before worldgen, or just embarking again as a different civ and hoping that they're more cultured.

This script helps with that. It can enable unknown crafting recipes, such as high boots, face veils, or thongs, after embark. It can also enable items that can normally never be crafted by your civilization, like exotic weapons. Currently it's limited to weapons and armour, but it could probably be extended to things like instruments with some extra effort (instruments are procedurally generated, so they work slightly differently).

It should work with most raw mods, though I've only tested it with [Sver's combat rework](http://www.bay12forums.com/smf/index.php?topic=170474.0).


lethosor:
I don't think "recipe" is a great term - I've never seen it in the raws, and the only place it's used in df-structures is for food/meals. Maybe "resources" or "itemdefs" would work, although the latter might be more confusing.

brndd:
Recipe came to me mostly from other games, where the term "crafting recipe" is a common term. DF handles the concept quite abstractly with there not really being a term for this kind of thing. I considered "knowledge" since what the script technically does is adds the knowledge of (or familiarity with) these items to the civ, but "add-knowledge" sounds much too ambiguous for what it does in practice.

From the perspective of most players familiar with RPGs or Minecraft or something, I think "recipe" will make sense, because the primary reason they'd probably want to use the script is to make the items craftable (though that's not the only thing it does; since it makes your civ know about the items, they'll also start appearing with the caravan and such). It's true though that it is a word not used in DF itself.

brndd:
gonna have to close and reopen this because I fail at Git, sorry.

PR #61, opened 2018-06-05
Deathcause

PatrikLundell:
One case of df-structures issue 247 (although no list has been seen yet...).
I've tested it on a random DFFD fortress and have seen it behave the same as the original version on most of the things in the refuse/corpse stockpile as well as on the dead units list.

The "not at hist fig" message wasn't produced in any of the cases, however, nor did I find any body parts belonging to living creatures to actually test the flag change.

However, during the testing I found using deathcause on a living unit resulted in it being dead due to an unknown cause, which led to the changes to the display_death_unit operation, and that change worked correctly, i.e. claiming the unit wasn't dead, compared to the original, which (still, of course), claimed an unknown death cause. Given that the flag change was applied in that branch, it's likely it works for body parts of live units as well. The live unit body part text could benefit from the addition of race, given that it's present everywhere else, but I don't want to mess with that part of the script since I don't have a case to test it on.

### DIFF
--- a/deathcause.rb
+++ b/deathcause.rb
@@ -19,9 +19,10 @@ def display_death_event(e)
 end
 
 def display_death_unit(u)
+    str = "The #{u.race_tg.name[0]}"
+    str << " #{u.name}" if u.name.has_name
+    
     if not u.flags2.killed and not u.flags3.ghostly
-        str = "The #{u.race_tg.name[0]}"
-        str << " #{u.name}" if u.name.has_name
         str << " is not dead yet !"
         
         puts str.chomp(',')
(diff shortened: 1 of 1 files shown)


PatrikLundell:
Realized slaughter of non hist fig units could be identified as a death cause. While testing that I encountered a refusal to report the death cause of the unit selected in the unit lists, presumably because there was also an item found somehow. Worked around that by checking if a unit has been found before going down the no item found branch.

PatrikLundell:
I think it would still be better than if I did it as my Windoze search function is unreliable. A week ago or so I searched raws for something, and a short while later I searched for something else that should be there, found nothing, and went back to the previous search which then found nothing. It eventually went back to finding things again, but I definitely can't trust it.

When it comes to the deathcause script it should be noted that all I know about ruby is what's in that script file, so it could definitely benefit from a sanity check.

PR #62, opened 2018-06-06
Dead misuse

PatrikLundell:
Adjusted usage of flags1.dead to use flags2.killed in scripts that appear to use the flag to target dead unit rather than inactive ones.
The DT script has updated the comments produced to better reflect the current understanding, but not any address info.

The scripts have gone through rudimentary testing to see that they still work (to e.g. catch typing errors), but not anything beyond that.

It can be noted that gui/unit-info-viewer.lua produced a nonsensical output of strength:agility being in cubic decimeters, resulting in a 1.5:1.1 cubic meter dwarf, but I don't intend to do anything about it beyond this comment.

Relates to df-structures issue 247.
Link:  DFHack/df-structures#247

### DIFF
--- a/emigration.lua
+++ b/emigration.lua
@@ -66,7 +66,7 @@ function canLeave(unit)
     
     return dfhack.units.isOwnRace(unit) and  --  Doubtful check. naturalized citizens
            dfhack.units.isOwnCiv(unit) and   --  might also want to leave.
-           not dfhack.units.isActive(unit) and
+           dfhack.units.isActive(unit) and
            not dfhack.units.isOpposedToLife(unit) and
            not unit.flags1.merchant and
            not unit.flags1.diplomat and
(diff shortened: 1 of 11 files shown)


emigration.lua
```diff
@@ -56,25 +56,27 @@ function desert(u,method,civ)
...
-        or dfhack.units.isCitizen(unit)
-        or dfhack.units.isSane(unit)
-        or unit.profession ~= 103
-        or not dfhack.units.isDead(unit)
-            then return false end
-    return true
+    
+    return dfhack.units.isOwnRace(unit) and  --  Doubtful check. naturalized citizens
+           dfhack.units.isOwnCiv(unit) and   --  might also want to leave.
+           not dfhack.units.isActive(unit) and
```
lethosor: Isn't this still reversed? Active units can leave, but inactive units can't.

PatrikLundell: Yes it is...

PR #64, opened 2018-06-07
item-descriptions editor's pass

brndd:
Fixed some grammar errors and improved a few of the descriptions.


item-descriptions.lua
```diff
@@ -36,24 +36,23 @@ vanilla item in the game.
...
---luacheck: global
 descriptions = {
-    AMULET = {  "An item of jewellery worn around the neck for it's aesthetic value.",
+    AMULET = {  "An item of jewellery worn around the neck for its aesthetic value.",
                 "An amulet does not interfere with wearing other equipment."},
     ANIMALTRAP = {
                 "This tiny trap is used by your trappers to catch vermin. Some dwarves",
                 "like vermin as pets - if your cats don't get them first. May be built",
-                "from wood or metal. Catching vermin requires trap to be set with bait."},
+                "of wood or metal. Catching vermin requires trap to be set with bait."},
```
lethosor: "built of" just doesn't sound right to me. It's not the building material in its original form any more. Maybe "made of" would be better.

item-descriptions.lua
```diff
@@ -36,24 +36,23 @@ vanilla item in the game.
 
 If "raw/scripts/item-descriptions.lua" exists, it will entirely replace this one.
 Instead, mods should use "raw/scripts/more-item-descriptions.lua" to add content or replace
-descriptions on a case-by-case basis.  If an item description cannot be found in
+descriptions on a case-by-case basis. If an item description cannot be found in
 the latter script, view-item-info will fall back to the former.
 ]]
 
 -- see http://dwarffortresswiki.org/index.php/cv:Item_token
---luacheck: global
```
lethosor: Keep this in. It's necessary for df-luacheck stuff that @BenLubar is working on, apparently.

item-descriptions.lua
```diff
@@ -463,124 +473,120 @@ descriptions = {
...
                 "reduces their effectiveness. Mauls use and train the hammerdwarf",
-                "skill. Being foreign weapons, dwarves cannot forge mauls."},
+                "skill. Being foreign weapons, mauls cannot be forged by dwarves."},
     ITEM_WEAPON_MORNINGSTAR = {
                 "A morningstar is an edged weapon that consists of a spiked ball mounted",
-                "on the end of a handle. Despite similarities to a mace in appearance,",
-                "a morningstar's size and contact area are closer to those of a war",
-                "hammer. Specifically, a morningstar is 25% larger than a war hammer",
+                "on the end of a shaft. Despite similarities to the mace in appearance,",
+                "a morningstar's size and contact area are closer to those of the war",
```
lethosor: "a" and "the" aren't used consistently

item-descriptions.lua
```diff
@@ -590,53 +596,54 @@ descriptions = {
...
-                "and must be powered for operation.  Millstones are made of stone."},
+                "and must be powered for operation."},
     ORTHOPEDIC_CAST = {
                 "Casts are made from plaster, and are used to keep broken bones in",
                 "place until they are healed. Applying a cast requires a bucket,",
                 "cloth and a water source."},
     PIPE_SECTION = {
                 "An enormous piece of pipe, it is a part of a screw pump."},
     QUERN = {   "A hand-operated mill for plants, grains, and seeds. It mills plants",
-                "much slower than a millstone. Must be built from stone."},
```
lethosor: Is this not true? http://dwarffortresswiki.org/index.php/DF2014:Quern says they must be built at a mason's workshop, and there's not a page for the quern item itself.

Of course, you could say that this isn't necessary in the description that's already built.

item-descriptions.lua
```diff
@@ -590,53 +596,54 @@ descriptions = {
...
+                "The tanned hides of animals are flexible enough to be made into an",
                 "assortment of goods for military and civilian use. Leather can also",
                 "be used to decorate items with sewn images at a Leather Works. Armor",
-                "and shields made from leather are not terribly effective, but are",
+                "and shields made of leather are not terribly effective, but are",
                 "still better than nothing at all."},
-    SLAB = {    "A memorial stone, used to quiet restless ghost when engraved with",
-                "the name of the deceased and built."},
+    SLAB = {    "A memorial stone, used to calm a restless ghost when engraved with",
+                "the name of the deceased and constructed."},
```
lethosor: "built" seems appropriate to me since it's the "b: Building" menu. Constructed could imply crafted at a workshop too.

item-descriptions.lua
brndd: Good catch. Apparently I based the edit off of some old version of the file, and this got lost in transition.

item-descriptions.lua
```diff
@@ -30,7 +30,7 @@ if not moduleMode then
     print("scripts/item-descriptions.lua is a content library; calling it does nothing.")
 end
 
-local help --[[
```
lethosor: Heh, it's supposed to be `local help = [====[`, then `]====]` at the end. But since this script doesn't even attempt to print help text, I would just move any useful content from here to the documentation above and get rid of the rest.

item-descriptions.lua
lethosor: Yeah, I figured that was the most likely cause.

item-descriptions.lua
brndd: I felt it was a bit superfluous because the player wouldn't be reading this description unless they'd already built a quern.

item-descriptions.lua
brndd: Further cleaned up this and a lot of other things in many of the other weapon descriptions. I'm not sure what the grammatical term for this kind of article use is, but you can also see it in eg. [the Wikipedia article for the war hammer](https://en.wikipedia.org/wiki/War_hammer): "**A war hammer** is a late medieval weapon [...] whose design resembles **the hammer**."

item-descriptions.lua
brndd: I wasn't sure what the deal with this was, because it was one of the things that had changed between the latest master branch and the revision of the file I was editing (because being a dummy I didn't pull before working on the file). I ended up just going with what was currently in.

item-descriptions.lua
```diff
@@ -36,14 +36,15 @@ vanilla item in the game.
...
 If "raw/scripts/item-descriptions.lua" exists, it will entirely replace this one.
 Instead, mods should use "raw/scripts/more-item-descriptions.lua" to add content or replace
-descriptions on a case-by-case basis.  If an item description cannot be found in
+descriptions on a case-by-case basis. If an item description cannot be found in
 the latter script, view-item-info will fall back to the former.
 ]]
 
 -- see http://dwarffortresswiki.org/index.php/cv:Item_token
---luacheck: global
+---luacheck: global
```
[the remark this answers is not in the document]
brndd: dang it, the - denoting a removed line in the diff confused me. fixed.

(fwiw you should have edit rights to the branch so feel free to change stuff like this yourself if you want to)

item-descriptions.lua
```diff
@@ -41,7 +41,7 @@ the latter script, view-item-info will fall back to the former.
 ]]
 
 -- see http://dwarffortresswiki.org/index.php/cv:Item_token
----luacheck: global
+--luacheck: global
```
BenLubar: This should be on the line immediately before the variable. It won't silence the warning if there's a blank line in between.

brndd: Fixed... I think.

PR #68, opened 2018-07-04
Improve digfort empty cell handling

suokko:
If csv has extra empty columns digfort may try to look far outside
designated area. To avoid going outside map limits empty cells needs to
be checked before accessing.

To help user understand the issue I decided to add error message when
designation goes outside map limits.

Fixes 1179

### DIFF
--- a/digfort.rb
+++ b/digfort.rb
@@ -55,6 +55,8 @@ end
 
 offset = [0, 0]
 tiles = []
+max_x = 0
+max_y = 0
 planfile.each_line { |l|
     if l =~ /#.*start\s*\(\s*(-?\d+)\s*[,;]\s*(-?\d+)/
         raise "Error: multiple start() comments" if offset != [0, 0]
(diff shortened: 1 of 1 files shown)


suokko:
dfhack bug https://github.com/DFHack/dfhack/issues/1179

digfort.rb
```diff
@@ -70,20 +72,36 @@
...
 y = df.cursor.y - offset[1]
 z = df.cursor.z
 starty = y - 1
 
+map = df.world.map
+
+if x < 0 or y < 0 or x+max_x >= map.x_count or y+max_y >= map.y_count
+    max_x = max_x + x + 1
+    max_y = max_y + y + 1
+    raise "Position would designated outside map limits. Selected limits are from (#{x+1}, #{y+1}) to (#{max_x},#{max_y})"
```
lethosor: Changed to "designate"

PR #77, opened 2018-08-17
Added armoks-blessing documentation & missing class type check

mattregul:


armoks-blessing.lua
```diff
@@ -162,7 +186,7 @@ function adjust_all_dwarves(skillname)
             elevate_attributes(v)
             rejuvenate(v)
             if skillname then
-                if skillname=="Normal" or skillname=="Medical" or skillname=="Personal" or skillname=="Social" or skillname=="Cultural" or skillname=="MilitaryWeapon" or skillname=="MilitaryAttack" or skillname=="MilitaryDefense" or skillname=="MilitaryMisc" then
+                if skillname=="Normal" or skillname=="Medical" or skillname=="Personal" or skillname=="Social" or skillname=="Cultural" or skillname=="MilitaryWeapon" or skillname=="MilitaryUnarmed" or skillname=="MilitaryAttack" or skillname=="MilitaryDefense" or skillname=="MilitaryMisc" then
```
lethosor: This should be replaced with
```lua
if df.job_skill_class[skillname] then
```
(you can add `~= nil` for readability if you prefer, although even in the case for `Normal`, where df.job_skill_class.Normal == 0, 0 is treated as true in conditionals in Lua).

PR #79, opened 2018-08-27
Added extinguish.lua

AtomicChicken:
This script kills fires affecting units, items, buildings and plants, replacing burning tiletypes with the appropriate extinguished types where necessary and resetting temperatures to a safe level. A target location can be selected using the cursor or specified via its x,y,z coordinates. Alternatively, the -all argument can be used to put out fires across the entire map.

closes DFHack/dfhack#1358

### DIFF
--- a/extinguish.lua
+++ b/extinguish.lua
@@ -1,15 +1,14 @@
 -- Puts out fires.
 -- author: Atomic Chicken
+--@ module = true
 
 local utils = require 'utils'
 local validArgs = utils.invert({
-'all',
-'location',
-'help'
+  'all',
+  'location',
+  'help'
 })
 local args = utils.processArgs({...}, validArgs)
-
---@ module = true
 
 local usage = [====[
 
@@ -31,8 +30,8 @@ Arguments::
 ]====]
 
 if args.help then
- print(usage)
- return
+  print(usage)
+  return
 end
 
 if args.all and args.location then
@@ -77,7 +76,7 @@ function extinguishContaminant(spatter)
 end
 
 function extinguishItem(item)
-  if item.flags.on_fire == true then
+  if item.flags.on_fire then
     item.flags.on_fire = false
     item.temperature.whole = 10050
     item.temperature.fraction = 0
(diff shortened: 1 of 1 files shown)


extinguish.lua
```diff
@@ -0,0 +1,187 @@
...
+Arguments::
+
+    -all
+        extinguish everything on the map
+
+    -location [ x y z ]
+        extinguish only at the specified coordinates
+]====]
+
+if args.help then
```
lethosor: The 1-space indentation here should be changed to 2 spaces to match the rest of the file

extinguish.lua
```diff
@@ -0,0 +1,187 @@
...
+
+local utils = require 'utils'
+local validArgs = utils.invert({
+'all',
+'location',
+'help'
+})
+local args = utils.processArgs({...}, validArgs)
+
+--@ module = true
```
lethosor: Personally, I'd move this after the comments at the top of the file, but this is close enough.

extinguish.lua
```diff
@@ -0,0 +1,187 @@
...
+
+function extinguishContaminant(spatter)
+-- reset temperature of any contaminants to prevent them from causing reignition
+-- (just in case anyone decides to play around with molten gold or whatnot)
+  spatter.temperature.whole = 10050
+  spatter.temperature.fraction = 0
+end
+
+function extinguishItem(item)
+  if item.flags.on_fire == true then
```
lethosor: Don't really need `==true`

extinguish.lua
```diff
@@ -0,0 +1,187 @@
...
+    end
+  end
+  local campfires = df.global.world.campfires
+  for i = #campfires-1,0,-1 do
+    if same_xyz(pos, campfires[i].pos) then
+      extinguishTile(x,y,z)
+      campfires:erase(i)
+    end
+  end
+  local units = dfhack.units.getUnitsInBox(x,y,z,x,y,z)
```
lethosor: Too bad there isn't a better API for this

PR #81, opened 2018-10-11
Adds lua files for optimizing dwarves in fort-mode

cppcooper:
The core script could probably be renamed to be more appropriate to its function. Suggestions welcome.

The script was created to optimize dwarves for fort-mode work. The core function takes dwarves and allocates a "dorf_job" to each dwarf. This decision takes into account current counts for each dorf_job, in relation to how many *should* be allocated based on the working population size.
**dorf_job:**
- a custom profession name
- a specific set of required labors
- a specific set of probabilistic labors
  - these labors are applied based on individual probabilities which also act as enforced ratios
- a specific set of dorf_types

**dorf_types:**
- a probability of being applied to any dwarf during the `pimpem` process
- some types include specific labors with a specific range of values for randomly assigning skill level for said labor
- a specific set of attributes with a specific attribute level

**attribute levels:**
```lua
--probability is used for generating all dwarf stats, some jobs include dorf_types which will upgrade particular stats
attrib_levels = { -- prob,      avg,    std deviation
    incompetent =   {p=0.01,    100,    20},
    verybad =       {p=0.02,    250,    25},
    bad =           {p=0.04,    450,    30},
    average =       {p=0.21,    810,    60},
    good =          {p=0.28,    1350,   75},
    verygood =      {p=0.22,    1700,   42},
    superb =        {p=0.12,    1900,   88},
    amazing =       {p=0.06,    2900,   188},
    incredible =    {p=0.03,    3800,   242},
    unbelievable =  {p=0.01,    4829,   42}
}
```

The script relies on persistent data saved in with the user's world data to know how many dwarves are allocated to a job, and what jobs/professions are allocated to a single dwarf. This is important for maintaining ratios after resetting a subset of the fort's population.

The main script `pimp-it.lua` relies on a secondary file `dorf_tables.lua` in order to know what it is doing with the dwarves. The user can use the script `pimp-it.lua` to manually apply additional skills to one or many dwarves; in order to know exactly what a dorf_job or profession or dorf_type does the user would ultimately need to look at `dorf_tables.lua`.

The below information is help information shown via command-line, it mainly deals with how to use the script. It does not explain what types, professions, or jobs do. Suggestions for improving usability are welcome here.

```lua
[====[
usage: pimp-it [-help|-select]
               -select <sel-opt> -<command> <args>
============
pimp-it script
~~~~~~~~~~~~
To use this script, you need to select a subset of your dwarves. Then run commands on those dwarves.
Examples:
  [DFHack]# pimp-it -select [ jobs Trader Miner Leader Warden ] -applytype adaptable
  [DFHack]# pimp-it -select all -clear -pimpem
  [DFHack]# pimp-it -select pimped -reroll
  [DFHack]# pimp-it -select named -reroll inclusive -applyprofession RECRUIT
~~~~~~~~~~~~
 select options:
   (protected is the only option which will select PROTECTED dwarves)
    all         - selects all dwarves.
    highlighted - selects only the in-game highlighted dwarf (from any screen).
    named       - selects dwarves with user-given names.
    unnamed     - selects dwarves without user-given names.
    employed    - selects dwarves with custom professions. Excludes pimped dwarves.
    pimped      - selects dwarves based on session data. Dwarves who have been pimped, should be listed in this data.
    unpimped    - selects any dwarves that don't appear in session data.
    protected   - selects any dwarves which use protection signals in their name or profession. (ie. {'.', 'c', 'j', 'p'})
    unprotected - selects any dwarves which don't use protection signals in their name or profession.
    drunks      - selects any dwarves which are currently zeroed, or were originally drunks as their profession.
    jobs        - selects any dwarves with the listed job types. This will only match with custom professions, or pimped dwarves (for pimped dorfs see: dorf_jobs in dorf_tables.lua).
                - usage `-select [ jobs job1 job2 etc. ]` eg. `-select [ jobs Miner Trader ]`
~~~~~~~~~~~~
Commands will run on the selected dwarves
 available commands:
    clear              - zeroes selected dwarves. No attributes, no labours. Assigns 'DRUNK' profession.
    reroll <inclusive> - zeroes selected dwarves, then rerolls that dwarf based on its job. Ignores dwarves with unlisted jobs.
                       - optional argument: inclusive. Only performs the reroll, will no zero the dwarf first. Benefit: stats can only go higher, not lower.
    pimpem             - performs a job search for unpimped dwarves. Each dwarf will be found a job according to the job_distribution table in dorf_tables.lua
    applyjobs          - applies the listed jobs to the selected dwarves. list format: `[ job1 job2 jobn ]` brackets and jobs all separated by spaces.
                       - see dorf_jobs table in dorf_tables.lua for available jobs.")
    applyprofessions   - applies the listed professions to the selected dwarves. list format: `[ prof1 prof2 profn ]` brackets and professions all separated by spaces.
                       - see professions table in dorf_tables.lua for available professions.
    applytypes         - applies the listed types to the selected dwarves. list format: `[ type1 type2 typen ]` brackets and types all separated by spaces.
                       - see dorf_types table in dorf_tables.lua for available types.
~~~~~~~~~~~~
    Other Arguments:
      help - displays this help information.
      debug - enables debugging print lines

No dorfs were harmed in the building of this help screen.
]====]
```

### DIFF
--- a/dorf_tables.lua
+++ b/dorf_tables.lua
@@ -25,11 +25,11 @@ job_distributions = {
 }
 
 attrib_levels = {
-	incompetent =	{p=0.02,	100,	20},
-	verybad =		{p=0.04,	250,	25},
-	bad =			{p=0.08,	450,	30},
-	average =		{p=0.18,	810,	60},
-	good =			{p=0.24,	1350,   75},
+	incompetent =	{p=0.01,	100,	20},
+	verybad =		{p=0.02,	250,	25},
+	bad =			{p=0.04,	450,	30},
+	average =		{p=0.21,	810,	60},
+	good =			{p=0.28,	1350,   75},
 	verygood =		{p=0.22,	1700,   42},
 	superb =		{p=0.12,	1900,   88},
 	amazing =		{p=0.06,	2900,   188},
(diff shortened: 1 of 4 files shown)


dorf_tables.lua
```diff
@@ -0,0 +1,381 @@
+-- dorf_tables has job distribution configurations, random number information for attributes generation, job configurations, profession configurations, dorf type configurations
+-- usage: loaded by pimp-it.lua
+-- by josh cooper(cppcooper) [created: 12-2017 | last edited: 10-2018]
+local help = [====[
```
lethosor: Add: `--@ module = true` somewhere near the top

dorf_tables.lua
```diff
@@ -0,0 +1,381 @@
...
+
+dorf_tables
+===========
+Data tables for pimp-it.lua.
+
+Usage: load inside secondary script (pimp-it.lua)
+
+]====]
+--local _ENV = mkmodule('dorf_tables')
```
lethosor:
```suggestion

```

dorf_tables.lua
```diff
@@ -0,0 +1,381 @@
...
+        p = 0.4242,
+        attribs = {STRENGTH={'good'},AGILITY={'average'},ENDURANCE={'good'},RECUPERATION={'average'},FOCUS={'verygood'}},
+        skills  = {DODGING={5,10},GRASP_STRIKE={5,10},STANCE_STRIKE={5,10},WRESTLING={5,10},SITUATIONAL_AWARENESS={5,10}}},
+    soldier = {
+        p = 0.3333,
+        attribs = {STRENGTH={'verygood'},AGILITY={'verygood'},ENDURANCE={'verygood'},RECUPERATION={'verygood'},FOCUS={'superb'}},
+        skills  = {DISCIPLINE={7,14},SITUATIONAL_AWARENESS={5,10},MELEE_COMBAT={4,6},RANGED_COMBAT={4,6},ARMOR={4,6},HAMMER={4,6},CROSSBOW={4,6},COORDINATION={4,6},BALANCE={4,6},MILITARY_TACTICS={5,8}}}
+}
+
+--return _ENV
```
lethosor:
```suggestion

```

(not needed for reqscript/script_environment)

pimp-it.lua
```diff
@@ -0,0 +1,1060 @@
...
+]====]
+
+print("v1.1")
+utils ={}
+utils = require('utils')
+json = require('json')
+local rng = require('plugins.cxxrandom')
+local engineID = rng.MakeNewEngine()
+print("Loading data tables..")
+local dorf_tables = dfhack.script_environment('dorf_tables')
```
lethosor:
```suggestion
local dorf_tables = reqscript('dorf_tables')
```
This will work with the `--@module = true` above

pimp-it.lua
```diff
@@ -0,0 +1,1060 @@
...
+            bAlreadySetProf2 = true
+            dwf.profession2 = df.profession[prof]
+        end
+        --These are required professions for this job class
+        ApplyProfession(dwf, prof, 11, 17)
+    end
+        
+    -- Loop tertiary professions
+    -- Sort loop (asc)
+    --[[]]
```
lethosor: This kind of stuff is hard to parse, even with syntax highlighting. Can you just remove this and the matching ``--]]`` below?

lethosor:
Sorry for the long delay. This is rather large, so I'm not sure how much of a code review I can do. I can go more into detail if you want. As long as you're able to maintain this, that's good with me.

Regarding naming, I would stick with `dwarf`, not `dorf`, everywhere. Also, the helper file should probably not go in the root scripts folder, because then it'll show up in `ls` and people might think it's a script. We don't have a great system for helper files yet, but try the check that https://github.com/DFHack/scripts/blob/master/item-descriptions.lua does near the top - it's at least more user-friendly.

pimp-it.lua
cppcooper: Are you asking to remove the comments or the `--[[]]` part?

lethosor: Just the --[[]] (line 459)

cppcooper: Okay, looking at it I get what you mean. A disabled commented out code block if you will. It's removed now. For the next push.

dorf_tables.lua
cppcooper: So presumably this is responsible for the variable in the example you linked to:
```lua
if not moduleMode then
    print("scripts/item-descriptions.lua is a content library; calling it does nothing.")
end
```

lethosor: It allows loading the script as a module, which will set moduleMode when it happens. It will also set dfhack_flags.module, which is equivalent.

pimp-it.lua
cppcooper: done, this is in the next commit

dorf_tables.lua
cppcooper: Added it for the next commit

cppcooper:
Yea no worries, I know it's huge. That's why I waited until after my exams were done to worry about it.

When I was testing the script, following the implementation of your suggested changes, I noticed there was a bug. I made a comment on the line that fixes it. If you read that line it probably won't encourage faith in my ability to maintain this code.

Regardless of that though, people reporting such bugs will only improve the maintainability imho. Which is why I wanted this pull request.

I'll commit, probably, in the next day or two.

I've implemented most of the changes, but I need to properly consider the dorf->dwarf change. As I recall the choice of dorf had a purpose in distinguishing between a constructs or concepts of my own making and the games. Such as dorf_jobs, though I should have probably moved that scheme to the professions table.

cppcooper:
Really the only thing left, I think, is the file naming, and maybe a comment or two that I've overlooked.

I honestly don't like either file name but I can't seem to think of anything better. Something with the same sort of implication as "pimping" [think pimp my ride, not pimp this person]. Something less ambiguous.. refine maybe?? I'd go with optimize if it was 100% optimizing work loads.

`fort-dwarfer` perhaps? It probably doesn't matter, I'll defer to the opinions of others.

lethosor:
`max-dwarf`? `optimize-dwarf` (maybe long)? `opt-dwarf` (maybe ambiguous)? `superdwarf` is already taken, but this is probably less-intended for use on single dwarves, so it probably doesn't matter.

cppcooper:
Perhaps `dwarf-op` [ie. dwarf-optimizer(or some variant), or dwarf-operations] I think it fits pretty well.

cppcooper:
Tested the refactor and everything still functions. There is a [pull request here](https://github.com/DFHack/dfhack/pull/1435) which prevents a bit of spam initiated by this script.

The forced pushes were catching a few stragglers, and also fixing the title underline thing which gets treated as an error by Travis =/

PR #82, opened 2018-10-17
Adds tame.lua a script for taming/training animals

cppcooper:
This script takes the highlighted unit and attempts to set the level of training, and toggles the `tame` flag as necessary.

```
tame
====
Tame and train animals.

Usage: `tame -read` OR `tame -set <level>`

:0: semi-wild
:1: trained
:2: well-trained
:3: skillfully trained
:4: expertly trained
:5: exceptionally trained
:6: masterfully trained
:7: tame
:8: wild
:9: wild
```

### DIFF
--- a/tame.lua
+++ b/tame.lua
@@ -1,0 +1,38 @@
+-- tame and train animals {set animals as Wild through to Masterfully Trained}
+
+local help=[====[
+tame
+====
+Tame and train animals.
+
+Usage: tame -set <level>
+
+:0: wild
+:1: trained
+:2: well-trained
+:3: skillfully trained
+:4: expertly trained
+:5: exceptionally trained
+:6: masterfully trained
+:7: tame
+:8: semi-wild
+]====]
+
+local utils = require('utils')
+local selected = dfhack.gui.getSelectedUnit()
+local validArgs = utils.invert({'set'})
+local args = utils.processArgs({...}, validArgs)
+--[
+if args.set and tonumber(args.set) then
+    local level = tonumber(args.set)
+    if level < 0 or level > 8 then
+        print(help)
+        error("range must be 0 to 9")
+    end
+    selected.flags1.tame = level ~= 0
+    selected.training_level = level
+else
+    print(help)
+end
+--]]
+
--- a/tame.lua
+++ b/tame.lua
@@ -5,9 +5,9 @@ tame
 ====
 Tame and train animals.
 
-Usage: tame -set <level>
+Usage: `tame -read` OR `tame -set <level>`
 
-:0: wild
+:0: semi-wild
 :1: trained
 :2: well-trained
 :3: skillfully trained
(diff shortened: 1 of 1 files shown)


tame.lua
```diff
@@ -0,0 +1,38 @@
...
+local utils = require('utils')
+local selected = dfhack.gui.getSelectedUnit()
+local validArgs = utils.invert({'set'})
+local args = utils.processArgs({...}, validArgs)
+--[
+if args.set and tonumber(args.set) then
+    local level = tonumber(args.set)
+    if level < 0 or level > 8 then
+        print(help)
+        error("range must be 0 to 9")
```
lethosor: `qerror` is generally more user-friendly for user-run scripts. (The message here should also have `8` instead of `9`.)

tame.lua
```diff
@@ -0,0 +1,38 @@
...
+:6: masterfully trained
+:7: tame
+:8: semi-wild
+]====]
+
+local utils = require('utils')
+local selected = dfhack.gui.getSelectedUnit()
+local validArgs = utils.invert({'set'})
+local args = utils.processArgs({...}, validArgs)
+--[
```
lethosor: Might want to remove these?

tame.lua
```diff
@@ -0,0 +1,38 @@
...
+:0: wild
+:1: trained
+:2: well-trained
+:3: skillfully trained
+:4: expertly trained
+:5: exceptionally trained
+:6: masterfully trained
+:7: tame
+:8: semi-wild
+]====]
```
lethosor: Looking at the animal_training_level enum (I got this from df/animal_training_level.h, but you could also use `lua @df.animal_training_level` for similar output):
```
        SemiWild, // 0, 0x0
        Trained, // 1, 0x1
        WellTrained, // 2, 0x2
        SkilfullyTrained, // 3, 0x3
        ExpertlyTrained, // 4, 0x4
        ExceptionallyTrained, // 5, 0x5
        MasterfullyTrained, // 6, 0x6
        Domesticated, // 7, 0x7
        Unk8, // 8, 0x8
        /**
         * Seems to be used as default when not flags1.tame
         */
        WildUntamed, // 9, 0x9

```
so I think your names for at least `0` and `8` might be wrong here, unless the enum values have changed in DF. Can you double-check those?

tame.lua
cppcooper: I'll replace it.

tame.lua
cppcooper: Yea I'll change things up, I was only trying to match the order of this table as closely as possible. Although now that I think about it, it certainly would make more sense to have semi-wild be at 0. Though I'm not sure what to do with `Unk8` should I just use it as Wild, or both 8 and 9 as Wild =/
image

tame.lua
cppcooper: Presumably you mean the defunct comments from lines 25 and 37

AtomicChicken:
The modtools/create-unit script used to have problems with "tame" animals exhibiting hostility towards fortress citizens, which required alterations such as setting the unit civ id, and adding entity link membership to the histfig info. This is especially important with regards to naturally hostile creatures (megabeasts, demons, etc). It'd probably be best to take a look at modtools/create-unit and modify this script accordingly to avoid running into issues down the line.

tame.lua
lethosor: Oh, yeah, the wiki table is probably in a more human-friendly order.

cppcooper:
So I don't understand why the travis build is failing. This message it generates doesn't make a lot of sense to me. I don't know what formatting changes it expects.
```
../../../dfhack/scripts/tame.lua:6: WARNING: undefined label: tame -read (if the link has no caption the label must precede a section header)
```

tame.lua
```diff
@@ -0,0 +1,49 @@
+-- tame and train animals {set animals as Wild through to Masterfully Trained}
+
+local help=[====[
+tame
+====
+Tame and train animals.
+
+Usage: `tame -read` OR `tame -set <level>`
```
lethosor: All of these backticks should be double backticks

tame.lua
```diff
@@ -0,0 +1,49 @@
...
+            print(help)
+        end
+        selected.flags1.tame = level ~= 8 and level ~= 9
+        if selected.flags1.tame then
+            selected.training_level = level
+        else
+            selected.training_level = 9
+        end
+    elseif args.read then
+        print("tame:",selected.flags1.tame)
```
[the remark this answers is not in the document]
cppcooper: `print(x,y)` indents between x and y

tame.lua
```diff
@@ -0,0 +1,49 @@
...
+
+:0: semi-wild
+:1: trained
+:2: well-trained
+:3: skillfully trained
+:4: expertly trained
+:5: exceptionally trained
+:6: masterfully trained
+:7: tame
+:8: wild
```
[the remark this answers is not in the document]
cppcooper: When if it becomes used the script will need to be updated. Right now I have it consistent to its own documentation, but as you surely saw I don't use 8 either. I use it only to set the tame flag.

tame.lua
cppcooper: Done. It'll be in the next commit.

tame.lua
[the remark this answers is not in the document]
lethosor: You're right, I thought it was concatenation.

PR #83, opened 2018-10-19
Adds a script for querying data structures recursively

cppcooper:
Query is a script useful for finding and reading values of data structure fields. Purposes will likely be exclusive to writing lua script code. 

This script will recursively search an input for fields matching your query. Matching fields will be printed along with their values. When a match is a table the entire table will not be printed, only the reference to it.

```lua
function Query(table, query, parent)
    if not parent then
        parent = ""
    end
    for k,v in safe_pairs(table) do
        if not tonumber(k) and type(k) ~= "table" and not string.find(tostring(k), 'script') then
            if string.find(tostring(k), query) then
                print(parent .. "." .. k .. ":", v)
            end
            --print(parent .. "." .. k)
            if not string.find(parent, tostring(k)) then
                if parent then
                    Query(v, query, parent .. "." .. k)
                else
                    Query(v, query, k)
                end
            end
        end
    end
end
```
   

When performing table queries, use dot notation to denote sub-tables. The script has to parse the input string and separate each table.

example: "query -unit -query profession" will print all fields matching '*profession*' inside the highlighted unit. The values input to -unit or -table must be global identifiers
(ie. _G[identifier] must exist).

Usage: ``query -unit <global_value> -query <query>`` or ``query -table <global_value> -query <query>``


query.lua
```diff
@@ -0,0 +1,50 @@
...
+        return function() end
+    end
+end
+
+function Query(table, query, parent)
+    if not parent then
+        parent = ""
+    end
+    for k,v in safe_pairs(table) do
+        if not tonumber(k) and type(k) ~= "table" and not string.find(tostring(k), 'script') then
```
cppcooper: This line should probably have a comment on it explaining why these checks are made. The reason is essentially to prevent infinite recursion.

query.lua
```diff
@@ -1,128 +1,128 @@
...
```
cppcooper: This line should probably have a comment on it explaining why these checks are made. The reason is essentially to prevent infinite recursion.

lethosor:
Neat idea. There is a `printall_recurse` function in dfhack.lua as of around 0.44.12 (can't remember if it's in r1 or just the develop branch). I don't know if it's easy to use for your purposes, though. At any rate, this is something people would definitely find useful for research/development, so :+1:.

lethosor:
FYI: I'm going to move this to the devel folder (i.e. `devel/query`) to be consistent with other dev-focused scripts.

cppcooper:
Ya that makes sense to put in devel. Line 34 being in the merge makes me happy on a fundamental level =)
cheers

PR #84, opened 2019-01-08
bodyswap: made script useable in arena mode

AtomicChicken:
As per the PR title, this minor change makes it possible to bodyswap in arena mode when controlling a unit. Testing has shown that this script can be used copiously in arena mode with no apparent ill effects. Also included is a trivial change in documentation.

### DIFF
--- a/bodyswap.lua
+++ b/bodyswap.lua
@@ -39,7 +39,7 @@ if args.help then
   return
 end
 
-if not df.global.gamemode == df.game_mode.ADVENTURE then
+if df.global.gamemode ~= df.game_mode.ADVENTURE then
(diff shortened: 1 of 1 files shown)


bodyswap.lua
```diff
@@ -35,11 +35,11 @@ Arguments::
...
 
 if args.help then
- print(usage)
- return
+  print(usage)
+  return
 end
 
-if not dfhack.world.isAdventureMode() then
+if not df.global.gamemode == df.game_mode['ADVENTURE'] then
```
warmist:
```suggestion
if not df.global.gamemode == df.game_mode.ADVENTURE then
```
IMHO is better style

PR #86, opened 2019-01-26
create-unit: several fixes

AtomicChicken:
- `he` was being passed on to createFigure() as nil, resulting in missing civilisation entity associations in the historical figure data

- Also removed the `compute_health` flag, thereby fixing an issue wherein newly spawned citizens were being immediately carted off to hospital. This flag causes the generation of the unit's health structure to occur a tick after being spawned, which results in the generation of a "Recover Wounded" job if the unit in question is a fortress citizen. `unit.health` is very often nil until a unit is injured, and `compute_health` doesn't appear to be useful for provoking body size recalculation or the like, so the inclusion of this flag appears to have been unnecessary.

- In addition, made nameless creatures created in arena mode retain their race-number name ("Chicken 12"), so as to fix the issue of these creatures having a blank display name in this mode. (Nameless creatures outside of the arena display their caste name ("hen") as expected).

- Fixed creatures of the incorrect race/caste being spawned in arena mode after having used the in-game menu to create creatures the normal way. This was due to `df.global.world.arena_spawn.filter` and `df.global.world.arena_spawn.type` not being cleared. Also made the script ignore pre-selected interaction effects (vampirism, etc) in a similar manner.

- Added ability to specify `-name \\LOCAL` to use the entityRawName associated with the group entity in fort mode.

### DIFF
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -593,7 +593,6 @@ for n = 1,spawnNumber do
   u.flags2.calculated_bodyparts = false
   u.flags3.body_part_relsize_computed = false
   u.flags3.size_modifier_computed = false
-  u.flags3.compute_health = true
   u.flags3.weight_computed = false
 
   if age or age == 0 then
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -309,15 +309,12 @@ function createNemesis(trgunit,civ_id,group_id)
 
   nem.save_file_id=-1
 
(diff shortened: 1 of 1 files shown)


lethosor:
It looks like you deleted your fork at some point since making this PR. That does make it a bit harder to test things locally, so hopefully that isn't part of your normal workflow, but I was able to [work around](https://help.github.com/en/articles/checking-out-pull-requests-locally) it.

PR #88, opened 2019-02-09
Added unretire-anyone.lua

AtomicChicken:
Running this script during adventure mode setup pulls up a gui from which the user can select a nemesis to add to the "Specific Person" list on the menu. This effectively allows the user to "unretire" and play as any viable historical figure in adventure mode.


lethosor:
Had fun melting things as a dragon. +1

PR #90, opened 2019-03-24
Script for printing average Focus per need

previsualconsent:
filters for dwarves and sorts the list by average(need_level)*average(focus_level)

### DIFF
--- a/allneeds.lua
+++ b/allneeds.lua
@@ -1,0 +1,84 @@
+-- print sum of all needs
+--[====[
+
+allneeds
+=========
+Show which neesd are high for all dwarfs
+
+]====]
+local utils = require('utils')
+local validArgs = utils.invert({})
+local args = utils.processArgs({...}, validArgs)
+
+local ENUM = {}
+
+ENUM[0]  = "Socialize"
+ENUM[1]  = "DrinkAlcohol"
+ENUM[2]  = "PrayOrMedidate"
+ENUM[3]  = "StayOccupied"
+ENUM[4]  = "BeCreative"
+ENUM[5]  = "Excitement"
+ENUM[6]  = "LearnSomething"
+ENUM[7]  = "BeWithFamily"
+ENUM[8]  = "BeWithFriends"
+ENUM[9]  = "HearEloquence"
+ENUM[10] = "UpholdTradition"
+ENUM[11] = "SelfExamination"
+ENUM[12] = "MakeMerry"
+ENUM[13] = "CraftObject"
+ENUM[14] = "MartialTraining"
+ENUM[15] = "PracticeSkill"
+ENUM[16] = "TakeItEasy"
+ENUM[17] = "MakeRomance"
+ENUM[18] = "SeeAnimal"
+ENUM[19] = "SeeGreatBeast"
+ENUM[20] = "AcquireObject"
+ENUM[21] = "EatGoodMeal"
+ENUM[22] = "Fight"
+ENUM[23] = "CauseTrouble"
+ENUM[24] = "Argue"
+ENUM[25] = "BeExtravagant"
+ENUM[26] = "Wander"
+ENUM[27] = "HelpSomebody"
+ENUM[28] = "ThinkAbstractly"
+ENUM[29] = "AdmireArt"
+
+
+local need = {}
+local focus = {}
+local n = 0
+local n_needs = 0
+for _, unit in ipairs(df.global.world.units.all) do
+    -- local unit = dfhack.gui.getSelectedUnit(true)
+    if not unit then
+        qerror('A unit must be specified or selected.')
+    end
+    if unit.status.current_soul and unit.race == 572 then
+        n = n+1
+         local mind = unit.status.current_soul.personality.needs
+          for k,v in pairs(mind) do
+              n_needs = n_needs +1
+              if need[v.id] == nil then
+                  need[v.id] = {0, 0}
+              end
+              need[v.id] = {need[v.id][1] + v.need_level, need[v.id][2] + v.focus_level}
+          end
+    end
+    -- break
+end
+
+function compare(a,b)
+    return a[1]*a[2] > b[1]*b[2]
+end
+sorted = {}
+i = 1
+for k,v in pairs(need) do 
+    sorted[i] = {v[1], v[2], ENUM[k]}
+    i = i + 1
+end
+
+table.sort(sorted, compare)
+print(string.format("%20s %8s %8s", "Need", "Weight", "Focus"))
+for k,v in ipairs(sorted) do 
+    print(string.format("%20s %8.1f %8.1f",  v[3], v[1]/n, v[2]/n))
+end
--- a/allneeds.lua
+++ b/allneeds.lua
@@ -6,10 +6,6 @@ allneeds
 Show which neesd are high for all dwarfs
 
 ]====]
-local utils = require('utils')
-local validArgs = utils.invert({})
-local args = utils.processArgs({...}, validArgs)
-
 local ENUM = {}
 
 ENUM[0]  = "Socialize"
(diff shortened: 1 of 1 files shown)


janxious:
5mon old fort with basically nothing in it:

```
[DFHack]# allneeds
                Need   Weight    Focus
        DrinkAlcohol      8.0    185.8
        StayOccupied      1.8    382.2
       PracticeSkill      1.6     50.9
              Wander      0.1   -189.9
               Argue      0.8   -285.7
      LearnSomething      1.1   -350.2
        CauseTrouble      0.9   -523.6
          TakeItEasy      0.7  -1169.7
     MartialTraining      0.9  -1500.7
     ThinkAbstractly      0.9  -1550.8
          BeCreative      1.1  -1367.7
               Fight      1.0  -1725.4
         EatGoodMeal      1.0  -1788.7
           MakeMerry      1.0  -1788.7
           Socialize      1.3  -1528.4
          Excitement      1.1  -1900.1
           AdmireArt      1.4  -1526.8
       AcquireObject      1.2  -2088.2
       BeWithFriends      1.8  -2183.1
       BeExtravagant      1.7  -2511.6
        HelpSomebody      2.2  -2702.6
        BeWithFamily      2.0  -3577.3
      PrayOrMedidate      3.8  -5974.7
         CraftObject      5.0  -6486.8
```

Fresh embark, never unpaused:
```
[DFHack]# allneeds
                Need   Weight    Focus
       HearEloquence      0.1     -0.3
          BeCreative      0.6     -1.1
       BeExtravagant      0.6     -1.1
               Fight      0.9     -1.7
           MakeMerry      0.9     -1.7
     MartialTraining      0.9     -1.7
          Excitement      0.9     -1.7
       AcquireObject      1.0     -2.0
        CauseTrouble      1.0     -2.0
      LearnSomething      1.0     -2.0
          TakeItEasy      1.0     -2.0
               Argue      1.0     -2.0
     ThinkAbstractly      1.1     -2.3
        HelpSomebody      1.1     -2.3
           Socialize      1.3     -2.6
      PrayOrMedidate      1.3     -2.6
         EatGoodMeal      1.4     -2.9
       BeWithFriends      1.4     -2.9
       PracticeSkill      1.7     -3.4
        StayOccupied      1.9     -3.7
        BeWithFamily      2.0     -4.0
           AdmireArt      2.0     -4.0
         CraftObject      5.0    -10.0
        DrinkAlcohol     10.0    -20.0
```

150yo stable fort:
```
[DFHack]# allneeds
                Need   Weight    Focus
        DrinkAlcohol      9.6     73.1
           Socialize      1.4     27.9
           SeeAnimal      0.0      8.3
       SeeGreatBeast      0.0      8.3
     SelfExamination      0.0      3.5
       HearEloquence      0.0    -21.8
           AdmireArt      1.8    -13.5
     UpholdTradition      0.0   -507.9
        StayOccupied      2.0    -54.2
         MakeRomance      0.1  -1949.7
          TakeItEasy      1.0   -332.5
        CauseTrouble      1.0   -355.3
               Argue      1.1   -459.5
       PracticeSkill      1.8   -336.0
           MakeMerry      1.1   -644.2
      LearnSomething      1.7   -578.1
          BeCreative      1.0  -1636.2
          Excitement      1.5  -1776.6
               Fight      1.1  -2593.9
     ThinkAbstractly      1.0  -3832.4
        HelpSomebody      1.5  -3617.1
       BeExtravagant      1.2  -5103.3
     MartialTraining      1.0 -18065.7
              Wander      0.9 -18666.0
       AcquireObject      1.1 -17857.1
         EatGoodMeal      1.2 -41473.5
      PrayOrMedidate      4.4 -13128.0
        BeWithFamily      1.9 -36493.9
       BeWithFriends      1.8 -49158.0
         CraftObject      4.6 -27852.6
```

allneeds.lua
```diff
@@ -0,0 +1,81 @@
...
+
+allneeds
+========
+Show which needs are high for all dwarfs.
+Weight and Focus are divided by the number of dwarves that have that need.
+
+]====]
+local ENUM = {}
+
+ENUM[0]  = "Socialize"
```
lethosor: You should use `df.need_type` instead - you're building your own copy of it here. (Side note: all of the built-in enum types provided by DFHack can be indexed by both ID and name. You're indexing by ID here, which is common, i.e. `df.need_type[0] == "Socialize"`, but the reverse, `df.need_type.Socialize == 0`, also works.)

allneeds.lua
```diff
@@ -0,0 +1,81 @@
...
+ENUM[27] = "HelpSomebody"
+ENUM[28] = "ThinkAbstractly"
+ENUM[29] = "AdmireArt"
+
+
+local need = {}
+local n = 0
+for _, unit in ipairs(df.global.world.units.all) do
+    if not unit then
+      qerror('Unit not real.')
```
lethosor: This check isn't really necessary. For one thing, so many tools iterate over the units list without checking for null pointers that it's practically guaranteed that there won't be null pointers. And even if there are, since you're aborting the entire script at this point, you might as well just not have the check. Trying to access a field of `nil` will raise an error anyway, with more descriptive information, and won't crash DF (unlike a similar loop in C++, of which there are many in DFHack plugins). And `Unit not real` isn't really an appropriate description of the situation here. I'd just take out the check.

allneeds.lua
```diff
@@ -0,0 +1,81 @@
+-- print sum of all needs
+--[====[
+
+allneeds
+========
+Show which needs are high for all dwarfs.
```
lethosor: Should be `dwarves` instead.

allneeds.lua
```diff
@@ -0,0 +1,81 @@
...
+
+
+local need = {}
+local n = 0
+for _, unit in ipairs(df.global.world.units.all) do
+    if not unit then
+      qerror('Unit not real.')
+    end
+    -- Select valid dwarf
+    if unit.status.current_soul and unit.race == 572 then
```
lethosor: I'm guessing 572 is the creature raw ID for dwarves in the specific world you were testing? This is **not** guaranteed to be consistent in any other save. It might be usually consistent if you're using the same raws, but any mods will break this check entirely.

There are many functions in the [Units module](https://dfhack.readthedocs.io/en/stable/docs/Lua%20API.html#units-module) that you can use to check traits of units. I would guess `isCitizen` is probably what you actually want here, although perhaps `isDwarf` if you want to exclude non-dwarven citizens from this check.

allneeds.lua
```diff
@@ -0,0 +1,81 @@
...
+
+-- for sorting output
+function compare(a,b)
+    return a[1]*a[2] > b[1]*b[2]
+end
+
+sorted = {}
+i = 1
+for k,v in pairs(need) do
+    sorted[i] = {v[1], v[2], v[3], ENUM[k]}
```
lethosor: This appears to be the only place you use ENUM - once you've taken the definition out, replacing ENUM with df.need_type should produce identical behavior.

lethosor: I would use `table.insert` instead of maintaining the separate `i` counter.

allneeds.lua
lethosor: Actually, I'm curious now - how did you come up with this list? I was able to find [need_type](https://github.com/DFHack/df-structures/blob/0.44.12-r2/df.units.xml#L467) almost immediately in the XML files by searching for `Socialize` and `DrinkAlcohol`, and it looks like the names all match, so clearly you got them from somewhere... but I'm not sure where that was, and if it wasn't making the type (`need_type`) clear.

previsualconsent: So procrastination let me stumble upon this abandoned side project. 

I believe I found that list, but since I do not really understand how lua works at all (my experience is with python and C++) I made it work. 

I think the issue was I didn't know how to print the name of the enum as a string. 

maybe i'll fix this up over the summer in case others find it useful

lethosor: No worries, I've been getting side-tracked too. Like I said, most of the things I had comments on were specific to the "DFHack way" of doing things - the script itself looks good. If I get around to this before you do, I can certainly help clean it up (I'm just focused on getting a stable release finalized at the moment).

myk002:
@previsualconsent This still looks like an innovative and useful script. Are you interested in fixing this up for merge?

myk002:

previsualconsent:
It's been a while since if looked at any of this code. Feel free to
implement the comments

myk002:
I'll see if I can find it a new owner. Thanks for the idea of this script!

myk002:
merged in #538

PR #92, opened 2019-04-22
Fix & Update for v0.44.12

vjek:
Old version would cause a consistent DF crash whenever it was used, and then a particular dwarf's thoughts and preferences were examined.  This was largely due to not handling music, dances, and poetic values properly.

This version fixes that, and adds several new features.  Notable is the ability to put together preference profiles pretty easily, as demonstrated by the 'goth' option.  If there's enough interest, I'll add a preference profile generator with persistent storage.
I also updated the documentation block so that it reflects reality, now.  The 'list' option is very handy, and enumerates all the songs, dances, and poems in the world, among other preference types.

@lethosor adds: Fixes DFHack/dfhack#1438


pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
+        local unit = dfhack.gui.getSelectedUnit()
+        if unit==nil then
+            print ("No unit available!  Aborting with extreme prejudice.")
+        return
+        end
+        clear_preferences(unit)
+        brainwash_unit(unit,profile)
+        return
+    end
+    if opt=="one" then
```
lethosor: This block looks like it's copied directly from above section with one word changed (given that the `return` isn't indented in the same place). Some refactoring to avoid duplicated code is probably in order.

Incidentally, I would use `elseif` instead of a bunch of independent `if`s to make your intentions clearer.

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
+        brainwash_unit(unit,profile)
+        return
+    end
+    if opt=="all" then
+        local profile="IDEAL"
+        clearpref_all_dwarves()
+        adjust_all_dwarves(profile)
+        return
+    end
+    if opt=="goth_all" then
```
lethosor: Again, duplicated from the block above.

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
-clearpref_all_dwarves()
-adjust_all_dwarves()
+pss_counter=31415926
+utils = require 'utils'
+printflag=0
+build_all_lists(printflag)
+
+local opt = ...
+
+if opt and opt ~= "help" then
```
lethosor: You don't really need the outer `if` at all here. If `opt` isn't set, it will end up being `nil`, which will compare not equal to "list" or any of the other strings. If you replace all the `if`s below (except the first, of course) with `elseif`, the logic will work properly.
```lua
if opt == "list" then
  ...
elseif opt == "clear_one" then
  ...
else
  -- show help
end
```

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
+end -- end func build_all_lists
+
 -- ---------------------------------------------------------------------------
 -- main script operation starts here
 -- ---------------------------------------------------------------------------
-clearpref_all_dwarves()
-adjust_all_dwarves()
+pss_counter=31415926
+utils = require 'utils'
+printflag=0
```
lethosor: You should make `printflag` a boolean. Then you can just use `if printflag` without the `==1` bit. (Perhaps you've noticed that `0` evaluates to true in conditions in Lua? Only `false` and `nil` evaluate to false, so use true/false for things like this.)

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
+    for k=0,#vec-1 do
+        name=vec[k].id
+        list_of_plants[name]=k
+        list_of_plants_string=list_of_plants_string..name..","
+    end
+    if printflag==1 then
+        print("\nTYPE 2,5 PLANTS:"..list_of_plants_string) --    printall(list_of_plants)
+    end
+-- ------------------------------------
+    list_of_items={} -- Type 4 "Likes armor.." (TODO need recursive material decode lists?)
```
lethosor: Should probably get rid of this if you're not using it.

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
+    list_of_items_string=""
+    -- [lua]# @dfhack.matinfo.decode(31,1)
+    -- <material 31:1 CREATURE:TOAD_MAN:GUT>
+    -- [lua]# @dfhack.matinfo.decode(0,1)
+    -- <material 0:1 INORGANIC:GOLD>
+    for k,v in ipairs(df.item_type) do
+        list_of_items[v]=k
```
lethosor: Side note: you didn't need `list_of_items` in the first place. `df.item_type` and any other DF enum can be indexed by both key and value. Example:
```
[lua]# ~df.item_type[0]
BAR
[lua]# ~df.item_type.BAR
0
```

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
         if v.race == df.global.ui.race_id then
             print("Adjusting "..dfhack.TranslateName(dfhack.units.getVisibleName(v)))
-            brainwash_unit(v)
+            brainwash_unit(v,profile)
         end
     end
 end
+-- ---------------------------------------------------------------------------
+function build_all_lists(printflag)
+    list_of_inorganics={} -- Type 0 "Likes iron.."
```
lethosor: I commented [below](https://github.com/DFHack/scripts/pull/92/files#diff-d46c24224e6cb0ecfa2dab848b3c2190R186) before I saw these `list_of_*` tables in multiple places. Please remove them if you're not going to use them - they're not necessary in most cases, either, given how DF enum types can be indexed.

pref-adjust.lua
```diff
@@ -1,80 +1,125 @@
...
+    if #df.global.world.musical_forms.all then
+        insert_preference(unit,10,0) -- same goes for music
+    end
+    if #df.global.world.dance_forms.all then
+        insert_preference(unit,11,0) -- and dancing
+    end
+end -- end GOTH profile
+
+    prefcount = #(unit.status.current_soul.preferences)
+    print ("After adjusting, unit "..dfhack.TranslateName(dfhack.units.getVisibleName(unit)).." has "..prefcount.." preferences")
```
lethosor: Everywhere you use `TranslateName` (I see 6 places), you should make sure that when you print to the console (like here) you wrap the output in `dfhack.df2console`, like so:
```lua
print(dfhack.df2console(text_with_translated_name))
```
If you're on Windows, this won't make a difference, because the console there is CP437 encoded (same as DF). But on other platforms, any accented/special characters in names will display incorrectly unless you do this.

pref-adjust.lua
```diff
@@ -1,80 +1,125 @@
...
+end -- end of function brainwash_unit
 -- ---------------------------------------------------------------------------
 function clear_preferences(v)
```
lethosor: I know you didn't change this, but might as well - you're reassigning `v` to `unit`, then using `unit` only once in this function. Just name the parameter `unit` in the first place, and it'll be a lot clearer.

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
 -- main script operation starts here
 -- ---------------------------------------------------------------------------
-clearpref_all_dwarves()
-adjust_all_dwarves()
+pss_counter=31415926
+utils = require 'utils'
+printflag=0
+build_all_lists(printflag)
+
+local opt = ...
```
lethosor: This is not obvious that it assigns only the *first* argument to the script to `opt`. Something like this, albeit ugly, is more explicit:
```lua
local opt = ({...})[1]
```
or alternatively:
```lua
select(1, ...)
```
(the last one is probably better)

lethosor:
Since the link in the review didn't produce any link in the reported issue:
Fixes DFHack/dfhack#1438

pref-adjust.lua
```diff
@@ -1,80 +1,125 @@
...
+        -- mattype for some is non zero, those non-iorganice like creature:gazelle:hoof is 42,344
+    end
+
+    if mytype == 2 then
+        consumable_type=val1[1]
+        consumable_name=val1[2]
+        utils.insert_or_update(unit.status.current_soul.preferences, { new = true, type = 2 , item_type = consumable_type , poetic_form_id = consumable_type, musical_form_id = df.consumable_type, dance_form_id = consumable_type, item_subtype = dfhack.matinfo.find(consumable_name).subtype , mattype = dfhack.matinfo.find(consumable_name).type , mat_state = 0, matindex = dfhack.matinfo.find(consumable_name).index , active = true, prefstring_seed = pss_counter }, 'prefstring_seed')
+    end
+
+    if mytype == 1 or (mytype >= 3 and mytype <= 11) then
```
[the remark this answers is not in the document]
lethosor: Turns out the type here is `unit_preference.T_type`, so `1` is `df.unit_preference.T_type.LikeCreature`. (The reverse is also true, which can help convert numbers into names: `df.unit_preference.T_type[1]` is `"LikeCreature"`.) It's a "nested type", which is admittedly harder to find than a type that's directly under `df`. The list of values is here in the XML files: https://github.com/DFHack/df-structures/blob/0.47.04-r1/df.units.xml#L2167-L2180

lethosor:
@vjek evidently I never got around to this; I'm going to fix up a couple issues I noticed, and I'm planning on getting this into the next release. Thanks for fixing the issues here!

I'm also leaving some comments as I notice things. These are just for your reference - hopefully they're useful for your future script development.

Also, if you plan on making future changes to this script, you'll want to update it with DFHack's copy once this is merged; otherwise, we'll probably run into some merge conflicts later.

pref-adjust.lua
```diff
@@ -1,80 +1,165 @@
...
+            new = true,
+            type = preftype,
+            item_type = val1,
+            creature_id = val1,
+            color_id = val1,
+            shape_id = val1,
+            plant_id = val1,
+            poetic_form_id = val1,
+            musical_form_id = val1,
+            dance_form_id = val1,
```
lethosor, line 78: These 8 fields are a union: https://github.com/DFHack/df-structures/blob/0.47.04-r1/df.units.xml#L2184-L2193
Essentially, they all refer to the same location in memory, just under different names, so you only need to set one of them. (I assume you did it this way because you noticed that they always have the same value - that's true, but you only need to set one of them to make that happen.)

pref-adjust.lua
```diff
@@ -1,80 +1,165 @@
...
+    insert_preference(unit,2,{df.item_type.DRINK,"WEED_RAT:DRINK"})
+    insert_preference(unit,2,{df.item_type.DRINK,"SLIVER_BARB:DRINK"})
+    insert_preference(unit,2,{df.item_type.PLANT,"TUBER_BLOATED:STRUCTURAL"})
+    insert_preference(unit,3,list_of_creatures.ELF)
+    insert_preference(unit,3,list_of_creatures.HUMAN)
+    insert_preference(unit,3,list_of_creatures.DWARF)
+    if list_of_creatures.DEMON_1 and df.global.world.raws.creatures.all[list_of_creatures.DEMON_1].prefstring[0] then
+        insert_preference(unit,1,list_of_creatures.DEMON_1)
+    end
+    if #df.global.world.poetic_forms.all then
```
lethosor, line 146: This doesn't do what you expect, because Lua is weird and treats `0` as `true`:
```lua
> if 0 then print('wat') end
wat
```
I'll add an explicit `> 0` check to these

pref-adjust.lua
lethosor: It's actually not *quite* as simple as I made it out to be - only the last 3 fields are int32s, so setting any of the first 5 fields will only set the lower two bytes of the last 3. `assign()` also only appears to assign the first field passed to it that's in the union (i.e. only the value you give for `item_type` matters here), so I think the safest option here (albeit confusing) is to assign to `poetic_form_id` only. I'll leave a comment to that effect.

pref-adjust.lua
lethosor: I missed the fact that the `list` option uses these. This makes a lot of sense, actually!

pref-adjust.lua
lethosor: For consistency with the others, which don't correspond to DF enums, I think this is fine.

pref-adjust.lua
```diff
@@ -91,16 +136,201 @@ function clearpref_all_dwarves()
...
+        print("\nTYPE 11 DANCES:"..list_of_dances_string) --    printall(list_of_dances)
+    end
+end -- end func build_all_lists
+
 -- ---------------------------------------------------------------------------
 -- main script operation starts here
 -- ---------------------------------------------------------------------------
-clearpref_all_dwarves()
-adjust_all_dwarves()
+pss_counter=31415926
```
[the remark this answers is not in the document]
lethosor: I see now that it was used before, and it's basically a unique preference ID. No need to change

pref-adjust.lua
```diff
@@ -1,80 +1,157 @@
...
+To see what values can be used with each type of preference, use
+``pref-adjust list``.  Optionally, a single dwarf or all dwarves can have
+their preferences cleared manually with the use of ``pref-adjust clear``
+and ``pref-adjust clear_all``, respectively. Existing preferences are
+automatically cleared, normally.
 
 ]====]
 
+utils = require 'utils'
+pss_counter = pss_counter or 31415926
```
lethosor, line 34: This is one potential issue I noticed - the seed would reset every time the script runs, but it's preserved across runs, so using `pss_counter or [initial value]` will only set it to the initial value if it wasn't defined in an earlier run of the script.

PR #93, opened 2019-05-12
Added rivers to the exportlegends custom xml export.

JapaMala:
This will be used in my new tool, but obviously other programs can use it.

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -173,6 +173,24 @@ function export_more_legends_xml()
     end
     file:write("</underground_regions>\n")
 
+    file:write("<rivers>\n")
+    for riverK, riverV in progress_ipairs(df.global.world.world_data.rivers, 'river') do
+        file:write("\t<river>\n")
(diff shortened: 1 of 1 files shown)


exportlegends.lua
```diff
@@ -173,6 +173,24 @@ function export_more_legends_xml()
     end
     file:write("</underground_regions>\n")
 
+    file:write("<rivers>\n")
+    for riverK, riverV in progress_ipairs(df.global.world.world_data.rivers, 'river') do
+        file:write("\t<river>\n")
+        file:write("\t\t<name>"..dfhack.df2utf(dfhack.TranslateName(riverV.name, 1)).."</name>\n")
```
lethosor: :+1: for `df2utf`

lethosor: Worth noting that you're printing the English names, so I don't think you strictly need this (but it is a good idea). I assume you knew this and don't want the dwarven names instead, yes?

JapaMala: Yeah, I just want the English names. Honestly, I don't even know if strictly need them at all, but it's good to have them

PR #94, opened 2019-05-21
Add special case to exterminate to only target hostile members of a race

janxious:
Add a new special case caste called `enemy` so one can target non-fortress guests i.e. goblin invaders  without the tedium of doing 1 by 1 or killing citizens of your fort that happen to be goblins.

I have done limited testing of this with invaders and undead and things but this is feature complete I think.

Also it updates the help text to indicate how to use this feature.

### DIFF
--- a/exterminate.rb
+++ b/exterminate.rb
@@ -34,6 +34,7 @@ Ex::
 
     exterminate gob
     exterminate gob:male
+    exterminate gob:enemy
 
 To kill a single creature, select the unit with the 'v' cursor and::
 
@@ -117,6 +118,7 @@ when nil
 when 'help', '?'
     puts <<EOS
 Kills all creatures of a given race.
+You may add a 'caste' to the race to kill all males, females, or a special case: enemy.
 With no argument, lists possible targets with their head count.
 With the special argument 'this', kill only the currently selected creature.
     Alternatively, 'him', 'her', 'it', 'target', and 'selected' do the same thing.
@@ -129,6 +131,7 @@ The special final argument 'butcher' will mark the targets for butchering instea
 
 Ex: exterminate gob
     exterminate gob:male
+    exterminate gob:enemy
     exterminate elve magma
     exterminate him
     exterminate pig butcher
@@ -167,7 +170,9 @@ else
 
     race_nr = df.world.raws.creatures.all.index { |cr| cr.creature_id == raw_race }
 
-    if caste
+    if caste == "enemy"
+        caste_nr = true
+    elsif caste
         all_castes = df.world.raws.creatures.all[race_nr].caste.map { |c| c.caste_id }
         raw_caste = df.match_rawname(caste, all_castes)
         if not raw_caste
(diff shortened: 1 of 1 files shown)


JapaMala:
Can it check of the entity is at war?

janxious:
@JapaMala as an addition, or instead of the hostile check?

I think it would be as easy as checking the stuff returned from https://github.com/DFHack/dfhack/blob/315852a2513679ca1afe0d53b34974cefd946f89/plugins/ruby/unit.rb#L100

exterminate.rb
```diff
@@ -167,24 +170,30 @@
...
+        caste_nr = false
     end
 
     count = 0
-    df.world.units.active.each { |u|
+    df.world.units.active.each do |u|
         if u.race == race_nr and checkunit[u]
-            next if caste_nr and u.caste != caste_nr
+            if caste_nr
+                next if caste == "enemy" && df.unit_hostile(u)
```
lethosor: Should this be `!df.unit_hostile(u)`?

janxious: It should. This is what I get for editing in the browser.

exterminate.rb
```diff
@@ -188,7 +188,7 @@
     df.world.units.active.each do |u|
         if u.race == race_nr and checkunit[u]
             if caste_nr
-                next if caste == "enemy" && !df.unit_hostile(u)
+                next if caste == "enemy" && !df.unit_ishostile(u)
```
lethosor: Nice, I missed this (good thing I'm slow at merging PRs, I guess)

janxious: no joke 🙈

janxious: Not the being slow thing, just missing easy things.

lethosor:
As a side note, for working on scripts, you can edit `dfhack-config/script-paths.txt` and point it to your local clone of this repo to avoid having to copy files manually (if you were doing that). It'll pick up changes to scripts as soon as you save and re-run them in DFHack, which hopefully makes testing easier.

janxious:
that would be good, but my dev env and play env are separate :/

janxious:
Another interesting wrinkle for this script is if you have living and undead creatures of a race on map, you can target them with enemy, but if you only have undead you cannot. So I have dwarfs that are alive and a bunch of undead so I can `exterminate dwarf:enemy` and just kill the undead dwarfs and leave the rest of the undead alone, but I *cannot* `exterminate troll:enemy` because they are undead and the way it's written precludes targeting races that are not expected by the script because there are living ones. I think changing that is out of scope, but it's a thing I ran across in testing that confused me.

janxious:
I recently tested this again against some dwarf invaders and it worked as expected. I think this is shippable.

lethosor:
Looks good! I made a couple tweaks in 04de918 - particularly the `slain 0  YAK` message had two spaces in there, because it was trying to print the caste, so I added that. I don't think I have any saves with sieges/invaders handy, so thanks for the update.

janxious:
👍 thanks @lethosor!e

PR #97, opened 2019-06-26
fix tame to work with autobutcher

pseudodragon:
needed to set the civ_id on the tamed animal or it won't be seen by autobutcher.

### DIFF
--- a/tame.lua
+++ b/tame.lua
@@ -34,6 +34,7 @@ if selected ~= nil then
         selected.flags1.tame = level ~= 8 and level ~= 9
         if selected.flags1.tame then
             selected.training_level = level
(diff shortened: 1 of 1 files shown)


lethosor:
Worth noting that this isn't reversible (i.e. once you've tamed an animal, un-taming the same animal won't clear their civ ID), but that's probably fine.

pseudodragon:
> Worth noting that this isn't reversible (i.e. once you've tamed an animal, un-taming the same animal won't clear their civ ID), but that's probably fine.
we could add 
            selected.civ_id = -1
after 
            selected.training_level = 9

lethosor:
That will have side effects if the animal *was* part of a civ before, though.

PR #99, opened 2019-07-15
modtools/create-unit: Tidied up, with some minor additions

AtomicChicken:
- Rewrote and tidied up various sections of the script, moving everything into module-friendly functions (including a central processNewUnit function). Removed clutter and some bits of unnecessary code.
- Now enables the default labours for adult units with CAN_LEARN.
- Now sets historical figure orientation.
- Now clears the arena spawn equipment list before unit creation to avoid unintentionally spawning units with extra items or skills. This is carried out even outside of arena mode, to eliminate the possibility of interference by other scripts making use of arena spawning. The equipment lists are repopulated once unit creation is complete, returning the arena creature creation menu to its original state.
- Moved the spawn loop to the points between equipment data clearance and repopulation, to avoid having to needlessly repeat this process when using the -quantity arg to spawn multiple units simultaneously.
- Removed the dubious inclusion of unidentified nemesis flags, which I've concluded are unnecessary (having observed several historical units in all modes, including fortress citizens, with none of these flags set). As a result, created units no longer unintentionally exhibit a glowing creature tile.
- When asked about their family in adventure mode, created units now state that they "have no family to speak of" instead of describing [the historical figure with an ID of 0] as their head of the family. (Makes reference to `family_head_id` from DFHack/df-structures#325).


modtools/create-unit.lua
```diff
@@ -37,64 +27,94 @@ Creates a unit.  Usage::
...
-end
---]]
+local utils = require 'utils'
 
-local utils=require 'utils'
+function createUnit(raceStr, casteStr, pos, age, domesticate, civ_id, group_id, entityRawName, nickname, quantity, flagSet, flagClear)
+--  creates the desired unit(s) at the specified location
+--  returns a table containing the created unit(s)
+  if not pos then
+    qerror("Location not specified!") -- check repeated for module usage
```
lethosor: This is breaking `spawnunit`. Haven't investigated further, but maybe `pos` used to default to the cursor position?

AtomicChicken: The createUnitInner function in modtools/create-unit works by simulating the arena mode spawning process roughly as follows: 
1) Populate the arena spawn data as desired
2) Set up a dwarf mode viewscreen
3) Set the ui mode to "look around", causing the cursor to appear (at its last position)
4) Switch to arena mode
5) Move the cursor to the target spawn location
6) Open the arena creature creation menu, with the target creature pre-selected
7) Pass the equivalent of the "Enter" key to spawn the unit
8) Process the unit as desired
9) Loop steps 6-8 if creating multiple units simultaneously
10) Return everything to its original state

As you can see, the position at which the unit is spawned depends entirely on the cursor location prior to opening the creature creation menu. If the target location were to be left unspecified, the cursor would be left untouched, causing the spawn location to default to wherever it was that the cursor was left prior to running the script.

Since such a scenario is generally not desirable, I added a pos check within the function itself just in case it’s ever called by other scripts. `spawnunit` is failing at this check as it does not require the user to pass location data to `modtools/create-unit`, instead seeming to rely on users having selected the spawn location via the cursor prior to running the script  in the event that this information is omitted (despite this not being explained in the documentation). On that note, I have not figured out how to correctly specify a target location using x, y, z coordinates in `spawnunit`, despite this purportedly being possible according to the documentation.

I’m not sure what `spawnunit` is meant to be used for (indeed, I was unaware of its existence prior to reading the above), but given its limited capabilities as a front-end, my initial impulse is to just have it include a cursor check and do away with its defective(?) location input (since end-users are unlikely to fumble around with coordinates in lieu of using the cursor, and modders are better suited to using the proper modtools script for their purposes). Thoughts?

lethosor: The original spawnunit took race, caste, and name, and created a unit at the selected cursor position. The new one is just a wrapper around this script which passes through any extra arguments. If always passing a position is necessary, that shouldn't be too hard, I guess.

PR #100, opened 2019-07-29
Fix error in gui/companion-order's Leave command

Atkana:
Not sure if it was a typo or if things were changed between versions of DF, but this line was causing an error when the Leave command was executed. I've changed it to the correct value, so the script no longer crashes when that command is used.

### DIFF
--- a/gui/companion-order.lua
+++ b/gui/companion-order.lua
@@ -360,7 +360,7 @@ end},
         v.relationship_ids.GroupLeader=-1
         local u_nem=dfhack.units.getNemesis(v)
         if u_nem then
-            u_nem.relationship_ids.GroupLeader=-1
+            u_nem.group_leader_id=-1
         end
         if t_nem and u_nem then
             for k,v in pairs(t_nem.companions) do
--- a/gui/companion-order.lua
+++ b/gui/companion-order.lua
@@ -369,6 +369,18 @@ end},
                     break
                 end
             end
+            for k,v in pairs(t_nem.figure.histfig_links) do
+                if df.histfig_hf_link_companionst:is_instance(v) and v.target_hf == u_nem.figure.id then
+                    t_nem.figure.histfig_links:erase(k)
+                    break
+                end
+            end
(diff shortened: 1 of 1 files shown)


warmist:
Huh... 
It seems that the commit before you did exactly the reverse. Are you sure you are at last version of dfhack? See commit: #9558feba77c57168a7970e1ebcbc1c7b3a82ab97

Atkana:
I'm currently running version `0.44.12-r2`, if that's the most up to date version.

From what I can tell, the changes in that old commit are correct for the instances where manipulations to _units_ were changed. Both _units_ and _nemesis_ had a value called `group_leader_id` and some DF update (?) presumably changed it so now _units_ have `relationship_ids.GroupLeader` instead, while _nemesis_ still kept `group_leader_id`. It just happens that this manipulation to a _nemesis_ was accidentally included in the commit (presumably it was just a find and replace :P).

lethosor:
Yeah, probably a faulty find and replace. The unit change happened around 0.43.05 (in 2016) from what I remember, though, so it's not anything that new.

lethosor:
Yeah, that was me in https://github.com/DFHack/scripts/commit/9558feba77c57168a7970e1ebcbc1c7b3a82ab97#diff-6c7bc3a240451d6a415637094cd5f270. Must have thought it was a unit because it started with "u", I guess.

Atkana:
Spotted that companion links weren't being removed when using leave, so I've added that in to. If I figure out agreements before this pull goes through, I'll add something for ending the adventuring agreement, too.

lethosor:
No rush. There are some other things still to do for r3: https://github.com/orgs/DFHack/projects/7. Just let me know when this is ready.

Atkana:
Hehe, seeing this go through reminded me I had a bit more concerning agreements to add, but nothing that actually made the agreement count as ended in the log screen (so not really vital)... I tried for a few hours to figure out how an agreement counts as concluded but eventually gave up. If I had to guess, it might be something relating to an agreement's party anon1, but I couldn't work out how to access what was stored in it properly :/

PR #101, opened 2019-08-01
Added modtools/create-tree.lua and gui/create-tree.lua

AtomicChicken:
Written with this thread in mind: http://www.bay12forums.com/smf/index.php?topic=174381.0

`modtools/create-tree.lua` is based on modtools/create-unit.lua, and works via similar arena mode logic. It is intended for modding use, and is module-friendly.

`gui/create-tree.lua` is a user-friendly front for the former script, using the cursor to determine spawn location and allowing the user to manually select the desired tree from a list gui (and also input its age).


lethosor:
If there are dialogs involved, I might recommend `gui/spawn-tree`, similar to `gui/create-unit`

AtomicChicken:
Got it; I'll move it to the gui folder.

PR #102, opened 2019-08-11
Added combat-harden.lua

Atkana:
This script can be used to alter the combat_harden value of a unit (or a whole fort), which effects the emotions that a unit feels when witnessing death (to the point where they feel nothing).

(I originally made this to publish with my own list of scripts, but I figured it was something that might have wide enough appeal to go into DFhack proper.)

### DIFF
--- a/combat-harden.lua
+++ b/combat-harden.lua
@@ -1,0 +1,131 @@
+-- Sets a unit's combat-hardened value to a given percent
+--@ module = true
+
+local help = [====[
+
+combat-harden
+=============
+Sets the combat-hardened value on a unit, making them care more/less about seeing corpses.
+Requires a value and a target.
+
+Valid values:
+
+:value <0-100>:
+    A percent value to set combat hardened to.
+:tier <1-4>:
+    Choose a tier of hardenedness to set it to.
+      1 = No hardendness.
+      2 = "is getting used to tragedy"
+      3 = "is a hardened individual"
+      4 = "doesn't really care about anything anymore" (max)
+
+If neither are provided, the script defaults to using a value of 100.
+
+Valid targets:
+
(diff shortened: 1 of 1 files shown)


combat-harden.lua
```diff
@@ -0,0 +1,131 @@
...
+local help = [====[
+
+combat-harden
+=============
+Sets the combat-hardened value on a unit, making them care more/less about seeing corpses.
+Requires a value and a target.
+
+Valid values:
+
+:value <0-100>:
```
lethosor: My thoughts (for when I'm back online and can merge this, but feel free to change this too if you're monitoring this PR): it might not be clear that you need the `-` in front of the arguments here (`-tier` instead of `tier`, `-all`, etc.), so I would specify that. A bunch of the modtools use a different style for this ([example](https://github.com/DFHack/scripts/blob/master/modtools/create-item.lua)) although I don't really like how that displays in the [HTML docs](https://dfhack.readthedocs.io/en/stable/docs/_auto/modtools.html#modtools-create-item). Adding the `-`s is probably enough, honestly, and I can take care of that when I'm able to merge this.

combat-harden.lua
```diff
@@ -0,0 +1,131 @@
...
+        for _, unit in pairs(df.global.world.units.active) do
+            table.insert(unitsList, unit)
+        end
+    elseif args.citizens then
+        -- Technically this will exclude insane citizens, but this is the
+        -- easiest thing that dfhack provides
+        
+        -- Abort if not in Fort mode
+        if not dfhack.world.isFortressMode() then
+            return
```
lethosor: An error might be helpful (e.g. `qerror("Not in fortress mode")`)

PR #103, opened 2019-08-11
Update create-unit.lua to use nemesis ID

Atkana:
Previously, when the script created a historical figure + nemesis record, the historical figure's `unit_id2` record was being set to the unit's id, rather than the nemesis id.


warmist:
I've checked this. Seems you are right. Would be nice to rename unit_id2 to nemesis_id in structures.

PR #105, opened 2019-09-26
Unit editor with profiles

acattaneo88:
At the moment you can alter attributes, skills, preferences, beliefs, goals and facets. They can be assigned one at a time or you can define a profile and apply it as a whole to the unit. More modules providing more editing functionality can be added (if they expose a common function so that the profile script can call them).

The most important script (which represents the main idea behind this pull request) is the `assign-profile` script: everything else is just modules (but they can also be used as standalone scripts) that this script call.

Some scripts in this repository have overlapping scopes, I think we should keep and maintain only one version.

English is not my first language, but I've tried to write clear and complete documentation. If anything is not clear, please ask.

The included profiles and modules are just an example of what can be done, and I would appreciate your contribution in creating more profiles.


acattaneo88:
TODO: implement changes suggested by @lethosor . Will be back soon,

assign-facets.lua
```diff
@@ -0,0 +1,217 @@
...
+    ``She very rarely develops negative feelings
+    toward things. She is rarely happy or enthusiastic,
+    and she is conflicted by this as she values
+    parties and merrymaking in the abstract.``
+
+    Note that the facets are compared to the beliefs,
+    and if conflicts arise they will be reported.
+]====]
+
+local valid_args = {
```
lethosor: There's a `utils.processArgs()` function that does something pretty similar to this - have you tried that?

acattaneo88: Yes, I thought about adding some of these scripts to gui/gm-unit: I will probably do it in the future, thanks for the suggestion.

And yes, I have tried the `utils.processArgs()` function but it took me a while to understand how to pass lists of different parameters to it: I had to analyse the source code to be sure and so I thought that maybe I should implement it in a different way. Anyway, I will be more than happy to change the arguments parsing function of my scripts to call the `utils` standard function in order to standardise the code between different scripts in this repository, if you prefer so.

lethosor: I just noticed there was a decent amount of argument-parsing code that you'd duplicated between scripts, and was thinking that could clean it up. It's possible that yours behaves differently from processArgs, though, but it looks pretty similar.

acattaneo88: Ok, I cleaned the argument-parsing code a little bit and performed a few tests: everything seems to be working nicely.
At the moment, I am not so confident about the `assign-profile.lua` script: it feels like it needs more tests, but the other scripts are independent from one another and can be used.

lethosor:
Thanks for the changes! I meant to get to this about two weeks ago but got wrapped up in other stuff. This is looking good; I might make one or two tweaks, but the docs look _great_, which is a plus.

PR #108, opened 2019-10-31
Add various tools, and update gui/gm-unit

Atkana:
I recently got through updating a bunch of my modtools + old edits to gui/gm-unit and thought they might be worth submitting. Only now have I noticed that some of them are similar to those submitted in another pull.

### modtools/set-need
Provides command line and script function options for editing a unit's needs.
### modtools/set-belief
Provides command line and script function options for editing a unit's values. I used the term "belief" instead of the game's usage of "value" because of the latter's use as a generic programming term (which might cause some confusion as to what the script does). The randomisation function doesn't follow the proper base game method (I don't know how it works), but it's good enough for a start - anyone who knows how it works is welcome to implement it / tell me.
### modtools/set-personality
Provides command line and script function options for editing a unit's personality facets. Same case for randomisation as `modtools/set-belief`.
### modtools/pref-edit
Provides command line and script function options for editing a unit's preferences.
### set-orientation
Provides command line and script function options for editing a unit's sexual orientation.
### Edits to gui/gm-unit
Added: Personality editor, belief editor, orientation editor, Body appearance editor, colors editor. Originally, the latter two were supposed to be separate scripts/modtools, but I couldn't decide how I wanted to make it, so for now those two functions are hardcoded into that editor. In the future, I might still make them and make a new edit to separate it all out.

Also sorry if I screwed up updating my master fork, I barely know how to use github :P


modtools/pref-edit.lua
```diff
@@ -0,0 +1,326 @@
...
+--@ module = true
+
+local help = [====[
+
+modtools/pref-edit
+==================
+Add, remove, or edit the preferences of a unit.
+Requires a modifier, a unit argument, and filters.
+
+:unit <UNIT ID>:
```
warmist: Afaik `[]` is used for optional arguments

gui/gm-unit.lua
```diff
@@ -613,6 +613,700 @@ function editor_wounds:init( args )
 end
 add_editor(editor_wounds)
 
+-- ATKANA EDIT START
+local setbelief = dfhack.reqscript("modtools/set-belief")
+local setpersonality = dfhack.reqscript("modtools/set-personality")
+local setneed = dfhack.reqscript("modtools/set-need")
+local setorientation = dfhack.reqscript("set-orientation")
```
warmist: any reason this is not same as others? (i.e. modtools folder)

gui/gm-unit.lua
```diff
@@ -613,6 +613,700 @@ function editor_wounds:init( args )
...
+    end
+    table.insert(choices, {text = self:beautifyString(df.appearance_modifier_type[modifier.entry.type]) .. ": " .. currentValue, value = currentValue, modifier = modifier})
+  end
+  
+  self.subviews.modifiers:setChoices(choices)
+end
+
+function editor_body_modifier:init(args)
+  self.target_unit = args.target_unit
+  self.partChoice = args.partChoice
```
warmist: In cases like this, you can use `class.ATTRS={name=default}` with DEFAULT_NIL if you want it to be nil by default. That way if you pass something to `args` it gets auto-set to a value. More about it here: https://dfhack.readthedocs.io/en/stable/docs/Lua%20API.html#class

modtools/pref-edit.lua
Atkana: I checked the style guide but couldn't find anything about using square brackets (though did find I'm not sticking to them exactly, oops :P). It's kinda confusing as to whether the unit id is an optional addition because of how the scripts handle missing/invalid targets.

gui/gm-unit.lua
Atkana: I didn't include set-orientation as a modtool basically as a whim (and I don't fully know what counts as a modtool and what doesn't). If it seems like it should belong as a modtool (or that the others shouldn't), somebody more knowledgable about them is welcome to change them.

gui/gm-unit.lua
Atkana: Ahah, so that's how that works. I'll remember this for the future!

modtools/pref-edit.lua
lethosor: Yeah, that's probably something we could put in our documentation (I'll try to find a good place for it)

warmist: It's more of a "unix" convention or just general commandline tool convention.

Atkana:
Is there something I need to do to get rid of the red x?

warmist:
According to the [build log](https://travis-ci.org/DFHack/scripts/builds/605568149?utm_source=github_status&utm_medium=notification) set-orientation.lua:24 there is a unexpected intendation and there are some trailing whitespaces in set-need.lua:554 and gm-unit.lua: lines 679, 988, 1138, 1240.

Atkana:
I thought lua didn't care about whitespace, how come it's a problem? Also the error isn't really helpful to me because I don't know what expected / unexpected indentation is. I've edited it to match as closely to all the other scripts I've submitted but it still isn't working, so I have no idea why that one in particular is causing a problem.

warmist:
AFAIK first one is documentation generation warning.
Second ones are our whitespace requirements. I'm not sure why project have these requirements but usually it's for tidyness and cleaner commits

lethosor:
Yeah, the current error is from Sphinx (parsing the documentation as RST). I agree, I don't see what's wrong with it, but I'll look into it.
```
Warning, treated as error:
../../../dfhack/scripts/modtools/pref-edit.lua:25: ERROR: Unexpected indentation.
```

modtools/set-belief.lua
```diff
@@ -0,0 +1,460 @@
+-- Changes the beliefs (values) of units.
```
lethosor: Just realized this sounds similar to `assign-belief` from #105 - haven't really looked into how similar they are, but we might need to consolidate them.
None of your other scripts sound as similar, but there could be some overlap too.

Atkana: Hehe yeah, that's the one I was alluding to :P

lethosor: Since neither are merged yet, I'm thinking I'll take the ones from #105 and make sure any extra features from these make it in too - sound ok?

Atkana: If you want, but you'd be having to add quite a lot of features. It's awkward that they both have overlap and dependencies on themselves. It might be dumb but you could always include both, it's not like the regular scripts aren't without redundancies (or at least some will be with the inclusion of either of these) :P

gui/gm-unit.lua
lethosor: Generally if they are end-user-facing they shouldn't be in modtools (my rule of thumb is "end users shouldn't have to type 'modtools' into the DFHack console). Looks like most of these are at least somewhat user-facing, so I would probably move them out.

modtools/set-belief.lua
lethosor: The gm-unit features are neat, and I don't feel like reimplementing those using the other scripts, so I'll merge this as-is and we can refactor it later.

lethosor:
Another update: looks like the addition of `NONE` values of `value_type` and `personality_facet_type` in https://github.com/DFHack/df-structures/pull/331 broke set-belief and set-personality, respectively.

PR #109, opened 2019-11-02
export-dt-ini: add more size_info fields

cvuchener:
Rename size_info to size_cur and add size_base.

required by https://github.com/Dwarf-Therapist/Dwarf-Therapist/pull/175


lethosor:
Is this because the layout of size_info changed between some releases? (I'm not entirely sure I understand the bug - sorry if I'm asking something you've already answered.)

Also, https://github.com/Dwarf-Therapist/Dwarf-Therapist/pull/175 looks like it still includes size_info in the ini files - did you mean to get rid of it here?

cvuchener:
I needed size_base for fixing the weapon check. I changed size_info to size_info.size_cur, because it is read as an int not the structure. I don't know why it was this way.

size_info is kept for compatibility as older DT versions may automatically download newer memory layouts. It will be deleted later.

Or maybe I should keep size_info for the name of size_info.size_cur? At first I forgot about this issue, so I just renamed it. Then I realized the problems it would cause and readded it. Now that I am thinking about it, I think I should be more conservative with names in the memory layouts. Hold the merge, I will revert the name.

cvuchener:
updated with old name reverted and a bit more explanations in the commit message.

PR #110, opened 2019-11-16
Added deep-embark.lua

AtomicChicken:
I wrote this following a modder's request for a script that enables players to embark underground. It does this by detecting blocks corresponding to the specified depth (any of the three cavern layers or the underworld), finding one that has a sufficient number of valid walkable tiles in it, and teleporting starting units and equipment there. I developed a workaround for triggering the game to properly reveal tiles centered around the target location which involves placing a glowing barrier tiletype and making it disappear; I'm unaware if there's a more straightforward method to do this, but it works quite well. This script requires the user to set up an `onMapLoad.init` file as detailed in the `-help` printout.

### DIFF
--- a/modtools/deep-embark.lua
+++ b/modtools/deep-embark.lua
@@ -238,8 +238,8 @@ function moveEmbarkStuff(selectedBlock, embarkTiles)
       local flags = item.flags
       if item.age == 0 -- embark equipment consists of newly created items
       and item.pos.x >= x1 and item.pos.x <= x2
-      and item.pos.y >= y1 and item.pos.x <= y2
-      and item.pos.z >= z1 and item.pos.x <= z2
+      and item.pos.y >= y1 and item.pos.y <= y2
+      and item.pos.z >= z1 and item.pos.z <= z2
       and flags.on_ground
       and not flags.in_inventory
       and not flags.in_building
--- a/modtools/deep-embark.lua
+++ b/modtools/deep-embark.lua
@@ -1,0 +1,321 @@
+-- embark underground
+-- author: Atomic Chicken
+
+local usage = [====[
+
+deep-embark
+===========
+Moves the starting units and equipment to
+a specific underground region upon embarking.
+
+To use, create a file called "onMapLoad.init"
+in the DF raw folder and enter within it the name of
+this script followed by any of the args listed below.
+
+example:
+    modtools/deep-embark -depth CAVERN_2
+
+Usage::
+
+    -depth X
+        (obligatory)
+        replace "X" with one of the following:
+            CAVERN_1
+            CAVERN_2
+            CAVERN_3
+            UNDERWORLD
+
+    -atReclaim
+        including this arg will enable deep embarking
+        when reclaiming sites too
+
+]====]
+
+local utils = require 'utils'
+local teleport = reqscript("teleport").teleport
+
+function getFeatureID(cavernType)
+  local features = df.global.world.features
+  local map_features = features.map_features
+  if cavernType == 'CAVERN_1' then
+    for i, feature in ipairs(map_features) do
+      if feature._type == df.feature_init_subterranean_from_layerst
+      and feature.start_depth == 0 then
+        return features.feature_global_idx[i]
+      end
+    end
+  elseif cavernType == 'CAVERN_2' then
+    for i, feature in ipairs(map_features) do
+      if feature._type == df.feature_init_subterranean_from_layerst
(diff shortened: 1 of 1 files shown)


clinodev:
This is very, very cool! I wonder if it couldn't be expanded slightly to make it possible to invoke through the terminal in embarkation rather than solely through the onMapLoad.init file? I also wonder about the lack of a cavern discovery message (for spores,) but I note I also didn't get a magma discovery message and my test embark is able to build magma workshops immediately. I imagine you've tested it.

Regardless, this is a script I would have asked for years ago if I imagined it was possible! 

Thank you!

modtools/deep-embark.lua
```diff
@@ -0,0 +1,321 @@
...
+    local oldDO_MEGA = announcementFlags.DO_MEGA
+    announcementFlags.DO_MEGA = false -- turn off popup announcement
+    local oldPAUSE = announcementFlags.PAUSE
+    announcementFlags.PAUSE = false
+    local oldRECENTER = announcementFlags.RECENTER
+    announcementFlags.RECENTER = recenter and true or false
+    local oldA_DISPLAY = announcementFlags.A_DISPLAY
+    announcementFlags.A_DISPLAY = false
+    local oldD_DISPLAY = announcementFlags.D_DISPLAY
+    announcementFlags.D_DISPLAY = recenter and true or false -- won't recenter without an actual announcement
```
lethosor: Looking at https://dfhack.readthedocs.io/en/stable/docs/Lua%20API.html#typed-object-references, I think you could simplify this with:
```lua
local oldFlags = df.global.d_init.announcements.flags.ENDGAME_EVENT_1:new()
...
df.global.d_init.announcements.flags.ENDGAME_EVENT_1:assign(oldFlags)
```

AtomicChicken: A good idea; changed as advised.

AtomicChicken:
I've updated the script with a new arg and a minor fix. It makes reference to changes from DFHack/df-structures#329.

@clinodev, you're welcome! I have indeed tested the script, and can confirm that cavern spores will spread as expected. Regarding your request: I think I'll go ahead and make a separate terminal-friendly wrapper for this script (since modtools scripts aren't meant to be used in this manner).

modtools/deep-embark.lua
```diff
@@ -115,9 +122,46 @@ function getValidEmbarkTiles(block)
...
-function reveal(pos, recenter)
--- creates an unbound glowing barrier at the target location to trigger proper tile revelation when it disappears (fortress mode only)
--- if recenter is true, shifts the view to the target location
+function blockGlowingBarrierAnnouncements(recenter)
+--  temporarily disables the "glowing barrier has disappeared" announcement
+--  announcement settings are restored after 1 tick
+--  setting recenter to true enables recentering of game view to the announcement position
+  local announcementFlags = df.global.d_init.announcements.flags.ENDGAME_EVENT_1 -- glowing barrier disappearance announcement
+  local oldFlags = df.global.d_init.announcements.flags.ENDGAME_EVENT_1:new()
+  oldFlags:assign(announcementFlags) -- backup announcement settings
```
lethosor: I think :new() creates a copy, so you don't need the :assign() call (but I have yet to actually test this)

AtomicChicken: Looks like this is correct; I'll remove it.

lethosor:
Thanks! FYI, I tweaked some things: https://github.com/DFHack/scripts/compare/c9931ad...4be69fb
My general rule of thumb is that modtools is reserved for scripts that users will never run directly - this is a bit of a grey area, because mods definitely *could* use it, but it also works from the console directly (when run at the correct time).

AtomicChicken:
@lethosor 

This script as is **cannot** be run from the console directly, as it is written to perform all its checks and actions immediately upon loading the map (checking for a fortress age of 0 ticks to determine whether it's a fresh embark or not, for instance). Manually running the script from the console within this timeframe is impossible as fortress mode starts off unpaused. Moreover, the lack of qerror messages was intentional, as I visualised an embark location being preset for a particular mod, with the script being silently aborted when the player is not freshly embarking. This change will result in error messages every single time a fortress is reloaded and throughout adventure mode play.

A better approach might be to make a console-friendly front to be run whilst setting up the embark, which waits for the moment of embark to occur before immediately calling this script.

lethosor:
I should clarify that I ran it from the console on the "you have arrived" screen, and it seemed to work there, although I can't guarantee that there weren't any side effects.

PR #112, opened 2020-02-13
Added creature_raw names and flags

PatrikLundell:
Support for Legends Viewer (and its siblings). Note that no attempts to add 0.47.X additions has been made. That will have to be done by someone more familiar with what needs to be exported separately and what the standard XML provides just fine.

I forgot to add that this fixes DFHack/dfhack#1493.

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -194,6 +194,18 @@ function export_more_legends_xml()
     end
     file:write("</rivers>\n")
 
+    file:write("<creature_raw>\n")
+    for creatureK, creatureV in ipairs (df.global.world.raws.creatures.all) do
+        file:write("\t<creature>\n")
+        file:write("\t\t<creature_id>"..creatureV.creature_id.."</creature_id>\n")
+        file:write("\t\t<name_singular>"..creatureV.name[0].."</name_singular>\n")
+        file:write("\t\t<name_plural>"..creatureV.name[1].."</name_plural>\n")
+        for flagK, flagV in ipairs (df.creature_raw_flags) do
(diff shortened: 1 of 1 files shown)


PatrikLundell:
Hm, I don't think there's any need for any document changes, so I suspect it's not really my task to fix the document complaint.

lethosor:
Yeah, there was a new script added that doesn't have docs, so Travis is complaining about that - no need to worry about fixing it here

exportlegends.lua
```diff
@@ -194,6 +194,18 @@ function export_more_legends_xml()
     end
     file:write("</rivers>\n")
 
+    file:write("<creature_raw>\n")
+    for creatureK, creatureV in ipairs (df.global.world.raws.creatures.all) do
+        file:write("\t<creature>\n")
+        file:write("\t\t<creature_id>"..creatureV.creature_id.."</creature_id>\n")
+        file:write("\t\t<name_singular>"..creatureV.name[0].."</name_singular>\n")
+        file:write("\t\t<name_plural>"..creatureV.name[1].."</name_plural>\n")
```
lethosor: Just as a heads-up, it appears that these names can be generated, which means that they need to be escaped (df2utf and probably escape_xml to be safe) to avoid errors like these:
```
$ xmllint --noout medium-world-04702-00162-01-01-legends_plus.xml 
medium-world-04702-00162-01-01-legends_plus.xml:157043: parser error : Input is not proper UTF-8, indicate encoding !
Bytes: 0x98 0x6F 0x6E 0x61
		<name_singular>folly of �onaliitheca</name_singular>
		                        ^
```

PR #114, opened 2020-02-21
Adds/Updates scripts

cppcooper:
There is a problem with the documentation in basically all the scripts, I can't figure out how to resolve said issue. "Definition list ends without a blank line; unexpected unindent."
Honestly I am not entirely sure what it is even referencing, it seems to always point to lines 20~22 and I first thought it had to do with the `validArgs` list, and then the help documentation. Tried several ideas for solving it, the only thing that works is removing it all (I did not test extensively enough to determine which section's removal solves it though).

Added Scripts:
- animalcontrol.lua: helpful in managing butchery/gelding of animals
- dig.lua: simple script for hack-digging channels
- geld.lua: simple script for gelding,ungelding, and toggling the geld status of animals
- getwaves.lua: helpful in reading migration wave information
- settile.lua: simple script for changing the tileset of a tile (it may or may not change anything else as a consequence)
- tile-query.lua: helpful in querying tiles for information about their state
- ungeld.lua: simple script to ungeld animals
- install-script.sh: simple shell script to install a symbolic link into the installation directory (requires this directory to be specified in `install-location.txt`)

Updated Scripts:
- dwarf-op.lua => dwopit.lua: renamed obviously, and implements migration wave selection, improves ability to select protected dwarves and overall improves the code for checking such things, a few other minor modifications such as the minimum skill level added to dwarves (0->2)
- dorf_tables.lua: modified job distribution, professions for various jobs, and skills for various professions


lethosor:
> There is a problem with the documentation in basically all the scripts, I can't figure out how to resolve said issue. "Definition list ends without a blank line; unexpected unindent."

I think you need a blank line after the  `[====[` and before the `]====]`

> * install-script.sh: simple shell script to install a symbolic link into the installation directory (requires this directory to be specified in `install-location.txt`)

I would suggest using dfhack-config/script-paths.txt to configure other places to look for scripts instead (and I realize we should document this better)

cppcooper:
> > There is a problem with the documentation in basically all the scripts, I can't figure out how to resolve said issue. "Definition list ends without a blank line; unexpected unindent."
> 
> I think you need a blank line after the `[====[` and before the `]====]`

I'll try that out.

> > * install-script.sh: simple shell script to install a symbolic link into the installation directory (requires this directory to be specified in `install-location.txt`)
> 
> I would suggest using dfhack-config/script-paths.txt to configure other places to look for scripts instead (and I realize we should document this better)

I'm concerned that this file appears to be intended for something entirely different. Perhaps more importantly the shell script merely uses `$(cat install-location.txt)/hack/scripts/$file` and scripts-paths is intended for multiple paths, there would be no way to tell what path is the correct one.

lethosor:
> I'm concerned that this file appears to be intended for something entirely different. Perhaps more importantly the shell script merely uses `$(cat install-location.txt)/hack/scripts/$file` and scripts-paths is intended for multiple paths, there would be no way to tell what path is the correct one.

Not sure what you mean. What's your shell script supposed to do? I thought it was linking one script into hack/scripts, and my suggestion is that, instead of doing that, you add the path to the folder that has your development scripts in script-paths.txt. That will make DFHack look there for scripts too, removing the need to make any symlinks.

cppcooper:
> add the path to the folder that has your development scripts in script-paths.txt. That will make DFHack look there for scripts too

Oh I see, yes. With the + preceding that would certainly remove the need for such a script. Thanks.

cppcooper:
> I think you need a blank line after the `[====[` and before the `]====]`

I did this, but the build still fails.

animalcontrol.lua
```diff
@@ -0,0 +1,184 @@
...
+DT allows you to sort by various stats and flags - this script does provide
+many options for filtering animals. Additionally you can mark animals for
+slaughter or gelding, you can even do so enmasse if you so choose.
+
+Examples:
+  [DFHack]# animalcontrol -race 170
+  [DFHack]# animalcontrol -race 170 -male -notgelded -showstats
+  [DFHack]# animalcontrol -markfor gelding -id 1988
+  [DFHack]# animalcontrol -markfor slaughter -id 1988
+  [DFHack]# animalcontrol -gelded -markedfor slaughter -unmarkfor slaughter
```
lethosor: Oh, I missed the "definition list" part of the error - Sphinx might be complaining about these examples, the list below not having a blank line before it, or something else (I don't know if the line number is relative to the start of the docs or the file). I would advise pasting these docs into an online RST editor, like http://rst.ninjs.org/, and editing until it's valid.

cppcooper:
I fixed all the documentation errors caused by the help docs.
There is a new error regarding dwarf-op which I renamed to dwopit (to make it easier to type).

>../../../dfhack/docs/_auto/news.rst:23: WARNING: undefined label: dwarf-op (if the link has no caption the label must precede a section header)

The extent of my looking into this has revealed nothing. There are no mentions of `dwarf-op` in any script files. The only mention anywhere is in a changelog in the main dfhack repo.

animalcontrol.lua
```diff
@@ -0,0 +1,195 @@
...
+dwarf fortress which prevented the geld status of animals from updating. So
+I would be queuing animals to be gelded that had already been gelded.
+
+While not as powerful as Dwarf Therapist in managing animals - in so far as
+DT allows you to sort by various stats and flags - this script does provide
+many options for filtering animals. Additionally you can mark animals for
+slaughter or gelding, you can even do so enmasse if you so choose.
+
+Examples:
+  [DFHack]# animalcontrol -race 170
```
AtomicChicken: Most DFHack users have no idea how to obtain the numerical ID of a particular race, and this varies greatly when mods come into play. If this is intended to work as a standalone script, it might be better to allow users to specify a race by its raw-defined name. The following is a modified excerpt from modtools/create-unit.lua which exemplifies how this might be done:

```lua
local raceID
for i,c in ipairs(df.global.world.raws.creatures.all) do
  if c.creature_id == args.race then
    raceID = i
    break
  end
end
if not raceID then
  qerror('Invalid race: ' .. args.race)
end
```

cppcooper: It lists all information required for further command narrowing. I can give a shot at adding that code and see if it functions as required

lethosor:
> The extent of my looking into this has revealed nothing. There are no mentions of `dwarf-op` in any script files. The only mention anywhere is in a changelog in the main dfhack repo.

Yeah, this is why it's breaking - the documentation build process needs the documentation from the DFHack repo too in order to build properly, so it's pulling from there, and the changelog file you mentioned is linking to dwarf-op. It's easy enough to fix manually, but harder to get automated tests to handle it, so don't worry if that's the only issue.

Renaming scripts can be disruptive/confusing to people already using them, though - I'd encourage you to use the `alias` command to set up a personal alias that's easier to type, if you like (this can be done in dfhack.init).

cppcooper:
> > The extent of my looking into this has revealed nothing. There are no mentions of `dwarf-op` in any script files. The only mention anywhere is in a changelog in the main dfhack repo.
> 
> Yeah, this is why it's breaking - the documentation build process needs the documentation from the DFHack repo too in order to build properly, so it's pulling from there, and the changelog file you mentioned is linking to dwarf-op. It's easy enough to fix manually, but harder to get automated tests to handle it, so don't worry if that's the only issue.
> 
> Renaming scripts can be disruptive/confusing to people already using them, though - instead, I'd encourage you to use the `alias` command to set up a personal alias that's easier to type, if you like (this can be done in dfhack.init).

I'll look into that, for now I'm just gonna leave things as they are. Perhaps I should update the PR title to "WIP - *" though? Albeit I think that is the only thing left unless I think of more improvements for devel/query

cppcooper:
I believe the docs for, basically, all these scripts need markup added to them.. on the other hand they look fine in console now. So I am reluctant to clutter that with markup formatting.

Aside from that, I believe this is ready to be merged.

cppcooper:
I would for sure use multiple PRs if that wouldn't be such a pain in the ass for me to do from the code side.

Anyways, I've rewritten devel/query from the ground up essentially. It is super maintainable now.

It also can avoid crashing on misaligned structs (ie. '-safer'), and has several other new options. Options for walking blacklisted things like scripts, to perform less intelligent checks (eg. recursive structures [ie. field is a parent]), max table lengths to be walked/depth to go to, and better debugging options as well.

I think it is done, but I'll probably find some other way to improve it.

animalcontrol.lua
```diff
@@ -0,0 +1,211 @@
...
+    end
+    for k,v in safe_pairs(Units) do
+        if v.civ_id == df.global.ui.civ_id and v.flags1.tame then
+            if not (args.male or args.female) or args.male and v.sex == 1 or args.female and v.sex == 0 then
+                if not args.race or tonumber(args.race) == v.race then
+                    if not args.markedfor or (args.markedfor == "slaughter" and v.flags2.slaughter) or (args.markedfor == "gelding" and v.flags3.marked_for_gelding) then
+                        if not args.notmarkedfor or (args.notmarkedfor == "slaughter" and not v.flags2.slaughter) or (args.notmarkedfor == "gelding" and not v.flags3.marked_for_gelding) then
+                            if not args.gelded or v.flags3.gelded then
+                                if not args.notgelded or not v.flags3.gelded then
+                                    if not args.id or tonumber(args.id) == v.id then
```
lethosor: You might want to consider moving these checks into a separate function - 8 nested conditions is rather a lot. Could be something like this (noting that my checks are inverted from yours):
```lua
function canIncludeUnit(args, v)
    if args.race and tonumber(args.race) ~= v.race then 
        return false
    elseif args.markedfor == "slaughter" and not v.flags2.slaughter then
        return false
    ....
    elseif args.gelded and not v.flags3.gelded then
        return false
    ...
    else
        return true
    end
end
```

devel/query.lua
```diff
@@ -1,38 +1,244 @@
...
+this recursively until it has walked over everything it is allowed. Everything
+it walks over it checks against any (optional) string/value queries, and if it
+finds a match it then prints it to the console.
+
+You can control most aspects of this process, the script is fairly flexible. So
+much so that you can easily create an infinitely recursing query and/or potentially
+crash dwarf fortress and dfhack. In previous iterations memory bloat was even a
+concern, where RAM would be used up in mere minutes or seconds; you can probably
+get this to happen as well if you are careless with the depth settings and don't
+print everything walked over (ie. have a search term).
```
lethosor: It may be worth pointing to `kill-lua` as a safety mechanism here

dorf_tables.lua
```diff
@@ -1,52 +1,121 @@
 -- dorf_tables has job distribution configurations, random number information for attributes generation, job configurations, profession configurations, dwarf types(ie. attributes/characteristic) configurations
--- usage: loaded by pimp-it.lua
--- by josh cooper(cppcooper) [created: 12-2017 | last edited: 12-2018]
+-- usage: loaded by dwopit.lua
```
lethosor: 3 occurrences in this file should be updated to `dwarf-op`

dorf_tables.lua
```diff
@@ -1,52 +1,121 @@
...
+-~~~~~~~~~~
+The data tables defined are described below.
+
+job_distributions:
+    Defines thresholds for each column of distributions. The columns must
+    add up to the values in the thresholds row for that column.
+    Every other row references an entry in 'jobs'
+
+jobs:
+    Defines dwopit's nameable jobs. This is preferable to taking every
```
lethosor: If you wrap the script name in backticks, it will produce a link. For example (hopefully this works):
```suggestion
    Defines `dwarf-op`'s nameable jobs. This is preferable to taking every
```

lethosor:
Looks like some of BenLubar's luacheck changes introduced some conflicts here, by the way. They look straightforward enough to resolve with the "Resolve conflicts" button here or by merging locally (i.e. they're just comment changes, as far as I can tell)

animalcontrol.lua
```diff
@@ -0,0 +1,211 @@
...
+        break
+      end
+    end
+    if not raceID then
+      qerror('Invalid race: ' .. args.race)
+    end
+    args.race = raceID
+end
+
+function safe_pairs(item, keys_only)
```
lethosor, line 98: Is this copied from another script? (It's fine if it is, I just might move it to the utils module if it's useful.)

animalcontrol.lua
```diff
@@ -0,0 +1,211 @@
...
+bvalid=(args.all and not bfilters) or (not args.all and (bfilters or bcommands))
+
+if args.help or not bvalid then
+    print(help)
+else
+    count=0
+    if args.showstats then
+        print(string.format("%-20s %-6s %-9s %-9s %-5s %-22s %-8s %-25s %-7s %-7s %-7s %-7s %-7s %-7s"
+            , "animal type", "id", "unit id", "race id", "sex", "marked for slaughter", "gelded", "marked for gelding"
+            ,"str","agi","tgh","endur","recup","disres"))
```
lethosor, line 131: I'm not sure what some of these abbreviations mean - particularly, I'm guessing "tgh" is "toughness", but I haven't been able to figure out "disres" yet. I understand why you want to keep them short, so maybe you could put explanations in the script documentation.

animalcontrol.lua
```diff
@@ -0,0 +1,211 @@
...
+                        end
+                    end
+                end
+            end
+        end
+    end
+    if not args.id then
+        if args.showstats then
+            print(string.format("%-20s %-6s %-9s %-9s %-5s %-22s %-8s %-25s %-7s %-7s %-7s %-7s %-7s %-7s"
+                , "animal type", "id", "unit id", "race id", "sex", "marked for slaughter", "gelded", "marked for gelding"
```
lethosor, line 201: It may be worth having this defined in just one place instead of repeating it here

animalcontrol.lua
lethosor: This is still an issue - the docs look like this when rendered (below). I would suggest looking at other scripts for examples of how they format their docs, such as:
* [add-recipe](https://github.com/DFHack/scripts/blob/master/add-recipe.lua) ([rendered](https://docs.dfhack.org/en/latest/docs/_auto/base.html#add-recipe))
* [modtools/create-item](https://github.com/DFHack/scripts/blob/master/modtools/create-item.lua) ([rendered](https://docs.dfhack.org/en/latest/docs/_auto/modtools.html#modtools-create-item)) (this one is just a preformatted block of text)
* [modtools/pref-edit](https://github.com/DFHack/scripts/blob/master/modtools/pref-edit.lua) ([rendered](https://docs.dfhack.org/en/latest/docs/_auto/modtools.html#modtools-pref-edit))
* [modtools/pref-skill-change](https://github.com/DFHack/scripts/blob/master/modtools/skill-change.lua) ([rendered](https://docs.dfhack.org/en/latest/docs/_auto/modtools.html#modtools-skill-change))


For reference, I'm building the docs with `docs/build.sh` in the main DFHack repo, but we have a [new page](https://docs.dfhack.org/en/latest/docs/Documentation.html) that goes into some more detail about building the docs. 

image

dig.lua
```diff
@@ -0,0 +1,50 @@
+-- digs a square channel around the tile selected (use: dig x y z, dig 3 3 3 -> 7x7x3 channel)
```
lethosor, line 1: This script actually conflicts with the `dig` plugin - although that plugin doesn't implement a `dig` command, any links in the documentation are ambiguous:
```
scripts/dig.lua:4: WARNING: duplicate label dig, other instance in dfhack/docs/Plugins.rst
```

geld.lua
```diff
@@ -0,0 +1,133 @@
...
+
+function exists(thing)
+    if thing then return true else return false end
+end
+
+function FindBodyPart(unit,newstate)
+    bfound=false
+    for i,wound in ipairs(unit.body.wounds) do
+        for j,part in ipairs(wound.parts) do
+            if type(unit.body.wounds[i].parts[j].flags2.gelded) ~= "nil" then
```
lethosor, line 67: flags2 appears to be of type wound_damage_flags2, which always has a gelded field present, so it can never be nil: https://github.com/DFHack/df-structures/blob/2d67b03a1c5c701e65fb483b9d6c1d3067f990cb/df.units.xml#L1344-L1348
I think this should just check for parts where gelded is different from `newstate` and set them to `newstate`, but I might be missing something

settile.lua
```diff
@@ -0,0 +1,18 @@
+-- sets the selected tile's 'tiletype' ie. changes the tileset you see
```
lethosor, line 1: The `tiletypes` command should be able to do this already, right?

tame.lua
```diff
@@ -21,9 +21,14 @@ Usage: ``tame -read`` OR ``tame -set <level>``
 
 local utils = require('utils')
 local selected = dfhack.gui.getSelectedUnit()
-local validArgs = utils.invert({'set','read'})
+local validArgs = utils.invert({'set','read','help'})
 local args = utils.processArgs({...}, validArgs)
 
+if args.help then
+    print(help)
+    do return end
```
lethosor, line 29: Just `return` here would be enough - `do return end` is only necessary in rare cases where you have code *between* an `return` statement and `end`

ungeld.lua
```diff
@@ -0,0 +1,70 @@
...
+if not unit then
+    qerror("Invalid unit selection.")
+end
+
+unit.flags3.gelded = false
+function FindBodyPart(unit)
+    bfound=false
+    for i,wound in ipairs(unit.body.wounds) do
+        for j,part in ipairs(wound.parts) do
+            if type(unit.body.wounds[i].parts[j].flags2.gelded) ~= "nil" then
```
lethosor, line 56: Same comment as in `geld.lua`

ungeld.lua
```diff
@@ -0,0 +1,70 @@
+-- ungelds animals
```
lethosor, line 1: This script seems to duplicate `geld.lua` quite a bit - I would suggest using `reqscript` here to import geld and run it with the ungeld option. It will be easier to maintain the scripts that way. You'll need to add module support to geld as described here for it to work: https://docs.dfhack.org/en/stable/docs/Lua%20API.html?highlight=module#scripts

animalcontrol.lua
cppcooper: Disease resistance

animalcontrol.lua
cppcooper: Probably. I am not sure if the strings are different in any way, but if they are the same you're right. I was likely just being lazy.

dig.lua
cppcooper: Indeed. I'm planning write a channel-safety plugin. Something to manage channel designations safely for the dumb dorfs.. to prevent them from digging from the bottom up - something my dwarves seem really keen on doing.

geld.lua
cppcooper: I'll have to double check. I know there were possibly a redundancy in there, but otherwise I know I was very careful to manage the geld state appropriately to avoid creating edge cases.

settile.lua
cppcooper: Probably, I found the tile commands after writing these simple ones. I think I had previously overlooked it(them?) as the script name(s) seemed unrelated to whatever I was doing at the time.

tame.lua
cppcooper: I know, just being safe; I copy paste code around a lot. Just avoiding high risk code from breaking if I move if for "reasons"

ungeld.lua
cppcooper: Good idea. I was thinking about converting them into a single plugin, but this would be less work for sure.

animalcontrol.lua
cppcooper: The safe pairs I believe is. I think I found it in a file you wrote back in 2017, I use it in a lot of the scripts I work on.

cppcooper:
I've been without internet for a bit, but should be back up in the next day or so. I can start taking a look into this stuff, but I might be busier than can be accurately anticipated for the next several weeks and on. I'll probably be able to tackle it before 2021, and hopefully slide into that channel safety plugin so I can get back to playing dwarf fortress. I've got a dozen or two layers of designated dig plans for my newest fort - no earth has been broken yet - and as soon as I get that plugin done and enabled I'm gonna unleash my dwarven minions on the plan designations.

lethosor:
All right, thanks for the update - I'm fine with addressing some of my comments here, as long as you're not also doing it at the same time. I'm also not sure when that'll happen (due to a combination of other PRs and real-life stuff) so I'll ping you here (or on IRC?) if/when I'm working on it.

cppcooper:
Pinging on here is probably better, as I am more likely to check my email than IRC.

lethosor:
@cppcooper I'm planning on getting to this this week. Do you have any unpushed changes I should be aware of?

cppcooper:
I don't think so. If I do have changes they are obscured by a boat load of file permission changes from setting up an NFS.

dig.lua
```diff
@@ -0,0 +1,50 @@
...
+--print(block.tiletype[x][y])
+--block.tiletype[x][y] = 32
+
+if width and height then
+    if depth then
+        for iz=pos.z, pos.z-depth, -1 do
+            for ix=pos.x-width, pos.x+width, 1 do
+                for iy=pos.y-height, pos.y+height, 1 do
+                    block = dfhack.maps.ensureTileBlock(ix,iy,iz)
+                    block.tiletype[ix%16][iy%16] = 32
```
lethosor, line 34: The `32`s in this script should be replaced with `df.tiletype.OpenSpace` (which you can figure out with `print(df.tiletype[32])`)

lethosor:
@cppcooper I'm planning on pushing my fixes directly to this branch (`fork-dev`) to keep them organized under this PR. Hopefully that doesn't cause any issues. In the event that it does, for reference, this branch was at 4280483be8fb425fd0b24f27866aa1952b502ef3 before I started making changes.

Also, I noticed that you added a `-list` argument to `dorf_tables`, allowing it to actually be used as a script - cool! I had planned to take this in a slightly different direction - https://github.com/DFHack/dfhack/pull/1612 adds support for putting helper files like this in the `internal` folder, which causes them to not show up in the documentation or in `ls`'s output. They can still be run from the console, but I'm thinking of moving the `-list` option to dwarf-op itself and having that call out to `internal/dorf_tables` internally, so that people can find out about the `list` option more easily.

In addition to that, I'll probably also make some other changes/fixes/etc, so be sure to update your fork with this repo's master branch once this PR is merged (otherwise we'll probably run into some pretty nasty merge conflicts later). I'll try to split up my changes into relatively-clean commits, but feel free to comment on them if you have questions about what I'm doing.

animalcontrol.lua
lethosor: I fixed this in 988d7d1 and was able to keep pretty much the same formatting in the source code. It looks like this now (rendered as HTML):

image

animalcontrol.lua
lethosor: Looks like regular pairs() or ipairs() should be fine here, since you only use it on world.units.active, which always supports those functions

animalcontrol.lua
```diff
@@ -0,0 +1,213 @@
...
+                                            elseif mark == "slaughter" then
+                                                --print("slaughter",state)
+                                                v.flags2.slaughter = state
+                                            end
+                                        end
+                                        attr = v.body.physical_attrs
+                                        if v.sex == 1 then
+                                            if args.showstats then
+                                                print(string.format("%-20s %-6s %-9d %-9d %-5s %-22s %-8s %-25s %-7d %-7d %-7d %-7d %-7d %-7d"
+                                                    ,name,v.id,k,v.race,sex
```
lethosor, line 171: By the way, the `unit id` column you're including here is actually the index of that unit in world.units.active, which is not the same as the unit ID (your `id` column)

dig.lua
lethosor: I spotted a couple more UX-related issues with this one, so I'm thinking of leaving it out for now, since they're probably non-trivial to address:
- It doesn't generate floors
- It doesn't reveal tiles that it cleared, so it's a bit hard to tell what it did if you're using it in an unrevealed area

I would definitely be interested in seeing what sort of channel-safety tool you come up with, though!

geld.lua
```diff
@@ -0,0 +1,133 @@
...
+    print(string.format("unit %s gelded.",unit.id))
+end
+
+if unit.flags3.gelded then
+    Geld(unit)
+else
+    Ungeld(unit)
+end
+
+function state(st)
```
lethosor, line 127: This function needs to be defined before other code above it uses it (unless that code moves to its own function which is called later).

This is tricky to notice because scripts' global environments persist, so if you run `geld` successfully once, `state()` will be defined for all future runs (unless you restart DF or run `devel/clear-script-env geld` like I did).

geld.lua
```diff
@@ -0,0 +1,133 @@
...
+    for i,wound in ipairs(unit.body.wounds) do
+        if wound.id == 1 and #wound.parts == 0 then
+            utils.insert_or_update(unit.body.wounds[i].parts,{ new = true, body_part_id = 1 }, 'body_part_id')
+        end
+    end
+end
+
+function Geld(unit)
+    if not FindBodyPart(unit,true) then
+        utils.insert_or_update(unit.body.wounds,{ new = true, id = 1 }, 'id')
```
lethosor, line 90: For the `id` field, you should use `unit.body.wound_next_id` (and then increment it) to ensure that you avoid ID conflicts for units with existing wounds.

geld.lua
```diff
@@ -0,0 +1,133 @@
...
+unit=nil
+
+if args.unit then
+    id=tonumber(args.unit)
+    if id then
+        for _,unit_ in pairs (df.global.world.units.active) do
+            if unit.id == id then
+                unit=unit_
+            end
+        end
```
lethosor, line 40: I'm replacing this with just `unit = df.unit.find(id)`, which is faster (since it does a binary search). Incidentally, your implementation mixed up `unit` and `unit_` on line 37

ungeld.lua
lethosor: Refactoring geld to support this looked a bit complex, so I just made ungeld use `dfhack.run_script()` instead.

cppcooper:
Cool. I was thinking I'd pull dfhack up to the front while things are quiet. I assume you're testing as you go. Is there anything left in this PR? Or should I start on my channel safety plugin?

lethosor:
I still have a couple things to work through (I haven't gotten to getwaves yet, and I'll probably also update the changelog in this PR, although I could do that after merging too).

lethosor:
Ok, I'm done with this now (i.e. I'll stop touching your branch). Thanks! I left out clear-channel and settile for now, but we can discuss what the best approach for them is.

Regarding documentation: I really appreciate all the detail you put into the docs in these scripts - they are much more detailed than some existing scripts, which is great. Formatting, however, ended up taking me a lot longer than I expected for me to fix. 

I think I might have linked to it earlier, but we have a [documentation page](https://docs.dfhack.org/en/latest/docs/Documentation.html) now that (hopefully) explains how to build the docs (I'm open to feedback, particularly when it comes to getting this working with just the scripts repo). I would appreciate it if you could build the docs locally in the future and make sure they look good. "Good" is subjective, of course, and there is a lot of variety in existing scripts' docs, but the original docs here seemed to have not been RST-formatted, which led to weird-looking HTML output: https://github.com/DFHack/scripts/pull/114#discussion_r449679118. There is value to docs that look good in the source code too, especially when you're using them for in-script help messages. I took a couple approaches to preserving your multi-column layout to an extent, while also making the HTML version more readable. There are some other scripts that use different approaches (you can take a look at https://docs.dfhack.org/en/latest/docs/_auto/base.html for some examples), so you're welcome to adopt a different approach from one of them too.

It would also be helpful if you could indicate what you'd like in the changelog. Your commit messages were pretty helpful in this regard, but there were a lot - that's definitely better than one huge commit, but I also might have missed something worth mentioning in the changelog.

I hope some of my review comments are useful for future script development - feel free to let me know if you have any questions (I wrote some of them late at night...).

cppcooper:
Thank you, I'll take a look at the documentation on the documentation. I feel so meta right now.

I presume I'll find information on how to build the docs, or where to find that html output.

PR #115, opened 2020-02-22
Add `list` option to repeat.lua

TymurGubayev:
This allows viewing what commands are set to repeat (by their names).

The idea is to help novice users (like me) to explore what a preinstalled `dfhack` (like from _Lazy Newb Pack_) is doing.

### DIFF
--- a/repeat.lua
+++ b/repeat.lua
@@ -14,6 +14,7 @@ Usage examples::
 
     repeat -name jim -time delay -timeUnits units -command [ printArgs 3 1 2 ]
     repeat -time 1 -timeUnits months -command [ multicmd cleanowned scattered x; clean all ] -name clean
+    repeat list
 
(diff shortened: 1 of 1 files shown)


repeat.lua
```diff
@@ -33,6 +34,8 @@ Arguments:
     ``...`` specifies the command to be run
 ``-cancel NAME``
     cancels the repetition with the name NAME
+``list``
```
lethosor: I'd suggest making it `-list` for consistency with the other arguments (although I'd be okay with it taking both `list` and `-list`)

TymurGubayev: I've changed it to `-list` but also slightly changed it's semantic: it doesn't make `repeat` ignore everything else anymore, instead, the list is printed after a command is added or canceled (if any).

PR #117, opened 2020-02-22
deep-embark: implemented proper console support

AtomicChicken:
- the script can now be run directly from the DFHack console at any point whilst setting up an embark
- modders should now call this from onLoad.init instead of onMapLoad.init
- a previously set depth can now be overridden by running the script from console again
- added `-clear`, which re-enables normal surface embarks
- updated documentation


clinodev:
Yay, AtomicChicken! Cheers for this!

PR #118, opened 2020-02-24
export-dt-ini: change offset name to reflect changed type

cvuchener:
https://github.com/DFHack/df-structures/pull/366 re-add creature_interaction_effect_body_transformationst.race which was removed in https://github.com/DFHack/df-structures/commit/c18cbeb28a8fb4f67e63db0547ac4b221c6903b6 but as a vector instead of an int32_t. Change the name of the offset so DT can see the changed type and adapt.


cvuchener:
I completely forgot I already did this one.

lethosor:
Heh, I was wondering why the diff was empty when I merged this (I was going top-down, so merged #140 first)

PR #119, opened 2020-02-24
once-per-save.lua: like multicmd but only if not already ran

TymurGubayev:
Runs commands the way `multicmd` does, but only unless not already ran once in current save.

Use this in `onMapLoad.ini` with f.e. `ban-cooking`

    once-per-save ban-cooking tallow; ban-cooking honey; ban-cooking oil; ban-cooking seeds; ban-cooking brew; ban-cooking fruit; ban-cooking mill; ban-cooking thread; ban-cooking milk;

Only successfully ran commands are saved.

I would love to make this per-fortress instead of per-save but have no idea how.

### DIFF
--- a/once-per-save.lua
+++ b/once-per-save.lua
@@ -21,7 +21,8 @@ Parameters:
 
 ]====]
 
-local STORAGEKEY = 'once-per-save'
+local STORAGEKEY_PREFIX = 'once-per-save'
+local storagekey = STORAGEKEY_PREFIX .. ':' .. tostring(df.global.ui.site_id)
 
 local args = {...}
 local rerun = false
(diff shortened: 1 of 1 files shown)


once-per-save.lua
```diff
@@ -0,0 +1,57 @@
...
+--rerun commands  ignore saved commands
+--reset           deleted saved commands
+
+]====]
+
+local STORAGEKEY = 'once-per-save'
+
+local args = {...}
+local rerun = false
+if args[1] == "--help" or args[1] == "-?" or args[1] == "?" then
```
lethosor: A lot of our up-to-date Lua scripts use `utils.processArgs()` - for a consistent experience, it might be better to use that.

lethosor:
Perhaps attaching the site ID (`df.global.ui.site_id`) to the storage key would work to make this per-fort, but that probably wouldn't re-trigger for reclaims.

once-per-save.lua
TymurGubayev: The problem I see with `utils.processArgs()` is we wouldn't be able to call `once-per-save ban-cooking tallow` anymore, only `once-per-save -command [ ban-cooking tallow ]`. I don't like added verbosity in this case.

lethosor: Ah, good point. I do think supporting arguments prefixed with a single `-` would be good, though. (I'm personally a fan of `--`, but I'd rather have DFHack scripts be consistent.)

TymurGubayev: added single-dash alternatives, but I don't know how do I make reStructuredText generated by sphinx behave nicely in this case, so I kept double dashes in parameter list.

TymurGubayev:
> Perhaps attaching the site ID (`df.global.ui.site_id`) to the storage key would work to make this per-fort, but that probably wouldn't re-trigger for reclaims.

did that + fortress_age to maybe somewhat help with reclaimed fortresses.

lethosor:
Nice strategy. Have you checked to see if fortress_age updates every in-game year? If so, this might re-run commands when loading a fortress after at least a year has passed.

TymurGubayev:
I checked: I have a fortress with `fortress_age` of 48720 which is just over a year (it's increased by 1 every 10 ticks and there are 403200 ticks per year).

I think I figured the unretirement out: an unretired fort will either have smaller `fortress_age` or `0`, in which case we check world date since every new game the world is advanced by two weeks. If `fortress_age` is `0` but date is bigger than previous, it's an unretired fortress.

TymurGubayev:
yeah, there is an `undefined label: on-new-fortress` now...

lethosor:
Yeah, that's not present in this PR so the link won't work. Not sure why the Github action isn't working... it should resolve itself once #121 is merged, anyway.

PR #120, opened 2020-02-25
fix adv-max-skills

kremlin-:


lethosor:
This is for 0.47, right?

kremlin-:
> This is for 0.47, right?

yeah, latest

lethosor:
Thanks! (Letting us know what build you're using, more specifically than "latest", would be helpful, but I know roughly when this changed.)

PR #121, opened 2020-02-25
onNewFortress: runs commands only in a newborn fortress

TymurGubayev:
The script only checks if `df.global.ui.fortress_age` is `0` and if it isn't, does nothing.

Use this in `onMapLoad.init` with f.e. `ban-cooking`:

    onNewFortress ban-cooking tallow; ban-cooking honey; ban-cooking oil; ban-cooking seeds; ban-cooking brew; ban-cooking fruit; ban-cooking mill; ban-cooking thread; ban-cooking milk;

This may make #119 unnecessary.

### DIFF
--- a/on-new-fortress.lua
+++ b/on-new-fortress.lua
@@ -4,9 +4,9 @@ local HELP = [====[
 
 on-new-fortress
 ===============
-Runs commands like ``multicmd``, but only in a newborn fortress.
+Runs commands like :ref:`multicmd`, but only in a newborn fortress.
 
-Use this in ``onMapLoad.init`` with f.e. ``ban-cooking``::
+Use this in ``onMapLoad.init`` with f.e. :ref:`ban-cooking`::
 
   on-new-fortress ban-cooking tallow; ban-cooking honey; ban-cooking oil; ban-cooking seeds; ban-cooking brew; ban-cooking fruit; ban-cooking mill; ban-cooking thread; ban-cooking milk;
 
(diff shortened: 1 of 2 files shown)


AtomicChicken:
I would definitely add a dfhack.world.isFortressMode() check to this, since onMapLoad.init runs in adventure mode too.

lethosor:
I'd rather have this be named `on-new-fortress` - there aren't any scripts in this repo that use camel case currently. Good idea, though! It's your choice as to what to do with #119.

TymurGubayev:
> I would definitely add a dfhack.world.isFortressMode() check to this, since onMapLoad.init runs in adventure mode too.

did that. I also loaded adventure mode for the very first time :)

>It's your choice as to what to do with #119.

I don't see how `once-per-save` is useful when there is `on-new-fortress`, which doesn't mean there actually is no use. I would keep it.

on-new-fortress.lua
```diff
@@ -4,9 +4,9 @@ local HELP = [====[
 
 on-new-fortress
 ===============
-Runs commands like ``multicmd``, but only in a newborn fortress.
+Runs commands like :ref:`multicmd`, but only in a newborn fortress.
 
-Use this in ``onMapLoad.init`` with f.e. ``ban-cooking``::
+Use this in ``onMapLoad.init`` with f.e. :ref:`ban-cooking`::
```
lethosor: You don't actually need the `:ref:` for links to other docs (in fact, I didn't know that worked). Just single backticks will produce links in RST (double backticks are code markup)

TymurGubayev: reST is new to me and so far strange.

PR #122, opened 2020-02-27
Added resurrect-adv.lua

AtomicChicken:
Written following a request on the forums. Essentially a wrapper for `full-heal.lua`, which most adventure mode players have difficulty using on their player character once dead as it's impossible to select the unit and few know how to obtain the relevant unit ID.

Alternatively, we might modify the latter script to allow it to target adventurers specifically (maybe allow users to specify 'ADVENTURER' in the `-unit` argument or add a separate arg).


AtomicChicken:
Huh, I legitimately don't recall creating that other branch. `keep_corpse` wasn't intended here, so I'd go with this one.

PR #123, opened 2020-02-27
WIP filter reports out from announcelist screen

TymurGubayev:
The script filters reports out from announcelist screen (`a`), which doesn't prevent them from
entering gamelog.txt.

Example usage: in announcements.txt add `D_D` to regular and combat conversations like this.
```
  [REGULAR_CONVERSATION:A_D:D_D]
  [CONFLICT_CONVERSATION:A_D:UCR_A:D_D]
```
Restart the game.

Now there are conversations in the reports, which is a bit too much. Luckily,
this script deals with exact that problem (`filter-reports REGULAR_CONVERSATION CONFLICT_CONVERSATION`).

And finally you would need another tool to do something about all the stuff
that's in the gamelog now. Like Soundsense or Announcement Window. But the filters are missing as of now.

Alternative is adapting [it_was_inevitable](https://github.com/BenLubar/it_was_inevitable/blob/master/game.go), maybe both.

### DIFF
--- a/filter-reports.lua
+++ b/filter-reports.lua
@@ -1,0 +1,101 @@
+-- filters reports out from announcelist screen (``a``).
+
+local HELP = [====[
+
+filter-reports
+==============
+filters reports out from announcelist screen (``a``). Doesn't prevent them from
+entering gamelog.txt.
+
+Example usage: in your announcements.txt, find following lines::
+
+  [REGULAR_CONVERSATION:A_D]
+  [CONFLICT_CONVERSATION:A_D:UCR_A]
+
+and add ``D_D``, like this::
+
+  [REGULAR_CONVERSATION:A_D:D_D]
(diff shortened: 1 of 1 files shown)


filter-reports.lua
```diff
@@ -0,0 +1,101 @@
...
+    log(id, df.announcement_type[report.type], report.text)
+
+    if not suppressing[report.type] then return end
+
+    local vec = df.report.get_vector()
+    -- TODO: loop because I don't know if current id will
+    -- always be the last one.
+    for i = #vec-1,0,-1 do
+        if vec[i].id == id then
+            -- TODO: do we need to set `world.status.display_timer = 0` here?
```
lethosor: You probably want to set display_timer to 0 when the last report is one that shouldn't be displayed.

PR #124, opened 2020-02-28
modtools/create-item: support stated params

kremlin-:
modtools/create-item specifies -matchingGloves and -matchingShoes params for creating item pairs, this commit properly support them on dfhack 0.47.02-alpha0

### DIFF
--- a/modtools/create-item.lua
+++ b/modtools/create-item.lua
@@ -44,6 +44,8 @@ local validArgs = utils.invert({
  'item',
 -- 'creature',
 -- 'caste',
+ 'leftHand',
+ 'rightHand',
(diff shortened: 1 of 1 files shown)


modtools/create-item.lua
```diff
@@ -44,8 +44,8 @@ local validArgs = utils.invert({
  'item',
 -- 'creature',
 -- 'caste',
- 'leftHand',
```
lethosor: This will break for anyone using these arguments - in the interests of compatibility, it might be best to keep them. I don't think having support for both will break anything from a quick glance through.

kremlin-: readded

PR #125, opened 2020-03-02
Added script to change heavy to light aquifers

PatrikLundell:


lightaqonly.lua
```diff
@@ -0,0 +1,34 @@
+-- Changes heavy aquifers to light pre embark
+local help = [====[
+
+lightaqonly
+======
```
lethosor: I think Sphinx is complaining about this line not matching the length of the script name
```suggestion
===========
```

PatrikLundell: Thanks. I'd never have figured out the weird doc complaints looking like doc generation refused to be started had anything to do with lines not being of the desired length!

lethosor: I usually find the GitHub output a bit easier to navigate than Travis:
image

lightaqonly.lua
```diff
@@ -0,0 +1,32 @@
...
+
+  if dfhack.isMapLoaded () then
+    qerror ("Error: This script requires a world to be loaded, but not a map.")
+  end
+
+  for i = 0, df.global.world.world_data.world_width - 1 do
+    for k = 0, df.global.world.world_data.world_height - 1 do
+      local tile = df.global.world.world_data.region_map [i]:_displace (k)
+      
+      if tile.drainage % 20 == 7 then
```
lethosor: This is some interesting math - do you know what the values of `drainage` represent? I don't think it could be an enum or a bitfield going by this. Just curious why e.g. changing `7` to `8` and `27` to `28` do the same thing.

PatrikLundell: Yes, Drainage is one of the PSV values, one that has the "normal" range of 0 - 100 (and so, essentially a percentage), which influences what the biome is. The math is straight from Toady's answer to my question in the latest FotF, which essentially is that it's a crude "RNG" to get heavy aquifers about 5 % of the time, basically because there were no spare bits available. I did check that adding 1 did not result in the crossing of a biome boundary for any of the values, and I believe I could have removed one instead.

lethosor: A link to either FOTF or the wiki to clarify where these numbers came from would be great

AtomicChicken:
This is just my two cents, but I think that the purpose of this script might not be immediately obvious to someone scrolling through the scripts list. Perhaps it could be named 'light-aquifers-only.lua' to avoid confusion?

PatrikLundell:
I'd happily rename it with underscores (I don't like dashes, as that's usually not valid names, even though it works with Lua), but there seems to be a general pathological fear of excessive typing...

lethosor:
@PatrikLundell I've fixed this, but just so you know: your last commit, 7771eab, only removed the file - it didn't add it by its new name. (`git mv` will ensure that this happens properly, by the way.)

PatrikLundell:
Thanks. I'm considering updating the script to work after embark, but then just affect that embark, so the behavior would differ depending on when it's invoked. It would be possible to have the embark invocation affect the whole world as well, but I think a local effect for a local invocation is better.
However, it requires that the name of the heavy/light aquifer flag has stabilized and been imported into the main DFHack repo.

PR #126, opened 2020-03-03
exportlegends: fix export of deities, export deities and weapons for Entity type 'militaryunit'

McArcady:
- deities are listed in historical_entity:unknown1b:deities rather than :worship
- export deities and weapons for entity type 'militaryunit' (= mercenary group)


McArcady:
merged with #137

PR #129, opened 2020-03-05
[TO BE CANCELLED] exportlegends: export profession for Entity type 'guild'

McArcady:
remark: 
This code is not very clean: I had to interpret `struct historical_entity:T_anon_1` as an array of int32_t through the pointer (lightuserdata) contained in vector `historical_entity:anon_1`. 

[EDIT]
fixed by using new `guild_professions` from in DFHack/df-structures@1cb6a65
Thanks guys for the feedback -

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -350,6 +350,11 @@ function export_more_legends_xml()
                 end
             end
         end
+        if entityV.type == df.historical_entity_type.Guild then -- Get profession
+           for professionK,professionV in pairs(entityV.guild_professions) do
+              file:write("\t\t<profession>"..df_enums.profession[professionV.profession]:lower().."</profession>\n")
+           end
+        end
         for id, link in pairs(entityV.entity_links) do
             file:write("\t\t<entity_link>\n")
                 for k, v in pairs(link) do
--- a/add-recipe.lua
+++ b/add-recipe.lua
@@ -36,7 +36,7 @@ end
   --1: the currently known recipes to the player civ
   --2: the native (non-exotic) recipes to the player civ
   --3: all possible recipes including exotic
-local categories = {
+local categories = { --as:{1:'number[]',2:'number[]',3:'df.itemdef[]'}[]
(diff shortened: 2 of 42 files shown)


BenLubar:
I'd recommend changing df-structures so you don't have to do that madness.

exportlegends.lua
```diff
@@ -346,6 +346,13 @@ function export_more_legends_xml()
                 end
             end
         end
+        if entityV.type == df.historical_entity_type.Guild then -- Get profession
+           for anonK,anonV in pairs(entityV.anon_1) do
```
lethosor: Agreed - anon names are considered unstable (i.e. the numeric suffixes could change at any time) and should not be used in scripts. Have you figured out what this array contains?

lethosor: Turns out that this has since broken - as of https://github.com/DFHack/df-structures/commit/2f93b7188d37be889bb5dc83a758dd13b72d6353, anon_1 through anon_6 point to [these vectors](https://github.com/DFHack/df-structures/blob/2f93b7188d37be889bb5dc83a758dd13b72d6353/df.entities.xml#L662-L667), and `guild_professions` was identified in https://github.com/DFHack/df-structures/commit/1cb6a65f788e0c52674130685a12976f39979905

McArcady: fixed by using new vector 'guild_professions' from df.entities.xml

lethosor:
https://github.com/DFHack/scripts/pull/129/commits/63b2ea8f8aae85d870d1a0c3b7e6601742b8d659 is the only commit you care about, right? Looks like you rebased or something and added around 35 other commits that are duplicates of ones already present in this repo. (Do you remember what you did? I can try to help you back out of it)

McArcady:
Sorry for that! I rebased and push -f like a troll, that was a mistake.
Maybe it's simpler if we use this new PR #133.
Then we cancel this one if its ok with you

McArcady:
merged with other PR's in #137

lethosor:
Yup, #137 looks good to me, thanks

PR #130, opened 2020-03-14
Update advfort.lua to new version of df

warmist:
Update and fix small issue.

### DIFF
--- a/gui/advfort.lua
+++ b/gui/advfort.lua
@@ -280,7 +280,7 @@ function addJobAction(job,unit) --what about job2?
     --local pos=copyall(job.pos)
     unit.path.dest:assign(pos)
     --job
-    local data={type=df.unit_action_type.Job,data={Job={x=pos.x,y=pos.y,z=pos.z,timer=10}}}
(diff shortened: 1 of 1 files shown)


gui/advfort.lua
```diff
@@ -280,7 +280,7 @@ function addJobAction(job,unit) --what about job2?
     --local pos=copyall(job.pos)
     unit.path.dest:assign(pos)
     --job
-    local data={type=df.unit_action_type.Job,data={job={x=pos.x,y=pos.y,z=pos.z,timer=10}}}
+    local data={type=df.unit_action_type.Job,data={Job={x=pos.x,y=pos.y,z=pos.z,timer=10}}}
```
lethosor: Mentioned this in https://github.com/DFHack/dfhack/issues/1520 - I think it would be better for us to change the name in structures back instead of fixing the scripts that use this.

lethosor: As of https://github.com/DFHack/df-structures/commit/37415475903f0d524d19c50b4be057a481554af7, this part can be changed back

PR #131, opened 2020-03-17
handled new compounds, warning on legends compound elements

PatrikLundell:
Probably only dependent on what's accepted in df-structures, but cannot rule out that something is in the pull request.

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -566,7 +566,7 @@ function export_more_legends_xml()
                         df.history_event_entity_dissolvedst:is_instance(event) or
                         df.history_event_item_stolenst:is_instance(event)) and k == "circumstance" then
                     file:write("\t\t<circumstance>\n")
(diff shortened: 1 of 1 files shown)


exportlegends.lua
```diff
@@ -746,6 +790,8 @@ function export_more_legends_xml()
                     file:write("\t\t<"..k..">"..df_enums.profession[v]:lower().."</"..k..">\n")
                 elseif df.history_event_change_creature_typest:is_instance(event) and (k == "old_race" or k == "new_race")  and v >= 0 then
                     file:write("\t\t<"..k..">"..escape_xml(df.global.world.raws.creatures.all[v].name[0]).."</"..k..">\n")
+                elseif tostring(v):find ("<") then
```
lethosor: This is a good check! I think it might be better to log to the console instead of writing this to the file, though - people might not notice if it's only written to the file. (Maybe once for each unique value of k?)

PR #132, opened 2020-03-20
exportlegends: fix export of <reason> and <circumstance>

McArcady:
historical_event type `item_stolen`:
fix export of `<reason>` and `<circumstance>` following their change of type in DFHack/df-structures@3834b0a9


McArcady:
already fixed by 2b4b9ae7

PR #134, opened 2020-03-22
exportlegends: export new fields for regions: evilness and force_id

McArcady:
evilness is good/evil (default=neutral)
force_id is hfid of the associated 'force'


PatrikLundell:
If you're going to export evilness, then I'd include the Neutral case as well, causing the tag to always be present. Note, though, that spreading evil causes the base region evilness to be less important than it was.

Also, rather than creating new pull requests for each change, I'd update the pending one, allowing me to use a single branch for exportlegends. If the previous pull request has been accepted, I'd just create a new one for the new update. Note that the experts may have a different opinion, in which case you should listen to them...

McArcady:
merged with #137

PR #138, opened 2020-03-28
added export of vague relationship data.

PatrikLundell:
NOTE: This script relies on mapping made in currently pending DFHack/df-structures#386 being available.

The indication that a DFHack beta may be imminent resulted in a desire to ensure these additions have a chance to be considered, even though the structures work hasn't been processed yet.

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -811,6 +811,29 @@ function export_more_legends_xml()
         end
     end
     file:write("</historical_events>\n")
+    file:write("<historical_event_relationships>\n")
+    for ID, set in progress_ipairs(df.global.world.history.relationship_events, 'relationship_event') do
+        for k = 0, set.next_element - 1 do
+            file:write("\t<historical_event_relationship>\n")
+            file:write("\t\t<event>"..set.event[k].."</event>\n")
+            file:write("\t\t<relationship>"..set.relationship[k].."</relationship>\n")
+            file:write("\t\t<source_hf>"..set.source_hf[k].."</source_hf>\n")
+            file:write("\t\t<target_hf>"..set.target_hf[k].."</target_hf>\n")
+            file:write("\t\t<year>"..set.year[k].."</year>\n")
+            file:write("\t</historical_event_relationship>\n")
+        end
+    end
+    file:write("</historical_event_relationships>\n")
+    file:write("<historical_event_relationship_supplements>\n")
+    for ID, event in progress_ipairs(df.global.world.history.relationship_event_supplements, 'relationship_event_supplement') do
+        file:write("\t<historical_event_relationship_supplement>\n")
+        file:write("\t\t<event>"..event.event.."</event>\n")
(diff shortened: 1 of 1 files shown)


exportlegends.lua
```diff
@@ -811,6 +811,29 @@ function export_more_legends_xml()
...
+        for k = 0, set.next_element - 1 do
+            file:write("\t<historical_event_relationship>\n")
+            file:write("\t\t<event>"..set.event[k].."</event>\n")
+            file:write("\t\t<relationship>"..set.relationship[k].."</relationship>\n")
+            file:write("\t\t<source_hf>"..set.source_hf[k].."</source_hf>\n")
+            file:write("\t\t<target_hf>"..set.target_hf[k].."</target_hf>\n")
+            file:write("\t\t<year>"..set.year[k].."</year>\n")
+            file:write("\t</historical_event_relationship>\n")
+        end
+    end
```
lethosor, line 825: This section increases the size of the generated XML file by around 25% on a medium-sized world for me. Out of curiosity, is this new information in 0.47? Have tools been looking to add this? Is there any way to reduce/avoid the inner loop (probably unlikely)?

I'm fine with adding it, especially if it's useful to legends viewers - I'd just prefer not to get into the habit of making large additions like this often, since disk space is a concern for some. (Now that I think about it, maybe allowing partial exports is a more future-proof option, but that would be a project for another PR.)

PatrikLundell: In the cases I've seen, the Plus file is a few hundred MB, while the vanilla file weighs in at the GB scale.

Yes, it's new to 0.47. Vague relationships are war buddy, childhood friend, etc. (plus some not so vague things like ex lovers and lieutenants, apparently). The info would allow Viewers to provide info that DF's Legends Mode provides, but which currently isn't available externally, namely the year of the establishment and the circumstance under which it is formed in some cases (scholarly artistic), such as e.g. "due to a shared interest in Geology during a seminar in Seminartown in 273". Tools haven't requested it, at least partially because the identification of the data was made very recently, and probably isn't known yet by the tool maintainers.
The one thing that could be done would come at an execution cost, namely cull all the entries where both parties are culled histfigs by checking if both fail to be found.

The data is organized in chunks (the outer loop) in which relationships are stored (each entry in the inner loop). The chunks don't serve any logical purpose, but "only" manage memory allocation. Technically all the entries could be stored in a single vector (with a struct, rather than parallel arrays: I've no idea why Toady likes those, but I'm sure he's got a reason).

lethosor: Wow, I haven't seen a DF file that big yet, but I heard Toady added stuff to the vanilla export in 0.47.
Thanks for the info + updates - I'm going to merge, and hopefully get a beta out the door

PatrikLundell: My latest 33*129 400 year old PSV world weighs in at over 4 GB, which is about twice that of what others generated from that PSV were (all for 0.47.04), so yes, they're large.

PR #139, opened 2020-03-29
Added reveal-hidden-units.lua

AtomicChicken:
A simple script which exposes all sneaking units. This is primarily intended to help adventure mode players deal with [an annoying bug in the new version](http://www.bay12games.com/dwarves/mantisbt/view.php?id=11299), but it's also useful for uncovering ambushers in fortress mode.

### DIFF
--- a/reveal-hidden-units.lua
+++ b/reveal-hidden-units.lua
@@ -1,0 +1,30 @@
+-- Expose all sneaking units.
+-- author: Atomic Chicken
+
+--[====[
+
+reveal-hidden-units
+===================
+This script exposes all units on the map who
+are currently sneaking or waiting in ambush
+and thus hidden from the player's view.
+
+Usable in both fortress and adventure mode.
+
+]====]
+
+local count = 0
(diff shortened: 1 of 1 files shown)


reveal-hidden-units.lua
```diff
@@ -0,0 +1,30 @@
...
+
+local count = 0
+for _, unit in ipairs(df.global.world.units.active) do
+  if unit.flags1.hidden_in_ambush then
+    unit.flags1.hidden_in_ambush = false
+    count = count + 1
+  end
+end
+
+if dfhack.is_interactive() then
```
lethosor: Are you trying to suppress output when this is invoked from an init script or something?

AtomicChicken: More specifically, I was considering the possibility of somebody deciding to use this with the repeat script (as the referenced bug occurs quite frequently in combat) and added the check to prevent it from spamming the console in such a case.

lethosor: Makes sense!

PR #141, opened 2020-04-04
fixes exportlegends maps error 1540

thurin:
fixes https://github.com/DFHack/dfhack/issues/1540


lethosor:
I forgot to check/mention this earlier - I've added a changelog specific to the scripts repo, so in the future I'd like to have PRs update that too. You can see my update here: bd7cc53
In this case, because the error was only present in pre-release builds, I've included the `@` to prevent it from showing up in the r1 changelog when that's released. More details here: http://docs.dfhack.org/en/stable/docs/Compile.html#building-the-changelogs

PR #142, opened 2020-04-08
cleaned up interaction string export

PatrikLundell:
fix DFHack/dfhack#1534

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -741,10 +741,17 @@ function export_more_legends_xml()
                         file:write("\t\t<caste>"..(df.global.world.raws.creatures.all[event.race].caste[v].caste_id):lower().."</caste>\n")
                     end
                 elseif k == "interaction" and df.history_event_hf_does_interactionst:is_instance(event) then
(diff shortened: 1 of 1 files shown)


exportlegends.lua
```diff
@@ -741,10 +741,17 @@ function export_more_legends_xml()
                         file:write("\t\t<caste>"..(df.global.world.raws.creatures.all[event.race].caste[v].caste_id):lower().."</caste>\n")
                     end
                 elseif k == "interaction" and df.history_event_hf_does_interactionst:is_instance(event) then
-                    file:write("\t\t<interaction_action>"..df.global.world.raws.interactions[v].str[3].value.."</interaction_action>\n")
-                    file:write("\t\t<interaction_string>"..df.global.world.raws.interactions[v].str[4].value.."</interaction_string>\n")
+                    if #df.global.world.raws.interactions[v].sources > 0 then
```
lethosor, line 744: To answer your question from https://github.com/DFHack/dfhack/issues/1534

> Does DF use a canonical order of tokens internally? If not, any user generated raws may result in the wrong token being exported.

This is just the order that lines are found in the raw files, so yeah, this was very fragile before. Thanks for fixing!

PR #143, opened 2020-04-09
Add ref-target support to gui/gm-editor

lethosor:
Depends on dfhack/dfhack#1543

This currently replaces the current `i` key functionality with a call to `push(ref_target_type.find(id))` if the ref-target type is available; otherwise, the current functionality with a dropdown allowing a type to be selected is preserved. I'm debating whether to allow something like `shift-i` to continue to show the dropdown in all cases, in case users really want to override the ref-target or something.

Screenshot: 
image

### DIFF
--- a/gui/gm-editor.lua
+++ b/gui/gm-editor.lua
@@ -5,15 +5,15 @@
 gui/gm-editor
 =============
 This editor allows to change and modify almost anything in df. Press :kbd:`?` for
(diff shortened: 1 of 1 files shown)


lethosor:
I did it anyway :)
(but I also added a bit of detail that should help explain why looking up e.g. a `caste_raw` doesn't work yet)

PR #144, opened 2020-04-15
adapted to changed structures

PatrikLundell:
https://github.com/DFHack/df-structures/pull/389


PatrikLundell:
Testing this provides an indirect test that the structures changes work...

lethosor:
In case you're interested, I realized after merging this that the enum attribute from dfhack/df-structures#389 can be used in place of the conditionals now, so I did that in 068ae13427c7bad0c084c21a1ac83eec91395efa

PatrikLundell:
Yes, I realized that as well, and had planned to do that in an hour or so. Thanks for being faster!

PR #145, opened 2020-04-16
modtools/create-unit: added -equip

AtomicChicken:
Improvement:
- added `-equip`, which allows items to be created and automatically equipped onto units as part of the spawning process.

This lets arena mode logic handle everything. Item sizes, which body part a particular item should be placed on, the handedness of gloves/footwear, the way clothes are layered relative to one another, assessing whether or not a unit is physically able to equip a particular item, etc. As such, this approach has clear advtanges over `modtools/create-item` + `modtools/equip-item`, at least with respect to equipping newly spawned units. Arena mode does however come with some limitations at present, such as an inability to set item quality or to place arrows into quivers.

I tried to make this work in a manner that should be easy for raw modders to understand, but the resultant code isn't particularly elegant. A review would be useful.

### DIFF
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -987,7 +987,7 @@ local validArgs = utils.invert({
   'duration',
   'locationRange',
   'locationType',
-  'equip'
+  'equip',
 })
 
 if moduleMode then
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -69,10 +69,32 @@ Creates a unit.  Usage::
         set the birth date of the unit by current age
         chosen randomly if this argument is omitted
 
+    -equip [ ITEM:MATERIAL:QUANTITY ... ]
+        The specified items will be created and equipped onto the unit
+            via the same logic used in arena mode. Such items will
+            therefore be sized to match the unit and equipped in a
+            manner that the game considers to be appropriate.
+        Note that this currently comes with some limitations,
+            such as arrows not being placed in quivers
+            and an inability to specify item quality.
+        Item quantity defaults to 1 if omitted.
+        When spaces are included in the item or material name,
+            the entire item description should be enclosed in
+            quotation marks. This can also be done to increase
+            legibility when specifying multiple items.
+        examples:
+            -equip [ RING:CREATURE:DWARF:BONE:3 ]
+                3 dwarf bone rings
+            -equip [ ITEM_WEAPON_PICK:INORGANIC:IRON ]
+                1 iron pick
+            -equip [ "ITEM_SHIELD_BUCKLER:PLANT:OAK:WOOD" "AMULET:AMBER" ]
+                1 oaken buckler and 1 amber amulet
+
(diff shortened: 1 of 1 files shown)


modtools/create-unit.lua
```diff
@@ -631,6 +682,88 @@ function nameUnit(unit, entityRawName)
...
+        if dfhack.items.getSubtypeDef(i, subType).id == itemTypeStr then
+          return i, subType
+        end
+      end
+    end
+  end
+  qerror("Invalid item type: " .. itemTypeStr)
+end
+
+function extractEquipmentDetails(equipmentStr)
```
lethosor, line 706: Is this the part you were talking about potentially cleaning up?
If anything, I would try to be consistent with createitem and modtools/create-item (it looks like the latter uses `dfhack.items.createItem()`). If there's a way to make certain items more convenient for people familiar with the raws, we ought to do it everywhere.

modtools/create-unit.lua
```diff
@@ -843,7 +986,8 @@ local validArgs = utils.invert({
   'quantity',
   'duration',
   'locationRange',
-  'locationType'
+  'locationType',
+  'equip'
```
lethosor, line 990: Not important, but using trailing commas (in this case, after `'equip'`) can clean up the git history a bit. As an example, because of the comma you added to the `locationType` line, someone looking to find when that was added with `git blame` would end up here first.
(Admittedly, we don't really have any style guidelines when it comes to this yet, but I thought I'd mention it.)

modtools/create-unit.lua
AtomicChicken: Yep, this is the part I was referring to. The code worked well enough during testing; I was just wondering whether there might be a better way to do this. `dfhack.items.createItem()` in itself isn't particularly useful here as the arena handles item creation for us; we just need to provide the desired item type, material and quantity. `modtools/create-item` obtains these from the user through multiple separate args, but it was written with the purpose of creating a single item per run. Here, I thought it would make more sense for everything to be specified in a single arg to make it easier to create units equipped with several items, such as an entire set of armour, as otherwise the modder would have to jump back and forth between the different args (making editing a nightmare). I tried to make the input of this new arg resemble the format used in raws, but I wouldn't recommend doing this everywhere as extracting the individual components from the input string adds an extra layer of complexity.

modtools/create-unit.lua
AtomicChicken: Makes sense; I'll edit this one and keep it in mind for any future scripts. I'm surprised I hadn't seen this mentioned before; most of the scripts I've come across in this repo seem to lack trailing commas in this context.

modtools/create-unit.lua
lethosor: Yeah, I guess I was thinking of the type/material parsing from the other plugins, not using `createItem()` itself. Makes sense why you'd want a single argument per item here, and I don't really have a good suggestion for reducing complexity. Feel free to merge when you think it's ready.

PR #146, opened 2020-04-20
handled single caste creatures

PatrikLundell:
Fix DFHack/dfhack#1550

### DIFF
--- a/catsplosion.lua
+++ b/catsplosion.lua
@@ -78,7 +78,7 @@ for id in pairs(creatures) do
             preg.colors:assign(female.appearance.genes.colors)
             female.pregnancy_genes = preg
             female.pregnancy_timer = math.random(1, 100)
-            female.pregnancy_caste = 1
+            female.pregnancy_caste = unit.caste  -- To handle female only species
             total_created = total_created + 1
         end
     end
--- a/catsplosion.lua
+++ b/catsplosion.lua
@@ -78,7 +78,7 @@ for id in pairs(creatures) do
             preg.colors:assign(female.appearance.genes.colors)
             female.pregnancy_genes = preg
             female.pregnancy_timer = math.random(1, 100)
-            female.pregnancy_caste = unit.caste  -- To handle female only species
+            female.pregnancy_caste = female.caste  -- To handle female only species
             total_created = total_created + 1
         end
     end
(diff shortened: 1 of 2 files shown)


catsplosion.lua
```diff
@@ -78,7 +78,7 @@ for id in pairs(creatures) do
             preg.colors:assign(female.appearance.genes.colors)
             female.pregnancy_genes = preg
             female.pregnancy_timer = math.random(1, 100)
-            female.pregnancy_caste = 1
+            female.pregnancy_caste = unit.caste  -- To handle female only species
```
lethosor, line 81: Pretty sure this should be female.caste

lethosor: Fixed

PatrikLundell:
You're right. It should have been female.caste. What puzzles me is that it did actually work when I tested it with a spawned harpy, so there must have been a global "unit" female variable hanging around from some poorly written script in my environment at the time.

lethosor:
Globals are isolated per-script, so I doubt that was it. Were you testing in the Lua interpreter? That automatically defines `unit` as the currently-selected unit, if one is selected.

PatrikLundell:
Yes, it was from the interpreter, and the unit was selected.

lethosor:
Ah. Maybe we should improve the docs and/or the banner shown when the Lua interpreter starts, then

PatrikLundell:
No, sorry. It was from the console, but not the interpreter, but the unit was selected.

lethosor:
Oh, like running `lua unit.pregnancy_caste = foo` from the console? Yeah, that wouldn't give you a banner. Docs could stand to be improved anyway if this isn't mentioned, though.

PatrikLundell:
No, I tested the catsposion script itself from the console.

lethosor:
Weird, so did I.
The only other explanation I can think of: since global environments persist per-script, maybe at one point in your testing you had defined a global `unit` in catsplosion.lua (which can be easy to do by leaving out `local` in an assignment). That would still be defined in future runs of the script until you quit DF (or clear its globals with devel/clear-script-env).
At any rate, it works now, but if you run into weirdness with globals affecting other scripts, let us know.

PatrikLundell:
Nope, the only line I changed was the one I botched up... I cannot rule out having run something else, though.

PR #148, opened 2020-04-24
Removed unneeded tabs in xml

ralpha:
The claims-tag closes on the same line.

Examples from before change:
```
<claims>                </claims>
```
or
```
<claims>20,7|20,9|21,7|21,9|21,10|           </claims>
```


lethosor:
For reference, some Git commands display your commit message on a single line because it just has a single newline between the first and second lines. A double newline (i.e. one blank line after the first line) would fix that. Not a big deal, just a heads-up.

PR #150, opened 2020-04-25
Make <item_subtype> optional instead of -1 when no subtype

ralpha:
Before:
```
<item_type>weapon</item_type>
<item_subtype>short sword</item_subtype>
...
<item_type>earring</item_type>
<item_subtype>-1</item_subtype>
```

With update:
```
<item_type>weapon</item_type>
<item_subtype>short sword</item_subtype>
...
<item_type>earring</item_type>
```


ralpha:
Yeah when I was looking closer at some parts of the code it seems there are more events that use the "-1" as a value that is not known.
This is not how it should be I think, but it depends on the backwards compatibility that is wanted.
Other "-1" values exist like: `file:write("\t\t<position>-1</position>\n")`
but others display noting if the value is "-1"

PR #152, opened 2020-04-26
Exports legends into folders

ralpha:
Fixes https://github.com/DFHack/dfhack/issues/1416
If no folder name given, default is of similar pattern to `region1-00125-01-01`

`exportlegends all` will use default path
`exportlegends all myfolder_name` will create and use the 
"myfolder_name" folder
`exportlegends all .` will not put the files in a folder. (tested on 
linux)

This works for all arguments.

Tested on Linux and works fine.
**TODO: Test this on windows systems.** It should work but it should also be tested.

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -1114,17 +1114,15 @@ end
 function create_folder(folder_name)
     if folder_name == "-00000-01-01" then
         qerror('"'..folder_name..'" is the default foldername, this folder will not be created as you are probably not in the legends screen.')
-        return false
     end
     -- check if it is a file, not a folder
     if dfhack.filesystem.isfile(folder_name) then
-      qerror(folder_name..' is a file, not a folder')
-      return false
+        qerror(folder_name..' is a file, not a folder')
     end
     if dfhack.filesystem.exists(folder_name) then
-      return true
+        return true
     else
-      return dfhack.filesystem.mkdir(folder_name)
+        return dfhack.filesystem.mkdir(folder_name)
     end
 end
 
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -144,10 +144,14 @@ setmetatable(df_enums, {
 function export_more_legends_xml()
     local problem_elements = {}
 
+    -- Move into the save folder
+    if not move_to_save_folder() then
+        qerror('Could not move into the save folder.')
+    end
     local filename = df.global.world.cur_savegame.save_dir.."-"..get_world_date_str().."-legends_plus.xml"
     local file = io.open(filename, 'w')
+    move_back_to_main_folder()
     if not file then
-      move_back_to_main_folder()
       qerror("could not open file: " .. filename)
     end
 
@@ -1000,10 +1004,15 @@ end
 
 -- export information and XML ('p, x')
 function export_legends_info()
+    -- Move into the save folder
+    if not move_to_save_folder() then
+        qerror('Could not move into the save folder.')
+    end
     print('    Exporting:  World map/gen info')
     gui.simulateInput(vs, 'LEGENDS_EXPORT_MAP')
     print('    Exporting:  Legends xml')
     gui.simulateInput(vs, 'LEGENDS_EXPORT_XML')
+    move_back_to_main_folder() -- Move back out of the save folder
     print("    Exporting:  Extra legends_plus xml")
     export_more_legends_xml()
 end
@@ -1015,7 +1024,6 @@ function export_detailed_maps()
             if not vs then
                 local legends_vs = dfhack.gui.getViewscreenByType(df.viewscreen_legendsst, 0)
                 if not legends_vs then
-                    move_back_to_main_folder()
                     qerror("Could not find legends screen")
                 end
 
@@ -1024,24 +1032,23 @@ function export_detailed_maps()
 
             vs = dfhack.gui.getViewscreenByType(df.viewscreen_export_graphical_mapst, 0)
             if not vs then
-                move_back_to_main_folder()
                 qerror("Could not find map export screen")
             end
             
             vs.sel_type = i - 1
-            print('    Exporting map: ' .. MAPS[i])
+            -- Move into the save folder
             if not move_to_save_folder() then
-                qerror('The '..folder_name..' folder was created successfully. But could not move into the folder')
+                qerror('Could not move into the save folder.')
             end
+            print('    Exporting map ' ..i.. '/' ..#MAPS..': '.. MAPS[i])
             gui.simulateInput(vs, 'SELECT')
             while dfhack.gui.getCurViewscreen() == vs do
                 script.sleep(10, 'frames')
             end
-            move_back_to_main_folder()
+            move_back_to_main_folder() -- Move back out of the save folder
         end
-        print("    Done exporting")
+        print("    Done exporting.")
     end)
-
 end
 
 -- export site maps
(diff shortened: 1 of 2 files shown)


neo6633:
Im loll in discord.

I just replaced the lua file to your new one.

\PeridexisErrant's Starter Pack 0.47.04-r03\Dwarf Fortress 0.47.04\hack\scripts\exportlegends.lua

replaced to your new version. 

https://raw.githubusercontent.com/ralpha/scripts/e02cb4beb8b2fd52c4569a4063f9df879dd9b513/exportlegends.lua

copied here.


and I runed "exportlegends info" 

I think new folder appeared. But there is error message in DFhack command line.

here is "https://pastebin.com/a80SijiG"

Actually im not a coder but just let you know.

OS : Windows 10

lethosor:
Your DFHack version is too old - identity_type was added after 0.47.04-beta1 was released. You could upgrade DFHack manually, or wait for a new pack to include a 0.47.04-r1 (which should be out soon).

ralpha:
@neo6633 You could try with `exportlegends maps` or `exportlegends sites` these don't export the legends_plus.xml and will thus should not give you any errors.

ralpha:
Got conformation from @neo6633 that it worked fine on Windows after updating DFHack.

ralpha:
Updated the script to be more robust when changing folders.
This is makes the chance of something going wrong because of an error
less likely to effect DFHack.

This is also ready to be merged.

Tested it on: 
* Linux 64bit with dfhack latest version compiled from source.
* Windows 10 64bit fresh install of DF 0.47.04 + DFHack 0.47.04-r1 (download from release page, precompiled) (tested in VM)

exportlegends.lua
```diff
@@ -47,6 +47,33 @@ local MAPS = {
...
+function get_world_date_str()
+    local month = dfhack.world.ReadCurrentMonth() + 1 --days and months are 1-indexed
+    local day = dfhack.world.ReadCurrentDay()
+    local year_str = string.format('%0'..math.max(5, string.len(''..df.global.cur_year))..'d', df.global.cur_year)
+    local date_str = year_str..string.format('-%02d-%02d', month, day)
+    return date_str
+end
+
+-- set current directory as main folder
+local main_folder = dfhack.filesystem.getcwd()
```
lethosor, line 60: I actually ran into a 9-year-old DFHack core bug when I was testing this - `dfhack.getDFPath()` is supposed to return the path to the DF root, but on Linux it was essentially returning `getcwd()` instead. With the [fix](https://github.com/DFHack/dfhack/commit/56e43a0dde023c5a4595a22b29d800153b31e3c4), I would use getDFPath instead, since that's really what you care about.

exportlegends.lua
```diff
@@ -988,19 +1023,31 @@ function export_detailed_maps()
...
+            -- Move into the save folder
+            if not move_to_save_folder() then
+                qerror('Could not move into the save folder.')
+            end
+            print('    Exporting map ' ..i.. '/' ..#MAPS..': '.. MAPS[i])
             gui.simulateInput(vs, 'SELECT')
             while dfhack.gui.getCurViewscreen() == vs do
                 script.sleep(10, 'frames')
             end
+            move_back_to_main_folder() -- Move back out of the save folder
```
lethosor, line 1048: Another thing that would be useful here, from the [Lua API docs](https://docs.dfhack.org/en/0.47.04-r1/docs/Lua%20API.html#locking-and-finalization), is `dfhack.with_finalizer`. That will allow you to specify a callback and a cleanup function (`move_back_to_main_folder`) that will *always* get called, even if the callback throws an error. I think that would be a good idea here because DF will probably break down the line if the working directory isn't restored properly. (Not that the script is throwing any errors currently, but if it ever does in the future, we could run into issues.)

In this case, you'd probably want to wrap the entire anonymous function passed to script.start:

```lua
script.start(dfhack.with_finalizer(move_back_to_main_folder, function()
  ...
end))
```

(`dfhack.with_finalizer(move_back_to_main_folder, export_detailed_maps)` from the caller would *not* work, since export_detailed_maps is asynchronous)

exportlegends.lua
ralpha: Yeah already had the code in there but did not seem to work, but did not know it was an issue.

exportlegends.lua
ralpha: Making it future proof is always a good plan.
Changed it, had to make it a bit bigger, but added comments to make everything clear.

ralpha:
I resolved the changes you asked for.
I also added some more documentation and added a catch so it does not create a folder when no world is loaded.

exportlegends.lua
```diff
@@ -1076,8 +1110,12 @@ function export_site_maps()
     end
 end
 
--- check if a folder with this name could be created or already exists
+-- Check if a folder with this name could be created or already exists
 function create_folder(folder_name)
+    if folder_name == "-00000-01-01" then
+        qerror('"'..folder_name..'" is the default foldername, this folder will not be created as you are probably not in the legends screen.')
+        return false
```
lethosor, line 1117: `qerror()` raises an error (similar to `error()`), so you don't really need the `return false` here (but it doesn't hurt anything).

exportlegends.lua
```diff
@@ -18,15 +18,30 @@ Options:
 :maps:   Exports all seventeen detailed maps
 :all:    Equivalent to calling all of the above, in that order
 
+Argument:
+:folder_name: (optional) The name of the folder where all the files will be saved.
```
lethosor, line 22: Turns out this doesn't render well (Sphinx doesn't warn about it, but it looks like this):
image

I can fix it, but we also have a new guide up for building the docs, which I'd recommend for testing changes locally: https://docs.dfhack.org/en/latest/docs/Documentation.html

exportlegends.lua
```diff
@@ -47,6 +77,31 @@ local MAPS = {
     "Diplomacy",
 }
 
+-- Get that date of the world as a string
+-- Format: "YYYYY-MM-DD"
+function get_world_date_str()
+    local month = dfhack.world.ReadCurrentMonth() + 1 --days and months are 1-indexed
+    local day = dfhack.world.ReadCurrentDay()
+    local year_str = string.format('%0'..math.max(5, string.len(''..df.global.cur_year))..'d', df.global.cur_year)
```
lethosor, line 85: One last bit of advice, in case it's useful in the future: the `5` in `%05d` is a minimum length, not a minimum and maximum, so you can avoid calculating the length of cur_year here:
```lua
> string.format('%05d', 1)
00001
> string.format('%05d', 125)
00125
> string.format('%05d', 123456)
123456
> string.format('%05d', 1234567890)
1234567890
```

ralpha: Yeah good point, this was just moved from the original code and did not change it. I though there was some reason behind it and did not look much more into it.
Original location: https://github.com/DFHack/scripts/blob/0.47.04-r1/exportlegends.lua#L118-L121
Looks much better in your changes below.

lethosor: Oops, evidently I didn't look at the full diff. Makes sense!

exportlegends.lua
ralpha: Thanks for the link. Next time I add some documentation I'll take a look.

ralpha:
@lethosor Thanks for all the help with getting this into DFHack. (and all the other very nice work you are doing for DF Hack)
This is a very welcome feature especially since this is useful for the [DF Storyteller](https://dfstoryteller.com/) application I'm currently finishing (will be out in a week or 2). (if you want to test it out let me know ;) )
This is also the reason why I made some other changes to the `exportlegends.lua` file in the past and will probably do in the future.

PR #153, opened 2020-04-27
removed some 'null' entries, converted enum numbers to strings

PatrikLundell:
Fixes DFHack/dfhack#1554.
Fixes DFHack/dfhack#1557

It turned out "df_enums" can't handle enum types whose definitions are buried inside structs, so that case was handled separately.

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -157,7 +157,9 @@ function export_more_legends_xml()
         file:write("\t\t<name>"..escape_xml(dfhack.df2utf(dfhack.TranslateName(mountainV.name,1))).."</name>\n")
         file:write("\t\t<coords>"..mountainV.pos.x..","..mountainV.pos.y.."</coords>\n")
         file:write("\t\t<height>"..mountainV.height.."</height>\n")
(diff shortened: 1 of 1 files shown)


PatrikLundell:
Fix DFHack/dfhack#1557

lethosor:
According to https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword, it needs to be in the PR description (i.e. the first post in the thread), and I suspect the word "issue" in the middle also won't work.
I have a "linked issues" option at the side but it only picks up on issues in this repo... I'll add it to the first post and hope that works.

PatrikLundell:
Taking a cue from Toady's XML, I've changed a number of boolean tags into valueless tags that are present only when true. "is_volcano" for mountain peaks is a good case for this use, and cutting down the size of the creature flag output to only mention the flags that are set also seemed like a good idea (and it's also more similar to the raw format).

Note that you probably can't use this as a blanket approach, as some booleans probably should be explicit regardless of their value.

lethosor:
I'm a little concerned that the flag changes might break existing tools, but they make sense, and the space savings for creature flags could be significant. I'll put a note in the changelog.

PR #154, opened 2020-05-11
Added an argument to remove-stress.lua to set stress to value

Bumber64:
Added argument -value [ integer ] to set stress to this value (if current stress is higher) instead of -1,000,000.

### DIFF
--- a/remove-stress.lua
+++ b/remove-stress.lua
@@ -8,8 +8,8 @@ remove-stress
 Sets stress to -1,000,000; the normal range is 0 to 500,000 with very stable or
 very stressed dwarves taking on negative or greater values respectively.
 Applies to the selected unit, or use ``remove-stress -all`` to apply to all units.
-Using the argument ``-value [ 0 ]`` will reduce stress to the value 0 instead of -1,000,000.
-
+Using the argument ``-value 0`` will reduce stress to the value 0 instead of -1,000,000.
+Negative values must be preceded by a backslash (\): ``-value \-10000``
 ]====]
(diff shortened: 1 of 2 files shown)


remove-stress.lua
```diff
@@ -8,42 +8,55 @@ remove-stress
...
     local args = utils.processArgs({...}, validArgs)
+    local stress_value = -1000000
 
     if args.help then
         print(help)
         return
     end
+    
+    if args.value then
+        if args.value[1] then
```
lethosor, line 45: Instead of making `-value` take an array containing a single item, I would just make it take a number (`-value x`) and check for `if tonumber(args.value) ~= nil` here

Bumber64: How do I handle negative values? Using `-value -1` throws an error for an invalid argument.

lethosor: [set-belief](https://github.com/DFHack/scripts/blob/47d9b3c9390d21ecc49ddaba477b16071884ee87/modtools/set-belief.lua#L44) (which uses processArgs) mentions escaping negative values with `\`, so I assume that would work (but it might actually need to be `\\` - I haven't tested)

lethosor:
FYI, this repo has a changelog now - in the future, it would be great if you could update that too. I realize it wasn't exactly explained before, but hopefully https://docs.dfhack.org/en/latest/docs/Contributing.html and/or https://docs.dfhack.org/en/latest/docs/Documentation.html will go into more detail about it soon.

PR #155, opened 2020-05-18
Fixes moddable-gods up

Putnam3145:
Removed unused-but-nevertheless-required domain and description arguments, added support for neuter gender, removed a line that just wasn't doing anything anymore, made "depictedAs" actually do something

EDIT: May be a crash upon use, don't merge just yet.

### DIFF
--- a/modtools/moddable-gods.lua
+++ b/modtools/moddable-gods.lua
@@ -66,6 +66,7 @@ local race
 for k,v in ipairs(df.global.world.raws.creatures.all) do
     if v.creature_id == args.depictedAs or v.name[0] == args.depictedAs then
         race = k
+        break
     end
 end
 
--- a/modtools/moddable-gods.lua
+++ b/modtools/moddable-gods.lua
@@ -104,4 +104,4 @@ for _,sphere in ipairs(args.spheres) do
 end
 df.global.world.history.figures:insert('#',godFig)
 
-print(args.name.first_name .. " created as historical figure " .. tostring(godFig.id))
+print(godFig.name.first_name .. " created as historical figure " .. tostring(godFig.id))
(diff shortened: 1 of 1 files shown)


modtools/moddable-gods.lua
```diff
@@ -90,13 +94,14 @@ godFig.id = df.global.hist_figure_next_id
 df.global.hist_figure_next_id = 1+df.global.hist_figure_next_id
 godFig.info = df.historical_figure_info:new()
 godFig.info.spheres = {new=true}
-godFig.info.secret = df.historical_figure_info.T_secret:new()
-
+godFig.info.known_info = {new=true}
```
lethosor, line 97: I would keep this as this, actually (I think it's clearer and more consistent with the rest of the script) - or is there a difference I'm missing?
```suggestion
godFig.info.known_info = df.historical_figure_info.T_known_info:new()
```

modtools/moddable-gods.lua
```diff
@@ -62,10 +56,20 @@ if args.gender == 'male' then
  gender = 1
 elseif args.gender == 'female' then
  gender = 0
+elseif args.gender == "neuter" then
+ gender = -1
```
lethosor, line 60: I realize this is unrelated, but if these are caste raw indices, you might want to check the caste raws to make sure you're getting the gender right. I'm not sure if there's an easy way to do that, though.

Putnam3145: Nope, these are directly setting sex--caste is always set to "0", which is *usually* the FEMALE caste on most creatures, but I don't think it'll cause any odd behavior

lethosor: Gotcha. We really should make that a proper enum at some point (or if it is already, a lot of occurrences don't use it).

PR #157, opened 2020-06-19
Changed width so a four-letter TZ name can fit

brickviking:
I noticed that whenever I selected a game to load, the timestamp had one letter truncated; in my case the box looked much like this:
DwarfLoadCutout

The timestamp is meant to be NZST, not NZS.

### DIFF
--- a/gui/load-screen.lua
+++ b/gui/load-screen.lua
@@ -369,7 +369,7 @@ function load_screen:onDismiss()
 end
 
 load_screen_options = gui.FramedScreen{
-    frame_width = 40,
+    frame_width = 42,
     frame_height = 6,
     frame_title = "",
     frame_inset = 1,


ab9rf:
Hm. This raises a question for me of how long a timezone descriptor can be. There are several abbreviations that are five characters long (e.g. AZOST for Azores Summer Time) and it's also possible that the %z descriptor will return the time zone _name_ (such as "America/Chicago") instead of an abbreviation.

lethosor:
Bumping up the max width by 2 (as this PR does) should work for 5-character descriptors too. This script [uses `%c`](https://github.com/DFHack/scripts/blob/47d9b3c9390d21ecc49ddaba477b16071884ee87/gui/load-screen.lua#L481), though, which is very implementation-defined (it's supposed to use the user's locale settings and so on). It would probably be best to resize the dialog automatically, although I don't think our widget system supports that cleanly yet.

brickviking:
There's a further problem with the locale en_US; a timestamp from "date +%c" results in the following:
`bash$ LC_ALL=en_US date +%c`
`Sun 21 Jun 2020 04:42:25 PM NZST (33 chars with terminating LF, 34 with CRLF)`
`123456789012345678901234567890123`

lethosor:
You're right, the timezone isn't particularly important, but when I originally wrote this, I was trying to avoid making decisions about what date format to use, and `%c` seems to be the only specifier Lua supports that uses a system-defined date format. I'm a little wary of overriding it by setting LC_ALL (which could affect other things in DF/DFHack as well).

brickviking:
I did some informal testing on my Fedora machine, so I don't know how well this will translate into the Windows world. Checking what %Z says in the case of every major timezone in /usr/share/zoneinfo/ seems to indicate that you'll either get a 3–5-letter abbreviation EST/AEST/AZOST or a 5-character tz-offset i.e. +0545 for Asia/Kathmandu. I wasn't able to duplicate the Azores output for summer time, though I suspect that could be due to lack of understanding.

lethosor:
Thanks! It turned out that this screen already resized itself based on the length of the save folder's name, so I just extended that logic to also cover date strings in 49ef34f33bd713bcbbfd71753afd341f73796bfa.

PR #159, opened 2020-07-19
quickfort script skeleton

myk002:
DFHack/dfhack#499 High-level structure, help text, and input validation.

Each PR for this script should be release-ready in that functionality will have been tested and unimplemented features will display explanatory error messages instead of just erroring out.

I have PRs queued up for settings management, file listing, tokenization, and dig blueprint functionality. I'll submit PRs one a time so review feedback can be focused and so no two open PRs depend on one another (github doesn't really support a chain of PRs that build on one another -- the diffs will be against the target repo instead of the parent PR)

### DIFF
--- a/quickfort.lua
+++ b/quickfort.lua
@@ -99,9 +99,20 @@ files are described in the files themselves.
 
 -- only initialize our globals once
 if not initialized then
-initialized = true
+
+local utils = require('utils')
+local quickfort_common = require('hack.scripts.internal.quickfort.common')
+local quickfort_dig = require('hack.scripts.internal.quickfort.dig')
+local quickfort_build = require('hack.scripts.internal.quickfort.build')
+local quickfort_place = require('hack.scripts.internal.quickfort.place')
+local quickfort_query = require('hack.scripts.internal.quickfort.query')
 
 local function do_reset()
+    reload('hack.scripts.internal.quickfort.common')
+    reload('hack.scripts.internal.quickfort.dig')
+    reload('hack.scripts.internal.quickfort.build')
+    reload('hack.scripts.internal.quickfort.place')
+    reload('hack.scripts.internal.quickfort.query')
     initialized = false
 end
 
@@ -145,25 +156,18 @@ For more info, see: https://docs.dfhack.org/en/stable/docs/_auto/base.html#quick
 ]]
 end
 
-local settings = {
-    blueprints_dir = 'blueprints',
-    force_marker_mode = false,
-    force_interactive_build = false,
-}
-
 local function do_set(args)
     if #args == 0 then
         print('active settings:')
-        printall(settings)
+        printall(quickfort_common.settings)
         return
     end
     if #args ~= 2 then
         error('error: expected "quickfort set [<key> <value>]"')
     end
-    -- TODO: validate and set settings
+    print('NOT YET IMPLEMENTED')
+    print(string.format('would set %s to "%s"', args[1], tostring(args[2])))
 end
-
-local utils = require('utils')
 
 local valid_list_args = utils.invert({
     'l',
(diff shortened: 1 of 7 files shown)


quickfort.lua
```diff
@@ -87,5 +100,94 @@ files are described in the files themselves.
...
+end
+
+end -- if not initialized
+
+-- main
+local argv = {...}
+local action = table.remove(argv, 1)
 
-print('Given arguments: ',...)
+if action == 'set' then
```
warmist, line 219: Not sure what is consensus on big if/elseif (i.e. switch statements) but another way to do this is a table with functions
```lua
switch={ list=do_list, run=do_command, orders=do_command}
switch[action](argv)
```

myk002: done. thanks!

myk002:
travis docs check should pass after DFHack/dfhack#1612 is merged

lethosor:
Actually, the approach I adopted in https://github.com/DFHack/scripts/commit/ad67c79f8074a491c144144b2825cd05c80a2236 is what scripts should use for their own modules now. Among other things, it respects [script paths](https://github.com/DFHack/dfhack/blob/develop/dfhack-config/script-paths.txt), while the mkmodule approach could result in the main script being loaded from another folder but the modules being loaded from hack/scripts/internal (I ran into this issue yesterday when updating advfort).

Not sure why the line numbers in the comments on https://github.com/DFHack/scripts/pull/159/commits/5177a127b32631fbcedc4b0e93e0c5c05a124388 are all wrong... I'll try to track that down. (FYI, we've switched from Travis to GitHub Actions, but the `travis` folder name is still used because several things depend on it.)

myk002:
done. i'm liking reqscript better than require -- it doesn't need an explicit reload

lethosor:
Sure enough - I was about to comment that I didn't think it would handle reloading cleanly, but it turns out that there's a [check](https://github.com/DFHack/dfhack/blob/0.47.04-r1/library/lua/dfhack.lua#L543) for the file's modification time that I forgot about. Glad it works!

I'll merge the DFHack PR and see if that helps with the CI output here.

myk002:
thanks! test are passing now.

myk002:
Would it be possible to merge this PR? I'm worried about the PRs getting too far behind where I'm actually writing code -- suggested changes may end up causing merge conflicts.

lethosor:
My only hesitation is that I'm wrapping up 0.47.04-r2 (hopefully by around the end of this week, if all goes well), and I wouldn't want to include *just* this PR in a new release, since it looks like it'll mostly print "not implemented yet" messages. That's fine, and I definitely don't want to set a deadline for what you're doing, but I also don't want to confuse people by having this end up in a release in its current state. Some possibilities include:

* We could merge this as-is, then deal with leaving it out of r2 later if the timing doesn't work out. Since some people read the docs, it would basically involve removing the script and changelog entry, then adding them back, which may end up creating merge conflicts with future changes.
* We could merge this into a separate `quickfort` branch - I feel like I suggested this for another PR that I can't find now, but it might be a better fit for this PR, given that it depends more directly on future work that you have planned.
* We could make a separate release branch (git-flow-style), although I don't think we've done that before in this repo, and it could get messy if a lot of other pre-release changes need to be merged in (there are some other PRs in this repo targeting r2).
* We could wait for your next PR before merging this one. If you base it on this one (without rebasing), there shouldn't be merge conflicts, although it would be confusing to have an open PR be a subset of another one.

Thoughts? I know at least part of this complexity is due to our release process being messy; I'm fine with merging this and figuring out how best to handle it when a release rolls around (i.e. option 1) if you would prefer that.

myk002:
My concern is just around not having the PRs lag too far behind the code. I've got 500+ lines for dig and another ~500 for place, not to mention the code for file input and tokenization. I'm starting to worry that if I get significant comments on those PRs, I'll have to rework a lot of code. I'm not very concerned about getting the script into the current release (interested beta testers can always pull from my repo), so if that's the blocker, I'm totally fine with a quickfort branch.

lethosor:
Got it, I'll set up a quickfort branch. We can discuss whether the version that's there is release-ready when a release rolls around.

PR #161, opened 2020-07-22
advfort: add auto/manual civ selection

warmist:
This is intended for better user experience as sometimes civ can have no recipes/iron/etc...
Fixes https://github.com/DFHack/dfhack/issues/1604


gui/advfort.lua
```diff
@@ -159,29 +160,45 @@ function deon_filter(name,type_id,subtype_id,custom_id, parent)
...
+            settings.unsafe=true
+        elseif v=="-s" or v=="--safe" then
+            settings.safe=true
+        elseif v=="-i" or v=="--inventory" then
+            settings.check_inv=true
+            settings.df_assign=false
+        elseif v=="-a" or v=="--nodfassign" then
+            settings.df_assign=false
+        elseif v=="-e" or v=="--entity" then
+            if #args>i and string.sub(args[i+1],1,1)~="-" then
```
lethosor, line 184: We do have `string.startswith` (custom to DFHack), by the way

PR #162, opened 2020-07-24
Add tools to add-recipe, some error message improvements

ElMendukol:
Right now add-recipe only lets you add recipes for weapons or armor, but not tools. I see no reason why this script shouldn't let you add recipes for tools, so I did just that.

It's particularly useful when you're playing with a modded civ, and then you find out only after starting a new fortress that you don't have stuff like minecarts or stepladders when you want them. It can also let you add recipes for procedurally-generated instruments.

Plus, I took the chance to improve the error messages a bit. Instead of letting the script throw normal LUA error messages when you give it bad arguments, I made it print how to use the script. Or a short error message when you get the name of an item wrong, so you at least know the script ran.

Also, I'm a noob at this, so sorry in advance if I screwed up something or if I didn't follow some guidelines I didn't know about

### DIFF
--- a/add-recipe.lua
+++ b/add-recipe.lua
@@ -149,7 +149,6 @@ function addSingleItem(itemstring)
   
   if (itemstring ~= nil) then
     itemType, itemId = string.match(itemstring, "(.*):(.*)")
-	if(verbose) then print("Item to add: " .. itemType .. ":" .. itemId) end
   else
     print("Usage: add-recipe single <item name> | Example: add-recipe single SHOES:ITEM_SHOES_BOOTS")
     return end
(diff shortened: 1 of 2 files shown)


ElMendukol:
Ah yes, I just had to forget something 😅

Also, I see that you had to adjust the style. I couldn't find a style guide for LUA anywhere in the docs, so I tried to follow what I saw in the scripts more or less. What style guidelines should I follow?

lethosor:
The only Lua bit I intentionally changed was moving `end` to its own line, which is what I'm used to seeing in scripts. (My editor stripped out some whitespace, but https://github.com/DFHack/scripts/commit/f028d38e4ee9d60b9295a07d23ddb8dcc5197699?w=1 or `git show -w` locally will hide those changes). The code block change was just something I thought looked a bit more consistent with other docs ([here's](https://docs.dfhack.org/en/latest/docs/_auto/base.html#add-recipe) how it looks now).

Other than that, we don't have a set style guide for Lua. https://docs.dfhack.org/en/latest/docs/Contributing.html has a couple points, some for C++, and https://github.com/DFHack/dfhack/pull/1538 is some work towards a standardized style for C++, but I'm not aware of a good Lua style checker (although I haven't looked very hard).

PR #163, opened 2020-07-29
Quickfort set

myk002:
DFHack/dfhack#499 Implement the quickfort set command and reading from the dfhack-config/quickfort/quickfort.txt config file

### DIFF
--- a/quickfort.lua
+++ b/quickfort.lua
@@ -147,6 +147,26 @@ local settings = {
     force_interactive_build = false,
 }
 
+local function set_setting(key, value)
+    if settings[key] == nil then
+        error(string.format('error: invalid setting: "%s"', key))
+    end
+    val = value
+    if type(settings[key]) == 'boolean' then
+        val = value == 'true'
+    end
+    settings[key] = val
+end
+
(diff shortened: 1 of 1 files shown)


quickfort.lua
```diff
@@ -151,6 +151,26 @@ For more info, see: https://docs.dfhack.org/en/stable/docs/_auto/base.html#quick
 ]]
 end
 
+local function set_setting(key, value)
+    if quickfort_common.settings[key] == nil then
+        error(string.format('error: invalid setting: "%s"', key))
```
lethosor, line 156: You could probably use [qerror](https://docs.dfhack.org/en/stable/docs/Lua%20API.html#exception-handling) for user-facing error messages like these

myk002:
yes, just setting them in memory. I note in the help text that changes made with quickfort set only apply to the current session and are not persistent.

quickfort.lua
myk002: thanks! done.

lethosor:
The help text was already present from an older PR, so I missed it :)

PR #164, opened 2020-07-29
Quickfort list

myk002:
DFHack/dfhack#499 Implement quickfort list functionality, including .csv file tokenization (so I can read the modeline in the first cell)

### DIFF
--- a/quickfort.lua
+++ b/quickfort.lua
@@ -395,7 +395,7 @@ local function do_command(in_args)
         end
         blueprint_file = blueprint_files[list_num]
         if not blueprint_file then
-            error(string.format('invalid list index: %d', list_num))
+            qerror(string.format('invalid list index: %d', list_num))
         end
         filename = blueprint_files[list_num].path
     end
--- a/quickfort.lua
+++ b/quickfort.lua
@@ -99,7 +99,8 @@ files are described in the files themselves.
 
 -- only initialize our globals once
 if not initialized then
-initialized = true
+
+local utils = require('utils')
 
 local function do_reset()
     initialized = false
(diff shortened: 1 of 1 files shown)


quickfort.lua
```diff
@@ -185,6 +185,157 @@ local function do_set(args)
...
+                 string.find(v.path, '[.]xlsx$')) then
+            if string.find(v.path, '[.]xlsx$') then
+                print(string.format(
+                        'skipping "%s": .xlsx files not supported yet', v.path))
+                goto skip
+            end
+            local modeline = scan_blueprint(v.path)
+            if not modeline then
+                print(string.format(
+                        'skipping "%s": no #mode marker detected', v.path))
```
lethosor, line 318: Do you have plans to support these later? I happened to have a very bare-bones CSV file I was testing digfort with yesterday, just containing a couple rows and columns:
```
d,h,d
h,d,h
d,h,d
```
I don't know how widespread files like these are, if at all (and I'm fine with not supporting them)

myk002: Could you test to see if it works in the original quickfort windows app? I *could* support it, and just assume that no header means #dig -- I don't *think* that's something that will bite us later..

At some point I'm thinking of supporting multiple "sheets" in the same .csv file, which would give them a lot of the advantages that .xlsx files have. each "sheet" would be separated by a modeline, but I don't think supporting missing initial modelines would interfere with that -- it's just subsequent "sheets" that would need one.

I'll code it up

myk002: we risk polluting the 'quickfort list' output with unrelated files, but how likely is it that users are storing unrelated .csv or .xlsx files in the blueprints folder?

I guess a greater risk is when we support .xlsx, users can't have non-blueprint sheets in their spreadsheet files. I myself usually keep a "Notes" sheet in my spreadsheet files. we could only support headerless .csv files and not headerless .xlsx sheets, but that might be confusing for users who convert their working .csv files into non-working .xlsx files

lethosor: I don't have ready access to Windows, sorry. All I know is that digfort accepts it. I suspect that people wouldn't have non-blueprints in the `blueprints` folder, particularly since it's new (to DFHack).

Multiple blueprints in a single file isn't something I had considered... if that makes it complicated to handle this, I'm fine with putting this off.

myk002: so I guess to prevent potential confusion later, I think I'd rather not implement modeless blueprints -- unless, of course, there is a clear signal from the community that this is wanted. I'm not seeing a lot of activity in the thread I started, so I'm not sure it's even worth asking now, but I expect more activity once this script gets into a release and starts seeing some real usage.

lethosor: Sounds good to me!

PR #167, opened 2020-07-31
Quickfort dig

myk002:
DFHack/dfhack#499 Implement dig mode logic for quickfort

Enforces the same restrictions and behavior as performing the actions from the UI


internal/quickfort/dig.lua
```diff
@@ -1,26 +1,569 @@
...
+local function is_hard(tileattrs)
+    return hard_natural_materials[tileattrs.material]
+end
+
+local function is_smooth(tileattrs)
+    return tileattrs.special == df.tiletype_special.SMOOTH
+end
+
+local function get_engraving(pos)
+    -- scan through engravings until we find the one at this pos
```
lethosor, line 75: I double-checked the engravings module; turns out that it's very old, not exposed to Lua, and also doesn't have something to do this. The buildings module does something similar with `findAtTile()`, but it's faster (since it's written in C++) and caches its results. You could try caching here, although I suspect most forts won't have *too* many engravings.

internal/quickfort/dig.lua
```diff
@@ -1,26 +1,569 @@
...
+    if ctx.flags.hidden then return false end
+    ctx.flags.traffic = values.traffic_low
+end
+
+local function do_traffic_restricted(ctx)
+    if ctx.flags.hidden then return false end
+    ctx.flags.traffic = values.traffic_restricted
+end
+
+-- add on the 'd' prefix to avoid spelling out reserved words (like 'or')
```
lethosor, line 443: Lua has an alternative syntax for using arbitrary expressions as table keys: `{['do'] = 2}`
(You would also need to index with `table['do']` instead of `table.do`, but you're doing that already)

internal/quickfort/dig.lua
```diff
@@ -1,26 +1,569 @@
...
+end
+
+local function clear_designation(flags, occupancy)
+    flags.dig = df.tile_dig_designation.No
+    flags.smooth = 0
+    occupancy.carve_track_north = 0
+    occupancy.carve_track_east = 0
+    occupancy.carve_track_south = 0
+    occupancy.carve_track_west = 0
+end
```
lethosor, line 101: I was thinking of suggesting that these be added to the maps module, similar to the existing `getTileDesignation()` and `getTileOccupancy()`. Those both take coordinates, though, although I think your context currently keeps track of that too.

Syncing scripts with core library changes is tricky, of course. Maybe once this is done, we can take a pass through and identify things that would be useful in the core library/modules.

internal/quickfort/dig.lua
myk002: agree -- I'll just add a comment that it could be made faster with caching and see if anybody complains about the slowness

internal/quickfort/dig.lua
myk002: Thanks! I should have remembered that from the Lua tutorial. done.

internal/quickfort/dig.lua
myk002: agreed. I'll add a comment here to remind us to do that

PR #171, opened 2020-08-03
Quickfort query

myk002:
DFHack/dfhack#499 Implement query mode logic for the quickfort script

I have a couple questions for this PR:

1) I use a "wiggle the cursor" technique to get the UI to recognize the building underneath the cursor after I set the cursor position. This feels hacky, though. Is there a better way to do this? I've tried:
- DwarfOverlay:simulateCursorMovement
- DwarfOverlay:selectBuilding
- setting the ui.main.mode to something else then back to QueryBuilding
but nothing else seems to work.

2) Is there currently a way to sleep in a Lua script? I could wire it up through LuaApi if not.

After this PR, the quickfort script can be a reasonable replacement for python quickfort. There are still some advanced features I have yet to support, like construction material labels and blueprint transformations, but all the basic functionality is there. I'm really surprised how productive I can be in Lua!

I need to write some more documentation, especially around file formats. Should I just keep on adding help text to the main quickfort.lua script, or is there a better place to put it?

### DIFF
--- a/internal/quickfort/query.lua
+++ b/internal/quickfort/query.lua
@@ -30,18 +30,7 @@ end
 
 local function move_cursor(screen, overlay, pos)
     overlay:moveCursorTo(pos)
-    -- wiggle the cursor so the building under the cursor gets properly selected
-    -- is there a better way to do this? I've tried:
-    -- - DwarfOverlay:simulateCursorMovement
-    -- - DwarfOverlay:selectBuilding
-    -- - setting the ui.main.mode to something else then back to QueryBuilding
-    if pos.y > 0 then
-        gui.simulateInput(screen, 'CURSOR_UP')
-        gui.simulateInput(screen, 'CURSOR_DOWN')
-    else
-        gui.simulateInput(screen, 'CURSOR_DOWN')
-        gui.simulateInput(screen, 'CURSOR_UP')
-    end
+    dfhack.gui.refreshSidebar()
 end
 
 local function handle_modifiers(token, modifiers)
@@ -53,7 +42,7 @@ local function handle_modifiers(token, modifiers)
         return true
     end
     if token_lower == '{wait}' then
-        print('{wait} not yet implemented')
+        print('{Wait} not yet implemented')
         return true
     end
     return false
--- a/internal/quickfort/query.lua
+++ b/internal/quickfort/query.lua
@@ -5,12 +5,86 @@ if not dfhack_flags.module then
     qerror('this script cannot be called directly')
 end
 
+local gui = require('gui')
+local guidm = require('gui.dwarfmode')
 local quickfort_common = reqscript('internal/quickfort/common')
 local log = quickfort_common.log
+local quickfort_aliases = reqscript('internal/quickfort/aliases')
+local quickfort_keycodes = reqscript('internal/quickfort/keycodes')
+
+local common_aliases_filename = 'hack/data/quickfort/aliases-common.txt'
+local user_aliases_filename = 'dfhack-config/quickfort/aliases.txt'
+
+local function move_cursor(screen, overlay, pos)
+    overlay:moveCursorTo(pos)
+    -- wiggle the cursor so the building under the cursor gets properly selected
+    -- is there a better way to do this? I've tried:
+    -- - DwarfOverlay:simulateCursorMovement
+    -- - DwarfOverlay:selectBuilding
+    -- - setting the ui.main.mode to something else then back to QueryBuilding
+    if pos.y > 0 then
+        gui.simulateInput(screen, 'CURSOR_UP')
+        gui.simulateInput(screen, 'CURSOR_DOWN')
+    else
+        gui.simulateInput(screen, 'CURSOR_DOWN')
+        gui.simulateInput(screen, 'CURSOR_UP')
+    end
+end
 
 function do_run(zlevel, grid)
-    local stats = nil
-    print('"quickfort run" not yet implemented for mode: query')
+    local stats = {
+        keystrokes={label='Keystrokes sent', value=0, always=true},
+        tiles={label='Settings modified', value=0},
+    }
+
+    quickfort_aliases.reset_aliases()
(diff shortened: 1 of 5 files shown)


internal/quickfort/query.lua
```diff
@@ -5,12 +5,117 @@ if not dfhack_flags.module then
...
+    return not flags.hidden and occupancy.building ~= 0
+end
+
+local function move_cursor(screen, overlay, pos)
+    overlay:moveCursorTo(pos)
+    -- wiggle the cursor so the building under the cursor gets properly selected
+    -- is there a better way to do this? I've tried:
+    -- - DwarfOverlay:simulateCursorMovement
+    -- - DwarfOverlay:selectBuilding
+    -- - setting the ui.main.mode to something else then back to QueryBuilding
```
lethosor, line 37: `dfhack.gui.refreshSidebar()` does pretty much exactly this (it uses the z axis, for no real reason other than the plugins that needed this before it was a library function did it that way).

internal/quickfort/query.lua
```diff
@@ -5,12 +5,117 @@ if not dfhack_flags.module then
...
+local function handle_modifiers(token, modifiers)
+    local token_lower = token:lower()
+    if token_lower == '{shift}' or
+            token_lower == '{ctrl}' or
+            token_lower == '{alt}' then
+        modifiers[token_lower] = true
+        return true
+    end
+    if token_lower == '{wait}' then
+        print('{wait} not yet implemented')
```
lethosor, line 56: Out of curiosity, what does this do? If it's a delay to allow DF to catch up, hopefully that won't be necessary with the script.

lethosor:
> 2. Is there currently a way to sleep in a Lua script? I could wire it up through LuaApi if not.

Depends on what you're trying to do. If you're trying to wait for the game to update, a raw `sleep()` call from Lua will freeze the game until it returns, because all Lua code (well, when invoked as a conventional script) suspends the DF game thread when it runs. There are a couple alternatives:
* If you want to run a callback after a certain amount of time, an `onupdate` hook could work. There's a [dfhack.timeout()](https://docs.dfhack.org/en/stable/docs/Lua%20API.html#core-interpreter-context) function available to Lua that's a bit more convenient.
* If you want a function to wait for a certain amount of time, then resume execution, coroutines could be an option. The `gui.script` module has some higher-level wrappers that things like `gui/create-item` use. However, doing anything significant may result in a confusing experience for console users, because the script will essentially be running in the background once you call yield(), so the console will accept commands again (and there's not an easy way to tell when the script is "done"). It's not terrible, but it tends to be something that GUI scripts use more than console-based scripts.

> I need to write some more documentation, especially around file formats. Should I just keep on adding help text to the main quickfort.lua script, or is there a better place to put it?

The main quickfort.lua documentation should be fine for now. Sub-sections are fine. I'd encourage you to build the docs locally if you haven't yet and you're making complicated changes, just to make sure they look okay. https://docs.dfhack.org/en/latest/docs/Documentation.html has some more detail about this.

internal/quickfort/query.lua
lethosor: Evidently I didn't read your comment very thoroughly before looking at the code. See below for ideas about what you could do for sleeping.

myk002: I'm actually not sure what the original intent was in python quickfort. It could be that -- allowing the UI to catch up, and I agree that is completely unnecessary now. The other, potentially still useful function is to allow a user to see what the current screen looks like at that point in the keystroke playback for debugging reasons. I don't know if anyone actually uses this, but I'm aiming for python quickfort compatibility so unless there's a reason *not* to do it, I plan to support it.

I'm thinking there is a greater opportunity to implement a "debug mode" that allows a user to set breakpoints, poke around the UI, then resume (or halt) playback. but that's for a future PR : )

internal/quickfort/query.lua
myk002: looking at the library code, that's hilarious we came up with the same "solution" : )  changed to use library version.

myk002:
freezing the UI might be sufficient, but if running a lua script suspends the game thread, am I right in thinking that the UI won't even process the simulateInput keystrokes until the script returns? If that's the case, I might leave {Wait} unimplemented until (and if) I get around to writing a blueprint debugger.

myk002:
Ok, I'll add everything to quickfort.lua. If the docs get too unwieldy (I am somewhat worried about quickfort documentation eventually dominating the scripts page), we can discuss moving them elsewhere.

lethosor:
> freezing the UI might be sufficient, but if running a lua script suspends the game thread, am I right in thinking that the UI won't even process the simulateInput keystrokes until the script returns? If that's the case, I might leave {Wait} unimplemented until (and if) I get around to writing a blueprint debugger.

simulateInput will work; users just won't see updates until DFHack returns control to DF.

The part of the DF main loop that deals with screens does roughly these things (not necessarily in this order):
* Checks for any outstanding keypresses, and calls viewscreen::feed() on the active screen with the keys if necessary
* Calls viewscreen::logic() on the active screen
* Calls viewscreen::render() if enough game ticks have passed (at least part of the rendering happens in another thread, but it's synchronized with the game thread to keep render() from accessing data in an inconsistent state)
* Calls SDL_NumJoysticks(), which DFHack intercepts and delays if any code (e.g. Lua scripts) that needs to access game data is waiting to run

You can call whatever viewscreen methods you like when DF is suspended, but other things that happen in the main loop (like checking for keypresses and updating the screen) won't happen until your code has returned control back to DF.

There is an edge case when dismissing screens (I think) where DF waits a tick before actually removing the screen from the stack (so you'll see a black frame sometimes). getCurViewscreen() might ignore this, but you might run into issues when dismissing screens and then interacting with their parents. Otherwise, most screen manipulation should work as you'd expect.

myk002:
if DF doesn't update the screen, then I think sleeping for {Wait} wouldn't serve its intended purpose. I'll leave it as a TODO for now, then. I'll come back to it if there is demand (I have no idea if anyone actually uses {Wait} -- in python quickfort, it never worked on anything but Windows anyway).

PR #172, opened 2020-08-03
implement stockpile setting init with query mode

myk002:
DFHack/dfhack#499

Once I implemented query mode in #171, I realized that I didn't have to refactor the stockpiles plugin in order to expose the logic to iterate through stockpile settings -- I could just generate a query blueprint and initialize the stockpiles through the UI!

Although initializing in C++ would execute faster and would avoid screen movement as we move the cursor around, there is a high risk that I would break something in the stockpiles plugin if I were to reorganize its thousands of lines of code. It may be worth it someday to do it "properly" in C++, but with query mode, that day doesn't have to be today.

### DIFF
--- a/internal/quickfort/place.lua
+++ b/internal/quickfort/place.lua
@@ -17,6 +17,7 @@ end
 require('dfhack.buildings') -- loads additional functions into dfhack.buildings
 local quickfort_common = reqscript('internal/quickfort/common')
 local quickfort_building = reqscript('internal/quickfort/building')
+local quickfort_query = reqscript('internal/quickfort/query')
 local log = quickfort_common.log
 
 local function is_valid_stockpile_tile(pos)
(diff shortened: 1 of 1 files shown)


internal/quickfort/place.lua
```diff
@@ -45,23 +46,23 @@ local function is_valid_stockpile_extent(s)
...
-    e={label='Gem'},
-    b={label='Bar/Block'},
-    h={label='Cloth'},
-    l={label='Leather'},
-    z={label='Ammo'},
-    S={label='Sheets'},
-    g={label='Finished Goods'},
-    p={label='Weapons'},
-    d={label='Armor'},
+    a={label='Animal', index=0},
```
lethosor, line 49: You could theoretically set `viewscreen_layer_stockpilest.layer_objects[0].cursor` to `index` directly, instead sending the down arrow repeatedly. One downside is that there's no guarantee that Toady won't move objects within `layer_objects` in the future, and there's little type-safety if that happens (it would only fail immediately if entry 0 changes from a `layer_object_listst` to a `layer_object_buttonst`).

Another even more fun approach would be to read tiles from the screen with `readTile()`, although that's probably way more brittle than your approach here).

Honestly, the stockpile screen is pretty stable, so your approach is fine. These are just a couple others that might be useful for other things.

myk002: interestingly, a screen scraping version of "readTile" is how python quickfort implemented some of its fancier features, like material autoselection. I think we'll have better options for our quickfort, though, since we have access to internal data structures.

If the UI changes or the lists change contents/order, a whole lot of query blueprints are going to break : p  I've tried to centralize a lot of the query logic in the common aliases file (next PR) so many of the fixes can be "soft" -- that is, we can fix blueprints by just updating the aliases that come with dfhack.

I'm watching what Toady is doing for the Steam UI. His attempts to make the UI cleaner might require a bunch of changes to this script and its data files.

I've been testing some heavy query blueprints, and so far I'm pretty satisfied with their performance -- I can toggle all the hundreds of tallow entries in a food stockpile in a fraction of a second, for example -- so I've been pretty happy with the simple approach of just sending keystrokes.

lethosor: That's a good point that I forgot about - any change to the screen would also break the arrow keys approach. It's hard to detect these sorts of things automatically, but hopefully we'll catch any issues that arise in future releases.

PR #174, opened 2020-08-05
Fix issues found while playtesting the quickfort script

myk002:
DFHack/dfhack#499

A hodgepodge of improvements and fixes for problems found while playtesting. Feedback requested for my decision to default stockpile wheelbarrows to 0. Is this controversial? The wiki suggests that wheelbarrows are a net loss for a fort. The player can still change the default in the quickfort.txt config file. The alternative is to default to 1 and require players who want to set it to 0 to use query mode. The problem with that, though, is the interface for setting wheelbarrows in query mode differs based on whether you have the DFHack tweak max-wheelbarrow enabled. And if you do have the tweak enabled, The issue with this, I've found, is that interacting with DFHack input prompts (like the one for setting the number of wheelbarrows) makes query mode unreliable. I'm documenting this in the user guide, but this is also an issue that I'd like to look into someday and fix.

Dig mode:
- allow additional tile shapes that were incorrectly disallowed for channeling
- mark the tile block as designated when we add a designation so the designation job scanner can create jobs for them
- initiate a job scan after designating tiles for dig

Build/Place mode:
- fix db definition for screw press (was miscategorized as a tool instead of a custom workshop)
- initialize max_general_orders when building a workshop so the blueprint-constructed buildings can take manager orders
- initiate a job scan after placing stockpiles/buildings
- rewrite extent splitting code to process too-large extents in arbitrary "chunks" instead of grids. This allows layouts like solid blocks of staggered workshops of the same type or "C"-shaped lines of 1x1 buildings like beds. The old algorithm would force buildings to align to the grid instead of respecting the layout from the blueprint author.
- initialize max containers for stockpiles
- make max stockpile containers configurable and default wheelbarrows to 0

Query mode:
- don't move the viewport in query mode. it happens too fast to see what's going on anyway and just looks like screen jitter.
- add shortcuts for ctrl ('!') and alt ('\~'). '!' for {Ctrl} is from python quickfort (though it was undocumented). '~' for {Alt} is new, but already in use elsewhere to represent alt
- make alias expansion log easier to read by adding spaces between keycodes

### DIFF
--- a/internal/quickfort/aliases.lua
+++ b/internal/quickfort/aliases.lua
@@ -8,11 +8,12 @@ end
 local quickfort_common = reqscript('internal/quickfort/common')
 local log = quickfort_common.log
 
--- special key sequences inherited from python quickfort. these cannot be
--- overridden with aliases
+-- special keycode shortcuts inherited from python quickfort.
 local special_keys = {
     ['&']='Enter',
     ['+']='{Shift}',
+    ['!']='{Ctrl}',
+    ['~']='{Alt}',
     ['@']={'{Shift}','Enter'},
     ['^']='ESC',
     ['%']='{Wait}'
--- a/internal/quickfort/aliases.lua
+++ b/internal/quickfort/aliases.lua
@@ -127,7 +127,7 @@ function expand_aliases(text)
     end
     local expanded_text = table.concat(tokens, '')
     if text ~= expanded_text then
-        log('expanded keys to: "%s"', expanded_text)
+        log('expanded keys to: "%s"', table.concat(tokens, ' '))
     end
     return tokens
 end
--- a/internal/quickfort/build.lua
+++ b/internal/quickfort/build.lua
@@ -389,8 +389,8 @@ local building_db = {
         type=df.building_type.Workshop, subtype=df.workshop_type.Custom,
         custom=0},
     wp={label='Screw Press',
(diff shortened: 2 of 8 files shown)


lethosor:
The number of wheelbarrows being configurable is already a plus - I'm not picky about the default.

> The alternative is to default to 1 and require players who want to set it to 0 to use query mode. The problem with that, though, is the interface for setting wheelbarrows in query mode differs based on whether you have the DFHack tweak max-wheelbarrow enabled.

Not quite sure I understand the issue - is this about overriding the number of wheelbarrows at the blueprint level? Looks to me like that's what place mode does. Could you have a special case to set the `max_wheelbarrows` field instead of relying on the UI, or is that tricky with the query mode architecture?

myk002:
> Not quite sure I understand the issue - is this about overriding the number of wheelbarrows at the blueprint level? Looks to me like that's what place mode does. Could you have a special case to set the `max_wheelbarrows` field instead of relying on the UI, or is that tricky with the query mode architecture?

You're right, there are several alternatives here. The one I was worried about was setting wheelbarrows via a query mode blueprint -- sending keycodes to the UI -- since the different methods of setting wheelbarrow counts would require different keycode sequences.
but you're absolutely right that if we want to make it configurable at a blueprint level instead of a global setting level, we can add syntax for the config to the place blueprint. If players request blueprint-level wheelbarrow config, I'll look into that. I think the global setting should cover most use cases.

PR #185, opened 2020-08-23
Quickfort remaining features

myk002:
DFHack/dfhack#499
the mesh of interdependent feature branches has become too complicated to merge individually, so here's a combined merge for the final milestones for the first public release of DFHack quickfort. Marked WIP until the other PRs (which this PR builds on) are merged. Changes
include:

Major features
--------------

- implement GUI interface for quickfort. can do everything except set settings. remembers last filter and selected position and restores them when GUI is reinvoked.

- allow multiple blueprints in a single .csv file or spreadsheet sheet. a new blueprint begins when a modeline is detected in the first column.

- new blueprint mode: zone. Like place mode but for zones. currently can't initialize zone settings (like pit vs pond). will address this later. tricky bit of this mode was ensuring a zone is not selected when it is deleted, since this causes a crash.

- new blueprint mode: meta. can run a series of related blueprints as if they were one. can also change z-levels and then run a specified blueprint.

- place and zone mode now recognize multi-type stockpiles and zones.

- implement new markers for modelines: hidden(), which hides the blueprint from list output unless the --hidden parameter is specified, and label(), which give the blueprint a name. both are primarily useful when writing meta blueprints. you can hide blueprints that are managed by a higher-level meta blueprint, and you need to give your blueprints labels so the meta blueprint can refer to them. also implemented message(), which displays a message to the player after the blueprint has been applied. these markers can be defined in anyhorder.

Bugfixes and minor changes
--------------------------

- remove the '+' alias for shift. it's too useful of a key to map to something else. also, it seems that the original quickfort only mapped + to shift when it is followed by {Enter}, which is a special case that we don't really want to support. {Shift}{Enter} already has the '@' shortcut.

- refactor screw pump database entries into a generator function

- initialize furnaces can accept workshop orders (we correctly initialized real workshops, but furnaces got missed)

- create quickfort setting 'buildings_use_blocks', which restricts anything that requires a generic building material to instead require a block

- refactor extent overwriting into a separate function to centralize memory leak protection

- only output one set of stats per quickfort command invocation instead of one set per (blueprint, zlevel) tuple.

- move some parsing logic from list.lua to parse.lua to centralize it

- split do_command() into a frontend and backend so meta blueprints can call the backend function directly

- refactored the list command into a frontend and backend so the gui can call the backend function directly

- handle multi-line quoted strings in .csv files


lethosor:
I'm not quite sure why GitHub thinks there are merge conflicts - maybe because of the other merged PRs? I was able to merge this branch into `quickfort` locally (and vice versa), although I don't want to push the merge up to your branch unless you want me to. I suspect however Git chooses to merge should be fine, but feel free to merge `quickfort` into this branch and double-check that it's doing the right thing.

myk002:
> * Blueprints in `blueprints/library` seem to no longer be listed by `quickfort list` - is this intentional?

intentional. I didn't want library files cluttering up the list output. players will *usually* want to just see their own blueprints. pass `-l` or `--library` to `quickfort list` to see library files.

> * Moving `dreamfort.csv` into `blueprints` gives a bunch of entries like this - is the `-n "nil"` intentional?

very unintentional. and unexpected. I've been testing this extensively while developing the gui. I may have a bad merge somewhere. I'll look into this.

myk002:
> feel free to merge `quickfort` into this branch and double-check that it's doing the right thing.

merged locally, and no conflicts that I could see. Let me retest a bit and address that -s "nil" business.

internal/quickfort/orders.lua
```diff
@@ -0,0 +1,161 @@
...
+        log('ordering %d %s', quantity, k)
+        stockflow.create_orders(order_spec.order, quantity)
+        table.insert(ctx.stats, {label=k, value=quantity})
+    end
+end
+
+function enqueue_orders(stats, buildings, building_db, ctx)
+    local order_specs = ctx.order_specs or {}
+    ctx.order_specs = order_specs
+    local reactions = stockflow.reaction_list -- don't cache this; it can reset
```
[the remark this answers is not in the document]
myk002: that's not a bad idea. no refactoring needed, and no need for the plugin to be enabled -- though can I access the stockflow lua module if the plugin isn't enabled? I'll play with this a bit. At least I can output a proper error message if I need stockflow and it isn't available.

internal/quickfort/orders.lua
```diff
@@ -115,18 +114,20 @@ local function get_num_items(b)
...
-    for k,order_spec in pairs(ctx.order_specs) do
+    for k,order_spec in pairs(ctx.order_specs or {}) do
         local quantity = math.ceil(order_spec.quantity)
         log('ordering %d %s', quantity, k)
         stockflow.create_orders(order_spec.order, quantity)
         table.insert(ctx.stats, {label=k, value=quantity})
     end
 end
 
+-- ensure we don't reinit this; it contains allocated memory
```
lethosor, line 125: You'll want to drop the `local` keyword to avoid reallocating memory - local variables don't persist across script runs, so `reactions` will always be nil before this line.

Granted, under normal circumstances, this file should only be "run" once by end-users, when the main quickfort entrypoint imports it. (Other things that would cause this to be re-run include editing this file or running `internal/quickfort/orders` from the console.) I'm pretty sure running just `quickfort` multiple times would only cause collect_reactions to be called once.

lethosor: I just realized that stockflow is probably regenerating this list on world load because there could be different reactions available in each world. A lot of the furniture reactions are hardcoded, though, but custom workshops could vary - do you support those?

This should be fine for the majority of furniture reactions, in any case. I'd be fine with refactoring stockflow.lua a bit later to support this use case better.

myk002: The only orders quickfort would ever generate will be for stuff that is buildable. can custom workshops produce buildable items? Is the list of what can be placed with the `b`uild menu variable?

lethosor: I was talking about building actual custom workshops... which I just realized you can't build through the manager. (They affect the `b-W` menu, and I saw support for building workshops in one of the library blueprints earlier, which is why I thought of custom workshops, but the presence of custom workshops alone won't affect what orders are available. So this is fine - I was confusing some unrelated things.

myk002: done

lethosor:
I see now that you were imitating the built-in shift+enter control in ListBox - I changed that too. I also took advantage of the `key_string()` wrapper in ef87b3b - turns out it was never documented, but it does simplify the common case of `key + ':' + label` a bit.

PR #186, opened 2020-08-27
Merge Quickfort into master

myk002:
Fixes DFHack/dfhack#499
I think the script is ready for users! Are there any outstanding changes you'd like to see before this is merged?


quickfort.lua
```diff
@@ -0,0 +1,203 @@
+-- DFHack-native implementation of the classic Quickfort utility (PRE-RELEASE)
+--[====[
+
+quickfort
+=========
+Processes Quickfort-style blueprint files. This is a pre-release -- not all
```
lethosor, line 6: Is this still accurate?

myk002: Ha! good catch. no. the script is feature complete. I'll submit a mini PR to fix this

PR #190, opened 2020-08-31
Quickfort: implement dig priorities

myk002:
implement setting the priority for dig designations. syntax for dig blueprints is now [letter][number] where if the letter is not specified, 'd' is assumed, and if number is not specified, 4 is assumed.

also fix a bug in has_designation where the uint32 values were interpreted as booleans instead of numbers when they were unset. changed code to be unambiguous.

edit: also fix a bug in my new fancy __index methods where we could overflow the stack if we tried to find a key that doesn't exist.

### DIFF
--- a/internal/quickfort/dig.lua
+++ b/internal/quickfort/dig.lua
@@ -510,7 +510,7 @@ local function extended_parser(_, keys)
             return nil
         end
     end
-    if priority and #code == 0 then code = 'd' end
+    if #code == 0 then code = 'd' end
     if not rawget(dig_db, code) then return nil end
     local custom_designate = copyall(dig_db[code])
     custom_designate.marker_mode = marker_mode
--- a/internal/quickfort/dig.lua
+++ b/internal/quickfort/dig.lua
@@ -492,6 +492,11 @@ local dig_db = {
     ['or']={action=do_traffic_restricted},
 }
 
+-- set default dig priorities
+for _,v in pairs(dig_db) do
+    if v.use_priority then v.priority = 4 end
+end
+
 -- handles marker mode 'm' prefix and priority suffix
 local function extended_parser(_, keys)
     local marker_mode = false
@@ -538,7 +543,6 @@ local function set_priority(ctx, priority)
     local pbse = get_priority_block_square_event(block_events)
     if not pbse then
         if priority == 4 then return end
-        print("creating event")
         block_events:insert('#',
                             {new=df.block_square_event_designation_priorityst})
         pbse = block_events[#block_events-1]
@@ -563,7 +567,7 @@ local function dig_tile(ctx, db_entry)
                         quickfort_common.settings['force_marker_mode'].value
                 ctx.occupancy.dig_marked = marker_mode
             end
-            if db_entry.use_priority and db_entry.priority then
+            if db_entry.use_priority then
                 set_priority(ctx, db_entry.priority)
             end
         end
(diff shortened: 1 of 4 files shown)


lethosor (commented):
I think I found an issue where designations with no priority are assigned a priority of 0, not 4, effectively giving them the highest priority (I think this is something that another tool changing designation priorities has run into before). Here's a test blueprint:
```
#dig
1,1,1
d,d,d
2,2,2
```
I'm expecting the middle row to be dug out last, but it's dug out first (followed by the top row, then the bottom).

A more complicated one, which seems to exhibit the same behavior of non-prioritized tiles being dug out first:
```
#dig
d,d4,d5
1,2,h3
4,d4,d
```

internal/quickfort/dig.lua
```diff
@@ -457,52 +457,115 @@ local function do_traffic_restricted(ctx)
...
+    end
+    if #code == 0 then code = 'd' end
+    if not rawget(dig_db, code) then return nil end
+    local custom_designate = copyall(dig_db[code])
+    custom_designate.marker_mode = marker_mode
+    custom_designate.priority = priority
+    return custom_designate
+end
+
+setmetatable(dig_db, {__index=extended_parser})
```
lethosor, line 521: Ah, extended_parser will never get called for keys that are already in `dig_db` (like "d" from my examples).
I'm not sure if you accounted for this - it looks like extended_parser is intended to handle keys without numbers, which I don't think it can be called with.

Update: I see set_priority() guarded by a check for dig_db[key].use_priority, but it doesn't seem to be called when the priority isn't specified, and it probably should be called with 4. I'm assuming you have a better idea of how to address this.

internal/quickfort/dig.lua
```diff
@@ -457,52 +457,115 @@ local function do_traffic_restricted(ctx)
...
+-- modifies any existing priority block_square_event to the specified priority.
+-- if the block_square_event doesn't already exist, create it iff we're setting
+-- the priority to any number but 4 (the default)
+local function set_priority(ctx, priority)
+    log('setting priority to %d', priority)
+    local block_events = dfhack.maps.getTileBlock(ctx.pos).block_events
+    local pbse = get_priority_block_square_event(block_events)
+    if not pbse then
+        if priority == 4 then return end
+        print("creating event")
```
lethosor, line 541: Probably should be a `log()` call

myk002: it was a leftover dev debugging line; removed.

internal/quickfort/dig.lua
myk002: it still gets called when there is an 'm' prefix for marker mode, so "md" would get here (and assign the default priority of 4).

I wasn't calling set_priority() when priority isn't specified since the game interprets lack of an event block as all priorities == 4, and I didn't think I needed to mess with them. but in this case what *might* be happening is an event block already exists with a 0 value and I'm not overwriting it? It's true that I hadn't considered that case. it could happen with non-zero values as well, I think. if a previous designation was at priority N, was cleared (which does not clear the event block, IIRC), and then a blueprint without an explicit priority was applied. I think I can fix this by assigning priority=4 to the db records.

lethosor: Yep, block events cover the whole block. You're not the first to run into issues caused by this. Looks good now!

lethosor (approved):

PR #191, opened 2020-09-03
timestream: The end of FPS death as we know it

daedsidog:
Proof of concept for https://github.com/DFHack/dfhack/issues/1638.

This is an expansion of [timestream](https://gist.github.com/IndigoFenix/cf358b8c994caa0f93d5). I fixed the idling bug and made it so that all units on the map and the calendar progress at a constant rate **at any given FPS.** This means that no matter if you have 100 FPS or 20 FPS, your gameplay will remain _mostly_ the same.

What I did was speed up action timers and the calendar based on the ratio between the desired FPS and the current FPS. Unlike `fastdwarf` which sets all actions to be instantaneous, this only speeds them up evenly for everyone.

At first I tried a lazy solution by applying the `DEBUG_MEGAFAST` flag on and off during different frames, but that resulted in the dwarves taking short breaks between their zooming about like insects. Nauseating. Speeding up the timers works like magic.

Another thing that this script doesn't do is deal with physics. Things like water flow, bolts, miasma, etc. are still at their normal speed.

Issues so far:
* ~I used the game's own FPS counter to calculate the average FPS, so naturally when you pause the game you're back at max FPS, so it would take a moment for the speed to pick up again, since the script thinks you're at 100 FPS and require no boost.~
* The current implementation is very inefficient. The script loops over all the creatures every tick and manually increments their timers. A vastly better way of doing this would be to either use some sort of callback to when the timers were set or to adjust the increment value that the game uses for the timers.
* Once you get below 10 FPS, there are diminishing returns and you won't get full 100 FPS experience. Although, during a ~300 invader siege my FPS dropped to about 6-7, but it played out as though it was around 30-50. The problem is that at that stage the GUI lag becomes the problem. I believe this is related to the 2nd issue, since if the script was more efficient this limit could be pushed lower.
* ~Speeds up too much during seasonal autosave.~
* Merchant wagons are not sped up so they take a very long time to reach the trade depot in very low FPS situations, which results in them hanging around sometimes for multiple seasons.
* Relationships between dwarves are formed slower since the tick rate remains the same yet time moves faster.
* ~Pregnancies take longer.~
* Crossbows are less effective in combat because ~reloading &~ projectile speed is slower relative to everything else.
* Everyone gets tired faster during combat because the stamina regeneration remains at the same rate, so exhaustion is much more common.

Let it also be known that I am also not very knowledgeable in the technical aspects of the game, so I may have missed things such as more timers or the underlying implications of speeding up the calendar.

### DIFF
--- a/timestream.lua
+++ b/timestream.lua
@@ -1,19 +1,13 @@
 -- speeds up the calendar, units, or both
 
 --[====[
-
 timestream
 ==========
 Controls the speed of the calendar and creatures. Fortress mode only. Experimental.
-
 The script is also capable of dynamically speeding up the game based on your current FPS to mitigate the effects of FPS death. See examples below to see how.
-
 Usage::
-
     timestream [-rate R] [-fps FPS] [-units [FLAG]] [-debug]
-
 Examples:
-
 - ``timestream -rate 2``:
     Calendar runs at x2 normal speed, units run at normal speed
 - ``timestream -fps 100``:
@@ -27,9 +21,7 @@ Examples:
 - ``timestream -fps 100 -units 2``:
     Activates a different mode for speeding up units, using the native DF
     ``debug_turbospeed`` flag (similar to `fastdwarf` 2) instead of adjusting
-    timers of all units. It may result in strange unit motion, so it is not
-    the default.
-
+    timers of all units. This results in rubberbanding unit motion, so it is not recommended over the default method.
 Original timestream.lua: https://gist.github.com/IndigoFenix/cf358b8c994caa0f93d5
 ]====]
 
@@ -65,7 +57,7 @@ local prev_time = df.global.enabler.clock
 local ui_main = df.global.ui.main
 local saved_game_frame = -1
 local frames_until_speeding = 0
-local speedy_frame_delta = desired_fps
+local speedy_frame_delta = desired_fps or DEFAULT_MAX_FPS
 
 local SEASON_LEN = 3360
 local YEAR_LEN = 403200
@@ -83,7 +75,9 @@ end
 simulating_desired_fps = true
 if desired_fps == nil then
     desired_fps = DEFAULT_MAX_FPS
-    simulating_desired_fps = false
+    if simulating_units ~= 1 and simulating_units ~= 2 then
+        simulating_desired_fps = false
+    end
 elseif desired_fps < MINIMAL_FPS then
     desired_fps = MINIMAL_FPS
 end
--- a/timestream.lua
+++ b/timestream.lua
@@ -20,15 +20,19 @@ local desired_fps = tonumber(args[2])
 local simulating_units = tonumber(args[3])  -- Setting this to 2 instead of 1 will use debug_turbospeed instead of adjusting the timers of all creatures. I don't quite like it because it makes everyone move like insects.
 local debug_mode = tonumber(args[4])
 
+local fps_samples = 10   -- Controls how many frames (fps_samples * desired_fps) inside each frame sum that are used to calculate avg_fps.
+local minimal_fps = 10 -- This ensures you won't get crazy values on pausing/saving, or other artefacts on extremely low FPS.
+local splicing_threshold = 0.25 -- This controls how generously the script decides to speed up units. Higher values mean that units will be sped up more often than not. This is only for when using debug_turbospeed.
+
+local sample_counter = 0
 local prev_tick = 0
 local ticks_left = 0
 local simulating_desired_fps = false
-local minimal_fps = 10 -- This ensures you won't get crazy values on pausing/saving.
 local prev_frames = df.global.world.frame_counter
 local frame_sum = 0
+local last_frame_sum = desired_fps * desired_fps * fps_samples
 local avg_fps = desired_fps
 local last_frame_sped_up = 0
-local splicing_threshold = 0.25 -- This controls how generously the script decides to speed up units. Higher values mean that units will be sped up more often than not. This is only for when using debug_turbospeed.
 
 if rate == nil then
     rate = 1
@@ -68,7 +72,6 @@ dfhack.onStateChange.loadTimestream = function(code)
                 print('Time running at x'..rate..".")
             else
                 print('Time running dynamically to simulate '..desired_fps..' FPS.')
-                prev_time = df.global.enabler.clock
                 simulating_desired_fps = true
                 prev_frames = df.global.world.frame_counter
                 rate = 1
@@ -196,17 +199,24 @@ function update()
             local counted_frames = df.global.world.frame_counter - prev_frames
             frame_sum = frame_sum + df.global.enabler.calculated_fps
             if counted_frames >= desired_fps then
-                avg_fps = frame_sum/counted_frames
+                avg_fps = (frame_sum + last_frame_sum)/(counted_frames + desired_fps * (sample_counter + fps_samples))
                 if avg_fps <= desired_fps then
                     rate = desired_fps/avg_fps  -- We don't want to slow down the game
+                else
+                    rate = 1
                 end
                 if debug_mode then
                     print("prev_frame: "..prev_frames..", avg_fps: " ..avg_fps.. ", rate: "..rate)
                 end
                 prev_frames = df.global.world.frame_counter
-                frame_sum = 0
+                sample_counter = sample_counter + 1
+                if sample_counter >= fps_samples - 1 then
+                    last_frame_sum = frame_sum
+                    frame_sum = 0
+                    sample_counter = 0
+                end
             end
-            if avg_fps > 0 and avg_fps < desired_fps then -- God forbid avg_fps is not positive...
+            if avg_fps > 0 then -- God forbid avg_fps is not positive...
                 if simulating_units == 2 then
                     local missing_frames = desired_fps - avg_fps
                     local speedy_frame_delta = desired_fps/missing_frames
--- a/timestream.lua
+++ b/timestream.lua
@@ -113,8 +113,8 @@ end
 
 function update()
     prev_tick = df.global.cur_year_tick
-    if last_frame + 1 == df.global.world.frame_counter then
-        if rate ~= 1 or simulating_desired_fps then
+    if rate ~= 1 or simulating_desired_fps then
+        if last_frame + 1 == df.global.world.frame_counter then
             timestream = 0
 
             --[[if rate < 1 then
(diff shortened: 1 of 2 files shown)


timestream.lua
```diff
@@ -0,0 +1,339 @@
...
+    debug_mode = false
+else
+    debug_mode = true
+end
+
+eventNow = false
+seasonNow = false
+timestream = 0
+counter = 0
+if df.global.cur_season_tick < 3360 then
```
lethosor, line 57: I realize you didn't write some of this, but it would be nice to use constants instead of magic numbers. I think I see how most of these correspond to units from https://dwarffortresswiki.org/index.php/DF2014:Time (I'm guessing `cur_season_tick` is `cur_year_tick / 10`?) but it would be easier to read through if they had names instead.

timestream.lua
```diff
@@ -0,0 +1,339 @@
...
+                        --df.global.cur_season_tick=event.season_ticks
+                        event.season_ticks = df.global.cur_season_tick
+                        eventNow = true
+                    end
+                    eventFound = true
+                end
+            end
+            if eventFound == false then eventNow = false end
+
+            if df.global.cur_season_tick >= 3359 and df.global.cur_season_tick < 6719 and month == 1 then
```
lethosor, line 152: Should this check for `>= 3360` instead? I ask because I'm seeing an assignment to 3360 below. (Same goes for the other checks).

Regarding constants, maybe something like `SEASON_LENGTH` could work, and could be reused as `SEASON_LENGTH - 1`, `SEASON_LENGTH * 2`, etc.

timestream.lua
```diff
@@ -0,0 +1,339 @@
...
+This is only for fortress mode.
+
+Usage:
+:timestream <scalar> <fps> <simulate units y/n>
+Examples:
+
+:timestream     2:          Calendar runs at x2 normal speed
+:timestream    -1 100:      Calendar runs at dynamic speed to simulate 100 FPS
+:timestream    -1 100 1:    Calendar & units are simulated at 100 FPS
+
```
lethosor, line 16: Might be nice to add a link to https://gist.github.com/IndigoFenix/cf358b8c994caa0f93d5 around here

lethosor (commented):
Neat!
Regarding performance, it's possible to write just the expensive parts as part of a plugin and call out to it from Lua. Not sure if it would fit better as its own plugin or part of `fastdwarf`.

timestream.lua
```diff
@@ -0,0 +1,339 @@
+-- speeds up the calendar, units, or both
+
+--[====[
+
+timestream
+==========
+This is only for fortress mode.
```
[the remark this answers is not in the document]
daedsidog: I added a check that would make sure it will only run once in fortress mode. I don't want to give an error because I'd like to have it run by default, unless there's a more elegant way of doing it.

timestream.lua
daedsidog: Done

timestream.lua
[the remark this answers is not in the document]
lethosor: Ok, I think that makes sense

timestream.lua
lethosor: That looks good, thanks!

timestream.lua
daedsidog: Was what I thought too at first, but then I figured he had some reason, so I didn't want to change something I didn't fully understand the consequences of.

timestream.lua
daedsidog: Done

daedsidog:
> Neat!
> Regarding performance, it's possible to write just the expensive parts as part of a plugin and call out to it from Lua. Not sure if it would fit better as its own plugin or part of `fastdwarf`.

I was really hoping you knew of a way to do this without looping over the timers, since that's the only real expensive part and even if made part it would still be inefficient.

lethosor:
fastdwarf does pretty much what you're doing here (except with a hardcoded `1` instead of `dec`):
https://github.com/DFHack/dfhack/blob/8728056674c7249f1a15f80c62e3ded16507bb56/plugins/fastdwarf.cpp#L118-L127

It may not be efficient in terms of operations required, but there can be a *lot* of overhead with accessing fields of DF structures from Lua, particularly in nested loops like here (one example Quietust brings up is that the `reveal` plugin, which runs pretty much instantly on most maps, can take more than a minute to complete when rewritten in Lua). Dealing with native Lua types from Lua is faster, and sometimes you can get a bit of a speedup by caching DF values in local variables, although the only opportunity I see to do that here is storing `action.type` in a local, which probably won't make much of a difference (but feel free to try it).

I wouldn't worry too much about this if you can still get a partial improvement in low-FPS situations, though. I could certainly try to expose part of `fastdwarf` to Lua (or help you with it), though, if you want to try that.

timestream.lua
lethosor: @IndigoFenix is fairly active on the forums (not sure about here), so you could ask there if you have an account. It's an old script, though.

daedsidog:
> fastdwarf does pretty much what you're doing here (except with a hardcoded `1` instead of `dec`):
> https://github.com/DFHack/dfhack/blob/8728056674c7249f1a15f80c62e3ded16507bb56/plugins/fastdwarf.cpp#L118-L127
> 
> It may not be efficient in terms of operations required, but there can be a _lot_ of overhead with accessing fields of DF structures from Lua, particularly in nested loops like here (one example Quietust brings up is that the `reveal` plugin, which runs pretty much instantly on most maps, can take more than a minute to complete when rewritten in Lua). Dealing with native Lua types from Lua is faster, and sometimes you can get a bit of a speedup by caching DF values in local variables, although the only opportunity I see to do that here is storing `action.type` in a local, which probably won't make much of a difference (but feel free to try it).
> 
> I wouldn't worry too much about this if you can still get a partial improvement in low-FPS situations, though. I could certainly try to expose part of `fastdwarf` to Lua (or help you with it), though, if you want to try that.

I don't have a problem with rewriting the entire thing in C++ provided it's worth the time. I did it in Lua instead because I really didn't expect it to turn out as it did, and it just expanded. The script stops simulating 100 FPS when I drop down below 10, which only happens during mass combat. Even then, you still get something like 60 FPS simulation. I'll do a small test to see the performance difference.

daedsidog:
I've compared the performance loss of this script to that of `fastdwarf`:

```
Normal..........................:    45 FPS
timestream -1 100 1:    36 FPS
fastdwarf 1 0...............:    40 FPS
```

lethosor:
Yeah, I don't really know exactly what's included in `units.active`, but I'm pretty sure it's a subset of `units.all`. It does tend to contain units with the `inactive` flag set (the apparent naming inconsistency is because this flag used to be named `dead`, so the vector name is older; however, not all units with this flag set were actually dead, so it was renamed to `inactive` as part of https://github.com/DFHack/df-structures/issues/247). `Units::isActive()` checks the `inactive` flag, so it'll usually narrow down the contents of `units.active`. I don't really have a better explanation at the moment.

Thanks for taking the time to benchmark this! I haven't taken the time to figure out exactly how to draw a comparison between this script and fastdwarf (e.g. I don't think that means the plugin is twice as fast as the script), but I would deem it "close enough", particularly if it results in a significant speedup overall. It's entirely possible that speeding up dwarves causes DF to perform more time-consuming operations more often, which would affect the numbers you see in a way that's hard to measure.

Anyway, I think I'd be fine with merging this given the above. I haven't actually had much time to test it out on my own, so I (or someone else) should probably do that first. Rewriting it in C++ or reusing some of fastdwarf would probably make it faster, but it's hard to say how much faster without actually doing it (which you are welcome to do, and I'd be willing to assist, but I also think the script's performance is good as-is).

timestream.lua
```diff
@@ -0,0 +1,393 @@
...
+
+--[====[
+
+timestream
+==========
+This only affects fortress mode.
+
+Usage:
+:timestream <scalar> <fps> <simulate units y/n>
+Examples:
```
lethosor, line 11: The formatting here is off - you need more blank lines for it to format correctly. This is what it currently looks like when built (see [here](https://docs.dfhack.org/en/latest/docs/Documentation.html) for information on building the docs):

image

I made a couple changes locally that made it look better (note that the `:` at the beginning of a line is only for "definition lists", which we use for some scripts' arguments, not for code blocks - those require `::` on the previous line, and need to be separated by a blank line):

```suggestion
Usage::

    timestream <scalar> <fps> <simulate units y/n>

Examples:

```


Here's what it looks like after my changes: 
image

timestream.lua
```diff
@@ -0,0 +1,393 @@
+-- speeds up the calendar, units, or both
+
+--[====[
+
+timestream
+==========
+This only affects fortress mode.
```
lethosor, line 7: Could you also add a description of what the script does? It can be short, and the comment on line 1 would be a good place to start.
It would also be good to mention that arguments can be skipped by passing `-1`, since there aren't examples for all of them.

timestream.lua
```diff
@@ -0,0 +1,393 @@
...
+            end
+        else
+            print('Time set to normal speed.')
+            loaded = false
+            df.global.debug_turbospeed = false
+        end
+            if debug_mode then
+            print("timestream: Debug mode is on.")
+        end
+    elseif code==SC_MAP_UNLOADED then
```
lethosor, line 124: Is this case deliberately empty?

timestream.lua
```diff
@@ -0,0 +1,393 @@
...
+local DEFAULT_MAX_FPS       = 100
+
+
+--- DO NOT CHANGE BELOW UNLESS YOU KNOW WHAT YOU'RE DOING ---
+
+args={...}
+local rate=tonumber(args[1])
+local desired_fps = tonumber(args[2])
+local simulating_units = tonumber(args[3])  -- Setting this to 2 instead of 1 will use debug_turbospeed instead of adjusting the timers of all creatures. I don't quite like it because it makes everyone move like insects.
+local debug_mode = tonumber(args[4])
```
lethosor, line 30: If these arguments are not all required, I might actually recommend using `utils.processArgs` instead. It would allow script invocations like this:
```
timestream -fps 200
timestream -rate 2
timestream -rate 3 -units
```
Arguments can be passed in without values, like `-units` or `-debug` - you would still need to convert from strings to the desired integer/boolean types manually, but I think it would be easier to handle this way, and easier to use. Searching for "processArgs" in this repo should turn up a lot of examples, but let me know if you have any questions about it.

lethosor (commented):
Just saw your change and managed to test a bit on my end - sorry for the delay there.

I'm testing on a fresh map and occasionally seeing an issue where tree-chopping designations created after running `timestream 5` or `timestream 5 -1 1` are not done by any dwarves at all. Rerunning timestream with a different speedup (like 4 or 6) causes the jobs to be done immediately. My first guess was that the `process_dig` global wasn't being set or cleared by DF properly, but it seems to be handled as expected. Do you have any ideas as to what could be causing this? I'm not really sure where to start looking.

timestream.lua
daedsidog: Done

timestream.lua
daedsidog: It's obsolete, I removed it.

timestream.lua
daedsidog: > Just saw your change and managed to test a bit on my end - sorry for the delay there.
> 
> I'm testing on a fresh map and occasionally seeing an issue where tree-chopping designations created after running `timestream 5` or `timestream 5 -1 1` are not done by any dwarves at all. Rerunning timestream with a different speedup (like 4 or 6) causes the jobs to be done immediately. My first guess was that the `process_dig` global wasn't being set or cleared by DF properly, but it seems to be handled as expected. Do you have any ideas as to what could be causing this? I'm not really sure where to start looking.
> 
> I was partly reminded by your most recent commit - are you still testing this? I'm fine with waiting if you're still performing some tests, but can also merge this whenever it's ready.

This might be related to the original `timestream` bug where units are strictly preoccupied with leisure activities and therefore don't do anything else. This is why the original script didn't work after the tavern system was introduced. I thought I fixed this bug because I haven't encountered it since, but frankly I have only tried messing around with the regular calendar speeding up only a handful of times, and strictly tested the FPS simulation. I haven't tested what could happen at extremely high values, either.

As for the main event, aka the FPS simulation; I've been running a fort for 70 years now (longest I've ever had, would've quit due to the FPS death otherwise) and found no significant bugs other than the ones I pointed out in the main post, which are not gamebreaking by any means, so to me I think it's good to go. I haven't tested the plain portion of the script, though.

timestream.lua
daedsidog: Done, thanks.

timestream.lua
lethosor: Yeah, I wasn't sure how extreme my test was. I'll see if I have a low-FPS fort sitting around, but that sounds good to me.

Any thoughts on the argument handling comment? The main issue I can think of with `processArgs` is that it doesn't handle negative numbers well, but you seem to just be using them to indicate "unspecified" here, which wouldn't be necessary with `processArgs`.

daedsidog: I'll implement it when I get the time.

lethosor: Would you be okay with me changing argument handling after I merge? Asking just in case you've already started or have something else in mind.

daedsidog: > Would you be okay with me changing argument handling after I merge? Asking just in case you've already started or have something else in mind.

(Un?)fortunately, I haven't had a lot of spare time recently so I haven't begun to change it. You have my gratitude.

lethosor:
FYI: planning to change argument handling and merge this shortly

timestream.lua
```diff
@@ -4,17 +4,21 @@
...
 
-    timestream <scalar> <fps> <simulate units y/n>
+    timestream <rate> <desired fps> <simulate units 1/0>
 
 Examples:
 
-:timestream     2:          Calendar runs at x2 normal speed
-:timestream    -1 100:      Calendar runs at dynamic speed to simulate 100 FPS
-:timestream    -1 100 1:    Calendar & units are simulated at 100 FPS
+    timestream     2:           Calendar runs at x2 normal speed, units run at normal speed
```
lethosor, line 17: Looks like you got rid of the leading colons here, which are needed to make it a "definition list" - it looks like this otherwise:
image

I will fix - just a heads-up for future work

lethosor:
I've updated argument handling and docs (and had to merge to resolve a changelog conflict to get tests to run).
If you could double-check that my argument handling changes didn't break anything, that would be great. I tried to limit my changes to the top of the file, since some of the logic using the arguments is a bit complicated. Specifically:
- `rate` still defaults to `-1` if not specified (since that was in examples, and appears to be handled differently from `nil`)
- `desired_fps` defaults to `nil` if not specified (since it was not passed as `-1` in any examples, and `nil` seems to be checked explicitly)
- `simulating_units` should always be a number - I checked with `-units`, `-units 1` (these two are equivalent), `-units 2`, and the absence of the argument
- `debug_mode` is always boolean (Lua unfortunately lacks a `toboolean()`...)

I'm pretty sure this should still work as intended, but there's a chance I broke something.

daedsidog:
I gave it a spin and broke it a few times when no FPS was being specified, but other than that everything, everything seemed to behave as it should.

Regarding `-units 2`, I didn't include it in the documentation because I didn't deem it feature-worthy. I think you were too kind with your description, so I changed it a bit.

That's pretty much it. Other than perhaps future improvements that may come up, there's not anything else left to say. The main issue as of now is having the entire world behave in sync with the same speed, which is really a minor nuisance at worst.

Thank you for your great support in this.

PS: Not sure why the docs build failed, something about bad indentation but I don't see the problem.

lethosor:
Yeah, RST is picky about newlines. Generally, they're needed before and after code blocks, and before and after lists. I just re-added all of the ones that 4bba3a0541e2edf6333709f1e2ad8ecd37d5410b removed. I *really* wish I could figure out a way to make line numbers in those errors correct. Right now, they're relative to the `--[====[` line (and 1-indexed), which makes it hard to count them exactly.

As for `-units 2`, I just thought the note would be better-suited for the docs than a comment. Thanks for rewording!

lethosor (approved):

PR #198, opened 2020-09-21
Add script to help "unstick" military dwarves' uniform equiping.

roccomoretti:
One of the long-standing issues with military dwarves is that they can often get "stuck" when putting on their uniform, such that they never actually equip their uniform items. This script is an attempt to address that, in a minimally "cheaty" way.

One cause is of the issue is that dwarves are not smart enough to remove a non-uniform piece from a body part to put on their uniform piece. This script can cause dwarves to drop onto the floor all items on a body part which is missing a uniform piece. Another cause is when the assigned item is held by someone or something else, so the script can also force the assigned items to be removed from whatever inventory/container they're currently in and be placed on the floor. The script does not attempt to assign an item if one isn't already (i.e. if the green check is missing from the military equip screen), and only prints info by default.

This script can achieve much the same thing as the "replace clothing" designation, but with the twin benefits that a) it will only cause disrobing for body parts which are missing uniform items (and not ones where uniform and clothing peacefully coexist) and b) it only functions where there is an assigned uniform item, and doesn't force your dwarves to go barefoot until the armorer finishes making high boots.

In my limited testing, the script is successful in prompting dwarves to pick up the needed armor, though it can take a while and several applications if the dwarf is missing many items.

The script is provisionally called "uniforce" (uniform + force), but that reflected mostly the initial rough concept, and may not be the best name for the current approach. Suggestions for a more appropriate name are welcome.

This is my first significant attempt at both Lua scripting and working with DFHack, so please let me know if there's anything I should be doing differently.

### DIFF
--- a/uniforce.lua
+++ b/uniforce.lua
@@ -1,0 +1,319 @@
+-- Force units to take on their uniform.
+local help = [====[
+
+uniforce
+========
+
+Prompt units to reevaluate their uniform, by removing/dropping potentially conflicting worn items.
+
+Unlike a "replace clothing" designation, it won't remove additional clothing
+if it's coexisting with a uniform item already on that body part.
+It also won't remove clothing (e.g. shoes, trousers) if the unit has yet to claim an 
+armor item for that bodypart. (e.g. if you're still manufacturing them.)
+
+By default it simply prints info about the currently selected unit,
+to actually drop items, you need to provide it the -drop option.
+
+The default algorithm assumes that there's only one armor item assigned per body part,
+which means that it may miss cases where one piece of armor is blocked but the other
+is present. The -multi option can possibly get around this, but at the cost of ignoring
+left/right distinctions when dropping items.
+
+In no cases should it cause a uniform item to be removed/dropped. 
+
+Targets:
+:(no target) Force the selected dwarf to put on their uniform.
+:-all        Force the uniform on all military dwarves
+
+Options:
+:(none)      Will simply show identified issues (dry-run)
+:-drop       Will cause offending items to be placed on ground under unit
+:-multi      Be more agressive in removing items, best for when uniforms have muliple items per body part
+]====]
+
+local utils = require('utils')
+
+local validArgs = utils.invert({
+  'all',
+  'drop',
+  'multi',
+  'help'
+})
+
+---- Debugging function
+
+local dumper = require('dumper')
+
+function debug_print_light(entry)
+  dfhack.println("----------------")
+  dfhack.println(tostring(entry))
+  for k,v in pairs(entry) do
+    dfhack.println('    '..k)
+  end
+  dfhack.println("----------------")
+end
+
+function debug_print(entry)
+  dfhack.println("----------------")
+  dfhack.println(tostring(entry))
+  for k,v in pairs(entry) do
+    dfhack.println('    '..k..':  '..tostring(v))
+  end
+  dfhack.println("----------------")
+end
+
+function find_item_by_ID(id)
+  for k,v in pairs(df.global.world.items.all) do
+    if v.id == id then
+      return v
+    end
+  end
+  return nil
+end
+
+function print_item_list(vector)
+  for i,v in pairs(vector) do
+    local item = find_item_by_ID(v)
+    dfhack.println(i.."  "..v.."   "..tostring(item).."  "..utils.getItemDescription(item))
+  end
+end
+
+function print_item_list_verbose(vector)
+  for i,v in pairs(vector) do
+    local item = find_item_by_ID(v)
+    debug_print_light(item)
+  end
+end
+
+-- Unused utility functions
+
+function val_in_vector( vector, value )
+  for i,v in pairs(vector) do
+    if v == value then
+      return true
+    end
+  end
+  return false
+end
+
+-- Functions
+
+function get_item_pos(item)
+  local x, y, z = dfhack.items.getPosition(item)
+  if x == nil or y == nil or z == nil then
+    return nil
+  end
+
+  -- TODO: Base this on valid world positions.
+  if not dfhack.maps.isValidTilePos(x,y,z) then
+    dfhack.println("NOT VALID TILE")
+    return nil
+  end
+  if not dfhack.maps.isTileVisible(x,y,z) then
+    dfhack.println("NOT VISIBLE TILE")
+    return nil
+  end
+--  if x < -2000 or y < -2000 or z < -2000 or
+--     x >  2000 or y >  2000 or z >  2000 then
+--     return nil 
+--  end
+  return {x=x, y=y, z=z}
+end
+
+function find_squad_position(unit)
+  for i, squad in pairs( df.global.world.squads.all ) do
+    for i, position in pairs( squad.positions ) do
+      if position.occupant == unit.hist_figure_id then
+        return position
+      end
+    end
+  end
+  return nil
+end
+
+-- This should be okay for dwarves, not sure if valid for others.
+PART_TO_POSITION = {}
+PART_TO_POSITION[ 0]="body"
+PART_TO_POSITION[ 1]="pants"
+PART_TO_POSITION[ 3]="head"
+PART_TO_POSITION[ 8]="gloves" -- also shield and weapon
+PART_TO_POSITION[ 9]="gloves" -- also shield and weapon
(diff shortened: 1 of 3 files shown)


uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+PART_TO_POSITION[15]="shoes"
+
+-- Convert the unit.inventory.body_part_id to a squad_position.uniform body position
+function body_part_to_body_position(part)
+  return PART_TO_POSITION[ part ]
+end
+
+-- Will figure out which items need to be moved to the floor, returns an item_id:item map
+function process(unit, args)
+  silent = args.all -- Don't print details if we're iterating through all dwarves
```
myk002, line 105: missing 'local'?

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+    if loc ~= 'shield' and loc ~= "weapon" then -- shields and weapons don't "cover" the bodypart they're assigned to.
+     covered[ worn_parts[ id ] ] = id
+    end
+  end
+
+  if multi then
+    covered = {} -- Don't consider current covers - drop for anything which is missing
+  end
+
+  -- Figure out body parts which should be covered but aren'y
```
myk002, line 181: aren'y

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+
+function do_drop( item_list )
+  if item_list == nil then
+    return nil
+  end
+
+  for id, item in pairs(item_list) do
+    local pos = get_item_pos(item)
+    if pos == nil then
+      dfhack.println("Could not find drop location for item #"..id.."  "..utils.getItemDescription(item))
```
myk002, line 216: dfhack.println is aliased to 'print', which is shorter and cleaner

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+
+local args = utils.processArgs({...}, validArgs)
+
+if args.help then
+    print(help)
+    return
+end
+
+if (args.drop or args.free) and df.global.ui.main.mode == df.ui_sidebar_mode.ViewUnits then
+  dfhack.println("Error: Cannot actually drop/free items when view-unit sidebar is open.")
```
myk002, line 239: you can set the mode and update the sidebar instead of failing here. just restore the original mode before returning. see https://github.com/myk002/scripts/blob/master/internal/quickfort/query.lua#L100 for an example

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+In no cases should the command cause a uniform item that is being properly worn to be removed/dropped.
+
+Targets:
+:(no target): Force the selected dwarf to put on their uniform.
+:-all:        Force the uniform on all military dwarves.
+
+Options:
+:(none):      Simply show identified issues (dry-run).
+:-drop:       Cause offending worn items to be placed on ground under unit.
+:-free:       Remove to-equip items from containers or other's inventories, and place on ground.
```
myk002, line 35: does this interact cleanly with items that are being hauled?

uniforce.lua
lethosor: Also: convention for fatal errors is `qerror`, which will throw an error without a traceback (and terminate the script)

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+    end
+  end
+else
+  local unit=dfhack.gui.getSelectedUnit(true)
+  if df.isvalid(unit) then
+    local to_drop = process(unit,args)
+    do_drop( to_drop )
+  else
+    dfhack.println("No unit is selected")
+  end
```
lethosor, line 257: You could also use qerror here, but my suggestion would be to remove the argument to getSelectedUnit, which will cause it to print an error message, then only continue if a unit is set:
```suggestion
  local unit = dfhack.gui.getSelectedUnit()
  if unit then
    local to_drop = process(unit,args)
    do_drop( to_drop )
  end
```

uniforce.lua
lethosor: `dfhack.printerr` could also be an alternative for messages that are errors (it does not have an alias, though). I'm guessing `qerror` would not be appropriate since you don't want to actually raise an error here (since you're processing a list).

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+
+  if not dfhack.maps.isValidTilePos(x,y,z) then
+    dfhack.println("NOT VALID TILE")
+    return nil
+  end
+  if not dfhack.maps.isTileVisible(x,y,z) then
+    dfhack.println("NOT VISIBLE TILE")
+    return nil
+  end
+  return {x=x, y=y, z=z}
```
lethosor, line 65: There's a built-in function for this: `xyz2pos(x, y, z)`

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+    for i, position in pairs( squad.positions ) do
+      if position.occupant == unit.hist_figure_id then
+        return position
+      end
+    end
+  end
+  return nil
+end
+
+-- This should be okay for dwarves, not sure if valid for others.
```
lethosor, line 88: Probably unlikely that it'll work for _all_ other races, although humans might be similar enough. It would be great if this could be looked up somehow, but I'm no expert on this. It appears that the caste raws have quite a bit of detail about body parts/layers/etc, accessible through e.g. `df.creature_raw.find(unit.race).caste[unit.caste].body_info` - maybe some of that is relevant. For instance, `body_info.body_parts[15]` of an unmodded dwarf has `category = "FOOT"` and `flags.STANCE = true` (as well as two unidentified flags set, 28 and 29).

lethosor (commented):
Cool! Thanks for handling documentation, and fixing the linter stuff.
Regarding naming, I'm also not sure what a good alternative to `uniforce` is. I thought of `reset-uniform`, but that might sound like `clear-uniform` to some people, which isn't what this is doing. We want it to be discoverable from the script listing, of course, so I think having the word `uniform` in there would be good (even though it's almost there, some people could latch onto the `force` part, since there's an existing `force` script). Naming is hard...

uniforce.lua
```diff
@@ -0,0 +1,259 @@
...
+    return nil
+  end
+  if not dfhack.maps.isTileVisible(x,y,z) then
+    dfhack.println("NOT VISIBLE TILE")
+    return nil
+  end
+  return {x=x, y=y, z=z}
+end
+
+function find_item_by_ID(id)
```
AtomicChicken, line 68: This function can be replaced with `df.item.find(id)`.

uniforce.lua
roccomoretti: It uses `dfhack.items.moveToGround()`, so it should work as cleanly as that does. (And if, for some reason, that function doesn't work cleanly, then I'd probably argue the fix should be in that function.)

roccomoretti:
I've updated the script based on the suggestions (thank you, all). The biggest change is probably the reworking to use the `bodyparts_that_can_wear(unit, item)` function rather than the uniform armor location designation.

I've also renamed the script `uniform-unstick`. Again, not necessarily the best name, but hopefully slightly better than it was.

uniform-unstick.lua
```diff
@@ -0,0 +1,277 @@
...
+
+function bodyparts_that_can_wear(unit, item)
+
+  local bodyparts = {}
+  local unitparts = df.creature_raw.find(unit.race).caste[unit.caste].body_info.body_parts
+
+  if item._type == df.item_helmst then
+    for index, part in pairs(unitparts) do
+      if part.flags.HEAD then
+        bodyparts[#bodyparts+1] = index
```
lethosor, line 91: `table.insert` is a built-in function that does what you want here (and below): 
```lua
        table.insert(bodyparts, index)
```

lethosor (commented):
Thanks for tackling this! `uniform-unstick` sounds good to me.

lethosor (approved):
Looks good, and thanks! I fixed a bug I spotted in a fort - hopefully that was all.

I also made a couple documentation tweaks - specifically, the lists starting with `:` (ReStructuredText calls them "definition lists") needed to have a blank line before them, to keep them from looking like this:
image

I wish there were a more obvious warning/error for things like these, but as-is, you'd have to build the documentation to catch them. We have a page [here](https://docs.dfhack.org/en/latest/docs/Documentation.html) that goes into some detail about how to do that, if you're interested, although it's a bit involved.

PR #203, opened 2020-10-01
Convert position args to number

xtmprsqzntwlfb:
Make teleport work with coordinates again.

### DIFF
--- a/teleport.lua
+++ b/teleport.lua
@@ -73,7 +73,7 @@ if args.showunitid or args.showpos then
  end
 else
  local unit = tonumber(args.unit) and df.unit.find(tonumber(args.unit)) or dfhack.gui.getSelectedUnit(true)
- local pos = not(not args.x or not args.y or not args.z) and {x=args.x,y=args.y,z=args.z} or {x=df.global.cursor.x,y=df.global.cursor.y,z=df.global.cursor.z}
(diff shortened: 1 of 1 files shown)


lethosor (approved):
Were you getting an error here? I tried `teleport -x 52 -y 43 -z 178` with scripts from 436b98f6b2dd8a8a845e5eccaeaecd5024dc1b9f and it worked for me.

Thanks for the PR, at any rate!

lethosor:
Oh, it looks like this is a cherry-pick of 2869085cf00b3d4c9716c6c197df11dc2448a1b8, right?
Judging by the date on that commit (Feb 13, 2017), you must have made it between DFHack 0.43.05-beta1 and 0.43.05-beta2, and the latter added support for implicit conversions to integers in https://github.com/DFHack/dfhack/commit/fa1adbbac4b3c1312587440684729652f89b3b73 to address many similar issues we were seeing in other Lua scripts/files, like https://github.com/DFHack/dfhack/issues/1044. That must be why I'm not seeing an error in 0.47.04-r2.
It looks to me like this is no longer necessary, then, but thank you for spotting it and making a PR!

xtmprsqzntwlfb:
Oh yes, in 0.47.04-r2 teleport works, this patch is indeed helpful only for for 0.43. Thank you!

PR #207, opened 2020-10-12
add flag to allow arbitrary query blueprints

myk002:
since I'm sure there are some out there that I didn't plan for

this is also useful for testing where I use query blueprints to mimic creating buildings from the UI

quickfort.txt config file modified in DFHack/dfhack#1670


lethosor (approved):
Thanks for the changelog update!

PR #210, opened 2020-10-23
[quickfort] Speed up first call to quickfort list

myk002:
This PR can be reviewed after #209 is merged.

On the first call to `quickfort list`, quickfort has to scan all the blueprints in the `blueprints/` directory tree to discover and cache the blueprint modelines. If the player's blueprint library is large, this can take a while.

This change significantly improves the performance of the first call to `quickfort list`. For the Starter Pack's blueprint corpus on my system (plus a few copies of dreamfort that I was testing), latency was reduced from 2.5 seconds to 0.5 seconds.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,6 +18,7 @@ that repo.
 - `quickfort`: fix misconfiguration of nest boxes, hives, and slabs that were preventing them from being built from build blueprints
 
 ## Misc Improvements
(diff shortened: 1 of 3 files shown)


myk002:
there is no other target branch to jiggle, so the commit list is sadly out of date (unless it autoupdates sometime after I write this). The relevant commits are just the last two: f518122 and 187c7a8

lethosor (approved):
Looks good (yeah, this repo just has one long-lived branch, but I can check diffs locally too)

PR #212, opened 2020-10-23
[quickfort] support for bookcases, display cases, and offering places

myk002:
Requires changes to dfhack/buildings.lua in DFHack/dfhack#1683


lethosor (approved):
Looks good - I'm assuming the DFHack PR was the only blocker, but I'll let you confirm

PR #216, opened 2020-11-04
[quickfort] display missing blueprint labels in dialog list output

myk002:
and only quote strings if they contain spaces. We only have 80 columns or so and we need to scrounge space.


myk002:
a fix for the gui dialog that I think we should include in -r4

lethosor (approved):

lethosor:
I happened to have a sheet with spaces in its name, and that shows up and is quoted now, so :+1:

(And I'm open to including pretty much anything in a release as long as it's not risky/broken/WIP)

myk002:
Thanks! I'm just trying to be frugal with PRs right now.. Self-imposed code freeze. I've seen too many "one last commit"s that end up biting me in the bum ^_^

PR #217, opened 2020-11-08
modtools/add-syndrome now allows adding by id

wolfboyft:


lethosor (approved):
Thanks! I assume this is necessary because not all syndromes have names (or unique names), yes?

FYI, `df.syndrome.find(ID)` is faster than a linear search when the ID is known, so I'll try to change to use that - should be pretty easy with how you've set it up.

lethosor:
For reference: f2c5775 is the change I made (as well as a changelog update in a2b0c7c)

PR #223, opened 2020-11-13
rewrite unsuspend and autounsuspend in lua

myk002:
Fixes DFHack/dfhack#1702
so we can properly check for buildingplan-managed buildings


autounsuspend.lua
```diff
@@ -0,0 +1,34 @@
...
+    print('syntax: autounsuspend [start|stop]')
+end
+
+local function stop()
+    dfhack.run_command{'repeat', '-cancel', job_name}
+    print('autounsuspend Stopped.')
+end
+
+local function start()
+    dfhack.run_command{'repeat', '-name', job_name, '-time', '1',
```
lethosor, line 22: I would actually suggest using the `repeat-util` module for this (which is what repeat uses) - it would avoid the need to construct these commands by hand.

And as a side note, `run_script()` has somewhat less overhead than `run_command()` (but only works for Lua scripts)

changelog.txt
```diff
@@ -19,6 +19,7 @@ that repo.
 - `quickfort`: fix valid placement detection for floor hatches, floor grates, and floor bars (they were erroneously being rejected from open spaces and staircase tops)
 - `quickfort`: fix query blueprint statistics being added to the wrong metric when both a query and a zone blueprint are run by the same meta blueprint
 - `quickfort`: display missing blueprint labels in gui dialog list
+- `autounsuspend` and `unsuspend`: now both leave buildingplan-managed buildings alone
```
lethosor, line 22: They worked before the buildingplan changes, right? Unless they were broken in r3, this can probably be left out of the changelog.

myk002: autounsuspend did, but unsuspend did not. updated changelog.

autounsuspend.lua
myk002: done. thanks!

changelog.txt
lethosor: Sure enough - thanks!

lethosor (approved):
Thanks for handling this!

PR #225, opened 2020-11-17
FIX: handle "flour" food type preference.

acattaneo88:
This small correction fixes a problem when adding flours as a food type preference to a unit.

Relevant forum post:
[http://www.bay12forums.com/smf/index.php?topic=110027.msg8205633#msg8205633](http://www.bay12forums.com/smf/index.php?topic=110027.msg8205633#msg8205633)


lethosor (approved):
Thanks for the fix! Left a comment regarding your `FIXME` note, but not urgent.

PR #226, opened 2020-11-19
Update the script to work with the new unit structure format.

acattaneo88:


assign-goals.lua
```diff
@@ -99,9 +99,10 @@ function assign(goals, unit, reset)
         assert(type(realized) == "boolean")
         goal = goal:upper()
         if df.goal_type[goal] then
-            utils.insert_or_update(unit.status.current_soul.personality.dreams,
-                                   { new = true, type = df.goal_type[goal], unk8 = realized and 1 or 0 },
+            local _, _, pos = utils.insert_or_update(unit.status.current_soul.personality.dreams,
+                                   { new = true, type = df.goal_type[goal] },
```
lethosor, line 103: Recursive assignment is supported here, so I would suggest adding `flags = {accomplished = realized}` to the table here instead.

acattaneo88: Done!

lethosor (approved):
Thanks for the fix!

lethosor:
Regarding d5abdb2 - was this a pull request to update your fork with the latest changes from this repo? (It appears to have been deleted from your fork, so I'm not quite sure.)

If that's the case, I would suggest the workflow described [here](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) to update branches of your fork when necessary. It will still require pulling from this repo and pushing to yours, but it will typically avoid the extra merge commit that pull requests create when they are merged through GitHub.

acattaneo88:
Yes it was, you are right. I will implement the workflow you suggested, thanks for the help!

PR #227, opened 2020-11-20
Update script to handle a "fresh" unit with no adaptation suffered yet.

acattaneo88:
I don't know if modifying others' scripts is considered impolite, if so I apologize. Feel free to disregard this PR.

I added some code to handle the case where the trait "CaveAdapt" does not exist: before my addition the script would have simply ignored the case (I think), while now it can add a new cave adaptation trait to the unit.
Also, "print" statements were being ignored, so I changed those too.

### DIFF
--- a/adaptation.rb
+++ b/adaptation.rb
@@ -64,15 +64,12 @@ set_adaptation_value = lambda { |u,v|
     next if u.flags2.killed
     u.status.misc_traits.each { |t|
         if t.id == :CaveAdapt
-            # TBD: expose the color_ostream console and color values of
-            # t.value based on adaptation level
             if mode == 'show'
                 if df.respond_to?(:print_color)
-                    print "Unit #{u.id} (#{u.name}) has an adaptation of "
+                    df.print_color(COLOR_WHITE, "Unit #{u.id} (#{u.name}) has an adaptation of ")
                     case t.value
                     when 0..399999
-                        #df.print_color(COLOR_GREEN, "#{t.value}\n")
-                        print "#{t.value}\n"
+                        df.print_color(COLOR_GREEN, "#{t.value}\n")
                     when 400000..599999
                         df.print_color(COLOR_YELLOW, "#{t.value}\n")
                     else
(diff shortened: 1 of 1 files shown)


adaptation.rb
```diff
@@ -64,15 +64,12 @@ def usage(s)
     next if u.flags2.killed
     u.status.misc_traits.each { |t|
         if t.id == :CaveAdapt
-            # TBD: expose the color_ostream console and color values of
-            # t.value based on adaptation level
             if mode == 'show'
                 if df.respond_to?(:print_color)
-                    print "Unit #{u.id} (#{u.name}) has an adaptation of "
+                    df.print_color(COLOR_WHITE, "Unit #{u.id} (#{u.name}) has an adaptation of ")
```
lethosor, line 69: This should use COLOR_RESET instead, which you'll probably need to define as `-1` above. White and black should be avoided because they are the default background colors of some terminals (e.g. the default color scheme of macOS's Terminal.app has a white background).

lethosor (commented):
Changes to scripts you didn't originally write are perfectly fine - that's part of the reason why we have a centralized repo for them!

If you have multiple changes in mind (seeing how you've submitted a few PRs recently), I would suggest making your changes on separate branches, e.g. branches named `adaptation` and `assign-goals` in your repo (or something more descriptive if you like). This would allow you to have multiple PRs with different changes open at the same time.

adaptation.rb
acattaneo88: I actually thought about using COLOR_RESET... I forgot to do it along the way... *facepalm

adaptation.rb
```diff
@@ -86,8 +83,22 @@ def usage(s)
                 t.value = v
                 num_set += 1
             end
+            return
```
lethosor, line 86: I'm getting this error here:
```
E: LocalJumpError: ../../code/dfhack/scripts/adaptation.rb:86: unexpected return
```
Are you trying to avoid checking subsequent entries in `misc_traits`? If so, you might want `break` instead.

I see that changing it to `break` results in duplicate output lines, so I'm guessing you're trying to skip the section below. If this `return` works on your end, I'm guessing the issue is that it doesn't work in Ruby 1.8, which is what we use on Linux. Upgrading Ruby to a consistent version on all platforms is a longstanding issue we should address, but I think a workaround in this case could be to set a flag here and check it before printing below.

acattaneo88: Fixed, I used a flag. Also tested on Linux and everything seems to be working now. Sorry about that.

lethosor: Looks good, thanks!

adaptation.rb
lethosor: By the way, looking at the [implementation](https://github.com/DFHack/dfhack/blob/fb44b26b47eca729646e316efc22cfe2a138691e/plugins/ruby/ruby.cpp#L575-L585), it appears that print_color does already reset the color - but I'm fine with explicitly resetting it for clarity between other print_color calls.

PR #229, opened 2020-11-21
assign-profile: clean up code and clarify documentation

acattaneo88:
There were some things I didn't like in the previous version of this script, and so I did a little bit of refactoring to update the code to what I think it's a more _dfhacky_ style and to improve (I hope) the documentation.

### DIFF
--- a/assign-profile.lua
+++ b/assign-profile.lua
@@ -64,8 +64,8 @@ Examples:
 
 * Loads and applies the profile called "ARCHER" in the provided (fictional) json,
   keeping all the old characteristics but the attributes and the skills, which
-  will be reset (and then, if the profile provides some attributes values, those
-  new values will be applied)::
+  will be reset (and then, if the profile provides some attributes or skills values,
+  those new values will be applied)::
 
     assign-profile -file /hack/scripts/military_profiles.json -profile ARCHER -reset [ ATTRIBUTES SKILLS ]
 
--- a/assign-profile.lua
+++ b/assign-profile.lua
@@ -104,7 +104,7 @@ local default_filename = "/hack/scripts/dwarf_profiles.json"
 --luacheck: in=string[],df.unit,bool[]
 function apply_profile(profile, unit, reset)
     assert(type(profile) == "table")
-    assert(not unit or type(unit) == "number" or type(unit) == "userdata")
+    assert(not unit or type(unit) == "number" or df.unit:is_instance(unit))
     assert(not reset or type(reset) == "table")
 
     reset = reset or {}
--- a/assign-profile.lua
+++ b/assign-profile.lua
@@ -36,49 +36,51 @@ Usage:
                     the profile to apply. It's the name of
                     the profile as stated in the json file.
 
-``-reset <list of characteristics>``:
+``-reset [ <list of characteristics> ]``:
                     the characteristics to be reset/cleared. If not present,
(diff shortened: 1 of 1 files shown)


assign-profile.lua
```diff
@@ -92,36 +94,25 @@ local scripts = {
...
+---             an argument to the ``assign`` function of the script related to the characteristic.
+---             See the called script documentation for more details.
+---   :unit: a valid unit id, a df.unit object, or nil. The called script has the responsibility to validate this value.
+---   :reset: nil, or a list of values. See this script documentation for accepted values.
 --luacheck: in=string[],df.unit,bool[]
-function apply_profile(profile, unit, reset_table)
-    assert(not profile or type(profile) == "table")
+function apply_profile(profile, unit, reset)
+    assert(type(profile) == "table")
     assert(not unit or type(unit) == "number" or type(unit) == "userdata")
```
lethosor, line 107: The last check could be `df.unit:is_instance(unit)` - this is safe for non-userdata types as well.

assign-profile.lua
```diff
@@ -155,63 +146,40 @@ end
...
-            while args[i + 1] and not contains(valid_args, args[i + 1]) do
-                i = i + 1
-                reset_table[args[i]:upper()] = true
-            end
-        else
-            qerror("'" .. arg .. "' is not a valid argument.")
+    local unit
+    if args.unit then
+        unit = tonumber(args.unit)
+        if not unit then
```
lethosor, line 159: You could also use `df.unit.find(unit)` here to catch IDs that are valid numbers but don't refer to valid units (although I suspect the scripts you're calling out to do that already).

lethosor (approved):
Looks good (sorry for the delay) - I left a couple minor comments for other "DFHack-y" things that I thought of, but the improvements look good in general (glad you were able to mostly maintain the previous syntax with `processArgs()`!).

Any user-facing changes here that you think should be mentioned in the changelog?

assign-profile.lua
acattaneo88: Yes, the called scripts already do that in their `assign` function.

My reasoning is: the `assign` function in each script can be called from external scripts, so we should check if the arguments passed to it are valid.

The `main` function in each script exists to allow calling the scripts directly via the DFHack console, and so (in my opinion) should just quickly check the correct formatting of the arguments but should not be responsible for the actual validity of the values passed to it, because it's going to call the `assign` function anyway, which will perform the required checks.

Of course, I will modify the code as requested.

assign-profile.lua
acattaneo88: Fixed!
I should change it in my other scripts as well, so I will probably open a single PR with that single change for every script, otherwise I can open a PR for each script, as you prefer.

About the changelog: the main difference from the previous script is the way the list of "reset" options is passed while calling the script via the DFHack console (it now requires the arguments, if more than one, to be enclosed between square brackets). The documentation is changed accordingly, so if someone uses the old syntax to pass a list will receive an error, and in this case they will probably just go and check the documentation again to discover the new syntax. I don't think it's worth mentioning in the changelog.

assign-profile.lua
lethosor: Makes sense - thanks for the details!

assign-profile.lua
lethosor: A single PR for that sort of thing (assuming the diff isn't massive) is fine with me.

PR #230, opened 2020-11-25
Fixed advfort hatches (which were broken)

wolfboyft:
Also added support for pressure plates, maybe...?

### DIFF
--- a/gui/advfort.lua
+++ b/gui/advfort.lua
@@ -1109,7 +1109,7 @@ function LinkBuilding(args)
         end
     end
     if lever_bld==nil then
-        if bld:getType()==df.building_type.Trap and bld:getSubtype()==df.trap_type.Lever or bld:getSubtype()==df.trap_type.PressurePlate then
+        if bld:getType()==df.building_type.Trap and (bld:getSubtype()==df.trap_type.Lever or bld:getSubtype()==df.trap_type.PressurePlate) then
             lever_id=bld.id
             dfhack.gui.showAnnouncement("Selected trigger for linking",COLOR_YELLOW,true)
             return
--- a/gui/advfort.lua
+++ b/gui/advfort.lua
@@ -1091,7 +1091,7 @@ function fake_linking(lever,building,slots)
 
     lever.linked_mechanisms:insert("#",item2)
     --fixes...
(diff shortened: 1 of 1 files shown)


wolfboyft:
Oops, that should be "deselected trigger" not lever

wolfboyft:
RE pressure plates, a fortress-made water-triggered pressure plate just opened a hatch (fortress-linked, so idk if linking in adventure works but it very likely does) in adventure mode when i dropped some water on it, so it seems it's just the stepping-on-them-as-an-adventurer that doesn't work.

And building them won't work either because we can only build triggerless plates with advfort. So I'll maybe make a contribution for that sometime :>

warmist:
AFAIK stepping on pressure plates is intentionally disabled in adv mode. I think it's related to cage traps and stuff and/or your adventurer "knowing" the trap and intentionally not stepping on it.

wolfboyft:
Oops. Didn't mean to do that.
Anyway I'll fix the error the test mentions and the "lever" thing.

gui/advfort.lua
```diff
@@ -1109,16 +1109,17 @@ function LinkBuilding(args)
         end
     end
     if lever_bld==nil then
-        if bld:getType()==df.building_type.Trap and bld:getSubtype()==df.trap_type.Lever then
+        if bld:getType()==df.building_type.Trap and bld:getSubtype()==df.trap_type.Lever or bld:getSubtype()==df.trap_type.PressurePlate then
```
lethosor, line 1112: You should use parentheses here - `A and B or C` is equivalent to `(A and B) or C`, but you want `A and (B or C)`

wolfboyft: Oops, you're right.

wolfboyft:
Anything else need fixing?

lethosor:
Nope, I just haven't had much time to try this out. Looks simple enough from a static review, though. Have you tested linking doors and hatches as well as other building types?

wolfboyft:
Yee, & all seems good & cool!
Hatches and doors seem to be the only buildings that need a flag flipped to properly work with mechanisms (is bc they can work s/o)

lethosor (approved):

PR #231, opened 2020-12-04
Assert df.unit type using the specific is_instance check.

acattaneo88:
As the title says: given a "unit" passed as an argument to the `assign` function of each of these scripts, use `df.unit:is_instance(unit)` instead of `type(unit) == "userdata"` to assert its type.


lethosor (approved):
Looks good! (And thanks for switching over to dedicated branches for each PR - that's looking good too)

PR #234, opened 2020-12-11
[quickfort] gui dialog fixes

myk002:
- prevent two instances of the dialog from being opened at the same time
This can happen easily if you just run `quickfort gui` twice. One dialog will appear on top of the other and they must be dismissed separately. This is a particular problem when you try to run a query blueprint with a second dialog open, since that dialog will snarf all the keystrokes.

- prevent the details page from being opened or 'orders' from triggering if there is no list item selected
This can happen if the user puts in a search filter that doesn't match anything, then hits the right arrow key or alt-o

### DIFF
--- a/internal/quickfort/dialog.lua
+++ b/internal/quickfort/dialog.lua
@@ -207,7 +207,7 @@ local function dialog_command(command, text)
 end
 
 function do_dialog()
-    if blueprint_dialog then qerror('Quickfort gui dialog already open') end
+    if blueprint_dialog then blueprint_dialog:dismiss() end
     blueprint_dialog = BlueprintDialog{
         frame_title='Select quickfort blueprint',
         with_filter=true,
--- a/internal/quickfort/dialog.lua
+++ b/internal/quickfort/dialog.lua
@@ -15,6 +15,7 @@ local quickfort_parse = reqscript('internal/quickfort/parse')
 local min_dialog_width = 73
 
 -- persist these between dialog invocations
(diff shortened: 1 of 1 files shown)


internal/quickfort/dialog.lua
```diff
@@ -202,7 +207,8 @@ local function dialog_command(command, text)
 end
 
 function do_dialog()
-    local blueprint_dialog = BlueprintDialog{
+    if blueprint_dialog then qerror('Quickfort gui dialog already open') end
```
lethosor, line 210: It may be worth having this dismiss the existing dialog instead of raising an error - in the (unlikely) event that the dialog gets into a broken state and doesn't respond to `Esc`, users that don't know about `devel/pop-screen` might try running `quickfort gui` again to reset it. (`gui/dfstatus` takes a slightly different approach and dismisses its dialog when run a second time, but I don't think that makes as much sense here.)

myk002: This is a good idea. Thanks! I'll test, but I don't think the dismissed viewscreen on the stack is going to be a problem. I always call getViewscreen(true) to skip dismissed screens, and the df ui should work just fine with dismissed screens on the stack.

lethosor: Yeah, from a closer look at the logic, it looks like you're already feeding keys immediately after the dialog's `dismiss()` is called, so it would likely still be on the stack, and a second dismissed dialog on the stack shouldn't make a difference.

lethosor (approved):

PR #235, opened 2020-12-15
Add modtool: dynamic-adv-reactions

Atkana:
The tool gives options to modders to restrict adventurers from using particular adventurer reactions (normally in vanilla, absolutely all reactions that are flagged as for adventure mode are available to the player). It's, uh, a pretty niche tool, but provides some options for modding possibilities (and it's what my dumb brain decided it wanted me to make :P).

In the future when the structures for the adventure mode actions menu are more mapped out, the script's method of altering the menu options should probably be updated to something less hacky.

### DIFF
--- a/modtools/dynamic-adv-reactions.lua
+++ b/modtools/dynamic-adv-reactions.lua
@@ -1,0 +1,882 @@
+-- Tool to allow dynamically enabling / disabling adventure mode reactions based on circumstances
+--@ module = true
+-- Created for dfhack version 0.44.12-r3
+
+--[[ HOW IT WORKS:
+Commands / scripts register a set of conditional checks associated with a reaction (the set of conditions is referred to a "condition blob").
+When the player opens up the actions menu in game (the menu where they can perform reactions), the script checks if they satisfy one of the condition blobs associated with that reaction (that is, they satisfy every condition in any of the registered blobs for that reaction).
+If they don't satisfy all the conditions for any of the condition blobs, the reaction is removed from their list of reactions.
+]]
+
+--[[ TODO / WISHLIST:
+- Time: Available within certain hourly bounds, allowing for day/night requirements.
+- Season. Available during a particular season.
+- Weather. Available during certain weathers.
+]]
+
+local help = [====[
+
+modtools/dynamic-adv-reactions
+==============================
+Tool to allow dynamically enabling / disabling adventure mode reactions based on certain conditions.
+When a reaction is registered with this script, its usage by adventurers will be disabled by default. Each usage of this command creates its own set of conditions to check, and the reaction will be enabled for the player if they satisfy all the criteria of one of those sets of conditions.
+Registered reactions are cleared when the world is unloaded, and so it is best to include any commands for this within an `onLoad*.init` file.
+Checks with multiple instances of the same type of condition are possible if registered via code rather than with commands.
+
+:-reaction reactionCode:
+  (Required)
+  Specify the reaction the conditions will allow.
+  Examples:
+    MAKE_SHARP_ROCK
+    "ASSEMBLE STONE AXE"
+
+Conditions:
+Most conditions can optionally be a table of conditions, rather than a single argument. That condition is satisfied if any of them are met.
+
+:-creature token:
+  Require the adventurer to be the specified creature.
+  Can optionally be a table of arguments.
+  Example: DWARF
+
+:-caste token:
+  Require the adventurer to be the specified caste.
+  Can optionally be a table of arguments.
+  Examples:
+    FEMALE
+    DRONE
+
+:-creatureCaste tokens:
+  Require the adventurer to be the specified creature + caste combination.
+  Can optionally be a table of arguments.
+  Example: DWARF:MALE
+
+:-creatureClass class:
+  Require the adventurer to be a creature with the specified creature class.
+  Can optionally be a table of arguments.
+  Examples:
+    GENERAL_POISON
+    MAMMAL
+
+:-sphere sphere:
+  Require the adventurer to be a creature with the specified sphere.
+  Can optionally be a table of arguments.
+  Examples:
+    METALS
+    MUCK
+
+:-syndrome name:
+  Require the adventurer to be under the effects of the specified syndrome.
+  Can optionally be a table of arguments.
+  Examples:
+    "giant cave spider bite"
+    "night sickness"
+
+:-syndromeClass class:
+  Require the adventurer to be under the effects of a syndrome with the specified SYN_CLASS.
+  Can optionally be a table of arguments.
+  Example: RAISED_UNDEAD
+
+:-syndromeIdentifier identifier:
+  Require the adventurer to be under the effects of a syndrome with the specified SYN_IDENTIFIER.
+  Can optionally be a table of arguments.
+  Example: INEBRIATION
+
+:-secret id:
+  Require the adventurer to know the specified secret.
+  Can optionally be a table of arguments.
+  Since game-generated secrets have unique names, this is more likely for use with mod-made secrets, which keep the same ids between worlds.
+
+:-item token:
+  Require the adventurer to be holding the specified item in their hands.
+  Can optionally be a table of arguments.
+  Examples:
+    WEAPON:ITEM_WEAPON_PICK
+    BUCKET:NONE
+
+:-itemType token:
+  Require the adventurer to be holding the an item of the specified item type in their hands.
+  Can optionally be a table of arguments.
+  Examples:
+    WEAPON
+    COIN
+    BOX
+
+:-moon [ startPhase endPhase ]:
+  Require the moon to be within a set bound of phases.
+  See the wiki entry for what numbers correspond to which phases: https://dwarffortresswiki.org/index.php/DF2014:Syndrome#MOON_PHASE
+
+Example Usage:
+Restrict stone axe making to only drunken elves:
+  `modtools/dynamic-adv-reactions -reaction "ASSEMBLE STONE AXE" -creature ELF -syndromeIdentifier INEBRIATION`
+Only allow beings related to an appropriate to sharpen rocks:
+  `modtools/dynamic-adv-reactions -reaction MAKE_SHARP_ROCK -sphere [ EARTH MINERALS METALS ]
+]====]
+
+local utils = require "utils"
+local gui = require("gui")
+
+local validArgs = utils.invert({
+  "help",
+  "reaction",
+  "creature",
+  "caste",
+  "creatureCaste",
+  "creatureClass",
+  "sphere",
+  "syndrome",
+  "syndromeClass",
+  "syndromeIdentifier",
+  "secret",
+  "item",
+  "itemType",
+  "moon",
+})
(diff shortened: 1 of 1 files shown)


lethosor (commented):
There are a couple issues with the documentation that come up (mostly issues with ReStructuredText formatting) - unfortunately the GitHub actions output only listed one of them, which I fixed in 254f27309088cae50b2d7ae29eaabd6572f34cd3, but those likely won't show up on this PR until it's updated. For reference, here's the complete list I get locally from [building the documentation](https://docs.dfhack.org/en/latest/docs/Documentation.html#building-the-documentation):

```
scripts/modtools/dynamic-adv-reactions.lua:14: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:29: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:41: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:48: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:55: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:77: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:84: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:94: WARNING: Unexpected indentation.
scripts/modtools/dynamic-adv-reactions.lua:95: WARNING: Block quote ends without a blank line; unexpected unindent.
scripts/modtools/dynamic-adv-reactions.lua:96: WARNING: Inline interpreted text or phrase reference start-string without end-string.


scripts/modtools/dynamic-adv-reactions.lua:5: WARNING: undefined label: onload*.init (if the link has no caption the label must precede a section header)
scripts/modtools/dynamic-adv-reactions.lua:94: WARNING: undefined label: modtools/dynamic-adv-reactions -reaction "assemble stone axe" -creature elf -syndromeidentifier inebriation (if the link has no caption the label must precede a section header)
```

The line numbers appear to be relative to the `local help = [====[` line (since it's on line 17, subtract 16 from the numbers above). I'll comment below with some specifics.

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,882 @@
...
+When a reaction is registered with this script, its usage by adventurers will be disabled by default. Each usage of this command creates its own set of conditions to check, and the reaction will be enabled for the player if they satisfy all the criteria of one of those sets of conditions.
+Registered reactions are cleared when the world is unloaded, and so it is best to include any commands for this within an `onLoad*.init` file.
+Checks with multiple instances of the same type of condition are possible if registered via code rather than with commands.
+
+:-reaction reactionCode:
+  (Required)
+  Specify the reaction the conditions will allow.
+  Examples:
+    MAKE_SHARP_ROCK
+    "ASSEMBLE STONE AXE"
```
lethosor, line 31: This currently looks like this:
image

I suspect changing `Examples:` to `Examples::` will create a code block. Alternatively, you could make this a bulleted list by adding `-` or `*` before each line.

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,882 @@
...
+
+Conditions:
+Most conditions can optionally be a table of conditions, rather than a single argument. That condition is satisfied if any of them are met.
+
+:-creature token:
```
lethosor, line 36: I realize some existing documentation uses this formatting, but some combination of long option text, long descriptions, and spaces in the option names is making this almost unreadable when rendered:
image

Instead, I would probably recommend the approach used in scripts like https://github.com/DFHack/scripts/blob/master/armoks-blessing.lua and https://github.com/DFHack/scripts/blob/master/caravan.lua

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,882 @@
...
+    COIN
+    BOX
+
+:-moon [ startPhase endPhase ]:
+  Require the moon to be within a set bound of phases.
+  See the wiki entry for what numbers correspond to which phases: https://dwarffortresswiki.org/index.php/DF2014:Syndrome#MOON_PHASE
+
+Example Usage:
+Restrict stone axe making to only drunken elves:
+  `modtools/dynamic-adv-reactions -reaction "ASSEMBLE STONE AXE" -creature ELF -syndromeIdentifier INEBRIATION`
```
lethosor, line 110: Single backticks make links in RST (which is the cause of the last two warnings in my output in my earlier comment). For a code block, replace the `:` on the line above with `::`, and remove the backticks. You might also need to indent this block with 4 spaces instead of 2, and add a blank line on either side.

Same goes for the block below.

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,882 @@
...
+- Weather. Available during certain weathers.
+]]
+
+local help = [====[
+
+modtools/dynamic-adv-reactions
+==============================
+Tool to allow dynamically enabling / disabling adventure mode reactions based on certain conditions.
+When a reaction is registered with this script, its usage by adventurers will be disabled by default. Each usage of this command creates its own set of conditions to check, and the reaction will be enabled for the player if they satisfy all the criteria of one of those sets of conditions.
+Registered reactions are cleared when the world is unloaded, and so it is best to include any commands for this within an `onLoad*.init` file.
```
lethosor, line 23: Inline code blocks need double backticks (single backticks make links):
```suggestion
Registered reactions are cleared when the world is unloaded, and so it is best to include any commands for this within an ``onLoad*.init`` file.
```

Atkana:
My nemesis the doc formatter strikes again. I've attempted to make the suggested changes, hopefully that fixes it.

It sucks that you have to compromise the readability of the help text so much for the docs. It makes me wonder if somebody using the `-help` command in game will even be able to work out how the commands are supposed to work (without the usage hints at the bottom). It's enough that in the future I might consider having 2 separate help sections in scripts, one for the docs and one for the in-console docs...

lethosor:
RST is intended to be readable in source form - admittedly, the double backticks surrounding examples could be confusing if you haven't seen that before (but so could the colons you had earlier), and links to named anchors in other documents wouldn't be easy to follow in source form (although I don't see any here). I do think the extra newlines around code blocks improve readability in general.

One option RST provides is "custom directives", which would give us some more flexibility over how certain blocks are rendered. For instance, we would be able to start a usage block with `.. usage::`, or something similar, and the text nested under it could be written in a format that's slightly easier to write and read in source form (and also renders well in HTML form). Open to ideas, although I'm not sure if it'll take a lot of work to implement yet - are there any specific changes you made here that you think decrease readability?

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,892 @@
...
+* ``-itemType token``
+  Require the adventurer to be holding the an item of the specified item type in their hands.
+  Can optionally be a table of arguments.
+  Examples::
+
+    WEAPON
+    COIN
+    BOX
+
+``-moon [ startPhase endPhase ]``
```
lethosor, line 110: Missing a bullet here:
```suggestion
* ``-moon [ startPhase endPhase ]``
```

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,892 @@
...
+  Require the moon to be within a set bound of phases.
+  See the wiki entry for what numbers correspond to which phases: https://dwarffortresswiki.org/index.php/DF2014:Syndrome#MOON_PHASE
+
+Example Usage:
```
lethosor, line 114: You generally need a double-newline if you want things to display on a separate line. It looks like this now:
image

You could also make this a subheading by underlining it with `-----` (matching the length of the text) or bold by surrounding it with `**`:
```
Example Usage
-------------
```
or
```
**Example Usage:**

...
```

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,892 @@
...
+
+* ``-reaction reactionCode``
+  (Required)
+  Specify the reaction the conditions will allow.
+  Examples::
+
+    MAKE_SHARP_ROCK
+    "ASSEMBLE STONE AXE"
+
+Conditions:
```
lethosor, line 33: Similar subheading/newline comment here as below

modtools/dynamic-adv-reactions.lua
```diff
@@ -0,0 +1,892 @@
...
+
+modtools/dynamic-adv-reactions
+==============================
+Tool to allow dynamically enabling / disabling adventure mode reactions based on certain conditions.
+When a reaction is registered with this script, its usage by adventurers will be disabled by default. Each usage of this command creates its own set of conditions to check, and the reaction will be enabled for the player if they satisfy all the criteria of one of those sets of conditions.
+Registered reactions are cleared when the world is unloaded, and so it is best to include any commands for this within an ``onLoad*.init`` file.
+Checks with multiple instances of the same type of condition are possible if registered via code rather than with commands.
+
+* ``-reaction reactionCode``
+  (Required)
```
lethosor, line 26: Possibly worth noting that the newlines in the source here aren't rendered, although I personally think it looks decent as-is (so I'll leave it up to you whether you want to change this):
image

lethosor (commented):
There are still a couple minor docs formatting issues - have you been able to build them on your end? If not, you can download the "docs" archive from GitHub Actions (should be under the "checks" tab of the PR) and navigate to `docs/_auto/modtools.html` to see what this looks like.

(If there's a way we can improve the docs-building instructions, I'm open to suggestions - most of the complexity I'm aware of is due to there being several ways to install Python.)

modtools/dynamic-adv-reactions.lua
Atkana: Yeah, it looks fine enough without having to make it any messier that I won't change it.

modtools/dynamic-adv-reactions.lua
Atkana: I've side-stepped the issue by making those two examples bulleted instead :P

Atkana:
I couldn't comment on the doc building instructions, because I've not used them (the sole things I use for scripts + submitting is Notepad++ and the github website :P). I've never really intended to purposefully engage with the docs stuff, and have only really had to to get through the check fails whenever I submit a PR. Now that I know a bit about it, I know that I can just wrap everything in a code block in the future.
That tip about how to see the info on how a check failed is very useful - up until now it's just been some esoteric process where I just have to guess why an X is showing :P I should've hopefully fixed everything using that information.

lethosor:
Ah, yes, I suppose that should be mentioned in the docs - since I set up the system, I tend to not think as much about writing certain steps like that down. (You can also click on the X or checkmark, then click "Details" to get to the same place:)
image

As a heads-up, I'm trying to get r4 out soon, so I might need to delay looking this over and merging it until after r4 is out.

lethosor:
Wrote up a bit here: https://docs.dfhack.org/en/latest/docs/Documentation.html#github-actions
Ideally this would be part of a larger explanation of our continuous integration setup (maybe under the Contributing page?) but we don't quite have that yet.

PR #236, opened 2020-12-16
[quickfort] remove 'extents' hack

myk002:
now that Buildings::setSize doesn't unconditionally deallocate existing extents.

This fixes the issue where non-rectangular extent-based buildings had occupancy incorrectly set in non-extent tiles.

Builds on (and requires) DFHack/dfhack#1729

### DIFF
--- a/internal/quickfort/building.lua
+++ b/internal/quickfort/building.lua
@@ -364,7 +364,7 @@ end
 -- allocate and initialize extents structure from the extents_grid
 -- returns extents, num_tiles
 -- we assume by this point that the extent is valid and non-empty
-function make_extents(b, db)
+function make_extents(b)
     local area = b.width * b.height
     local extents = df.reinterpret_cast(df.building_extents_type,
         df.new('uint8_t', area))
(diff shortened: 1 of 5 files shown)


internal/quickfort/building.lua
```diff
@@ -364,25 +364,17 @@ end
...
         df.new('uint8_t', area))
     local num_tiles = 0
     for i=1,area do
         local extent_x = (i-1) % b.width + 1
         local extent_y = math.floor((i-1) / b.width) + 1
-        local is_in_stockpile = b.extent_grid[extent_x][extent_y]
-        extents[i-1] = is_in_stockpile and 1 or 0
-        if is_in_stockpile then num_tiles = num_tiles + 1 end
+        local is_in_extent = b.extent_grid[extent_x][extent_y]
+        extents[i-1] = is_in_extent and 1 or 0
```
lethosor, line 376: The possible values of `extents` were turned into an enum after you wrote this - it's not just 0/1, and 1 is named `building_extents_type.Stockpile` currently. It likely does also apply to zones - 2 is `Wall`, for instance, which isn't applicable here.
Just to make sure: this is only intended to support stockpiles and zones currently, yes?

lethosor (approved):

internal/quickfort/building.lua
myk002: also extent-based buildings like farm plots and dirt roads

lethosor: Oh, right. Those also have values of 1 when created by DF, so I think `Stockpile` is probably not the best name for this item (I just pulled it from what was previously a comment). Anyway, this was working before, and hasn't really changed, so I'm not anticipating any issues here.

PR #237, opened 2020-12-29
workorder-recheck - set work order to Checking

TymurGubayev:
This mini-script allows deactivating a work order, f.e. to stop the job cancellation spam from an order whose conditions are no longer met.

### DIFF
--- a/gui/workorder-recheck.lua
+++ b/gui/workorder-recheck.lua
@@ -1,0 +1,22 @@
+-- Makes selected `Active` work order `Checking`.
+--[====[
+
+gui/workorder-recheck
+====================
+Sets the status to ``Checking`` (from ``Active``) of the selected work order (in the j-m or u-m screens). This makes the manager reevaluate its conditions.
+
+Example keybinding (put it in your ``dfhack*.init``-file):
+ ``keybinding add Alt-A@jobmanagement/Main gui/workorder-recheck``
+]====]
+local scr = dfhack.gui.getCurViewscreen()
+if df.viewscreen_jobmanagementst:is_instance(scr) then
+    local orders = df.global.world.manager_orders
+    local idx = scr.sel_idx
+    if idx < #orders then
+        orders[idx].status.active = false
+    else
+        dfhack.printerr("Invalid order selected")
+    end
+else
+    dfhack.printerr('Must be called on the manager screen (j-m or u-m)')
+end
--- a/gui/workorder-recheck.lua
+++ b/gui/workorder-recheck.lua
@@ -15,8 +15,8 @@ if df.viewscreen_jobmanagementst:is_instance(scr) then
     if idx < #orders then
         orders[idx].status.active = false
     else
-        dfhack.printerr("Invalid order selected")
+        qerror("Invalid order selected")
     end
 else
-    dfhack.printerr('Must be called on the manager screen (j-m or u-m)')
+    qerror('Must be called on the manager screen (j-m or u-m)')
 end
--- a/gui/workorder-recheck.lua
+++ b/gui/workorder-recheck.lua
@@ -1,4 +1,4 @@
--- Makes selected `Active` work order `Checking`.
+-- Resets the selected work order to the `Checking` state
 --[====[
 
 gui/workorder-recheck
--- a/gui/workorder-recheck.lua
+++ b/gui/workorder-recheck.lua
@@ -3,10 +3,8 @@
 
 gui/workorder-recheck
 =====================
-Sets the status to ``Checking`` (from ``Active``) of the selected work order (in the j-m or u-m screens). This makes the manager reevaluate its conditions.
-
-Example keybinding (put it in your ``dfhack*.init``-file):
- ``keybinding add Alt-A@jobmanagement/Main gui/workorder-recheck``
+Sets the status to ``Checking`` (from ``Active``) of the selected work order (in the ``j-m`` or ``u-m`` screens).
+This makes the manager reevaluate its conditions.
 ]====]
 local scr = dfhack.gui.getCurViewscreen()
 if df.viewscreen_jobmanagementst:is_instance(scr) then
--- a/gui/workorder-recheck.lua
+++ b/gui/workorder-recheck.lua
@@ -2,7 +2,7 @@
 --[====[
 
 gui/workorder-recheck
-====================
+=====================
 Sets the status to ``Checking`` (from ``Active``) of the selected work order (in the j-m or u-m screens). This makes the manager reevaluate its conditions.
 
 Example keybinding (put it in your ``dfhack*.init``-file):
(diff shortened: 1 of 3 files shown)


lethosor:
Looks like the step of our CI that gives some slightly-more-readable documentation errors (like for the underline issue you fixed) was skipped because the main docs build failed for the same reason. I've fixed it, so in the future, you should get things like this, probably under the "Checks" tab:

image

Looks like a useful script in any case! I'll double-check on my end.

gui/workorder-recheck.lua
```diff
@@ -0,0 +1,22 @@
...
+if df.viewscreen_jobmanagementst:is_instance(scr) then
+    local orders = df.global.world.manager_orders
+    local idx = scr.sel_idx
+    if idx < #orders then
+        orders[idx].status.active = false
+    else
+        dfhack.printerr("Invalid order selected")
+    end
+else
+    dfhack.printerr('Must be called on the manager screen (j-m or u-m)')
```
lethosor, line 21: I would generally prefer `qerror` because it propagates errors if this script is called from another script. And I realize I did it this way in `gui/manager-quantity` instead - for a small standalone script like this, I think it's unlikely that another script would use it instead of just setting the relevant field directly, so I think this way is fine.

gui/workorder-recheck.lua
```diff
@@ -0,0 +1,22 @@
+-- Makes selected `Active` work order `Checking`.
+--[====[
+
+gui/workorder-recheck
+=====================
+Sets the status to ``Checking`` (from ``Active``) of the selected work order (in the j-m or u-m screens). This makes the manager reevaluate its conditions.
+
+Example keybinding (put it in your ``dfhack*.init``-file):
+ ``keybinding add Alt-A@jobmanagement/Main gui/workorder-recheck``
```
lethosor, line 9: If you want this keybinding in `dfhack.init-example`, I would be totally fine with that, since it's scoped to a single screen. Doing so would also make the keybinding show up here automatically in the docs, so you wouldn't need the example command. It would need to be done in the DFHack repo, so I could handle it separately after merging this. Let me know what you think.

I might suggest `Alt-R`, though, to go along with the name of this script.

As a general comment, changing to this would create a standalone code block, which is generally what we use for example commands (not necessary if we move it to dfhack.init-example, though):

```suggestion
Example keybinding (put it in your ``dfhack*.init``-file)::

    keybinding add Alt-A@jobmanagement/Main gui/workorder-recheck``
```

gui/workorder-recheck.lua
```diff
@@ -0,0 +1,22 @@
+-- Makes selected `Active` work order `Checking`.
```
lethosor, line 1: Maybe this?
```suggestion
-- 
```

(in the changelog, it would look like this:)
```
```suggestion
`workorder`: resets the selected work order to the ``Checking`` state
```

lethosor (commented):
Looks like it works on my end! A couple suggestions:
* I'm assuming you borrowed this from `gui/manager-quantity`, but one difference I see is that this script doesn't create any in-game UI elements (that script creates a dialog), so it should go in the top-level folder (i.e. drop the `gui/` prefix and move it up a folder)
* It would also be great if you could add this to the changelog - we have a changelog in this repo, which I think was added after your last PR. It would go under a "New Scripts" section under "Future", and using the first line of this script as a description would be fine.

gui/workorder-recheck.lua
TymurGubayev: > I would generally prefer `qerror` because it propagates errors if this script is called from another script. And I realize I did it this way in `gui/manager-quantity` instead

It was just copy&pasted, so yeah. `qerror` is actually nicer in this case too: it automatically adds script name as prefix.

gui/workorder-recheck.lua
TymurGubayev: Didn't know about `dfhack.init-example`, so please do add this keybinding there. And yes, Alt-R is probably better.

gui/workorder-recheck.lua
TymurGubayev: > * I'm assuming you borrowed this from `gui/manager-quantity`, but one difference I see is that this script doesn't create any in-game UI elements (that script creates a dialog), so it should go in the top-level folder (i.e. drop the `gui/` prefix and move it up a folder)

Yeah, I was pondering on this, my reason was the script doesn't accept any parameters, and for it to work you need to navigate the UI to the correct state. It is a design-decision, so just tell me what do you prefer.

> * It would also be great if you could add this to the changelog - we have a changelog in this repo, which I think was added after your last PR. It would go under a "New Scripts" section under "Future", and using the first line of this script as a description would be fine.

I've made a change in the changelog (not 100% sure it was correct though because I've added a `##New Scripts` line).

On another note,
https://github.com/DFHack/scripts/blob/671dd3a59c66c53eddcaffbe5dc0d086327235ff/changelog.txt#L4
this should probably be `https://docs.dfhack.org/en/latest/docs/Documentation.html#changelog-syntax`, because there is no `Compile.html#build-changelog`.

gui/workorder-recheck.lua
TymurGubayev: (not a native speaker here, so yeah :) )

gui/workorder-recheck.lua
lethosor: Yeah, that sounds right, but I'm not seeing the change in this PR - did you forget to add the file, maybe?

And yes, that link is out-of-date - thanks for catching it!

lethosor: Oh, you added it to a new `patch-2` branch. If you're making these changes on GitHub, you probably want to navigate to the right branch in your fork before editing files, which in this case would be https://github.com/TymurGubayev/scripts/tree/patch-1

I'll apply the change here - no worries

lethosor: > Yeah, I was pondering on this, my reason was the script doesn't accept any parameters, and for it to work you need to navigate the UI to the correct state. It is a design-decision, so just tell me what do you prefer.

I would prefer just `workorder-recheck` - I think this is more along the lines of things like `clear-smoke`, `geld`, `putontable`, - I can handle this.

lethosor (approved):

gui/workorder-recheck.lua
TymurGubayev: > Oh, you added it to a new `patch-2` branch: [44469d2](https://github.com/DFHack/scripts/commit/44469d2e7a55d627667d5031086c742c8e11cef8). If you're making these changes on GitHub, you probably want to navigate to the right branch in your fork before editing files, which in this case would be https://github.com/TymurGubayev/scripts/tree/patch-1

You are right, I'm doing (at least trying to do) it via GitHub, I haven't even realized I've missed the correct branch somehow. Sorry about that.

PR #239, opened 2021-01-11
[gui/quickfort] add script for fast access to the quickfort interactive dialog

myk002:
For people who look for gui scripts in the `gui/` directory. Otherwise quickfort's gui mode might get missed.

### DIFF
--- a/gui/quickfort.lua
+++ b/gui/quickfort.lua
@@ -1,0 +1,14 @@
+-- in-game dialog interface for the quickfort script
+--[====[
+gui/quickfort
+=========
+In-game dialog interface for the `quickfort` script.
+]====]
+
+local args = {...}
+
+if #args == 0 then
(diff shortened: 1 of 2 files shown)


gui/quickfort.lua
```diff
@@ -0,0 +1,14 @@
...
+--[====[
+gui/quickfort
+=============
+In-game dialog interface for the `quickfort` script.
+]====]
+
+local args = {...}
+
+if #args == 0 then
+    dfhack.run_command('quickfort', 'gui')
```
lethosor, line 11: I would probably go with `run_script` for Lua scripts (which reminds me that I need to finish https://github.com/DFHack/dfhack/pull/1717 at some point...). Likely a negligible performance difference in this case, though.

lethosor (approved):
Good idea!

PR #240, opened 2021-01-11
git ignore kdevelop project and build files

myk002:
does this need a changelog entry?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 # kdevelop
-.kdev4/
+*.kdev4
 build/
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# kdevelop
+.kdev4/
+build/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 # kdevelop
 *.kdev4
-build/


lethosor (commented):
I'm generally of the opinion that IDE-specific entries should be kept in a personal [global gitignore](https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer), but it looks like we have `*.kdev4` in `dfhack/.gitignore`, so consistency is good with me.

Out of curiosity, what is generating a `build` folder for you? I don't have one on my end.

(And I definitely agree that this doesn't need to be in the changelog; I still need to really spell out criteria for that, but I'd classify this as a "minor dev-facing change" that doesn't affect most users)

myk002:
> Out of curiosity, what is generating a `build` folder for you? I don't have one on my end.

kdev creates the dir when it imports the cmake project. perhaps it wouldn't create the dir if I just created it as a "raw" project though...

lethosor:
Ah, this repo can't actually be built with CMake by itself (i.e. running `cmake .` in the scripts folder won't work) so I wouldn't recommend creating a CMake project from just this repo. I'm curious what happens with a raw project - if that doesn't create a `build` dir, I'd probably remove that entry from `.gitignore`.

myk002:
done. It seems I still need to "identify" the project with CMakeLists, but I don't actually need to treat it as a cmake project like I originally assumed. I can change it to "generic project manager" which does not create and require the build dir.

lethosor (approved):

PR #242, opened 2021-01-11
[quickfort] allow more whitespace in blueprint modelines

myk002:
whitespace is now allowed between a marker name and the opening parenthesis in blueprint modelines. for example, `#dig start (5; 5)` is now valid (you used to be required to write `#dig start(5; 5)`)

### DIFF
--- a/internal/quickfort/parse.lua
+++ b/internal/quickfort/parse.lua
@@ -85,7 +85,7 @@ end
 
 local function parse_label(modeline, start_pos, filename, marker_values)
     local _, label_str_end, label_str =
-            string.find(modeline, '^%s+label(%b())', start_pos)
+            string.find(modeline, '^%s+label%s*(%b())', start_pos)
     if not label_str then
         return false, start_pos
     end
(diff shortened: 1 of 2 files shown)


internal/quickfort/parse.lua
```diff
@@ -102,11 +102,11 @@ end
 
 local function parse_start(modeline, start_pos, filename, marker_values)
     local _, start_str_end, start_str =
-            string.find(modeline, '^%s+start(%b())', start_pos)
+            string.find(modeline, '^%s+start%s*(%b())', start_pos)
     if not start_str or #start_str == 0 then return false, start_pos end
     local _, _, startx, starty, start_comment =
             string.find(start_str,
-                        '^%(%s*(%d+)%s*[;, ]%s*(%d+)%s*[;, ]?%s*(.*)%)$')
+                        '^%(%s*(%d+)%s-[;, ]%s-(%d+)%s-[;, ]?%s*(.*)%)$')
```
lethosor, line 109: If I'm reading https://www.lua.org/pil/20.2.html correctly, it looks like this didn't change any functionality. Is this just for readability?

myk002: Yes, this worked before, but I changed to the non-greedy matcher so it would be clearer (as much as regexes can be clearer) that the space in the `[;, ]` character class was a valid required character.

lethosor: Oh, I was just looking at the table, and missed the explanation of greediness below it - makes sense!

lethosor (approved):

PR #248, opened 2021-01-29
[quickfort] don't require a cursor for #notes blueprints

myk002:
this removes a point of friction when viewing blueprint help text, like the dreamfort walkthrough. no map changes can be made, so no cursor should be required.


lethosor (approved):
Cool! Was hoping for this when I was testing notes blueprints.

PR #249, opened 2021-01-29
[quickfort] add --dry-run option for running blueprints without changing game state

myk002:
useful for validating blueprints and getting statistics on what they would do without actually running them in your game


lethosor:
Could I suggest `--dry-run` here? I realize it would be a bit of a sweeping change if you want to also update variable names, but a handful of other scripts use `dry-run` already (and also git :) ), so it would be nice for consistency.

It looks like `-n` is already taken in this case, but `-d` or something else for the short-form option could work.

myk002:
consistency with other scripts is good. I'll make that change. I've been corrupted by Gentoo's --pretend option for package management.

myk002:
done

lethosor (approved):
Looks good! I wasn't able to look closely at everything, but it looks like any game-modifying actions are guarded by a `dry_run` check somewhere, so :+1:

PR #250, opened 2021-01-30
[quickfort] update command reference docs

myk002:
document dependency on the search plugin as per PR DFHack/dfhack#1760
also update the text in general, including some formatting errors.

### DIFF
--- a/quickfort.lua
+++ b/quickfort.lua
@@ -20,9 +20,8 @@ real" in Dwarf Fortress, and then export your map using the DFHack
 `blueprint` plugin for later replay. Blueprint files should go in the
 ``blueprints`` subfolder in the main DF folder.
 
-For more details on blueprint file syntax, see the `quickfort-user-guide` or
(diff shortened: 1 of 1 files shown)


quickfort.lua
```diff
@@ -35,19 +34,19 @@ Usage:
 **quickfort gui**
     Starts the quickfort dialog, where you can run blueprints from an
     interactive list.
-**quickfort list [-m|--mode <mode>] [-l|--library] [-h|--hidden] [search string]**
+**quickfort list [-m|-\-mode <mode>] [-l|-\-library] [-h|-\-hidden] [search string]**
```
lethosor, line 37: Good catch! I would ordinarily suggest preformatted text here (double backticks instead of asterisks), but it looks like that would wrap, and the console help text is coming from a different source (i.e. users wouldn't see the `-\-` when running `quickfort help` and try to copy it).

image

lethosor (approved):

PR #252, opened 2021-02-01
[quickfort] add feedback message for gui orders

myk002:
to indicate that the operation was successful and to report how many orders were added. Before, the operation was completely silent with no feedback to the user that something was done.


lethosor (approved):
Looks good!

PR #253, opened 2021-02-01
[quickfort] add flags for setting initial gui state

myk002:
for example: `quickfort gui -l dreamfort notes` will start the dialog pre-filtered for the dreamfort walkthrough blueprints.

### DIFF
--- a/gui/quickfort.lua
+++ b/gui/quickfort.lua
@@ -2,7 +2,20 @@
 --[====[
 gui/quickfort
 =============
-In-game dialog interface for the `quickfort` script.
+In-game dialog interface for the `quickfort` script. Any arguments passed to
+this script are passed directly to `quickfort`. Invoking this script without
+arguments is equivalent to running ``quickfort gui``.
+
+Examples:
+
+-------------------------------------  --------------------------------------
+Command                                Runs
+-------------------------------------  --------------------------------------
+gui/quickfort                          opens quickfort interactive dialog
+gui/quickfort gui                      same as above
+gui/quickfort gui --library dreamfort  opens the dialog with custom settings
+gui/quickfort help                     prints quickfort help (on the console)
+-------------------------------------  --------------------------------------
 ]====]
 
 local args = {...}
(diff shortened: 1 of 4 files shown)


lethosor (approved):
Nice! I tried `quickfort gui -l` once and was disappointed - works now :)

A related thing: `gui/quickfort` appears to intentionally only add the `gui` argument if *no* arguments are passed to it - e.g. `gui/quickfort` -> `quickfort gui`, but `gui/quickfort foo` -> `quickfort foo`, so in order to pass these new flags through, you need something like `gui/quickfort gui -l`. The second `gui` seems redundant - is there a reason why it's necessary, or could it be changed?
https://github.com/DFHack/scripts/blob/5b7f81d5e0d79530ea10cfb9016c0b107c4103da/gui/quickfort.lua#L13

myk002:
I didn't implement `gui/quickfort <search string>` because I wasn't sure how to handle ambiguous cases. Does `gui/quickfort help` mean `quickfort gui help` or `quickfort help`? I think the redundant-looking "gui" is necessary, unfortunately. I should document this in the gui/quickfort help text, though

lethosor:
Makes sense, having `help` work is important. Thanks for updating the docs!

gui/quickfort.lua
```diff
@@ -2,7 +2,20 @@
...
 gui/quickfort
 =============
-In-game dialog interface for the `quickfort` script.
+In-game dialog interface for the `quickfort` script. Any arguments passed to
+this script are passed directly to `quickfort`. Invoking this script without
+arguments is equivalent to running ``quickfort gui``.
+
+Examples:
+
+-------------------------------------  --------------------------------------
```
lethosor, line 11: Looks like tables only work with `=====`, not `-----` - fixed on my end

myk002: Argh thank you. I need to find a way to point sphinx to an external scripts dir when building.

lethosor: Not sure if Sphinx has a way to do that. I usually develop in the `scripts` submodule and use [script-paths.txt](https://docs.dfhack.org/en/stable/docs/Core.html#script-paths) to point DFHack to it.

PR #254, opened 2021-02-01
[quickfort] allow blueprints to specify bins/barrels/wheelbarrows for stockpiles

myk002:
in #place blueprints.

syntax documentation in DFHack/dfhack#1768

### DIFF
--- a/internal/quickfort/place.lua
+++ b/internal/quickfort/place.lua
@@ -74,11 +74,32 @@ local stockpile_db = {
 }
 for _, v in pairs(stockpile_db) do utils.assign(v, stockpile_template) end
 
+local function add_resource_digit(cur_val, digit)
+    return (cur_val * 10) + digit
+end
+
 local function custom_stockpile(_, keys)
-    local labels = {}
-    local indices = {}
+    local labels, indices = {}, {}
     local want_bins, want_barrels, want_wheelbarrows = false, false, false
+    local num_bins, num_barrels, num_wheelbarrows = nil, nil, nil
+    local prev_key, in_digits = nil, false
     for k in keys:gmatch('.') do
+        local digit = tonumber(k)
+        if digit and prev_key then
+            local db_entry = rawget(stockpile_db, prev_key)
+            if db_entry.want_bins then
+                if not in_digits then num_bins = 0 end
+                num_bins = add_resource_digit(num_bins, digit)
+            elseif db_entry.want_barrels then
+                if not in_digits then num_barrels = 0 end
+                num_barrels = add_resource_digit(num_barrels, digit)
+            else
+                if not in_digits then num_wheelbarrows = 0 end
+                num_wheelbarrows = add_resource_digit(num_wheelbarrows, digit)
+            end
+            in_digits = true
+            goto continue
+        end
         if not rawget(stockpile_db, k) then return nil end
         table.insert(labels, stockpile_db[k].label)
         table.insert(indices, stockpile_db[k].indices[1])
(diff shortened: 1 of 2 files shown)


internal/quickfort/place.lua
```diff
@@ -137,32 +166,35 @@ end
...
+    if db_entry.want_wheelbarrows or db_entry.num_wheelbarrows then
+        local max_wb = db_entry.num_wheelbarrows or
                 quickfort_common.settings['stockpiles_max_wheelbarrows'].value
-        if max_wb < 0 then
-            fields.max_wheelbarrows = 1
-        elseif max_wb >= ntiles then
-            fields.max_wheelbarrows = ntiles
+        if max_wb < 0 then max_wb = 1 end
+        if max_wb >= ntiles - 1 then
+            fields.max_wheelbarrows = ntiles - 1
```
lethosor, line 193: I think this is the case, but just to confirm - you are directly setting this field, not sending keypresses to the game, right? Asking because `tweak max-wheelbarrows` changes the in-game wheelbarrow UI (the vanilla UI uses `w` to cycle through numbers and only goes up to 3), and it's not always obvious that this is a DFHack feature.

myk002: That's right - the variable ui for setting wheelbarrows is the whole reason for this feature. This will set the wheelbarrow (and/or container) count programmatically

lethosor:
How should something like this be handled?
```
#place
`,`,`,`
`,f2,f,`
`,f,`,`
`,`,`,`
```

Currently the `f2` cell produces a separate 1-tile stockpile with 1 barrel allowed, and the other two cells are connected to form a 2-tile, 2-barrel stockpile. Given that removing the `2` produces a single stockpile, I feel like the container count should be ignored when determining whether stockpiles are unique. However, that would lead to ambiguity in cases like this:
```
#place
`,`,`,`
`,f2,f3,`
`,f,`,`
`,`,`,`
```
so I'm not really sure what makes sense here.

myk002:
Good point. So the options are:
1. Status quo, require the same number across all tiles of stockpiles. Pros: fits with current expectations for boundary detection, no confusion about whether adjacent tiles are part of the same stockpile (same text == same stockpile), no code changes. Cons: extra text in blueprint that doesn't really add value. Changing the number of containers assigned to a stockpile requires changing every tile of that stockpile.

2. Ignore numbers for the purposes of boundary detection. Pros: only add container numbers to a single time, making it easy to change and simplifying the text in all the other stockpile tiles. Cons: potential for confusion, both in visually determining boundaries and in the case you brought up where two tiles have conflicting numbers, also need to figure out how to implement (will need to keep aux metadata about the region. Text == type will no longer be enough).

Not obvious to me which route is better. Ease of editing and maintenance, less clutter vs visual clarity of meaning.

The same question will come up when I implement stockpile labels, which will allow disjoint sections to be considered part of the same stockpile. The problem gets even worse then because labels will be much longer than the numbers added here. The decision we make now will decide the implementation of that feature too.

I'm leaning towards ignoring the numbers for purposes of boundary detection, mostly for its decluttering property, which will become even more important with labels.

How to handle conflicting specifications, though? I'll come back and finish this thought in a bit.

myk002:
I should note that I expect most stockpile declarations to be in expansion syntax: `f(4x5)`, so while the decisions we make here will be far-reaching, whatever we go with is unlikely to affect many users.

myk002:
Ok, conflicting specifications. Again, choices here will propagate to the future label feature.

Option 1, log but ignore conflicts, just use one. Option 2, fail with parse error on conflict detection. Option 3, consider conflict as defining a separate stockpile.

I'm thinking among the lines of option 3, which I think is most flexible. For each stockpile, each tile can have either no metadata or the same metadata as was declared elsewhere in the same stockpile. This allows the convenient method of only defining the metadata once, i.e 
f3,f,f
f,f
f

But also allows two adjacent, non-expansion syntax stockpiles to be differentiated:
f3,f3,f3
f3,f3,f2
f3,f2,f2

Labels could make this a little more clear:
f3:label1,f:label1,f:label1
f:label1,f:label1,f2:label2
f:label1,f:label2,f:label2

And of course labels would be able to differentiate the piles even if they were both f2.

I think this will give the blueprint designer simple convenience in the more common case but will also provide the flexibility required for complex layouts. Thoughts?

myk002:
I've looked into the required code changes. It alters a basic assumption the code has had up to this point -- that the text is the boundary map key. This assumption goes deep enough that I think the planned changes are too risky to do without regression tests.

Since the proposed changes will be backwards compatible with the current implementation, how about if we merge this as-is, then I spend some time setting up automated regression tests, then make the changes after I'm more certain things won't break?

lethosor:
I figured it would be complicated to make every possibility work well. Thanks for looking into it! I think https://github.com/DFHack/dfhack/pull/1768/commits/3278519c98f4d364b8ff97a1b61d8a3ef8543ecb is plenty good enough for now. I also don't expect stockpiles specified with non-expansion syntax _and_ with custom container counts to be very common, so glad to hear that you agree.

lethosor (approved):

PR #255, opened 2021-02-05
[quickfort] enqueue manager orders for containers set in #place blueprints

myk002:
a follow on to #254

this implements generating manager orders for containers specified in `#place` blueprints.

For example, in a place blueprint like this:
```
#place
s2(3x3)
```
two wheelbarrows would be enqueued. For this blueprint:
```
#place
f10(1x9)
```
10 rock pots would be enqueued, even though the stockpile only has 9 tiles.

A stockpile that does not specify a specific number is initialized with the default number of containers, but no manager orders are generated. This prevents huge numbers of containers being generated for regular blueprints that don't do anything special with their stockpile definitions.

This PR also fixes some in-code documentation that was just plain incorrect.

More docs in DFHack/dfhack#1768

### DIFF
--- a/internal/quickfort/orders.lua
+++ b/internal/quickfort/orders.lua
@@ -131,7 +131,7 @@ end
 -- ensure we don't reinit this; it contains allocated memory
 reactions = reactions or stockflow.collect_reactions()
 
-function enqueue_orders(stats, buildings, building_db, ctx)
+function enqueue_building_orders(buildings, building_db, ctx)
     local order_specs = ctx.order_specs or {}
     ctx.order_specs = order_specs
     for _, b in ipairs(buildings) do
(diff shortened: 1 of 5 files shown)


internal/quickfort/orders.lua
```diff
@@ -166,3 +166,18 @@ function enqueue_orders(stats, buildings, building_db, ctx)
...
 end
+
+function enqueue_container_orders(ctx, num_bins, num_barrels, num_wheelbarrows)
+    local order_specs = ctx.order_specs or {}
+    ctx.order_specs = order_specs
+    if num_bins and num_bins > 0 then
+        inc_order_spec(order_specs, num_bins, reactions, "wooden bin")
+    end
+    if num_barrels and num_barrels > 0 then
+        inc_order_spec(order_specs, num_barrels, reactions, "rock pot")
```
lethosor, line 177: I'm wondering if these might make sense to make configurable, but I think you've chosen reasonable defaults here.

lethosor (approved):
Looks like it works on my end!

internal/quickfort/orders.lua
myk002: I made similar material choices for the regular building blueprint orders. In the docs, I just wrote that if the material doesn't make sense for your fort, go to the manager orders screen and replace the order with a similar one that makes sense. I can definitely make it more configurable if we get complaints, though.

PR #256, opened 2021-02-13
[digfort] add deprecation warnings to digfort

myk002:
step 2 in the DFHack/dfhack#1644 deprecation plan

### DIFF
--- a/digfort.rb
+++ b/digfort.rb
@@ -1,11 +1,16 @@
-# designate an area based on a '.csv' plan
+# designate based on a '.csv' plan (deprecated: please use DFHack quickfort)
 =begin
 
 digfort
 =======
 A script to designate an area for digging according to a plan in csv format.
 
-This script, inspired by quickfort, can designate an area for digging.
+**Digfort is deprecated.** Please use DFHack's more powerful `quickfort` command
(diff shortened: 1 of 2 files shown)


digfort.rb
```diff
@@ -42,6 +47,8 @@
 
 =end
 
+puts "The digfort script is deprecated. Please use DFHack's quickfort command instead. The digfort script will be removed in a future DFHack release."
```
lethosor, line 50: Can this mention "quickfort run"? I'm not sure how easy that would be to work in - I see it's in the comment above, but I don't think there's a way to easily view that from the DFHack console.

Maybe something simple like this could work. It would be cool if the filename could be included somehow too, but since it would need to be moved to the `blueprints` folder, that might end up causing more confusion.

```suggestion
puts "The digfort script is deprecated. Please use DFHack's quickfort command instead:\n`quickfort run example.csv`, where example.csv is in the blueprints folder.\nThe digfort script will be removed in a future DFHack release."
```

myk002: reworded a bit. I'll update the fortplan deprecation notice similarly.

lethosor (approved):
Looks good, thanks!

PR #257, opened 2021-02-18
[quickfort] allow aliases to be defined directly in blueprint files

myk002:
instead of requiring them to be added to a separate aliases.txt file. This reduces friction for sharing blueprints with custom aliases. It also makes testing much easier since you can define needed aliases along with the test blueprint. It makes alias development much easier for the same reason.

Syntax documentation in DFHack/dfhack#1779, but here's an example I've been using for testing. It scripts the setup of a new dreamfort map (meant to be run over the starting wagon):

- sets manager, bookkeeper, doctor, and broker nobles to the first dwarf
- names hotkeys and sets them to be zoomed to the correct z-level (x and y will need adjustment, and z may need adjustment too depending on number of soil layers, but the pre-naming is handy regardless)
- sets only farmers harvest, gather refuse from outside (incl. vermin), and turns off autoloom
- creates an "Inside" burrow
- sets all preset uniforms to replace clothing
- in the "Metal" uniform, replaces generic "upper body" armor with explicit breastplate + mail shirt, replaces "lower body" armor with greaves
- creates "Siege" alert and sets civilians to the "Inside" burrow
- deconstructs the starting wagon

```
#aliases
startnobles: ^n{Down 4}{togglesequence 8}^q
sethotkey: {fkey}zn{name}&
starthotkeys: >{F1}^H>{sethotkey fkey={F2} name=Farming}>{sethotkey fkey={F3} name=Industry}>{sethotkey fkey={F4} name=Services}{> 4}{sethotkey fkey={F5} name=Guildhall}>{sethotkey fkey={F6} name=Beds}{> 6}{sethotkey fkey={F7} name=Quarry}{< 14}^q
startorders: ^ohrov^Wl^^q
startburrows: ^wa&nInside&^^q
metalarmorsetup: {Right 2}{Down 2}&{Down}&{Left}&M{Right}{Down}&{Left}{Down}{Right}{Down}&{Left}{Down 2}L{Right}{Down 2}&{Left}{Down 3}&M{Right}{Down}&{Left 2}
startmilitary: ^mnr{Down}r{metalarmorsetup}{Down}racNSiege&{Right}&fd^q
#query
x{startnobles}{starthotkeys}{startorders}{startburrows}{startmilitary}
```

Once I refine this a bit more, I might release it with dreamfort

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,6 +16,7 @@ that repo.
 ## Misc Improvements
 - `gui/no-dfhack-init`: clarified how to dismiss dialog that displays when no ``dfhack.init`` file is found
 - `quickfort`: an active cursor is no longer required for running #notes blueprints (like the dreamfort walkthrough)
+- `quickfort`: you can now add alias definitions directly to your blueprint files instead of having to put them in a separate aliases.txt file. makes sharing blueprints with custom alias definitions much easier.
 
 # 0.47.05-beta1
 
--- a/internal/quickfort/aliases.lua
+++ b/internal/quickfort/aliases.lua
@@ -29,7 +29,8 @@ function reset_aliases()
     alias_stack = {}
 end
 
-local function push_aliases(aliases)
+-- overwrites metatable of passed-in aliases map
+function push_aliases(aliases)
     local prev = alias_stack
     setmetatable(aliases, {prev=prev,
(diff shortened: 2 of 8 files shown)


lethosor (approved):
Looks good!

PR #258, opened 2021-02-18
[quickfort] be flexible about start mode for #query blueprints

myk002:
Now any mode that has an active cursor is acceptable. Before, we required that players start in one of a short list of approved modes that we know how to get back to after the query blueprint is done, but that leads to a bad user experience when you happen to be in a mode that's not on the list. It's even worse when you run a meta blueprint that includes other non-#query blueprints and then runs the #query blueprint. The other blueprints will succeed, then the #query blueprint will fail, leaving a #meta sequence partially applied.

This new approach lets players start in any mode they want to, as long as it has a cursor. we send `ESC` repeatedly until we're back at the default screen, then we just send `q` once to get into query mode. easy peasy, and works everywhere. if we don't know how to get back to the start mode when we're done, we just stay in query mode. this may be slightly confusing for a player that started in, say, building placement mode, but it's a better user experience than failing the blueprint entirely.

If we get complaints about not getting back into all start modes, we can expand the table that records how to get back into various modes. we can still fail if we don't get to the main screen within 10 presses of `ESC`, but I don't know of any screen in DF where that will happen.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/no-dfhack-init`: clarified how to dismiss dialog that displays when no ``dfhack.init`` file is found
+- `quickfort`: you can now be in any mode with an active cursor when running #query blueprints (before you could only be in a few "approved" modes, like look, query, or place)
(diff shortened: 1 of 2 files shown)


lethosor (approved):
Looks good! I had been wondering if `d` mode should be supported, and it works now.

PR #259, opened 2021-02-18
[quickfort] #query blueprint sanity check refinements

myk002:
- check that the cursor is still on the tile at the end of configuration. this catches cases like configuring a stockpile to take from another, but the movement keys to get to that other stockpile are incorrect
- allow viewscreen focus string to change if the configuration causes the building to be destroyed (or cancels destruction of a building)


lethosor (approved):
Looks good! Just to check, have you run dreamfort under this? I feel like that would be better than my (minimal) tests.

myk002:
Yes, this has been in my local dev branch for at least one dreamfort run

lethosor:
Cool, and thanks for fixing the merge conflict!

PR #262, opened 2021-03-06
[test] install scripts/test into hack/scripts/test/scripts

myk002:
Fixes DFHack/dfhack#1690

Before we merge any scripts tests into the repo, there is one problem I think we need to solve.

Currently, automated CI tests are only run for DFHack/dfhack PRs. This means that a bad test can get checked into DFHack/scripts without blocking the PR that it's a part of, and once the scripts module gets updated in DFHack/dfhack, the bad test will block the next (likely unrelated) PR in DFHack/dfhack. This would be very frustrating for contributors to DFHack/dfhack since they would not be able to do anything about the failures blocking their PRs.

Can we implement test runs for the scripts repo? Or are there other ways to prevent bad tests from getting into the codebase?

Merging the scripts repo back into dfhack would solve the problem, of course, but then the current separation of access controls would be lost.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
                        PATTERN "scripts/test" EXCLUDE
 )
 
-install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/test/
-        DESTINATION ${DFHACK_DATA_DESTINATION}/scripts/test/scripts
-)
+if(BUILD_TESTS)
+    install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/test/
+            DESTINATION ${DFHACK_DATA_DESTINATION}/scripts/test/scripts
+    )
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,3 +6,7 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
                        PATTERN "3rdparty" EXCLUDE
                        PATTERN "scripts/test" EXCLUDE
 )
+
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/test/
+        DESTINATION ${DFHACK_DATA_DESTINATION}/scripts/test/scripts
+)
--- a/changelog.txt
+++ b/changelog.txt
@@ -53,6 +53,9 @@ that repo.
 - `setfps`: improved error handling
 - `unretire-anyone`: the historical figure selection list now includes the ``SYN_NAME`` (necromancer, vampire, etc) of figures where applicable
 
+## Internals
+- Install tests in the scripts repo into hack/scripts/test/scripts when the CMake variable BUILD_TESTS is defined
+
 # 0.47.05-r1
 
 ## Misc Improvements
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,6 +16,9 @@ that repo.
 ## Fixes
 - `quickfort`: allow machines (e.g. screw pumps) to be built on ramps just like DF allows
 
+## Internals
+- Install tests in the scripts repo into hack/scripts/test/scripts when the CMake variable BUILD_TESTS is defined
+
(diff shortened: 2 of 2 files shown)


lethosor:
Continuing from https://github.com/DFHack/scripts/pull/263#issuecomment-803501839:

> I think I'll split #262 into two parts to take the time pressure off. I can merge the exclude line now to keep scripts tests from affecting CI, then we can merge the scripts/test installation part in once CI is in place for the scripts repo.

My understanding is that CI only runs on scripts before they're installed, so #1792 should take care of any CI issues (i.e. your EXCLUDE change here only determines what is installed to hack/scripts, which isn't what `script-docs.py` looks at). I think we can keep this as one PR, unless I'm missing something.

As for CI, I'll continue https://github.com/DFHack/scripts/pull/263#issuecomment-803496301 here too. I'm thinking https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization might be a solution that could avoid duplication, but I haven't tried it yet. It would likely require some refactoring to avoid duplicating the docs+lint checks that are already running in the scripts repo. Definitely an issue that we should resolve - I trust that you are running these tests locally, but breakages will be harder to detect if they're not running under CI down the line.



> Merging the scripts repo back into dfhack would solve the problem, of course, but then the current separation of access controls would be lost.

Access controls are more of a side effect - the original reason for the split (although there may have been others that I'm forgetting) was to allow for script development without needing to clone the entire DFHack repo.

myk002:
> > I can merge the exclude line now to keep scripts tests from affecting CI
> 
>  (i.e. your EXCLUDE change here only determines what is installed to hack/scripts)

I meant more that checking in the EXCLUDE line now would prevent subsequent merges of script tests from affecting CI for later PRs in DFHack/dfhack.

> I trust that you are running these tests locally, but breakages will be harder to detect if they're not running under CI down the line.

Yeah, I'm absolutely running the tests locally before I merge them. My fear is that the tests might fail on other platforms that CI checks but I do not (hi windows). The unit tests I'm writing now are low risk, but the integration tests I plan to write later may have unexpected failure modes. Maybe I'm just being paranoid, but "don't break the build" is pretty deeply ingrained in my psyche.

> > but then the current separation of access controls would be lost.
> 
> Access controls are more of a side effect - the original reason for the split was to allow for script development without needing to clone the entire DFHack repo.

Ah, ok. Makes sense.

myk002:
I split the EXCLUDE line out into #265. we can merge #265 to unblock merging of #263 and wait on this PR until we have CI for scripts figured out

lethosor:
I haven't gotten tests under CI worked out for this repo yet, but I'm ok with merging this if you are (and if it works, of course). I'm ok running tests manually when they (or a script being tested) change, and changes will get caught within a day since we have submodules updating nightly now.

Incidentally, I've been meaning to move individual tests from `hack/scripts` to `hack/test`, since they aren't "executable" by themselves, but that's a problem for another day.

myk002:
> I haven't gotten tests under CI worked out for this repo yet, but I'm ok with merging this if you are (and if it works, of course). I'm ok running tests manually when they (or a script being tested) change, and changes will get caught within a day since we have submodules updating nightly now.

Ok. I'll merge it in. There is still a non-zero risk of a test passing on Linux but failing on Windows, but as you say, it will be caught within a day, and if it does cause too many problems, we can always revert this PR.

> Incidentally, I've been meaning to move individual tests from `hack/scripts` to `hack/test`, since they aren't "executable" by themselves, but that's a problem for another day.

This makes sense.

lethosor:
We don't have tests running in any CI on Windows yet, so that could also come up (Buildmaster only runs them on Linux currently).

myk002:
I didn't know that - why *don't* we run tests on windows? I didn't see anything platform specific in the tests. Or is it just that we haven't installed Python on the windows images?

lethosor:
It took a good amount of troubleshooting to get tests running in a headless Linux environment, and I expect even more on Windows. Basic output handling is one known obstacle so far (https://github.com/DFHack/dfhack/issues/1680). The Windows Buildmaster images are actually using Linux + WINE, which is technically possible to run DFHack under, but could cause issues of its own.

PR #263, opened 2021-03-06
[quickfort] refactor and unit test quickfort/parse.lua

myk002:
DFHack/dfhack#1690

depends on:
- DFHack/dfhack#1791 for recursive diffing support in expect.table_eq()
- DFHack/dfhack#1792 for ignoring the test dir for doc requirement
- DFHack/dfhack#1796 for the xlsxreader Lua class wrappers

this is my first attempt with tests in the scripts repo. I chose `parse.lua` to test first because there are no interactions with df game state and the IO can be easily abstracted.

unit testing, unsurprisingly, found a number of unhandled edge cases and highlighted poor design choices. I expect there to be a large CL like this for each major file I test. Would you like to review them, or are these ok for me to merge directly? They're on the edge of the "it's ok to merge small bugfixes" guidance you gave me earlier.

improvements and fixed bugs:
- factored out file IO into a proper Reader class and subclasses
- factored out duplicate functionality from marker parsers
- can now successfully read further cells in the same row after a multiline
  value in a previous cell
- better help text for bad markers
- better ability to handle extra space characters throughout
- don't stop reading a row at any comment, just at the '#' marker
- fix line number accounting in blueprints with implicit #dig modelines
- print warning when a bad alias is encountered
- properly detect and error on params with no values


internal/quickfort/reader.lua
```diff
@@ -0,0 +1,100 @@
...
+    if not tokens then return nil end
+    -- raw numbers can get turned into floats. let's turn them back into ints
+    for i,token in ipairs(tokens) do
+        local token = tonumber(token)
+        if token then tokens[i] = tostring(math.floor(token)) end
+    end
+    return tokens
+end
+
+unit_test_hooks = {
```
lethosor, line 95: `plugins/workflow.lua` guards this with `if dfhack.internal.IN_TEST` - if that works, that could be worth doing here. (It might require more changes to the runner to work reliably, e.g. how the runner currently overrides `require()` to always reload the module, so we can put it off)

lethosor:
> Would you like to review them, or are these ok for me to merge directly? They're on the edge of the "it's ok to merge small bugfixes" guidance you gave me earlier.

Once the DFHack dependencies are merged in, I'm fine if you merge this :)

I'll run through tests on my end in a bit, but I'm fine with the fixes and refactoring you have here. (I'm still looking into ways to get CI working in this repo without excessive duplication across repos)

myk002:
> Once the DFHack dependencies are merged in, I'm fine if you merge this :)

Cool. Thanks! I think I'll split #262 into two parts to take the time pressure off. I can merge the exclude line now to keep scripts tests from affecting CI, then we can merge the scripts/test installation part in once CI is in place for the scripts repo.

That way I can merge the unit tests as I write them. This is important because if I get too far ahead of upstream, I won't be able to promptly fix bugs that people report without causing merge conflicts with my unit testing branch. How does that sound?

lethosor:
I'm fully on board with having unit tests merged in as you write them, especially if it makes developing them more straightforward. I must have missed some of the CI discussion in #262; I'll follow up there.

internal/quickfort/reader.lua
myk002: I'll have to play with this a little bit. I know from my coverage reports that the quickfort scripts are not being reloaded on every test run.

on closer inspection, this is probably because I use `reqscript`, not `require`. I'll look into providing a similar override for `reqscript`.

myk002: done in DFHack/dfhack#1816

internal/quickfort/parse.lua
```diff
@@ -66,93 +118,92 @@ local function get_next_csv_token(line, pos, get_next_line_fn, sep)
...
         marker_values.starty = starty
         marker_values.start_comment = #start_comment>0 and start_comment or nil
-    else
+    elseif #start_str > 0 then
         -- the whole thing is a comment
-        _, _, start_comment = string.find(start_str, '^%(%s*(.-)%s*%)$')
-        marker_values.start_comment = start_comment
+        marker_values.start_comment = start_str
+    else
+        dfhack.printerr(string.format(
```
lethosor, line 186: I noticed some of these messages being printed when the tests in this PR are run. I believe that Buildmaster treats `printerr()` messages as test failures (maybe that could be adjusted), but in any case, they could visually be confused with test failure messages (I thought they were failures at first, and was trying to figure out why they weren't being recorded).

Options I thought of (open to others too) - these apply to all `printerr()` calls being tested, and may not apply to this one in particular:
- Should these be `error` or `qerror` calls? That would allow detecting these specific messages with `expect.error_match()`, for instance.
- Should we be detecting that calls to `printerr()` happen in tests? If so, should we patch `printerr()`
  - in all tests?
  - just in tests that are known to print errors?
  - in a wrapper like `expect.error_match()` to allow detecting specific error messages?
- If these errors aren't important to test, could they be behind a `verbose` flag or something similar to allow tests to hide them?

I'm not sure what the best option is, honestly. I know we have at least one test (`structures/misc:viewscreenDtors`) that prints progress information currently, but not any that print errors as far as I know, so we haven't had to think about this before.

myk002: These are non-fatal parsing errors (erm..I guess that would make them warnings not errors), which is why they don't currently use error or qerror, but they're still important for users to see so they can correct them or report them to the blueprint owner, which is why they don't use log() (== hidden behind a `verbose` flag).

Since these are "warnings" and not "errors", we *could* create a `dfhack.printwarn()` function to differentiate them, but then what would be the use case for `dfhack.printerr()`? when would you use it instead of `q/error()`?

It sounds perfectly reasonable to me that "unexpected" `printerr()` calls should cause CI to fail. And if it causes CI to fail, it should cause the test to fail so devs can catch the error before it gets to CI.

here's a possible plan:

- test harness intercepts calls to `dfhack.printerr`
- if `printerr` was called and the action was not acked/matched by the test, then the unit test fails
- possible API for acking/matching `printerr` messages:
  - expect.push_printerr_matcher(table or string or matcher fn or `true`, ...)
  - matchers are popped in FIFO order when `printerr` is called. if a table was pushed (key = string or matcher fn, value = number of expected matches), any matcher in the table can be matched next (and is subsequently decremented and removed when its count reaches 0) from the table; the table itself is popped when it is empty). if `true` is pushed, it matches any message. The matcher queue must be empty before the next call to expect.*. if unit test returns before all matchers are popped, test fails.

Sample code:

```
-- test passes
function test.get_next_csv_token_multi_line()
    local lines = {'first,"start, ', 'middle,', ' end ",second'}
    local i = 1
    local next_line_fn = function() i = i + 1 return lines[i] end

    lines[3] = nil
    expect.push_printerr_match('unterminated')
    -- parse.get_next_csv_token calls printerr()
    expect.nil_(parse.get_next_csv_token(lines[1], 7, next_line_fn))
    expect.eq(3, i)
end
```

```
-- test fails because printerr call was unmatched in next call to expect.*
function test.get_next_csv_token_multi_line()
    [...]

    lines[3] = nil
    -- parse.get_next_csv_token calls printerr()
    expect.nil_(parse.get_next_csv_token(lines[1], 7, next_line_fn))
    -- call expect.nil_() fails
    expect.eq(3, i)
end
```

```
-- test fails because printerr call was unmatched before test end
function test.get_next_csv_token_multi_line()
    [...]

    lines[3] = nil
    -- parse.get_next_csv_token calls printerr()
    parse.get_next_csv_token(lines[1], 7, next_line_fn)
    -- test fails in post-test verification that all printerr calls have been matched
end
```

```
-- test fails because printerr message was expected but never happened
function test.get_next_csv_token_multi_line()
    expect.push_printerr_match('unterminated')
    -- test fails in post-test verification that printerr matcher queue is empty
end
```

This API allows us to handle arbitrary sequences of `printerr` calls but avoids changing the syntax of any existing `expect` API calls... at a cost of having to learn a different API style and semantics just for `printerr` messages.

Questions:

- should we just patch the env when we load the unit under test and be careful not to invoke error paths when the UUT calls another file that calls `printerr`, or can we redefine dfhack.printerr in _G (the transitive dependency problem again)

lethosor: My understanding is that Buildmaster literally looks for red text to determine whether a test failed (well, in addition to exit codes of the test runner and all), so a `printwarn()` wouldn't help wit the issue here if it also prints red text (it may be useful semantically, but that's another matter).

I suppose one option is to change how Buildmaster parses test output. @BenLubar would know more about how feasible this is, but I think any output between `Running %d tests` and `Test summary` could be ignored, since the test runner should exit with an appropriate status if any tests fail. This alone wouldn't do anything about errors appearing in the test output and potentially confusing people running the tests, though.

In terms of intercepting and checking `printerr()` calls, I would prefer an API more similar to `expect.error_match` - I think it would be a simpler to implement and understand when used in tests (vs. the "push" semantics) since the lifetime of the matcher would be more well-defined, at the expense of being more verbose. This could be implemented by patching `dfhack.printerr()` temporarily. For instance:

```lua
expect.printerr_match(function()
  parse.get_next_csv_token(lines[1], 7, next_line_fn)
end, "unterminated")
```

As for detecting `printerr()` calls that were _not_ caught by a test: that isn't something I'd considered, but you're definitely correct that it will need to be addressed somehow if Buildmaster isn't changed. I think this is the reason for most of the complexity of the "push" implementation you described. I _think_ a simple solution to this could be to have `run_test()` also patch `dfhack.printerr()` with some logic to cause the current test to fail when called, and the patched printerr installed by `expect.printerr_match()` would take priority over this one. Unsure if this type of failure should count as "errored", "failed", or another state, though.

In any case, patching `dfhack.printerr()` is definitely something we can do from Lua. The `dfhack` table is shared globally (maybe not always a good idea, but it does make things like this easier!) so any scripts should pick up on changes to it immediately, as long as they aren't caching values from the table (e.g. `f = dfhack.printerr`, in which case I would be inclined to change the script to not do that).

myk002: SGTM. I'll give it a go.

myk002: wait, I spoke too hastily. there is another reason for the API I proposed beyond catching `printerr()` calls that were not caught by a test: in general, the `printerr()` is just a side effect when testing some particular code path. In the example in your comment, I'd still want to verify that `parse.get_next_csv_token()` returns nil. That's why I made the printerr matching out of band.

However, I think we could still chain the expect calls with the callback syntax you propose in order to keep the better-defined scoping. e.g.:

```
expect.printerr_match(
    'unterminated',
    function()
        expect.nil_(parse.get_next_csv_token(lines[1], 7, next_line_fn))
    end)

expect.printerr_match(
    'unterminated',
    function()
        expect.eq(expected, parse.get_next_csv_token(lines[1], 7, next_line_fn))
    end)
```

The callback syntax is much cleaner here compared to what we would have had to do if we had used the "flat" syntax of `expect.error_match(func, matcher, ...)` since we have two functions with varying numbers of arguments. We'd have had to use the more confusing style of `dfhack.call_with_finalizer(num_cleanup_args,always,cleanup_fn[,cleanup_args...],fn[,args...])`

What would be really nice is if we could use python's `with()` syntax. something like:
```
with (printerr_matcher('unterminated')):
    expect.nil_(parse.get_next_csv_token(lines[1], 7, next_line_fn))
```
but I'll settle for the callback style : )

myk002: I'd like to match the style of `function expect.error_match(func, matcher, ...)`, but I'm not sure if I can do that here. The call to the unit *must* be done within a callback, but there is only room for one function pointer in this API. For example, if we were to use an `error_match` style for the `printerr_match` call, it could look like this:

```
expect.printerr_match(
    expect.nil_,
    'unterminated',
    function()
        return parse.get_next_csv_token(lines[1], 7, next_line_fn)
    end
```
but it could also look like this:
```
expect.printerr_match(
    expect.eq,
    'unterminated',
    expected,
    function()
        return parse.get_next_csv_token(lines[1], 7, next_line_fn)
    end
```

how would we know to call the second parameter of `expect.eq` instead of comparing it as a literal?

myk002: if we can't come to a decision here, I could always use `print()` for my warnings instead of `dfhack.printerr()` as a stopgap solution.

lethosor: > wait, I spoke too hastily. there is another reason for the API I proposed beyond catching `printerr()` calls that were not caught by a test: in general, the `printerr()` is just a side effect when testing for some result. In the example in your comment, I'd still want to verify that `parse.get_next_csv_token()` returns nil. That's why I made the printerr matching out of band.

Yeah, I was expecting that. Wouldn't this work?

```lua
expect.printerr_match(function()
  expect.nil_(parse.get_next_csv_token(lines[1], 7, next_line_fn))
end, "unterminated")
```

Variables like `expect` and `parse` would be inherited from the parent scope, so this is functionally similar to the Python `with` statement that you brought up.

If you wanted to use the value outside of the `printerr_match` callback, it would get a bit messier:

```lua
local value
expect.printerr_match(function()
  value = parse.get_next_csv_token(lines[1], 7, next_line_fn)
  expect.nil_(value)
end, "unterminated")
-- do something with value
```

lethosor: Here's a naive implementation (it doesn't handle multiple `printerr()` calls currently):
```lua
function expect.printerr_match(func, matcher, ...)
    local args = {...}
    local printed_error
    local old_printerr = dfhack.printerr

    dfhack.with_finalize(
        function()
            dfhack.printerr = old_printerr
        end,
        function()
            dfhack.printerr = function(e)
                printed_error = e
            end
            func(table.unpack(args))
        end)


    if printed_error == nil then
        return false, 'no error printed by function call'
    elseif type(matcher) == 'string' then
        if not tostring(printed_error):match(matcher) then
            return false, ('error "%s" did not match "%s"'):format(printed_error, matcher)
        end
    elseif not matcher(printed_error) then
        return false, ('error "%s" did not satisfy matcher'):format(printed_error)
    end
    return true
end
```

and a test that uses it:
```lua
function test.printerr()
    expect.printerr_match(function()
        expect.nil_(dfhack.printerr('foo'))
    end, 'foo')
end
```

and both `expect.` calls seem to be tracked:
```
Running 1 tests
test passed: printerr_match_test:printerr

Test summary:
1/1 tests passed
2/2 checks passed
```

lethosor: A couple notes:
* This discussion is pretty tangential to the PR at this point. If you like, we can merge this PR and deal with how to handle `printerr()` calls afterwards. (I _think_ I have a decent idea of how to handle them, but I'd want to run it by you too.)
* Regarding how `expect.error` and `expect.printerr`-like functions use `...`: I am a little confused about how the current implementation is useful. Using `expect.eq` as the callback itself doesn't work well, as you noted. Maybe we should just make `...` behave like it does for `expect.eq`, i.e. adding additional information to test failure messages. Currently, `...` in `expect.error`/etc. does enable sharing code between tests a bit more concisely, e.g.:
    ```lua
    function shared_test_thing_not_callable(thing)
        expect.ne(thing, nil, 'extra failure message here')
        thing()
    end

    function test.number_not_callable()
        expect.error_match(shared_test_thing_not_callable, 'attempt to call a number', 3)
    end

    function test.string_not_callable()
        expect.error_match(shared_test_thing_not_callable, 'attempt to call a string', 'foo')
    end
    ```
    but this is equivalent to:
    ```lua
    function shared_test_thing_not_callable(thing)
        expect.ne(thing, nil, 'extra failure message here')
        thing()
    end

    function test.number_not_callable()
        expect.error_match(function() shared_test_thing_not_callable(3) end, 'attempt to call a number')
    end

    function test.string_not_callable()
        expect.error_match(function() shared_test_thing_not_callable('foo') end, 'attempt to call a string')
    end
    ```
    (and note that there is no support for an `'extra failure message here'` parameter to `error_match()`)

myk002: I've been coming to that opinion myself. the `(func, expected, func_args...)` syntax is confusing. converting it to expect a zero-arg callback would be an improvement. I think the better clarity is worth the increased verbosity.

lethosor: Yeah, I had a hard time getting `...` args passed through to the callback itself in my `expect_printerr` example. I think this would work instead, but it's harder to follow, and I don't think we need to preserve the current `...` behavior in any case.

```lua
function expect.printerr_match(func, matcher, ...)
    local printed_error
    local old_printerr = dfhack.printerr

    dfhack.with_finalize(
        function()
            dfhack.printerr = old_printerr
        end,
        function(...)
            dfhack.printerr = function(e)
                printed_error = e
            end
            func(...)
        end, ...)
```

myk002: i'm happy with getting this hashed out first. I don't have any bugs open at the moment.

How's this for an API? (may need rewording/reformatting for clarity):
```
-- matches messages output via dfhack.printerr() that happen during the
-- specified callback. the check passes if all printerr messages are matched by
-- specified patterns and no specified patterns remain unmatched.
-- if matcher is false, nil, or an empty table, matches 0 messages
-- if matcher is true, matches any 1 message
-- if matcher is a string, attempts to match 1 message with the pattern
-- if matcher is a table, it can contain a list of strings, which will be
-- matched in order, and/or it can contain a map of strings to numbers, which
-- will match the provided strings, in any order, the number of times specified.
-- if true is a key in the map, it can match any message. when this function
-- attempts to match a message, it will first try the next string in the list
-- (that is, the next numeric key). if that matcher doesn't exist or doesn't
-- match, it will try all string keys whose values are numeric and > 0. if none
-- of those match, it will check for a key equal to true with a value > 0.
-- if any matchers matched, they are removed from the list or have their value
-- decremented by 1, as appropriate.
function expect.printerr_match(matcher, func, comment)
```

`error_match` can have similar semantics and arguments, except it wouldn't need a table for a matcher since only one error can ever be thrown.

myk002: I put up code for discussion in DFHack/dfhack#1823

myk002:
merging this and my other queued-up refactor/unit test PRs. thank you so much for helping me get the infrastructure for this in place!

PR #264, opened 2021-03-09
Add mass-remove.lua, a gui script that helps with marking buildings/constructions for removal

teolandon:
I regularly make mistakes while planning buildings/constructions, and then have to remove them. While DF has a nice way of removing a lot of constructions at once, with the `d->n` menu, it does not have a similar menu for mass removal of buildings. And even if it did, it would probably be a different menu, meaning I'd have to navigate two different menus to remove pretty similar things.

This script allows the user to remove both buildings and constructions really fast. It can also 'unremove' things that are planned for removal, if the user accidentally marks the wrong thing for removal. On top of that, you can suspend/unsuspend jobs for
planned constructions. There are visual aids for what is and isn't marked for removal (since the game doesn't visually mark suspended constructions or buildings that are marked for removal. Refer to the initial commit description for details.

### DIFF
--- a/gui/mass-remove.lua
+++ b/gui/mass-remove.lua
@@ -6,6 +6,17 @@ gui/mass-remove
 ===============
 Allows removal of buildings/constructions and suspend/unsuspend using
 a box selection.
+
+The following marking modes are available.
+
+- Suspend (s): suspends the construction of a planned building/construction
+- Unsuspend (p): resumes the construction of a planned building/construction
+- Remove Construction (n): designates a construction (wall, floor, etc) for removal. Similar to the native Designate->Remove Construction menu in DF
+- Unremove Construction (c): cancels removal of a construction (wall, floor, etc)
+- Remove Building (x): designates a building (door, workshop, etc) for removal. Similar to the native Set Building Tasks/Prefs->Remove Building menu in DF
+- Unremove Building (b): cancels removal of a building (door, workshop, etc)
+- Remove All (a): designates both constructions and buildings for removal, and deletes planned buildings/constructions
+- Unremove All (u): cancels removal designations for both constructions and buildings
 ]====]
 
 local gui = require "gui"
@@ -15,18 +26,6 @@ local utils = require 'utils'
 
 MassRemoveUI = defclass(MassRemoveUI, guidm.MenuOverlay)
 
---[====[
-The actions that can be taken. Key bindings in parentheses.
-
---suspend (s): suspends the construction of a planned building/construction
---unsuspend (p): resumes the construction of a planned building/construction
---remove_n (n): designates a construction (wall, floor, etc) for removal. Similar to the native Designate->Remove Construction menu in DF
---unremove_n (c): cancels removal of a construction (wall, floor, etc)
---remove_x (x): designates a building (door, workshop, etc) for removal. Similar to the native Set Building Tasks/Prefs->Remove Building menu in DF
---unremove_x (b): cancels removal of a building (door, workshop, etc)
---remove_a (a): designates both constructions and buildings for removal, and deletes planned buildings/constructions
---unremove_a (u): cancels removal designations for both constructions and buildings
-]====]
 --used to iterate through actions with + and -
 local actions={"suspend", "unsuspend", "remove_n", "unremove_n", "remove_x", "unremove_x", "remove_a", "unremove_a"}
 local action_indexes=utils.invert(actions)
--- a/gui/mass-remove.lua
+++ b/gui/mass-remove.lua
@@ -207,7 +207,7 @@ function MassRemoveUI:renderOverlay()
     end
 
     --show box selection
-    if gui.blink_visible(250) and self.marking then
+    if not gui.blink_visible(500) and self.marking then
         local x_start, x_end = minToMax(self.mark.x, df.global.cursor.x)
         local y_start, y_end = minToMax(self.mark.y, df.global.cursor.y)
         paintMapTile(dc, vp, nil, self.mark, "+", COLOR_LIGHTGREEN)
--- a/gui/mass-remove.lua
+++ b/gui/mass-remove.lua
@@ -11,6 +11,7 @@ a box selection.
 local gui = require "gui"
 local guidm = require "gui.dwarfmode"
 local persistTable = require 'persist-table'
+local utils = require 'utils'
 
 MassRemoveUI = defclass(MassRemoveUI, guidm.MenuOverlay)
 
@@ -26,10 +27,9 @@ The actions that can be taken. Key bindings in parentheses.
 --remove_a (a): designates both constructions and buildings for removal, and deletes planned buildings/constructions
 --unremove_a (u): cancels removal designations for both constructions and buildings
 ]====]
-local actions={"suspend", "unsuspend", "remove_n", "unremove_n", "remove_x", "unremove_x", "remove_a", "unremove_a"}
 --used to iterate through actions with + and -
-local action_indexes={suspend=1,unsuspend=2,remove_n=3,unremove_n=4,remove_x=5,unremove_x=6,remove_a=7,unremove_a=8}
-local actions_length=tablelength(actions)
+local actions={"suspend", "unsuspend", "remove_n", "unremove_n", "remove_x", "unremove_x", "remove_a", "unremove_a"}
+local action_indexes=utils.invert(actions)
 
 MassRemoveUI.ATTRS {
     action="remove_a",
@@ -301,10 +301,10 @@ function MassRemoveUI:onInput(keys)
         self.action = "unremove_a"
         return
     elseif keys.SECONDSCROLL_UP then
-        self.action = actions[((action_indexes[self.action]-2) % actions_length)+1]
+        self.action = actions[((action_indexes[self.action]-2) % #actions)+1]
         return
     elseif keys.SECONDSCROLL_DOWN then
-        self.action = actions[(action_indexes[self.action] % actions_length)+1]
+        self.action = actions[(action_indexes[self.action] % #actions)+1]
         return
     end
 
--- a/gui/mass-remove.lua
+++ b/gui/mass-remove.lua
@@ -44,14 +44,6 @@ local function tablelength(T)
   return count
 end
 
---Needed because I couldn't find a direct way of getting a tile at certain coordinates.
-local function getTileAt(x, y, z)
-    local block = dfhack.maps.getTileBlock(x, y, z)
-    if block then
-        return block.designation[x%16][y%16]
-    end
-end
-
 --Helper to match a job of a particular type at tile (x,y,z) and run the callback function on the job.
 local function iterateJobs(jobType, x, y, z, callback)
     local joblist = df.global.world.jobs.list.next
@@ -104,10 +96,9 @@ end
 
 --Construction removals can either be marked as dig on the tile itself, or picked up as jobs. This function checks both.
 function MassRemoveUI:unremoveConstruction(x, y, z)
-    local tile = getTileAt(x,y,z)
-    tile.dig = df.tile_dig_designation.No
-
-    iterateJobs(df.job_type.RemoveConstruction, x, y, z, function(job) dfhack.job.removeJob(job) end)
+    local tileFlags, occupancy = dfhack.maps.getTileFlags(x,y,z)
+    tileFlags.dig = df.tile_dig_designation.No
+    dfhack.maps.getTileBlock(x,y,z).flags.designated = true
 end
 
 function MassRemoveUI:removeBuilding(x, y, z)
--- a/gui/mass-remove.lua
+++ b/gui/mass-remove.lua
@@ -82,14 +82,8 @@ local function paintMapTile(dc, vp, cursor, pos, ...)
     end
 end
 
-function MassRemoveUI:init()
-    self.saved_mode = df.global.ui.main.mode
-    df.global.ui.main.mode=df.ui_sidebar_mode.LookAround
-end
-
-function MassRemoveUI:onDestroy()
-    persistTable.GlobalTable.massRemoveAction=self.action
-    df.global.ui.main.mode = self.saved_mode
+function MassRemoveUI:onAboutToShow(parent)
+    gui.simulateInput(parent, df.interface_key.D_LOOK)
 end
 
 function MassRemoveUI:changeSuspendState(x, y, z, new_state)
@@ -270,12 +264,6 @@ function MassRemoveUI:onRenderBody(dc)
 end
 
 function MassRemoveUI:onInput(keys)
-    if df.global.cursor.x==-30000 then
-        local vp=self:getViewport()
-        df.global.cursor=xyz2pos(math.floor((vp.x1+math.abs((vp.x2-vp.x1))/2)+.5),math.floor((vp.y1+math.abs((vp.y2-vp.y1)/2))+.5), vp.z)
-        return
-    end
-
     if keys.CUSTOM_S then
         self.action = "suspend"
         return
--- a/gui/mass-remove.lua
+++ b/gui/mass-remove.lua
@@ -254,13 +254,13 @@ function MassRemoveUI:onRenderBody(dc)
 
     dc:seek(1,9)
     dc:pen(self:getColor("suspend")):key_string("CUSTOM_S", "Suspend"):newline(1)
-    dc:pen(self:getColor("unsuspend")):key_string("CUSTOM_P", "Unsuspend"):newline():newline(1)
+    dc:pen(self:getColor("unsuspend")):key_string("CUSTOM_SHIFT_S", "Unsuspend"):newline():newline(1)
     dc:pen(self:getColor("remove_n")):key_string("CUSTOM_N", "Remove Construction"):newline(1)
-    dc:pen(self:getColor("unremove_n")):key_string("CUSTOM_C", "Unremove Construction"):newline():newline(1)
+    dc:pen(self:getColor("unremove_n")):key_string("CUSTOM_SHIFT_N", "Unremove Construction"):newline():newline(1)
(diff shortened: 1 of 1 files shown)


gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+}
+
+--Helper functions.
+local function tablelength(T)
+  local count = 0
+  for _ in pairs(T) do count = count + 1 end
+  return count
+end
+
+--Needed because I couldn't find a direct way of getting a tile at certain coordinates.
```
myk002, line 47: I think what you're looking for is dfhack.maps.getTileFlags()

See https://github.com/DFHack/scripts/blob/19f84c9a8f4acac933d65c8dfe5e396f4caf84bd/internal/quickfort/dig.lua#L345 where quickfort does something similar

myk002:
This gives me an idea for a "clear" command for buildingplan that can destroy all planned buildings that don't have all items matched yet.

Quickfort already has an "undo" command that removes all buildings previously placed with a specified blueprint.

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+        dfhack.buildings.deconstruct(building)
+    end
+end
+
+function MassRemoveUI:unremoveBuilding(x, y, z)
+    local building = dfhack.buildings.findAtTile(x, y, z)
+    if building then
+        for i, job in ipairs(building.jobs) do
+            if job.job_type == df.job_type.DestroyBuilding then
+                building.jobs:erase(i)
```
lethosor, line 131: You probably also want to remove this from `world.jobs` too. I suspect that in this case, the job simply wouldn't be cancelled, but a DF crash can also result in some cases.

Actually, `dfhack.job.removeJob(job)` should remove the job from everywhere, so I'd recommend switching to that. Since you're starting with a reference to the building, I think the loop around here would still be needed.

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+                local building = dfhack.buildings.findAtTile(x, y, vp.z)
+                if building and dfhack.buildings.markedForRemoval(building) then
+                    paintMapTile(dc, vp, nil, xyz2pos(x, y, vp.z), "x", COLOR_LIGHTRED)
+                end
+            end
+        end
+    end
+
+    --show box selection
+    if gui.blink_visible(250) and self.marking then
```
lethosor, line 210: We don't have a standard here, but I would probably increase this to `400` or `500` since it's potentially a large blinking area.

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+--[====[
+The actions that can be taken. Key bindings in parentheses.
+
+--suspend (s): suspends the construction of a planned building/construction
+--unsuspend (p): resumes the construction of a planned building/construction
+--remove_n (n): designates a construction (wall, floor, etc) for removal. Similar to the native Designate->Remove Construction menu in DF
+--unremove_n (c): cancels removal of a construction (wall, floor, etc)
+--remove_x (x): designates a building (door, workshop, etc) for removal. Similar to the native Set Building Tasks/Prefs->Remove Building menu in DF
+--unremove_x (b): cancels removal of a building (door, workshop, etc)
+--remove_a (a): designates both constructions and buildings for removal, and deletes planned buildings/constructions
```
lethosor, line 26: Could you move this into the first block comment above? That would include it in the user-facing docs; it's not visible there currently, and I think it's useful enough to be there.

It would also need to be formatted as RST in that case. I think the only changes needed would be to
- replace `--` with `- ` (that includes a trailing space)
- replace the internal names with human-readable names (e.g. `remove_n` -> `Remove`), ideally ones that match the UI

...but I can help with any syntax issues that come up. https://docs.dfhack.org/en/stable/docs/Documentation.html explains how to build the docs, but it doesn't work well if you don't also have the DFHack repo cloned.

lethosor (commented):
Nice! I'm seeing a lot of familiar-looking code, so I'm guessing you adopted this from an example. That's good, and means it should work with plugins like TWBT out of the box (just curious, have you tried it with TWBT?).

Most of my comments are from a static review, although I did play around a bit with building (un)removal so far.

gui/mass-remove.lua
teolandon: I mainly want the box selection blinking so that you can see what is under it, so you know what you're removing. So if I match the `500` blink delay of the "marked buildings/constructions" overlay, it'll be hard to see if the buildings/constructions you're selecting are marked or not.

After playing with it for a bit, I decided to go for `if not gui.blink_visible(250)`, so that the two blinks alternate and you can see both the box selection and the marks. Give it a try and let me know if it looks ok!

Implemented in 439612e

gui/mass-remove.lua
teolandon: Absolutely! I got this re-formatted and moved up, and we can check to see if it builds the docs right. I got the DFHack repo cloned on my machine, but not CMake. Would just running Sphinx work in this case to check if it builds them right?

gui/mass-remove.lua
teolandon: Good catch. I adopted this from `gui/advfort`, but that's an adventure mode plugin so it might be fine there. I did manage to use `dfhack.job.removeJob(job)` after I discovered it (it's not really in the docs). I used it in `MassRemoveUI:unremoveConstruction`, but that method was written after `MassRemoveUI:unremoveBuilding`, so I didn't go back to change it.

P.S. I've been testing this on a real fort I've been playing in because I'm dumb, so I sure hope those dangling jobs don't crash my fort down the line.

Fixed in ea34775b

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+    dc:pen(COLOR_WHITE)
+    if self.marking then
+        dc:newline(1):key_string("LEAVESCREEN", "Cancel selection")
+    else
+        dc:newline(1):key_string("LEAVESCREEN", "Back")
+    end
+end
+
+function MassRemoveUI:onInput(keys)
+    if df.global.cursor.x==-30000 then
```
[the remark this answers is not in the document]
teolandon: It's not necessary after all, with the `D_LOOK` implementation. Amazing! This part was copied from `gui/stamper`, wasn't sure why it was there but decided to keep it.

Removed in 816bd92

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+end
+
+function MassRemoveUI:removeConstruction(x, y, z)
+    dfhack.constructions.designateRemove(x, y, z)
+end
+
+--Construction removals can either be marked as dig on the tile itself, or picked up as jobs. This function checks both.
+function MassRemoveUI:unremoveConstruction(x, y, z)
+    local tile = getTileAt(x,y,z)
+    tile.dig = df.tile_dig_designation.No
```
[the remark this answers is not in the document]
teolandon: Yay!

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+--remove_n (n): designates a construction (wall, floor, etc) for removal. Similar to the native Designate->Remove Construction menu in DF
+--unremove_n (c): cancels removal of a construction (wall, floor, etc)
+--remove_x (x): designates a building (door, workshop, etc) for removal. Similar to the native Set Building Tasks/Prefs->Remove Building menu in DF
+--unremove_x (b): cancels removal of a building (door, workshop, etc)
+--remove_a (a): designates both constructions and buildings for removal, and deletes planned buildings/constructions
+--unremove_a (u): cancels removal designations for both constructions and buildings
+]====]
+local actions={"suspend", "unsuspend", "remove_n", "unremove_n", "remove_x", "unremove_x", "remove_a", "unremove_a"}
+--used to iterate through actions with + and -
+local action_indexes={suspend=1,unsuspend=2,remove_n=3,unremove_n=4,remove_x=5,unremove_x=6,remove_a=7,unremove_a=8}
```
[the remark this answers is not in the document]
teolandon: Nifty! Thanks for the tip.

Implemented in cec8752

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+local actions_length=tablelength(actions)
```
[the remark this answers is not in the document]
teolandon: Oops. I'm guilty of moving that around to keep it with the rest of the helper functions right before pushing, without re-testing the script. My bad, and thanks for still testing the script out despite it not working by default.

Thanks for the tip with the `#` operator. As you might have figured out, I'm not that familiar with Lua.

Fixed in cec8752

gui/mass-remove.lua
teolandon: Huh, you can mutate the returned flag table, and that affects the tile itself. I assumed it would be a copy or something like that.

Thanks for the reference to quickfort, I picked up another trick from over there. Instead of searching for jobs that remove constructions, and canceling them, I can just set the dig flag to `No` and also set the block's `designated` flag to true, and the game will look for jobs associated with constructions that are not being removed anymore, and cancel them. Saves me a lot of work!

Implemented in cff1bde

teolandon:
@myk002 a `clear` command in buildingplan would be great! This script is good enough for me now, but it would be cool to do it right off of buildingplan plugin.

teolandon:
@lethosor You guessed right, I built most of this off of `gui/stamper`! While the main DFHack docs did help a lot, I think looking at other Lua scripts helped just as much, they're a great resource.

teolandon:
If you have any naming suggestions, I am happy to take them. I don't know if `mass-remove` or `mass-removal` is descriptive enough, or if something clever-er is available. Also if there are any better names or keybinds for the actions, let me know. `Unremove...` is a bit awkward, but I think it makes sense.

Also currently, both `remove` actions will also completely delete planned constructions/buildings, which is something that cannot be undone, like marking a building for removal. I'm not sure if that's intuitive or unexpected. I think it could be avoided by checking the right flag before removing. On the other hand, removing planned constructions/buildings is useful. Maybe it could have its own action associated with it.

gui/mass-remove.lua
```diff
@@ -0,0 +1,314 @@
...
+
+gui/mass-remove
+===============
+Allows removal of buildings/constructions and suspend/unsuspend using
+a box selection.
+
+The following marking modes are available.
+
+- Suspend (s): suspends the construction of a planned building/construction
+- Unsuspend (p): resumes the construction of a planned building/construction
```
myk002, line 13: perhaps a slightly more mnemonic lettering would be to use lower case for one action and the upper case of that letter for the corresponding "un" action. e.g. Suspend (s), Unsuspend (S), Remove Construction (n), Unremove Construction (N), Remove Building (x), Unremove Building (X), Remove All (a), Unremove All (A)

myk002:
> Also currently, both `remove` actions will also completely delete planned constructions/buildings, which is something that cannot be undone

With about a week's work, I think it could be done. You could very well keep track of the building types you destroy and reinstate them if the user requests an undo. you could even farm this functionality out to the blueprint plugin for saving state and the quickfort script for reapplying. quickfort does idempotent reapplications of #build blueprints, so it will only rebuild buildings that were completely destroyed and skip over the ones that you just need to unremove. with the current implementation, building material item filters set on buildingplan buildings would be lost if the filters on those building types have been changed since the building was originally planned, but it wouldn't be too much trouble to add the API to buildingplan for saving and specifying item filters.

this copy-paste mechanism could be done with a temporary blueprint file, or you could modify the blueprint plugin to provide a stream sink where you could save the output in a lua table. then we'd need to modify quickfort to read the blueprint from a table instead of a file. shouldn't be too hard, but might be out of scope for the initial version of this script : )

gui/mass-remove.lua
```diff
@@ -0,0 +1,314 @@
...
+--[====[
+
+gui/mass-remove
+===============
+Allows removal of buildings/constructions and suspend/unsuspend using
+a box selection.
+
+The following marking modes are available.
+
+- Suspend (s): suspends the construction of a planned building/construction
```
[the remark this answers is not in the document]
teolandon: I started doing this then realized that there's no indication on the screen about whether a building is buildingplan or not. So a user could try to unsuspend a bunch of buildings and not know why it's not working, because they don't realize it's because those buildings are from buildingplan.

I tried a bunch of things to indicate the buildings, wanted to do something similar to the `pathable` plugin, which tints the screen based on conditions, but it seems like it does that through an actual C++ call, and trying to replicate that in Lua didn't work for me.

It's probably not that bad though, as long as the user knows that buildingplan buildings cannot be suspended/unsuspended, they should figure out why their suspend/unsuspend actions aren't actually being applied.

teolandon:
@myk002 Yeah I didn't mean it can never be undone, just that it cannot be undone using the script currently, like the "mark for removal" actions can. It did pass my mind to keep track of removed buildings and allow the user to restore them, but in my own use of the script, I haven't accidentally deleted planned buildings I didn't mean to, so I decided to wait on that. I'm sure that if I do make a big mistake like that I'll be pretty motivated to make an undo for removals.

My bigger qualm with this was not the actual coding, but the presentation. I'd want a nice clear way of showing that there used to be a plan for a building here, and you removed it, but you can restore it if you want. And I'm just not that good at graphic design to know what to pick to show that. Like I already have blinking characters for the buildings that are marked for removal, and the plans that are suspended, so I can't make that a blinking character too, or it might be confused. Just some thoughts.

The blueprint + quickfort usage idea is very smart, would be very fun to try to incorporate that into the script.

gui/mass-remove.lua
teolandon: See, that's the sorta stuff that I won't come up with by myself because I'm not that good at UX. Love it, I'll change the keybindings real fast, very intuitive.

gui/mass-remove.lua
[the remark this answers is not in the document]
teolandon: Applied in 2896b67

gui/mass-remove.lua
teolandon: Applied in 480f35b

teolandon:
I started testing this out with different tilesets. I noticed that more expansive graphical tilesets that try to replace _everything_ (Meph for example), replace even the overlay characters I use to show which buildings/constructions are marked for removal. So instead of blinking an `s` for suspended buildings, it blinks a spider tile instead.

I'm not too well versed in this whole thing, so let me know if there's any way to bypass that and show actual text on the screen. I know TWBT stands for Text Will Be Text, and it prevents tiles from showing up in the status bar and all other places, but I don't know if there's any way to utilize it for the main screen. If not, no big deal.

myk002:
> My bigger qualm with this was not the actual coding, but the presentation. I'd want a nice clear way of showing that there used to be a plan for a building here, and you removed it, but you can restore it if you want.

At some point O(months from now), I hope to provide a solution for this problem in quickfort, showing a shadow preview of what would be applied. The same UX issues need to be solved, but at least we'd only have to solve them once. This is a thought for the future, though. This PR is good as-is.

gui/mass-remove.lua
```diff
@@ -0,0 +1,336 @@
...
+        local stile = vp:tileToScreen(pos)
+        if stile.z == 0 then
+            dc:map(true):seek(stile.x,stile.y):char(...):map(false)
+        end
+    end
+end
+
+function MassRemoveUI:init()
+    self.saved_mode = df.global.ui.main.mode
+    df.global.ui.main.mode=df.ui_sidebar_mode.LookAround
```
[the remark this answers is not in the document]
teolandon: Implemented in 88c7c72

gui/mass-remove.lua
lethosor: Whoops, sorry for the delay. Yes, `docs/build.sh` from the DFHack repo should work (on non-Windows), or you can run Sphinx manually - https://docs.dfhack.org/en/latest/docs/Documentation.html#building-the-documentation has more instructions.

lethosor: It looks like this, for reference (and it builds warning-free).

The keys did get out of sync, though (`p` changed to `Shift-s`, for instance) - It's probably fine to just remove these because they're on-screen when the script is running.

image

lethosor:
> I started testing this out with different tilesets. I noticed that more expansive graphical tilesets that try to replace _everything_ (Meph for example), replace even the overlay characters I use to show which buildings/constructions are marked for removal. So instead of blinking an `s` for suspended buildings, it blinks a spider tile instead.
> 
> I'm not too well versed in this whole thing, so let me know if there's any way to bypass that and show actual text on the screen. I know TWBT stands for Text Will Be Text, and it prevents tiles from showing up in the status bar and all other places, but I don't know if there's any way to utilize it for the main screen. If not, no big deal.

As far as I'm concerned, that's not a huge deal. The primary issue I'm concerned about is that scripts that draw to the map area *without* calling `:map(true)` will confuse TWBT, and potentially cause it to black out the entire map area. Based on what you described and the code I'm seeing, I'm pretty confident that's not an issue.

gui/mass-remove.lua
```diff
@@ -0,0 +1,350 @@
...
+
+-- Helper functions.
+local function tablelength(T)
+  local count = 0
+  for _ in pairs(T) do count = count + 1 end
+  return count
+end
+
+-- Helper to match a job of a particular type at tile (x,y,z) and run the callback function on the job.
+local function iterateJobs(jobType, x, y, z, callback)
```
lethosor, line 48: One thing I missed earlier: I think `utils.listpairs()` would handle the loop for you. It's fairly simple logic, though, so I won't ask you to change it.

lethosor (approved):
Thanks for the updates!

lethosor:
FYI: I made a couple more changes in the commits referenced above. Nothing huge, just wanted to keep you in the loop.

teolandon:
Wow, it's been about a month huh. Sorry, life got in the way and I didn't have time to work on this anymore, just got back to having free time and look at that, it's been merged already!

Thank you for following this up and making those changes in the end.

PR #265, opened 2021-03-22
[test] don't install scripts/test files

myk002:
this will prevent script tests from being installed. this will allow tests to be merged to master without letting them affect CI in DFHack/dfhack. once we get CI figured out for DFHack/scripts, then we can enable installation of the tests in #262 

DFHack/dfhack#1690

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,4 +4,5 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
                        PATTERN "*.rb"
                        PATTERN "*.json"
                        PATTERN "3rdparty" EXCLUDE
+                       PATTERN "scripts/test" EXCLUDE
 )


lethosor:
Replying to https://github.com/DFHack/scripts/pull/262#issuecomment-803524827 here:

> I meant more that checking in the EXCLUDE line now would prevent subsequent merges of script tests from affecting CI for later PRs in DFHack/dfhack.

My point was that I don't think we need this change for CI purposes, because this line determines what gets installed to `<DF>/hack/scripts` and things like the linters don't run on files in the DF folder.

However, looking at it again, I realize you mean that the test runner might pick up unit tests that get installed to `<DF>/hack/scripts/test` instead. I don't _think_ this would be the case, because tests are searched for in a separate "test path" (the --test-dir argument to test/main.lua) and not the script paths (`test/main` is an exception, hence dfhack/dfhack#1815). I don't think this would break anything, so merging it sooner than the rest of #262 is fine with me.

I was a bit concerned that the `scripts/test` pattern wouldn't work across platforms, or at all since the existing `3rdparty` entry is relative to the top of the scripts repo. But from https://cmake.org/cmake/help/latest/command/install.html, it looks like paths are guaranteed to be absolute and use forward slashes as path separators, so this should work. LGTM!

lethosor (approved):

myk002:
> However, looking at it again, I realize you mean that the test runner might pick up unit tests that get installed to `<DF>/hack/scripts/test` instead. I don't _think_ this would be the case, because tests are searched for in a separate "test path" (the --test-dir argument to test/main.lua) and not the script paths (`test/main` is an exception, hence [DFHack/dfhack#1815](https://github.com/DFHack/dfhack/pull/1815)).

That is exactly the issue I was worried about. I may be missing something, but isn't the test dir equal to the install path here?
`scripts` gets installed to `<DF>/hack/scripts` so `scripts/test` would end up in `<DF>/hack/scripts/test`, which is the same directory that dfhack (the repo)/test gets installed to. Hence, without this PR, tests from the scripts repo would get picked up by the test runner when it scans `<DF>/hack/scripts/test`.

> I was a bit concerned that the `scripts/test` pattern wouldn't work across platforms, or at all since the existing `3rdparty` entry is relative to the top of the scripts repo. But from https://cmake.org/cmake/help/latest/command/install.html, it looks like paths are guaranteed to be absolute and use forward slashes as path separators, so this should work. LGTM!

Cool, thanks! -- I'll merge it in

lethosor:
Ah, you're right: https://github.com/DFHack/dfhack/blob/d5d41cd374fdda6093f3be8b5b23f97e7af6c380/CMakeLists.txt#L504-L505

I usually run with a test folder that *is* isolated from the scripts folder, so I didn't realize that these tests would install into the test folder that CI uses (and without the leading `scripts/` prefix either). Whoops!

PR #267, opened 2021-04-04
Add usage to setfps.lua

britishben:
Add usage to setfps when called without an argument (or with `-help`). 

Current script produces a traceback error:
```D:\Desktop\DF\df_47_05_win/hack/scripts/setfps.lua:19: attempt to concatenate a nil value (local 'cap')
stack traceback:
        D:\Desktop\DF\df_47_05_win/hack/scripts/setfps.lua:19: in local 'script_code'
        D:\Desktop\DF\df_47_05_win\hack\lua\dfhack.lua:680: in function 'dfhack.run_script_with_env'
        (...tail calls...)```

### DIFF
--- a/setfps.lua
+++ b/setfps.lua
@@ -15,6 +15,10 @@ Usage::
 local cap = ...
 local capnum = tonumber(cap)
 
+if not cap or cap=='-help' then
+	qerror('Usage:\n\tsetfps <number>')
+end
+
 if not capnum or capnum < 1 then
     qerror('Invalid FPS cap value: '..cap)
 end
--- a/setfps.lua
+++ b/setfps.lua
@@ -16,7 +16,7 @@ local cap = ...
 local capnum = tonumber(cap)
 
 if not cap or cap=='-help' then
-	qerror('Usage:\n\tsetfps <number>')
+    qerror('Usage:\n\tsetfps <number>')
 end
 
 if not capnum or capnum < 1 then

### CODE REVIEW

File: setfps.lua [hunk 1]
```diff
@@ -15,6 +15,10 @@ Usage::
 local cap = ...
 local capnum = tonumber(cap)
 
+if not cap or cap=='-help' then
```
myk002, line 18: I'm not sure what the standard should be, but some scripts use two dashes for "long" parameters and accept "short" parameters with one dash. E.g. `-h` (one dash) and `--help` (two dashes).

lethosor: I've often just used `cap:match('help')` to catch everything.

britishben: Happy to change that - just went by the few other DFhack scripts I checked against. Typically the standard is `-h`/`--help` as @myk002 stated, but for instance [assign-skills](https://github.com/DFHack/scripts/blob/c44df3b0dfdc35230f91ff2d95a619d69ac88d84/assign-skills.lua#L41) only responds to `-help`.

```suggestion
if not cap or cap:match('help') then
```


Alternately, we could just combine it with the `if not capnumber` line, so anything that doesn't resolve to a number triggers the usage text - not sure it's useful to return the text if we're not going to do anything with it.

lethosor: Yeah, that's what I was thinking. Your alternative is fine with me, assuming that means combining the messages into something like this:
```
Invalid FPS cap value: <val>
Usage: setfps <number>
```
The original issue here is that `..cap` (line 23 currently) fails for things that can't be converted implicitly to strings, e.g. `nil`. Changing that to `..tostring(cap)` should work, though (maybe `tostring(cap or '')` to avoid `nil` showing up in the output).

lethosor (approved):

setfps.lua
```diff
@@ -15,6 +15,10 @@ Usage::
...
+    qerror('Usage:\n\tsetfps <number>')
```
lethosor, line 19: FYI: I think I'm going to change this to a `print()` call followed by a `return` (which exits the script without raising an error). That's a bit more consistent with other help text, which isn't usually printed as an error message (in red text).

PR #273, opened 2021-04-06
bodyswap: handle companions

AtomicChicken:
Requires DFHack/df-structures#426.

Fix: 
- Prior party members no longer tag along after bodyswapping and reloading the map.

Improvement:
- Companions of bodyswapping targets are now added to the adventurer party (when swapping into a member of a previous party, for example) and can be swapped into using the in-game tactical party system.

### DIFF
--- a/bodyswap.lua
+++ b/bodyswap.lua
@@ -76,8 +76,10 @@ function createNemesis(unit)
 end
 
 function isPet(nemesis)
-  if nemesis.unit and nemesis.unit.relationship_ids.Pet ~= -1 then
-    return true
+  if nemesis.unit then
+    if nemesis.unit.relationship_ids.Pet ~= -1 then
+      return true
+    end
   elseif nemesis.figure then -- in case the unit is offloaded
     for  _, link in ipairs(nemesis.figure.histfig_links) do
       if link._type == df.histfig_hf_link_pet_ownerst then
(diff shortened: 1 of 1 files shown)


bodyswap.lua
```diff
@@ -75,13 +75,64 @@ function createNemesis(unit)
...
+end
+
+function processNemesisParty(nemesis, targetUnitID, alreadyProcessed)
+-- configures the target and any leaders/companions to behave as cohesive adventure mode party members
+  local alreadyProcessed = alreadyProcessed or {}
+  alreadyProcessed[tostring(nemesis.id)] = true
+
+  local nemUnit = nemesis.unit
+  if nemesis.unit_id == targetUnitID then -- the target you're bodyswapping into
+    df.global.ui_advmode.interactions.party_core_members:insert('#', nemesis.figure.id)
```
lethosor, line 100: I don't have an adventure save handy to test this: do you have any idea if DF expects these vectors of IDs to be sorted? If so, `utils.insert_sorted()` might be a good idea in these cases.

AtomicChicken: My impression is that it doesn't based on the observation that the IDs in `party_extra_members` and `party_pets` are ordered according to the sequence in which the corresponding figures joined the party, though it's difficult to say for certain with `party_core_members` as additional figures aren't added to this list during normal gameplay.

lethosor: Ok! Hopefully DF is consistent, then.

lethosor (approved):
Managed to swap between party members a couple times - looks good!

PR #276, opened 2021-04-13
Add fix/drop-webs.lua

Bumber64:
A script to cause floating webs to fall

### DIFF
--- a/fix/webs.lua
+++ b/fix/webs.lua
@@ -4,7 +4,7 @@
 --[====[
 
 fix/webs
-======
+========
 Turns floating webs into projectiles, causing them to fall down to a valid surface
 Use ``fix/webs -all`` to turn all webs into projectiles, causing webs to fall out of branches, etc.
 
--- a/fix/webs.lua
+++ b/fix/webs.lua
@@ -1,0 +1,60 @@
+--Cause floating webs to fall
+--By Bumber
+--@module = true
+--[====[
+
+fix/webs
+======
+Turns floating webs into projectiles, causing them to fall down to a valid surface
+Use ``fix/webs -all`` to turn all webs into projectiles, causing webs to fall out of branches, etc.
+
+]====]
+local utils = require "utils"
+
+function fix_webs(air_only)
+    if not dfhack.isMapLoaded() then
+        qerror("Error: Map not loaded!")
+    end
+    
+    local count = 0
+    for i, item in ipairs(df.global.world.items.all) do
+        if item:getType() == df.item_type.THREAD and
+        item.flags.spider_web and
+        item.flags.on_ground and
+        not item.flags.in_job then
+            local valid_tile = true
+            
+            if air_only then
+                local tt = dfhack.maps.getTileBlock(item.pos).tiletype[item.pos.x%16][item.pos.y%16]
+                valid_tile = (tt == df.tiletype.OpenSpace)
+            end
+            
+            if valid_tile then
+                local proj = dfhack.items.makeProjectile(item)
+                proj.flags.no_impact_destroy = true
+                proj.flags.bouncing = true
+                proj.flags.piercing = true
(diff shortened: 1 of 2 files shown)


fix/webs.lua
```diff
@@ -0,0 +1,60 @@
...
+        end
+    end
+    print(tostring(count).." webs projectilized!")
+end
+
+function main(...)
+    local validArgs = utils.invert({"all"})
+    local args = utils.processArgs({...}, validArgs)
+    
+    if args.all then
```
myk002, line 51: You can simplify this to just `fix_webs(not args.all)`

fix/webs.lua
```diff
@@ -0,0 +1,60 @@
...
+]====]
+local utils = require "utils"
+
+function fix_webs(air_only)
+    if not dfhack.isMapLoaded() then
+        qerror("Error: Map not loaded!")
+    end
+    
+    local count = 0
+    for i, item in ipairs(df.global.world.items.all) do
```
myk002, line 20: Is there a more specific item list we can use here so we can skip obviously unrelated items?

lethosor: Just from the name: `df.global.world.items.other.ANY_WEBS` might do the trick

fix/webs.lua
```diff
@@ -0,0 +1,60 @@
...
+        if item:getType() == df.item_type.THREAD and
+        item.flags.spider_web and
+        item.flags.on_ground and
+        not item.flags.in_job then
+            local valid_tile = true
+            
+            if air_only then
+                local tt = dfhack.maps.getTileBlock(item.pos).tiletype[item.pos.x%16][item.pos.y%16]
```
myk002, line 28: dfhack.maps.getTileType(item.pos) might be easier to use here. See, for example, https://github.com/DFHack/scripts/blob/55ac6f7269a10a1b3e525da2c3f117675b529289/internal/quickfort/dig.lua#L542

lethosor:
I would probably name this something like `drop-webs` or `detach-webs`. Generally, `fix/` scripts are reserved for scripts that fix bugs, which this isn't really (as far as I know - let me know if I'm wrong!), and IMO `drop-webs` makes it clearer what this script actually _does_ to webs.

Bumber64:
> 
> 
> Generally, `fix/` scripts are reserved for scripts that fix bugs, which this isn't really (as far as I know - let me know if I'm wrong!)

Webs floating after trees are chopped down definitely seems like a bug.

lethosor:
I suppose that makes sense. I was thinking specifically of scripts that fixed reported DF bugs, but that isn't necessarily a requirement.

We do, however, also have `clear-webs` in #285 now, so I'm debating whether the inconsistency is worth resolving somehow. We could alias `drop-webs` to `fix/drop-webs` fairly easily, for example.

Bumber64:
Looks like it *is* a reported bug: http://www.bay12games.com/dwarves/mantisbt/view.php?id=10004

Would a better alias be "fix/floating-webs"?

lethosor:
Gah, I missed that. Thanks! I'll link to it in the docs.

My (mild) concern regarding the `fix/` prefix was that someone who remembers `clear-webs` might be confused by trying to run `drop-webs`, not realizing that it's `fix/drop-webs`. `fix/floating-webs` would have the same effect. It's a bit complicated by the fact that scripts in subfolders are documented on [separate pages like this one](https://docs.dfhack.org/en/stable/docs/_auto/fix.html), and automatic suggestions won't kick in if someone tries to run `drop-webs` in the console (because that only matches prefixes - running "fix/drop" in the console would give "Did you mean drop-webs?", but "drop" would not).

`ls -a` would still turn up `fix/drop-webs`, though, and this script will be documented in the changelog, so I think this is a fairly minor concern.

lethosor (approved):

PR #281, opened 2021-05-04
[gui/blueprint] minimal logic for initial script

myk002:
DFHack/dfhack#1842

I coded up a fully-featured interface, but I have several questions on the implementation, which I wanted to tackle in separate, focused code reviews. Moreover, I wanted to get a simple version of the interface merged first to unblock the changes I have queued for the blueprint plugin itself.

The command at the end that launches `blueprint` won't work, of course, until the commandline parsing changes to blueprint are merged from DFHack/dfhack#1849.

This PR also depends on DFHack/dfhack#1840 for fortress mode support for the tests.

Future PRs for `gui/blueprint` will support:
- getting into a cursor-supporting mode from any game screen and restoring the previous screen mode when the script exits
- mouse support for selecting boundary tiles
- a live status line showing the dimensions of the currently selected area
- edit widget for setting the blueprint name (including detection of name collision with existing blueprint files)
- widgets to configure which blueprint phases to output
- presetting config values from commandline parameters


test/gui/blueprint.lua
```diff
@@ -0,0 +1,105 @@
...
+        },
+        function()
+            load_ui()
+            expect.eq('dfhack/lua', dfhack.gui.getCurFocus(true))
+            guidm.setCursorPos({x=10, y=20, z=30})
+            send_keys('SELECT')
+            send_keys('CURSOR_RIGHT', 'CURSOR_DOWN', 'CURSOR_DOWN',
+                      'CURSOR_UP_Z', 'CURSOR_UP_Z', 'CURSOR_UP_Z')
+            send_keys('SELECT')
+            expect.ne('dfhack/lua', dfhack.gui.getCurFocus(true))
```
lethosor, line 56: Might want to check for absence of a `dfhack/lua` or `dfhack/` prefix here (I don't think `dfhack_viewscreen::is_instance()` is exposed to Lua...), because Lua screens can add things to the focus string if they want. Not really sure if that's a valid concern here, though.

Glad the (minimal) mocking support is useful!

myk002: done. and the mocking support has been great! it really simplifies the structure of these tests.

lethosor: Looks to me like you changed the other two calls to `expect.ne` but not this one.

gui/blueprint.lua
```diff
@@ -19,4 +19,159 @@ values set in the interface. See the `blueprint` documentation for information
...
+    if self.get_mark_fn() then
+        return 'Select the second corner'
+    end
+    return 'Select the first corner'
+end
+
+BlueprintUI = defclass(BlueprintUI, guidm.MenuOverlay)
+BlueprintUI.ATTRS {
+    presets={},
+    frame_inset=1,
```
lethosor, line 59: To allow more specificity in tests (and keybinding contexts, etc.) you could define a `focus_path` field here to cause `getCurFocus()` to return a more detailed string than "dfhack/lua". 

I _think_ this should work (untested):
```suggestion
    frame_inset=1,
    focus_path='blueprint',
```

Example change in tests:

```diff
-expect.eq('dfhack/lua', dfhack.gui.getCurFocus(true))
+expect.eq('dfhack/lua/blueprint', dfhack.gui.getCurFocus(true))
```

test/gui/blueprint.lua
myk002: done.

gui/blueprint.lua
```diff
@@ -19,4 +19,160 @@ values set in the interface. See the `blueprint` documentation for information
...
+    local dc = gui.Painter.new(self.df_layout.map)
+
+    if gui.blink_visible(500) then
+        local cursor = df.global.cursor
+        -- clip blinking region to viewport
+        local _,y_start,y_end = min_to_max(self.mark.y, cursor.y, vp.y1, vp.y2)
+        local _,x_start,x_end = min_to_max(self.mark.x, cursor.x, vp.x1, vp.x2)
+        for y=y_start,y_end do for x=x_start,x_end do
+            local pos = xyz2pos(x, y, cursor.z)
+            if x == cursor.x and y == cursor.y then
```
lethosor, line 126: Perhaps this could use `same_xyz()` instead (unless performance is a concern here)? The check for z wouldn't affect anything since you've set `pos.z = cursor.z` on the line above. It would also allow removing the goto/continue statement without negating the compound statement here. For instance:

```lua
if not same_xyz(pos, cursor) then
   local stile = .......
end

lethosor (commented):
Looking good, will try more along with https://github.com/DFHack/dfhack/issues/1842

gui/blueprint.lua
myk002: cleaner. i like. done : )

lethosor (approved):
This looks great! I was about to comment on a couple things (namely, how to name blueprints and the current behavior when launching with no cursor present) but I see you are already planning to address those. If this is blocking something, I am certainly fine with merging it. Sorry I took so long (IRL things...).

On the UX side of things: from the GUI behavior (just closing when done) and the command-line behavior (logging the command that ran), it wasn't clear to me _where_ the blueprint was saved, since I have several in the `blueprints` folder. I ended up using `ls -t blueprints` to figure it out. I assume that will be somewhat more clear when users can specify a name to save to, but maybe a confirmation message at the end (either a dialog or console message) could be good.

myk002:
> maybe a confirmation message at the end (either a dialog or console message) could be good.

+1 I actually have this ready to go as the next PR - I report the list of files that were created in a message box.

It involved actually creating a Lua entrypoint for the `blueprint` plugin so I could return the filenames, so now `gui/blueprint` invokes the plugin that way instead of via `dfhack.run_command()`. This was a big enough change that I wanted to separate it out from the current PRs, which I tried to keep as small as possible.

PR #283, opened 2021-05-13
Created set-timeskip-duration.lua

AtomicChicken:
This script allows the player to modify the length of the ~2 week timeskip that precedes the start of a new fortress/adventure mode game. This allows players to do things like retiring a fortress and letting history (slowly) run its course for a century or so before the next fortress/adventure playthrough, or making a new adventurer appear nearly instantaneously after the death/retirement of the previous one.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,7 +18,7 @@ that repo.
 - `devel/luacov`: generate code test coverage reports for script development. define the ``DFHACK_ENABLE_LUACOV=1`` environment variable to start gathering coverage metrics.
 - `gui/mass-remove`: mass removal/suspension tool for buildings and constructions
 - `reveal-hidden-sites`: exposes all undiscovered sites
-- `set-timeskip-duration`: changes the duration of the `Updating World` process preceding the start of a new game, enabling you to jump in earlier or later than usual
+- `set-timeskip-duration`: changes the duration of the "Updating World" process preceding the start of a new game, enabling you to jump in earlier or later than usual
 
 ## Fixes
 - `bodyswap`: stopped prior party members from tagging along after bodyswapping and reloading the map
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,7 +18,7 @@ that repo.
 - `devel/luacov`: generate code test coverage reports for script development. define the ``DFHACK_ENABLE_LUACOV=1`` environment variable to start gathering coverage metrics.
 - `gui/mass-remove`: mass removal/suspension tool for buildings and constructions
 - `reveal-hidden-sites`: exposes all undiscovered sites
-- `timeskip`: modifies the duration of the Upating World process, enabling you to jump into play earlier or later than usual
+- `set-timeskip-duration`: changes the duration of the `Updating World` process preceding the start of a new game, enabling you to jump in earlier or later than usual
 
 ## Fixes
 - `bodyswap`: stopped prior party members from tagging along after bodyswapping and reloading the map
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,6 +18,7 @@ that repo.
 - `devel/luacov`: generate code test coverage reports for script development. define the ``DFHACK_ENABLE_LUACOV=1`` environment variable to start gathering coverage metrics.
 - `gui/mass-remove`: mass removal/suspension tool for buildings and constructions
 - `reveal-hidden-sites`: exposes all undiscovered sites
+- `timeskip`: modifies the duration of the Upating World process, enabling you to jump into play earlier or later than usual
 
 ## Fixes
 - `bodyswap`: stopped prior party members from tagging along after bodyswapping and reloading the map
--- a/set-timeskip-duration.lua
+++ b/set-timeskip-duration.lua
@@ -1,18 +1,19 @@
--- modify the duration of the "Updating World" timeskip
+-- Change the duration of the "Updating World" timeskip preceding the start of a new game.
+-- Author: Atomic Chicken
 
 local usage = [====[
 
-timeskip
-========
+set-timeskip-duration
+=====================
 Starting a new fortress/adventurer session is preceded by
-an "Updating World" timeskip which is normally 2 weeks long.
+an "Updating World" process which is normally 2 weeks long.
 This script allows you to modify the duration of this timeskip,
 enabling you to jump into the game earlier or later than usual.
 
 You can use this at any point before the timeskip begins
 (for example, while still at the "Start Playing" menu).
 
-It is also possible to run the script during the timeskip,
+It is also possible to run the script while the world is updating,
 which can be useful if you decide to end the process earlier
 or later than initially planned.
 
@@ -54,12 +55,12 @@ Usage::
 
 Example::
 
-    timeskip -ticks 851249
+    set-timeskip-duration -ticks 851249
         Sets the end of the timeskip to
         2 years, 1 month, 9 days, 8 hours, 58 minutes, 48 seconds
         from the current date.
 
-    timeskip -years 2 -months 1 -days 9 -hours 8 -ticks 49
+    set-timeskip-duration -years 2 -months 1 -days 9 -hours 8 -ticks 49
         Does the same thing as the previous example
 
 ]====]
@@ -185,8 +186,8 @@ if args.hours then
   end
 end
 
-if ticks == 0 then
-  qerror("Duration not specified! Enter \"timeskip -help\" for more information.")
+if ticks == 0 and not (args.ticks and tonumber(args.ticks) == 0) then -- permit users to remove the timeskip altogether by setting it to 0 ticks
+  qerror("Duration not specified! Enter \"set-timeskip-duration -help\" for more information.")
 end
 
 ticks = math.floor(ticks) -- get rid of decimals the user may have inputted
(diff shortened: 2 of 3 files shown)


timeskip.lua
```diff
@@ -0,0 +1,146 @@
...
+            33600 ticks = 4 weeks = 1 month
+            403200 ticks = 12 months = 1 year
+
+    -clear
+        If the timeskip has not yet begun, this resets its duration
+        to the default value.
+
+Example::
+
+    timeskip -ticks 851249
```
myk002, line 37: wouldn't it be more natural for the user to input the desired time skip in calendar time instead of having to convert to ticks? Perhaps an API more like [`dfhack.timeout()`](https://docs.dfhack.org/en/latest/docs/Lua%20API.html#core-interpreter-context) would be better, like:

`timeskip 1 minute`
`timeskip 42 years`
`timeskip 2 years 1 month 9 days 8 hours 58 minutes 48 seconds`

generalized: `timeskip [<years> year[s]] [<months> month[s]] [<days> day[s]] [<minutes> minute[s]] [<seconds> second[s]]`

and then calculate the ticks (rounded up) from that?

AtomicChicken: I considered this whilst planning the script, but decided against it for the following reasons:

- Since 1 tick equates to 72 seconds of calendar time, allowing players to specify `seconds` and `minutes` would probably be misleading and confusing, as the majority of inputs would need to be heavily rounded to conform to this. As such, it seemed to me that specifying ticks directly is more intuitive and precise in this regard (in addition to being simpler to code).

- Calculating tick conversions is quite simple, especially since most users are likely to opt for simple durations along the lines of 5 years, unlike the convoluted value given in the example (which is really only there to better illustrate how ticks translate to calendar time). Of note, players familiar with modding should already have some understanding of this, as things like syndrome durations require time to be specified in ticks. With this in mind, I assumed that the suggested functionality probably wouldn't be worth the added complexity.

That said, I do understand that manual conversions can be a bit of a hassle. Perhaps as a middle ground we could make it possible to specify time in either `ticks` or the more convenient `years`, `months`, `days` and `hours`, omitting `seconds` and `minutes` under the expectation that players should use the `ticks` argument if they want further precision?

myk002: supporting `-years`, `-months`, `-days`, and `-hours` in addition to `-ticks` sounds perfectly reasonable to me.

lethosor:
This is great to have, thanks! Having encountered a similar script before, [updatefinish.lua](https://github.com/maxthyme/scripts/blob/acb44912a1ddf92aafb4133ed4dda425ab4d365f/updatefinish.lua), the name seemed different enough that I wasn't quite sure if it served the same purpose. Given that it appears to, I have a couple naming-related thoughts:
- "time" could also imply fortress mode time, as in [timestream.lua](https://github.com/DFHack/scripts/blob/master/timestream.lua). Granted, "time" also applies to the world overall, so I could see an argument for _either_ timestream or timeskip being confused with the other.
- I believe the game calls the process in question "Updating world", and the viewscreen is `viewscreen_update_regionst`, so perhaps the word "update" could be used in the script's name? However, I'm not sure that any of `update-skip`, `skip-update`, `update-time`, or `time-update` are particularly clear. The first two could easily refer to other kinds of updates, and the last two could easily refer to changing the in-game time in fortress/adventure mode.

I don't have a clear suggestion in mind, so I'm completely fine with keeping the `timeskip` name, but I also wanted to see if you had thoughts on this.

AtomicChicken:
I can see how `timeskip` is not the best name for this, not only because of its similarity to `timestream`, but also because it gives the false impression that the script is exclusively meant to hasten the 'Updating World' process, when in fact it can also be used to drastically prolong it (I chose the name with reference to 'time skip', the narrative trope of events rapidly unfolding off-screen, to imply relation to 'Updating World', rather than to explicitly suggest skipping time). `update-skip` and `skip-update` are similarly misleading so should probably be avoided. `update-time` and `time-update` are less problematic, but also quite vague as you say. Something like `update-world-duration` or `world-update-duration` is probably the most comprehensible choice (though admittedly a little bland). Thoughts?

lethosor:
Another option: `skip-world-update`, or more accurately, `adjust-world-update` (or `change-world-update`, etc.). Those are also rather long, but I haven't come up with one that's more accurate _and_ fewer than three words yet (e.g. `set-world-update-time` could perhaps be shortened to `set-update-time`). Naming things is hard. I don't think `timeskip` is a bad name either. If you're defining a "timeskip" in [the one-line script description](https://github.com/DFHack/scripts/pull/283/files#diff-bd2c0c075089386e0ae2f577830a6ffea34e73793e394e4b411d66d07142cb40R1) as the time taken for the world to update, then I think `timeskip` makes sense regardless of whether users extend or reduce the time. However, the name would just consist of a noun in that case, which doesn't always make it as clear _what_ the script does (and I think that's why I parsed `skip` as a verb earlier).

set-timeskip-duration.lua
```diff
@@ -0,0 +1,213 @@
...
+local scr = dfhack.gui.getCurViewscreen()
+if scr._type == df.viewscreen_update_regionst then
+  setTargetDate(scr, ticks)
+else
+  dfhack.onStateChange.TimeskipMonitor = function(event)
+    if event == SC_VIEWSCREEN_CHANGED then
+      local scr = dfhack.gui.getCurViewscreen()
+      if scr._type == df.viewscreen_update_regionst then
+        setTargetDate(scr, ticks)
+        dfhack.onStateChange.TimeskipMonitor = nil
```
lethosor, line 206: I noticed that this logic is in place to make the script work at most once per invocation - in my case, it caught me off guard because I had loaded one world, then realized it was the wrong world and decided to load a different world instead, and wound up advancing the second world 14 days instead of the time I expected.

A couple suggestions I thought of - what do you think is best here?

1. Document this behavior (I don't see it documented above, unless I missed something)
2. Add a print statement here, similar to a couple lines below, to indicate that the handler was removed
3. Add a `-persist` argument to make settings persist until explicitly cleared with `-clear`
4. Make settings persist by default until explicitly cleared with `-clear`

Some could be combined (e.g. 1 & 2, 1 & 3), and some are mutually exclusive (3 & 4). We probably can't satisfy everyone, but I just think the current behavior could be a bit more clear.

set-timeskip-duration.lua
```diff
@@ -0,0 +1,213 @@
...
+
+if args.hours then
+  if tonumber(args.hours) and tonumber(args.hours) >= 0 then
+    ticks = ticks + tonumber(args.hours)*50
+  else
+    qerror("Invalid -hours duration: " .. tostring(args.hours))
+  end
+end
+
+if ticks == 0 and not (args.ticks and tonumber(args.ticks) == 0) then -- permit users to remove the timeskip altogether by setting it to 0 ticks
```
lethosor, line 189: `-days 0` appears to trigger the error message (which I tried to use to end world advancement after having passed `-days 8`).

I think this could be simplified by initializing `ticks` to nil earlier (e.g. with just `local ticks`), then reducing this check to just `if ticks == nil`. That way, any valid argument that assigns a number to ticks would avoid triggering an error, and would allow ticks to be 0.

lethosor (commented):
I managed to test this - works as advertised! The name sounds good to me (it's long, but there's always the `alias` command for anyone who wants to abbreviate it). Noticed a couple minor things during testing - see comments.

changelog.txt
```diff
@@ -18,6 +18,7 @@ that repo.
 - `devel/luacov`: generate code test coverage reports for script development. define the ``DFHACK_ENABLE_LUACOV=1`` environment variable to start gathering coverage metrics.
 - `gui/mass-remove`: mass removal/suspension tool for buildings and constructions
 - `reveal-hidden-sites`: exposes all undiscovered sites
+- `set-timeskip-duration`: changes the duration of the `Updating World` process preceding the start of a new game, enabling you to jump in earlier or later than usual
```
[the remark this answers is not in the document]
AtomicChicken: Got it, thanks. I'm not familiar with RST and the docs build error wasn't particularly helpful at identifying the cause of the problem. Will edit accordingly.

set-timeskip-duration.lua
AtomicChicken: Option (4) sounds best to me, especially since we've already got a `-clear` option in place.

set-timeskip-duration.lua
AtomicChicken: This is another good suggestion, will implement.

set-timeskip-duration.lua
```diff
@@ -0,0 +1,214 @@
...
+(for example, while still at the "Start Playing" menu).
+
+It is also possible to run the script while the world is updating,
+which can be useful if you decide to end the process earlier
+or later than initially planned.
+
+Note that the change in timeskip duration will persist until either
+    the game is closed,
+    the "-clear" argument is used (see below),
+    or it is overwritten by setting a new duration.
```
lethosor, line 23: Just so you know, I'm adding bullets to this list. The indentation gets rendered in the HTML version, but single line breaks don't, so these lines are all getting combined into a single paragraph.

image

(Checking the output of the docs build through CI in this repo is a bit harder than in the main DFHack repo, unfortunately.)

lethosor (approved):
Looks great, thanks for addressing those things!

changelog.txt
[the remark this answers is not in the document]
lethosor: Thanks - the changelog is particularly confusing because it gets included in both `news.rst` and `news-dev.rst`, and due to being included, line numbers don't match up either.

PR #284, opened 2021-05-21
[quickfort] support --cursor option

myk002:
and alphabetize options in help text and in the code

inspired by the new `blueprint` `--cursor` option. it will also be useful for functional testing so we don't have to muck around with sidebar modes or the game cursor in the test code.

note that it's fine if the cursor is off the map. after cropping, there may still be in-bounds tiles

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -34,6 +34,7 @@ that repo.
 
 ## Misc Improvements
 - `devel/annc-monitor`: added ``report enable|disable`` subcommand to filter combat reports
+- `quickfort`: now supports the ``--cursor`` option for running a blueprint at specific coordinates instead of starting at the game cursor position
 - `quickfort`: now prints more helpful error messages for invalid modeline markers
 - `quickfort`: added support for extra space characters in blueprints
 - `quickfort`: now prints a warning when an invalid alias is encountered instead of silently ignoring it
--- a/internal/quickfort/command.lua
+++ b/internal/quickfort/command.lua
@@ -75,18 +75,30 @@ function finish_command(ctx, section_name, quiet)
     end
 end
 
+local function parse_cursor(arg)
+    local _, _, x, y, z = arg:find('^(%d+),(%d+),(%d+)$')
+    if not x then
+        qerror(('invalid argument for --cursor option: "%s"; expected format' ..
+                ' is "<x>,<y>,<z>", for example: "30,60,150"'):format(arg))
+    end
+    return {x=tonumber(x), y=tonumber(y), z=tonumber(z)}
+end
+
 function do_command(args)
     local command = args.action
     if not command or not command_switch[command] then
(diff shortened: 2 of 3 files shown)


lethosor (approved):
Looks good!

Interestingly, when placing buildings, the `resume` plugin only refreshes its list of suspended buildings when the sidebar mode changes, so planned buildings won't have a green "X" over them (when using that plugin) until the sidebar mode changes. Seems out of scope for quickfort to fix to me, but is there some "update buildings" flag that quickfort sets after it has placed buildings? Maybe `resume` could check that if there is.

myk002:
quickfort does call `dfhack.job.checkBuildingsNow()` after applying a `#build` blueprint. That sets `df::global::process_jobs` to `true`. `resume` could potentially check that flag.

myk002:
ok, after looking into `resume`, I came up with a few options for fixing the interaction:

1. if `df::global::process_jobs` flag is set, rescan for suspended buildings on every frame. this would fix the problem with very little code, but would increase the CPU usage when render is active since it would essentially remove its cache. we can't reset the `process_jobs` flag after seeing that it is true since we don't own the flag, and we can't just mark that we've seen a true value with our own flag since `quickfort` might be run multiple times and incrementally add more buildings to scan.
2. register a callback with the `Job` module that will be called when `Job::checkBuildingsNow()` is called. the callback would only be registered while `resume` is rendering
3. increment a counter whenever `Job::checkBuildingsNow()` is called. add a method to the `Job` module to retrieve the counter. `resume` can poll the counter before it renders on each frame. when the counter value changes, invalidate the cache.
4. add code to `quickfort` to call the `resume` plugin directly to invalidate its cache

(1) would certainly be easiest, but we'd have to test to make sure that the interface is still responsive with O(1M) buildings to scan.
(2) would be a solid, generalized solution, but introduces a new level of complexity. In particular, we'd have to be careful of plugins leaving stale callbacks registered when the plugin is reloaded.
(3) solves the stale callback problem of (2) at the cost of (very slightly) higher CPU consumption and the addition of a single state variable
(4) is a hard-coded solution to a general problem, and I consider it a hack, but it would fix this particular case, and this might be the only case we have. any other code path that calls `Job::checkBuildingsNow()` would not be covered (e.g. anything that flows through `linkBuilding` in `Buildings.cpp`, but I don't know if that matters.

so at the moment I'm leaning most towards options (1) or (3), but I could be swayed towards any of the options.

lethosor:
`process_jobs` sounds exactly like what I was thinking of. I didn't get a chance to reply earlier - I was thinking of something similar to (1), but only checking for "rising edges" (i.e. re-render in plugin_onupdate if process_jobs is true _and_ it was false on the last check).

lethosor:
At any rate, I think this PR itself is good to go. I can take a look at `resume` tomorrow unless you beat me to it :)

PR #285, opened 2021-05-21
Created clear-webs.lua

AtomicChicken:
Following a [request on the forums](http://www.bay12forums.com/smf/index.php?topic=178462.0), here's a simple script for clearing the map of webs, similar to `clear-smoke.lua`. I also made it reset the 'webbed' counter on units, for players looking to free their dwarves from giant cave spiders and the like.

### DIFF
--- a/clear-webs.lua
+++ b/clear-webs.lua
@@ -1,0 +1,32 @@
+-- Removes webs and frees webbed units.
+-- author: Atomic Chicken
+
+--[====[
+
+clear-webs
+==========
+This script removes all webs that are currently on the map,
+and also frees any creatures who have been caught in one.
+
+Note that it does not affect sprayed webs until
+they settle on the ground.
+
+Usable in both fortress and adventurer mode.
+
+]====]
+
+local webCount = 0
+for i = #df.global.world.items.other.ANY_WEBS-1, 0, -1 do
+  dfhack.items.remove(df.global.world.items.other.ANY_WEBS[i])
+  count = count + 1
+end
+
+for _, unit in ipairs(df.global.world.units.all) do
+  unit.counters.webbed = 0
+end
+
+if count == 0 then
+  print("No webs detected!")
+else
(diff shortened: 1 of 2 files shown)


clear-webs.lua
```diff
@@ -0,0 +1,32 @@
...
+they settle on the ground.
+
+Usable in both fortress and adventurer mode.
+
+]====]
+
+local webCount = 0
+for i = #df.global.world.items.other.ANY_WEBS-1, 0, -1 do
+  dfhack.items.remove(df.global.world.items.other.ANY_WEBS[i])
+  count = count + 1
```
myk002, line 21: are all instances of `count` supposed to be `webCount`?

clear-webs.lua
```diff
@@ -0,0 +1,32 @@
...
+end
+
+for _, unit in ipairs(df.global.world.units.all) do
+  unit.counters.webbed = 0
```
myk002, line 25: would be nice to detect how many units were freed so the player could tell if it worked

myk002: I also suspect that some players would just want this functionality, not the removal of all webs across the map. perhaps a flag to isolate the behaviors? I'm not sure what the defaults should be, though

lethosor: I would expect that just clearing this flag would result in the unit getting stuck again when they attempt to move, but I could be wrong.

clear-webs.lua
AtomicChicken: Clearly a typo inserted while I was wondering whether or not to add a unitCount. No idea how it slipped my notice.

clear-webs.lua
AtomicChicken: Simple enough, will do. 

When a non-WEBIMMUNE unit steps on a thick web, the webbed status is set and the web itself is destroyed. Freeing the unit hence does not necessitate the clearing of any webs, though they will of course remain susceptible to being caught in any surrounding webs later on.

lethosor (approved):

PR #288, opened 2021-05-25
[gui/blueprint] report created filenames

myk002:
depends on DFHack/dfhack#1858

### DIFF
--- a/test/gui/blueprint.lua
+++ b/test/gui/blueprint.lua
@@ -107,12 +107,6 @@ function test.cancel_selection()
         end)
 end
 
--- ensure we pause long enough to render a new frame
-local frames_to_wait = math.ceil(df.global.enabler.fps_per_gfps) + 1
-local function force_render()
-    delay(frames_to_wait)
-end
-
 local screen_width, screen_height = dfhack.screen.getWindowSize()
 local SPACE_ASCII = (' '):byte()
 local function get_screen_word(screen_pos)
@@ -127,7 +121,7 @@ end
 
 function test.render_labels()
     local view = load_ui()
-    force_render()
+    view:onRender()
     local action_label = view.subviews[3].subviews[1]
     local action_word_pos = {x=action_label.frame_body.x1+11,
                              y=action_label.frame_body.y1}
@@ -138,11 +132,11 @@ function test.render_labels()
     expect.eq('Back', get_screen_word(cancel_word_pos))
     guidm.setCursorPos({x=10, y=20, z=30})
     send_keys('SELECT')
-    force_render()
+    view:onRender()
     expect.eq('second', get_screen_word(action_word_pos))
     expect.eq('Cancel', get_screen_word(cancel_word_pos))
     send_keys('LEAVESCREEN') -- cancel selection
-    force_render()
+    view:onRender()
     expect.eq('first', get_screen_word(action_word_pos))
     expect.eq('Back', get_screen_word(cancel_word_pos))
     send_keys('LEAVESCREEN') -- leave UI
@@ -153,7 +147,7 @@ local X_ASCII = ('X'):byte()
 local function check_overlay(view, mark, cursor)
     guidm.setCursorPos(cursor)
     dfhack.gui.revealInDwarfmodeMap(cursor, true)
-    force_render()
+    view:onRender()
     local vp = view:getViewport()
     local z = cursor.z
     local upper_left = {x=math.min(mark.x, cursor.x),
--- a/test/gui/blueprint.lua
+++ b/test/gui/blueprint.lua
@@ -3,6 +3,7 @@ config = {
 }
 
 local b = reqscript('gui/blueprint')
+local blueprint = require('plugins.blueprint')
 local gui = require('gui')
 local guidm = require('gui.dwarfmode')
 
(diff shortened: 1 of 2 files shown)


test/gui/blueprint.lua
```diff
@@ -15,12 +16,18 @@ function test.fail_if_no_map_loaded()
         end)
 end
 
+-- ensure we pause long enough to render
+local frames_to_wait = math.ceil(df.global.enabler.fps_per_gfps)
+local function force_render()
```
lethosor, line 21: Is this intended to wait for screens to appear/disappear? If so, I would generally recommend checking explicitly for what you want (e.g. looping until `dfhack.gui.getCurViewscreen(false)` changes or waiting for an `SC_VIEWSCREEN_CHANGED` event).

That said, I think this should be pretty reliable - `fps_per_gfps` seems to be calculated once, and from what I remember about libgraphics, graphical frames are guaranteed to be synchronized with simulation frames at this frequency. So this should work for viewscreen changes, at least - still curious if you're using it for anything else.

Maybe we could use a "wait for screen" utility function or something to help with tests like these longer-term.

myk002: when waiting just one frame with `delay()`, I discovered that `onRenderBody()` was not being run reliably in the unit tests. I wanted to make sure layout and render code was always run so tests wouldn't be flaky. there is no screen transition (in general) that I'm waiting for. Just rendering.

lethosor: Does onRenderBody need to run for these tests? Generally, rendering should just read state, not change it - mutations are best left to feed()/logic(), since that allows for things like feeding keys that trigger multiple state changes between renders. There are definitely scripts that break this rule (well, recommendation), but it doesn't look to me like `gui/blueprint.lua` is one of them.

I asked about screens because I noticed several `send_keys()` calls interacting with dialogs, which call `delay()`, and likely need to wait for the dialog to open/close (since it's a separate screen). I haven't verified locally yet, though.

One other reason I can think of for waiting for rendering is if a test is actually reading from the screen to make sure certain things are rendered, but I don't see that here either.

Avoiding flakiness in tests is certainly good. My goal here is just to understand what's causing the flakiness (for instance, screens taking a while to show up, or mutations occurring in `render()` that get skipped sometimes). I've seen cases in other projects where `sleep(0.2)` is introduced to make tests pass more often, but when they start failing again due to the underlying slowdown increasing, the delay gets increased to `sleep(10)`, and that just slows down all cases even if significant delays only occur occasionally. That's not as relevant here, given that we do have strong guarantees about exactly how often `render()` gets called, and I think the cumulative extra delay here is so small that it doesn't really matter. I'd still like to poke at it a bit, personally, just to see if I can identify a source of some flakiness and come up with a way for tests to deal with it (since I anticipate more tests like these for other tools).

Speaking of that, tests for GUI tools have been pretty much nonexistent until now, so thanks for your work on these!

myk002: It's true that `onRenderBody` does not mutate state and will not affect the outcome of the unit tests. I still wanted to make sure the code was always covered, though, to ensure that it won't error out when it *is* run. I've learned (painfully) that in non-compiled languages, endless typos hide in uncovered code. I could (and perhaps should) write a test to ensure that the overlay is being rendered properly. That at least would make it more understandable why I'm forcing a render.

But your point about `force_render()` also being a mechanism for allowing screen state transitions is very valid. If I don't call `force_render()` (actually, I think just `delay()` was sufficient -- I had the test working with that before I noticed the inconsistency in the coverage metrics and created `force_render`) after the ENTER keystrokes in the happy path test, the gui does not change state in time and the test does not pass. Note that the focus string does not change -- `gui/blueprint` is still active -- it's just needs to be in a different state.

I was very happy to discover `fps_per_gfps`. I certainly didn't want to set a precedent for "sleep creep".

lethosor: > after the ENTER keystrokes in the happy path test, the gui does not change state in time and the test does not pass

Yeah, this is the bit I'm confused about - in cases where `BlueprintUI` is the only screen present, any state updates made by `onInput()` should be immediately reflected by the time `onInput()` returns. Since `onRenderBody()` doesn't change state, there should be no need to wait for it before calling `onInput()` to change state again. (There isn't really anything special about calling `onInput()` through `simulateInput()` vs. calling methods like `commit()` or `on_mark()` directly.)

In the "happy path" test, one exception I see is [here](https://github.com/DFHack/scripts/blob/fdb64f58064d2cd9ecd0ed00c0b3c562d1cfdcd6/test/gui/blueprint.lua#L63), which relies on another screen (the dialog) showing up in the screen stack. If it hasn't, `SELECT` might be sent to the BlueprintUI screen, likely causing the test to hang. Maybe this is the source of some confusion - I'm using "state" here to refer to state local to a specific viewscreen, not the state of all screens created by the script.

In any case, I'm being pickier about this than I need to be. I'll approve based on functionality and maybe revisit the test flakiness aspect later.

myk002: Picky is fine. I'd like to make sure I'm submitting good code too.

I'll look into this more, but I can say that the success message dialog was not the issue. I discovered the need for `delay()` post-`enter` before the success dialog existed.

myk002: and this is why I'm appreciative that you ask these kinds of questions. I could not reproduce the issues with not processing keystrokes when I took away the calls to `delay()`. I only need them, as you noted, for screen transitions. I don't have snapshots of my code from back then, so I can only assume there was some other bug in my code that was causing the problems that I have since fixed.

I added in unit tests for the rendering code, and discovered that even waiting for `fps_per_gfps` resulted in flaky tests. using `fps_per_gfps + 1` worked consistently (for me), but this is sleep creep territory and made me nervous.

I started investigating what it would take to add a `gframe` option to `dfhack.timeout`, but in the course of my code spelunking, I discovered that I could just call `View:onRender()` directly and avoid sleeping entirely!

lethosor: Oops, I forgot to reply here! I tested locally and noticed the same thing (tests passed with just two delays, after showing/hiding the dialog screen, but failed if I removed either delay). And calling render() directly is what I'd recommend as well for testing that behavior - it would ensure that nothing else overwrites screen data before the test reads from it.

lethosor (approved):
Looks good!

PR #289, opened 2021-06-07
[teleport] use new common functionality in the Units module

myk002:
requires DFHack/dfhack#1865

### DIFF
--- a/teleport.lua
+++ b/teleport.lua
@@ -29,23 +29,7 @@ Examples:
 ]====]
 
 function teleport(unit,pos)
- local oldOccupancy = dfhack.maps.getTileBlock(unit.pos).occupancy[unit.pos.x%16][unit.pos.y%16]
- local newOccupancy = dfhack.maps.getTileBlock(pos).occupancy[tonumber(pos.x)%16][tonumber(pos.y)%16]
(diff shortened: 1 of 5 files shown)


teleport.lua
```diff
@@ -29,23 +29,7 @@ Examples:
 ]====]
 
 function teleport(unit,pos)
```
lethosor, line 31: Looks like there are three scripts that import and call this script's `teleport()`: `deep-embark`, `flashstep`, and `gui/teleport`. Should we migrate those as well? I'm fine putting those in another PR if you prefer.

myk002: you mean have them call `dfhack.units.teleport()` directly instead of bouncing off `teleport.teleport()`? Yeah I could totally do that here. One sec.

myk002: Note that I'm still leaving the `teleport.teleport` function exported since out of tree scripts may still be calling it.

myk002: done

lethosor: Yep, 97da3a82adb45025a9d2cac40bbb143671befd99 is exactly what I was thinking of. Thanks!

lethosor (approved):
Thanks!

PR #290, opened 2021-06-09
[quickfort] convert reactions from a global to a singleton

myk002:
essentially so the file itself can be loaded without automatically calling stockflow methods. This lets the file be included in tests where the test harness is parsing test files before a fortress is loaded.

### DIFF
--- a/internal/quickfort/orders.lua
+++ b/internal/quickfort/orders.lua
@@ -130,12 +130,18 @@ function create_orders(ctx)
     end
 end
 
--- ensure we don't reinit this; it contains allocated memory
-reactions = reactions or stockflow.collect_reactions()
+-- we only need to init this once, even if a new save is loaded, since we only
+-- care about the built-in reactions, not the mod-added ones.
+-- note that we also shouldn't reinit this because it contains allocated memory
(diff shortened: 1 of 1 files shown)


internal/quickfort/orders.lua
```diff
@@ -130,12 +130,18 @@ function create_orders(ctx)
     end
 end
 
--- ensure we don't reinit this; it contains allocated memory
-reactions = reactions or stockflow.collect_reactions()
+-- we only need to init this once, even if a new save is loaded, since we only
+-- care about the built-in reactions, not the mod-added ones.
+-- note that we also shouldn't reinit this because it contains allocated memory
+local function get_reactions()
+    g_reactions = g_reactions or stockflow.collect_reactions()
```
lethosor, line 137: I may have already asked about this already, so apologies if this is a repeat question: should this be calculated on a per-save basis? Mods can add reactions, for instance.

+1 to this change in any case. `quickfort list` works on the title screen now :)

lethosor: Indeed, I missed that comment. Makes sense!

PR #291, opened 2021-06-11
Reimplements devel/query.lua

cppcooper:
New implementation should be far easier to maintain and understand (as a user mostly).
Features have been merged, expanded, etc.

help message:

edit: (updated documentation)
```
devel/query
===========
Query is a script useful for finding and reading values of data structure
fields. Purposes will likely be exclusive to writing lua script code,
possibly C++.

This script takes your data selection eg.{table,unit,item,tile} then recursively
iterates through it outputting names and values of what it finds.

As it iterates you can have it do other things, like search for a specific
structure pattern (see lua patterns) or set the value of fields matching the
selection and any search pattern specified.

If the script is taking too long to finish, or if it can't finish you should run
``dfhack-run kill-lua`` from another terminal.

Examples::

  devel/query -unit -getfield id
  devel/query -unit -search STRENGTH
  devel/query -unit -search physical_attrs -maxdepth 2
  devel/query -tile -search dig
  devel/query -tile -search "occup.*carv"
  devel/query -table df -maxdepth 2
  devel/query -table df -maxdepth 2 -excludekind s -excludetype fsu -oneline
  devel/query -table df.profession -findvalue FISH
  devel/query -table df.global.ui.main -maxdepth 0
  devel/query -table df.global.ui.main -maxdepth 0 -oneline
  devel/query -table df.global.ui.main -maxdepth 0 -1

**Selection options:**

``-unit``:              Selects the highlighted unit

``-item``:              Selects the highlighted item.

``-plant``:             Selects the highlighted plant.

``-building``:          Selects the highlighted building.

``-job``:               Selects the highlighted job.

``-workshopjob``:       Selects the highlighted job in a workshop.

``-tile``:              Selects the highlighted tile's block, and then
                        uses the tile's local position to index the data.

``-block``:              Selects the highlighted tile's block.

``-table <value>``:     Selects the specified table (ie. 'value').

                        Must use dot notation to denote sub-tables.
                        (eg. ``-table df.global.world``)

``-getfield <value>``:  Gets the specified field from the selection.

                        Must use in conjunction with one of the above selection
                        options. Must use dot notation to denote sub-fields.

**Query options:**

``-search <value>``:       Searches the selection for field names with
                           substrings matching the specified value.

``-findvalue <value>``:    Searches the selection for field values matching the
                           specified value.

``-maxdepth <value>``:     Limits the field recursion depth (default: 7)

``-maxlength <value>``:    Limits the table sizes that will be walked
                           (default: 257)

``-excludetype [a|bfnstu0]``:  Excludes data types: All | Boolean, Function,
                               Number, String, Table, Userdata, nil

``-excludekind [a|bces]``:     Excludes data types: All | Bit-fields,
                               Class-type, Enum-type, Struct-type

``-noblacklist``:   Disables blacklist filtering.

``-dumb``:          Disables intelligent checking for recursive data
                    structures(loops) and increases the -maxdepth to 25 if a
                    value is not already present

**Command options:**

``-showpaths``:        Displays the full path of a field instead of indenting.

``-setvalue <value>``: Attempts to set the values of any printed fields.
                       Supported types: boolean,

``-oneline``:          Reduces output to one line, except with ``-debugdata``

``-1``:                Reduces output to one line, except with ``-debugdata``

``-disableprint``:     Disables printing. Might be useful if you are debugging
                       this script. Or to see if a query will crash (faster) but
                       not sure what else you could use it for.
                       
``-debug <value>``:    Enables debug log lines equal to or less than the value
                       provided.

``-debugdata``:        Enables debugging data. Prints type information under
                       each field.

``-help``:             Prints this help information.
```


devel/query.lua
```diff
@@ -10,92 +10,61 @@ local validArgs = utils.invert({
...
- - Is printing keys allowed?
+Purposes will likely be exclusive to writing lua script code, possibly C++.
+
+This script takes your data selection eg.{table,unit,item,tile}
+then recursively iterates through it outputting names and values of what it finds.
+
+As it iterates you can have it do other things, like search for a specific structure pattern (see lua patterns)
+or set the value of fields matching the selection and any search pattern specified.
+
+If the script is taking too long to finish, or if it can't finish you should run ``dfhack-run kill-lua`` from a terminal.
```
myk002, line 51: Since you actually print this out to the console, be mindful of the 80 column standard for help text

devel/query.lua
```diff
@@ -107,148 +76,326 @@ or specifically what key to print inside a unit.
...
-``-disableprint``: Disables printing fields and keys. Might be useful if you are
-debugging this script. Or to see if a query will crash (faster) but not sure
-what else you could use it for.
+]====]
 
-``-help``: Prints this help information.
 
-]====]
 
 --[[ Test cases:
```
myk002, line 123: note that you can make actual automated tests out of these. You can add tests to `test/devel/query.lua` and run them with the DFHack `test` command. e.g. `test -d/path/to/your/scripts/repo -tquery`

devel/query.lua
```diff
@@ -316,440 +463,149 @@ function parseKeyString(t,str)
...
-    else
-        if args.query then
-            debugf(0,"regular query")
-            Query(selection, args.query, info)
-        else
-            debugf(0,"empty query")
-            Query(selection, '', info)
+function debugf(level,...)
+    if args.debug and level <= tonumber(args.debug) then
+        str=string.format(" #  %s",select(1, ...))
```
myk002, line 604: did you mean `local str`?

devel/query.lua
cppcooper: Alright, I'll do that. But the standard should really be increased to 120 or more. Who in 2021 keeps their consoles at 80 wide.

devel/query.lua
cppcooper: good to know

devel/query.lua
cppcooper: For that matter, the output of this command will not be pretty if the console is only 80 wide. So I feel as though maybe I shouldn't even commit changes to the help message.

devel/query.lua
```diff
@@ -325,6 +329,23 @@ function runOnce(caller)
...
 
+function toType(str)
+    if str ~= nil then
+        if str == "true" then
+            return true
+        elseif str == "false" then
+            return false
+        elseif tonumber(str) then
+            return tonumber(str)
+        elseif string.find(str, "nil") then
```
lethosor, line 340: Is there a reason you want to accept any string containing "nil" here? i.e. why not `elseif str == "nil"` like the cases for true/false above?

Part of me would also like to find a way to allow searching for strings that happen to also be valid booleans/numbers (e.g. the string "true"). I'm not sure if that's really important, but one way could be to allow an escape character (like `\`) at the start of `str`, and remove a single occurrence of it prior to calling `tostring()` below.

(Incidentally, the `tostring` call is redundant if `str` is already a string.)

devel/query.lua
```diff
@@ -620,6 +631,22 @@ function printField(path, field, value)
     end
 end
 
+function print_linked_list(name, next)
+    if string.find(name, "next") then
+        local JobListItem=next --danke F.F.
+        while(JobListItem.next) do
```
lethosor, line 637: You might find `for _, item in utils.listpairs()` useful here.

devel/query.lua
cppcooper: Oh yea, wups. I didn't even think about it.

Not sure I understand your idea. But if it puts the code for conversion inside the utils codebase then I think it's worth exploring; presumably allowing for args={*} to already be filled with correctly typed table entries.

About the str specifically I threw it through tostring just in case because I mostly just wasn't comfortable assuming it was in fact a string I like to think less and just be careful with dynamically typed languages.

devel/query.lua
cppcooper: Certainly. I've taken a crack at revising the `foreach(table, name, callback(field, value))` function to include it as a third pairs option.

```
elseif string.find(name,".*list") and table["next"] then
```
I use this check to decide on using that pairs function, but if there is a more precise and elegant way for this too I would gladly use it.

The formatting I've so far decided to use for this particular type of data is borrowing repetitions from regex
```lua
local m = tostring(field):gsub("<.*: ",""):gsub(">.*",""):gsub("0x", "0x "):gsub("%x%x%x%x%x%x","%1 ",1)
local s = string.format("next{%d}.item [next: %s]", index, m)
-- which outputs as lines like this
| next{34}.item [next: 0x7fff6f 3b5440]: ~  <job: 0x7fff6d86ca80> [struct]
```
All the options I thought made sense are here
```
string.format("*{%d}next.item", index), "[next.item]{%d}")
string.format("*[%d]next.item", index), "**next.item"
string.format("next.item{%d}", index), "*{r}.next.item"
```

devel/query.lua
```diff
@@ -186,27 +194,44 @@ function query(table, name, search_term, path)
...
+                return
+            end
             callback(field, value)
+            index = index + 1
         end
+        index = 0
         for field, value in utils.listpairs(table) do
             local m = tostring(field):gsub("<.*: ",""):gsub(">.*",""):gsub("%x%x%x%x%x%x","%1 ",1)
-            local s = string.format("next{%d}.item [next: %s]", index, m)
+            local s = string.format("next{%d}->item", index)
```
cppcooper, line 221: I changed the format again. Good, bad..? I don't know.

cppcooper:
```
| occupancy[12][15]
| occupancy: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  <tile_occupancy: 0x7fff1625e100> [bitfield]
```
If anyone could figure out how to avoid this. That is, how to merge the two lines into one. I'd be very grateful.
I'm pretty sure the trick would be to simply pattern match the field (eg. "occupancy") with the [last] word from the path (eg. "occupancy[12][15]"). I couldn't easily figure out how to match that pattern though.

myk002:
> ```
> | occupancy[12][15]
> | occupancy: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  <tile_occupancy: 0x7fff1625e100> [bitfield]
> ```
> 
> If anyone could figure out how to avoid this. That is, how to merge the two lines into one. I'd be very grateful.
> I'm pretty sure the trick would be to simply pattern match the field (eg. "occupancy") with the [last] word from the path (eg. "occupancy[12][15]"). I couldn't easily figure out how to match that pattern though.

where is the current code for this? what part of the pattern are you having trouble matching?

cppcooper:
> > ```
> > | occupancy[12][15]
> > | occupancy: ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~ ~  <tile_occupancy: 0x7fff1625e100> [bitfield]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If anyone could figure out how to avoid this. That is, how to merge the two lines into one. I'd be very grateful.
> > I'm pretty sure the trick would be to simply pattern match the field (eg. "occupancy") with the [last] word from the path (eg. "occupancy[12][15]"). I couldn't easily figure out how to match that pattern though.
> 
> where is the current code for this? what part of the pattern are you having trouble matching?

https://github.com/DFHack/scripts/blob/d5e378335746819252195148bffd967b4690e1b6/devel/query.lua#L623-L625

https://github.com/DFHack/scripts/blob/d5e378335746819252195148bffd967b4690e1b6/devel/query.lua#L597-L613

Well the regex for the pattern would be `[:alpha:]+\[[\d]+\]\[[\d]+\]` but lua doesn't use regex.

myk002:
> Well the regex for the pattern would be `[:alpha:]+\[[\d]+\]\[[\d]+\]` but lua doesn't use regex.

Yeah, Lua patterns are a bit funky. You can still match the pattern, though: '%a+%[%d+%]%[%d+%]'

cppcooper:
> > Well the regex for the pattern would be `[:alpha:]+\[[\d]+\]\[[\d]+\]` but lua doesn't use regex.
> 
> Yeah, Lua patterns are a bit funky. You can still match the pattern, though: '%a+%[%d+%]%[%d+%]'
> 
> More info here: https://www.lua.org/pil/20.2.html

Thanks, that worked.

As far as I know, this is now complete. But if there are some data types it can't iterate/print let me know and I'll fix it.. or try to.

devel/query.lua
```diff
@@ -1,254 +1,507 @@
...
+            return true
+        elseif string.find(field,"saves") then
+            return true
+        elseif string.find(field,"movie") then
+            return true
+        elseif string.find(field,"font") then
+            return true
+        elseif string.find(field,"texpos") then
+            return true
+        end
```
lethosor, line 479: Could this function also take the type of `t` into account, assuming it's the table that has these fields?

Also, can the `string.find()` calls be changed to string equality checks? (Not sure which of these are exact field names and which are prefixes/suffixes; if any are prefixes/suffixes and not substrings that could occur anywhere, you could use `string.startswith()` and `string.endswith()` respectively)

lethosor (commented):
How would you like to summarize this in the changelog? Even better if you can update `changelog.txt`. I'm mainly asking because I see a lot of added/removed options and that seems like something that should be mentioned.

devel/query.lua
```diff
@@ -1,254 +1,507 @@
...
+
+This script takes your data selection eg.{table,unit,item,tile} then recursively
+iterates through it outputting names and values of what it finds.
+
+As it iterates you can have it do other things, like search for a specific
+structure pattern (see lua patterns) or set the value of fields matching the
+selection and any search pattern specified.
+
+If the script is taking too long to finish, or if it can't finish you should run
+``dfhack-run kill-lua`` from another terminal.
```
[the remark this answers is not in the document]
cppcooper: Oh, single? Wups. I must have rewrote the whole line from scratch and mixed that up.

devel/query.lua
```diff
@@ -1,254 +1,507 @@
...
-A few important ones:
-
- - Is the data structure capable of being iterated?
- - Is the data structure pointing to a parent data structure?
- - Is the current level of recursion too high, and do we need to unwind it first?
- - Is the number of entries too high (eg. 70,000 table entries that would be printed)?
- - Is the data going to be usefully readable?
- - Does the field or key match the field or key query or queries?
- - Is printing fields allowed?
- - Is printing keys allowed?
```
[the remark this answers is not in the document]
cppcooper: To be honest, I had written that because the script wasn't well organized. This section helped me rewrite it.

I'll re-add it though, with the necessary changes.

devel/query.lua
cppcooper: I don't think I follow. You want to blacklist an entire table? Not data fields? Cause I don't think any of the matching data are tables. Maybe you could elaborate?

devel/query.lua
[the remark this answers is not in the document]
cppcooper: Let me know if the new revised version is adequate

devel/query.lua
lethosor: No, for example, I want to *only* exclude fields with "font" in their name if they are actually the fields that we want to exclude. If `t` is the object containing `field`, then checking the type of `t` would be a way to ensure this.

lethosor: As for my second point about `string.find()`: If you only want to exclude fields named exactly "movie", it would be better to check for `field == "movie"` rather than `string.find(field, "movie")`. If this is possible to change, it would _limit_ the possible false-positives excluded by this function - it would only exclude fields named exactly "movie" in some other type, rather than any field containing "movie" in its name. Since this isn't a complete fix, I would still prefer to have this check the type of the containing object.

cppcooper: I tried to re-find the problem fields and could not.. at first. So I played around and realized that the max table length defaults were keeping the problem fields from flooding the console. So I'll just remove this feature altogether as it seems a blacklist is no longer needed..

I could convert it to an opt-in option which takes a list of names to exclude. What do you think?

lethosor: If it's no longer necessary, it's fine with me to remove it. I was actually recently thinking that a list of fields to _include_ would be useful, but I suppose the inverse would be useful as well.

cppcooper: Well I think it'd be too complicated to do with fields via the `getfield` option.. but that could work pretty easily with the search option. I'll get about implementing this stuff once it's not 30-45 Celsius around here. Barely have the energy to sit

lethosor: No worries. I can remove this, maybe touch up the docs, and merge if you're ok with it as-is.

cppcooper: Unless I missed something this should be pretty decent now. I may, or may not make a secondary help doc sans RST formatting. The blacklist option has been removed.

cppcooper:
Unless I missed something this should be pretty decent now. I may, or may not make a secondary help doc sans RST formatting. The blacklist option has been removed.

lethosor (approved):

devel/query.lua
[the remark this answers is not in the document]
lethosor: Still looks like it's using double backticks - I can fix this

devel/query.lua
```diff
@@ -1,254 +1,558 @@
...
+``-building``
+  Selects the highlighted building.
 
-``-table <value>``:     Selects the specified table (ie. 'value').
+``-job``
+  Selects the highlighted job.
 
-                        Must use dot notation to denote sub-tables.
-                        (eg. ``-table df.global.world``)
+``-script``
```
lethosor, line 131: Looks like this should be `-module`

lethosor: Also, this option _only_ works for scripts, not modules, so I think it should likely be renamed back to `-script`

lethosor:
@cppcooper FYI, I made some followup changes in 2ed13a78a62b949c5393fd709dbe25db978537ae...3dddfb866f948ccecdea4b6009e66f24b5ac9eb0. You'll want to pull before making any future changes.

Thanks!

devel/query.lua
cppcooper: Hmm I thought dorf_tables was a module.

lethosor: I guess I can see how it makes sense either way. However, I was reading "script/module" as "a script that can be included with reqscript() *or* a library that can be included with require()". The latter is not supported but is arguably also a module. (I would tend to define a module as a file that can be included. This includes some scripts, like dorf_tables, but not all scripts, and also includes other types of files.)

cppcooper: Honestly I'm a bit confused still cause of the ```--@ module = true``` at the top of the file. I guess it's only needed for allowing the file to be imported and doesn't really mean the file is a module.. but maybe somewhere between

lethosor: Sure, you could say that `--@ module=true` makes a script a module. What I am saying is that there are _other things_ that are also "modules", such as anything you include with `require()`. The flag in question only supports modules that are scripts, not all types of modules.

PR #292, opened 2021-06-13
[gui/blueprint] get into an appropriate mode on display

myk002:
depends on DFHack/dfhack#1884 and DFHack/dfhack#1887

and restore previous mode on dismiss

The mode switching logic is very similar to what is used in quickfort's `query.lua` and what was copied to the `mass-remove` script. Does it belong in the library? If so, where?

### DIFF
--- a/gui/blueprint.lua
+++ b/gui/blueprint.lua
@@ -112,6 +112,9 @@ function BlueprintUI:onAboutToShow()
     end
 
     self.saved_mode = df.global.ui.main.mode
+    if dfhack.gui.getCurFocus(true):find('^dfhack/') then
+        self.saved_mode = df.ui_sidebar_mode.Default
+    end
     switch_ui_sidebar_mode(df.ui_sidebar_mode.LookAround)
 end
 
--- a/gui/blueprint.lua
+++ b/gui/blueprint.lua
@@ -78,10 +78,48 @@ function BlueprintUI:init()
     }
 end
 
+-- the sidebar modes that we know how to get back to
+local basic_ui_sidebar_modes = {
+    [df.ui_sidebar_mode.Default]='',
+    [df.ui_sidebar_mode.QueryBuilding]='D_BUILDJOB',
+    [df.ui_sidebar_mode.LookAround]='D_LOOK',
+    [df.ui_sidebar_mode.BuildingItems]='D_BUILDITEM',
+    [df.ui_sidebar_mode.Stockpiles]='D_STOCKPILES',
+    [df.ui_sidebar_mode.Zones]='D_CIVZONE',
+}
+
+-- Send ESC keycodes until we get to dwarfmode/Default and then enter the
+-- specified sidebar mode. If we don't get to Default after 10 presses of ESC,
+-- error. The target sidebar mode must be a member of basic_ui_sidebar_modes.
+local function switch_ui_sidebar_mode(sidebar_mode)
+    for i=1,10 do
+        if df.global.ui.main.mode == df.ui_sidebar_mode.Default and
+                dfhack.gui.getCurFocus(true) == 'dwarfmode/Default' then
+            if sidebar_mode ~= df.ui_sidebar_mode.Default then
+                gui.simulateInput(dfhack.gui.getCurViewscreen(true),
+                                basic_ui_sidebar_modes[sidebar_mode])
+            end
+            return
+        end
+        gui.simulateInput(dfhack.gui.getCurViewscreen(true), 'LEAVESCREEN')
+    end
+    qerror('Unable to get into target sidebar mode from current UI viewscreen.')
+end
+
 function BlueprintUI:onAboutToShow()
     if not dfhack.isMapLoaded() then
         qerror('Please load a fortress map.')
     end
+
+    self.saved_mode = df.global.ui.main.mode
+    switch_ui_sidebar_mode(df.ui_sidebar_mode.LookAround)
+end
+
+function BlueprintUI:onDismiss()
+    if not basic_ui_sidebar_modes[self.saved_mode] then
+        self.saved_mode = df.ui_sidebar_mode.Default
+    end
+    switch_ui_sidebar_mode(self.saved_mode)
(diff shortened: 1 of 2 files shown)


lethosor:
Shared logic could go in `library/lua/gui/dwarfmode.lua`, or `modules/Gui.cpp` if we want it accessible to C++ as well (there isn't a fortress-mode-specific module in C++)

test/gui/blueprint.lua
```diff
@@ -202,6 +193,43 @@ end
 
 --auto enter and leave cursor-supporting mode
 
+function test.restore_mode()
+    df.global.ui.main.mode = df.ui_sidebar_mode.Stockpiles
```
lethosor, line 197: I know switching to some modes directly with this approach can cause crashes (e.g. `q` mode skips DF's logic to initialize `selected_building`, which could be garbage). Could this be replaced with `send_keys('D_STOCKPILES')`?

test/gui/blueprint.lua
```diff
@@ -202,6 +193,43 @@ end
...
+    load_ui()
+    expect.eq(df.ui_sidebar_mode.LookAround, df.global.ui.main.mode)
+    send_keys('LEAVESCREEN') -- cancel out of ui
+    expect.eq(df.ui_sidebar_mode.Stockpiles, df.global.ui.main.mode)
+    send_keys('LEAVESCREEN') -- get back to Default mode
+end
+
+function test.restore_default_on_unknown_mode()
+    df.global.ui.main.mode = df.ui_sidebar_mode.Burrows
```
lethosor, line 206: Similarly, `send_keys('D_BURROWS')`?

test/gui/blueprint.lua
```diff
@@ -58,7 +49,7 @@ function test.minimal_happy_path()
             expect.table_eq({'2', '3', '-4', 'blueprint', '--cursor=10,20,33'},
                             mock_run.call_args[1])
             send_keys('SELECT') -- dismiss the success messagebox
-            delay() -- allow the messagebox and view to disappear
+            while view:isActive() do delay() end -- wait for view to disappear
```
lethosor, line 52: This could potentially cause tests to hang indefinitely (disregarding CI timeouts, of course) if the screen being tested doesn't dismiss itself properly. Should we have a `delay_until()` helper that times out after some number of ticks?

test/gui/blueprint.lua
myk002: I hadn't done it that way since it presupposes that the mode is in `Default`, which could make the test flaky if whatever test runs before it doesn't clean up properly. This is a non-issue for the first test in a suite since the test harness ensures the mode is in `Default` for fortress-mode tests, but individual tests within a suite do not guarantee that we end up in the same mode we started with.

I also thought there wasn't a risk of crash here because of the modes I chose (stockpiles here and burrows below) were "non-dangerous" (I specifically stayed away from `q`) so I didn't bother to copy in (or call) the generic mode-switching code in `gui/blueprint`.

Your point is well taken, though, that forcing a mode switch has proven to be trouble and we shouldn't encourage it. I'll move the generic mode-switching code to `dwarfmode.lua` and modify these tests to call it.

test/gui/blueprint.lua
myk002: good idea. I'll spin a PR.

test/gui/blueprint.lua
myk002: done

test/gui/blueprint.lua
myk002: done

test/gui/blueprint.lua
myk002: done

gui/blueprint.lua
```diff
@@ -82,6 +82,19 @@ function BlueprintUI:onAboutToShow()
...
+        self.saved_mode = df.ui_sidebar_mode.Default
+    end
+    guidm.enterSidebarMode(df.ui_sidebar_mode.LookAround)
+end
+
+function BlueprintUI:onDismiss()
+    if not guidm.SIDEBAR_MODE_KEYS[self.saved_mode] then
+        self.saved_mode = df.ui_sidebar_mode.Default
+    end
+    guidm.enterSidebarMode(self.saved_mode)
```
lethosor, line 97: I can't seem to get this to work - pressing `v`, then running `gui/blueprint`, then pressing `Esc` at any stage always seems to bring me back to the default sidebar mode, rather than the `v` mode. I don't think this is super important, but wanted to confirm that I wasn't missing something obvious.

test/gui/blueprint.lua
```diff
@@ -202,6 +193,44 @@ end
...
+    guidm.enterSidebarMode(df.ui_sidebar_mode.Default)
+    send_keys('D_BURROWS')
+    load_ui()
+    expect.eq(df.ui_sidebar_mode.LookAround, df.global.ui.main.mode)
+    send_keys('LEAVESCREEN') -- cancel out of ui
+    expect.eq(df.ui_sidebar_mode.Default, df.global.ui.main.mode)
+end
+
+function test.fail_to_find_default_mode()
+    df.global.ui.main.mode = df.ui_sidebar_mode.Burrows
```
lethosor, line 215: One more possible case where D_BURROWS could be used (I can handle this)

lethosor (approved):
Picked the wrong option - works well!

test/gui/blueprint.lua
myk002: D'oh I missed this one. Thanks for fixing!

gui/blueprint.lua
myk002: Yeah, not all modes are handled. We can extend the map in enterSidebarMode over time.

lethosor: Oh, I thought that was one that was handled. Looks like it's not. Thanks!

myk002: it totally should have been handled. Added in DFHack/dfhack#1889

PR #297, opened 2021-06-17
[gui/blueprint] show selected area dimensions when moving cursor

myk002:
In addition to the dynamic status line that displays the selected area after the initial mark has been set, this commit introduces auto-resize functionality for the action panel. The ResizingPanel subclass will also be used for the upcoming Phases panel, which changes height based on whether phases should be autodetected or set manually.

The sequence I have to follow to get the layout set seems somewhat inefficient and I'm wondering if there is a better way to do it. I have to set the component widget positions after I know their heights, i.e. in `BlueprintUI:postUpdateLayout()`, but the widgets won't know their heights until `BlueprintUI:updateSubviewLayout()` is called, but the widgets *also* need to be re-laid out after their positions have been set so their subframes can be recomputed. So I end up having to rerun layout methods until properties are done shuffling. I've played around with the sequence, but I haven't found anything better than what I already have. In particular, I can't move any logic to a `preUpdateLayout()` method (which could otherwise remove the need to lay out subcomponents twice) since frames don't exist until subcomponents are done with their initial layout. Layout doesn't happen so often that I'm too worried about the inefficiency, though.

### DIFF
--- a/test/gui/blueprint.lua
+++ b/test/gui/blueprint.lua
@@ -304,10 +304,10 @@ end
 -- live status line showing the dimensions of the currently selected area
 function test.render_status_line()
     local view = load_ui()
+    view:updateLayout()
     local status_label = view.subviews.selected_area
     local status_text_pos = {x=status_label.frame_body.x1,
                              y=status_label.frame_body.y1}
-    view:updateLayout()
     view:onRender()
     expect.false_(status_label.visible)
     guidm.setCursorPos({x=10, y=20, z=30})
--- a/test/gui/blueprint.lua
+++ b/test/gui/blueprint.lua
@@ -116,6 +116,7 @@ end
 
 function test.render_labels()
     local view = load_ui()
+    view:updateLayout()
     view:onRender()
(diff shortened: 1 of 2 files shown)


myk002:
As @warmist suggested on discord, I refactored the custom layout logic into an `updateLayout()` override. This at least gets rid of the janky repeated `updateLayout() updateSubviewLayout()` call sequences.

test/gui/blueprint.lua
```diff
@@ -110,26 +110,29 @@ local function get_screen_word(screen_pos)
...
+local function get_cancel_word_pos(cancel_label)
+    return {x=cancel_label.frame_body.x1+5, y=cancel_label.frame_body.y1}
+end
+
 function test.render_labels()
     local view = load_ui()
+    view:updateLayout()
     view:onRender()
-    local action_label = view.subviews[3].subviews[1]
+    local action_label = view.subviews.action_label
```
lethosor, line 121: +1 to using view IDs

lethosor (approved):

PR #304, opened 2021-06-30
[gui/blueprint] support editable names for blueprint files

myk002:
Add an edit widget so generated blueprint names are customizable. A warning is displayed if the generated files might overwrite existing blueprints. Hitting ESC while editing (or hitting ENTER while the edit field is blank) aborts the edit and restores the previous name. The blueprint name can be initialized on the commandline.

I felt like I had to get slightly creative to manage the state of the edit widget. The edit widget is not self-contained in that it can't manage it's own "active" state (it doesn't get notified of the activation hotkey if it isn't already active). I'm not sure if anything needs to change here right now, but it seems like "self-activating" widgets could use some more thought and API design.

There is a soft dependency on DFHack/dfhack#1892. Everything will work fine without that PR, but the tests that don't otherwise mock out `dfhack.filesystem.listdir_recursive()` will scan the real directory tree. This has no effect on the tests other than the (negligible) extra time it takes to do the IO, but my preference is to prevent tests from doing IO unless they absolutely need to.

### DIFF
--- a/gui/blueprint.lua
+++ b/gui/blueprint.lua
@@ -126,8 +126,9 @@ function NamePanel:detect_name_collision()
 
     local paths = dfhack.filesystem.listdir_recursive('blueprints', nil, false)
     for _,v in ipairs(paths) do
-        if v.path:startswith(name) and
-                v.path:sub(suffix_pos,suffix_pos):find('[.-]') then
+        if (v.isdir and v.path..'/' == name) or
+                (v.path:startswith(name) and
+                 v.path:sub(suffix_pos,suffix_pos):find('[.-]')) then
             self.has_name_collision = true
             return
         end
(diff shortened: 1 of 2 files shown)


gui/blueprint.lua
```diff
@@ -84,6 +85,67 @@ function ActionPanel:get_area_text()
...
+                  {text=self:callback('get_name_help', 2),
+                   pen=self:callback('get_help_pen')}}
+        },
+    }
+
+    self:detect_name_collision()
+end
+function NamePanel:detect_name_collision()
+    -- don't let base names start with a slash - it would get ignored by
+    -- the blueprint plugin later anyway
```
lethosor, line 116: Names that _end_ with a slash don't seem to be detected - I was able to save to `foo/` twice without getting a warning. I haven't looked into the logic much yet, but maybe `normalizePath()` in the loop below would help?

(This is very useful, though! I wouldn't have checked this case if not for https://github.com/DFHack/dfhack/pull/1891)

myk002: fixed. Good catch! For the record, we do call `normalizePath()` before comparing filenames. it even prevents you from entering a non-normalized path (try typing two slashes in a row)

lethosor (approved):

PR #307, opened 2021-07-03
[gui/blueprint] allow blueprint phases to be configurable

myk002:
add widgets that allow configuration of the blueprint phases

by default, phase labels are not shown (and their hotkeys are not active) and the 'phases' configuration is set to 'Autodetect'. Once the 'phases' hotkey is hit, 'Autodetect' toggles to 'Custom' and the individual phases are listed. Each can be toggled with their hotkey.

This is the last PR for `gui/blueprint` for milestone 1 of the blueprint overhaul plan (DFHack/dfhack#1842)

### DIFF
--- a/gui/blueprint.lua
+++ b/gui/blueprint.lua
@@ -147,6 +147,89 @@ function NamePanel:get_help_pen()
     return self.has_name_collision and COLOR_RED or COLOR_GREY
 end
 
+CycleHotkeyLabel = defclass(CycleHotkeyLabel, widgets.Label)
+CycleHotkeyLabel.ATTRS{
+    key=DEFAULT_NIL,
+    label=DEFAULT_NIL,
+    label_width=DEFAULT_NIL,
+    options=DEFAULT_NIL,
+    option_idx=1,
+    help=DEFAULT_NIL,
+    on_change=DEFAULT_NIL,
+}
+function CycleHotkeyLabel:init()
+    local contents = {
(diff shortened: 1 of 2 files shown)


gui/blueprint.lua
```diff
@@ -147,6 +147,89 @@ function NamePanel:get_help_pen()
...
+function CycleHotkeyLabel:cycle()
+    if self.option_idx == #self.options then
+        self.option_idx = 1
+    else
+        self.option_idx = self.option_idx + 1
+    end
+    if self.on_change then self.on_change() end
+end
+
+ToggleHotkeyLabel = defclass(ToggleHotkeyLabel, CycleHotkeyLabel)
```
lethosor, line 185: Maybe these would be good to move into the widgets library if they're generic. I had implemented something along these lines in https://github.com/lethosor/dfhack/commit/9390ba4da6a3c967f5f370202fa8d922987b5d70 (unmerged)

lethosor (approved):
Looks good!

gui/blueprint.lua
myk002: Yours look better designed, I think. Perhaps we could merge this as-is and we can migrate to your implementation once its merged?

lethosor: Ha, thanks, although it could probably stand to be revisited given that I last touched it in 2017. Agreed that this is good for now!

PR #310, opened 2021-07-10
New script: build-now - instantly completes pending building construction jobs

myk002:
Depends on DFHack/dfhack#1853
Fixes DFHack/dfhack#1857

`build-now` will instantly complete any non-suspended building construction job my mimicking standard game behavior and effects.

All buildings are processed by default, but the area of effect is configurable.

The following operations are handled:
- required building items are teleported to the target location and made part of the building
  - partially constructed buildings where some items have already been brought manually are correctly handled
  - items actively in transit are detached from the carrying unit and teleported the rest of the way
  - items stored in other buildings as clutter are disassociated from the other building and teleported to the target building
  - constructions make use of the `no_build_item` flag (a la the [cleanconst](https://github.com/DFHack/dfhack/blob/develop/plugins/cleanconst.cpp) plugin) and destroy the component item
- if a building requires architecture (such as bridges, furnaces, or wells), the architecture tasks are marked as complete, but no unit will gain experience for it, and no unit is recorded as the architect or builder. I have been unable to find any place where these unit references are used by the game (or by DFHack), though.
- buildings are set to their maximum build stage so they can be used normally
- tile occupancy is initialized
- constructions set tiles below them to be "Inside"
  - walls additionally set their own tile to be "Inside"
- wall and ramp constructions properly modify the tile above if it is open
- walls are connected to adjacent smooth walls (reusing functionality provided by the new `dig-now` plugin)
- items at build site moved out of the way

### DIFF
--- a/build-now.lua
+++ b/build-now.lua
@@ -1,8 +1,8 @@
 -- instantly completes unsuspended building construction jobs
 --[====[
 
-modtools/build-now
-==================
+build-now
+=========
 
 Instantly completes unsuspended building construction jobs. By default, all
 buildings on the map are completed, but the area of effect is configurable.
--- a/build-now.lua
+++ b/build-now.lua
@@ -500,6 +500,21 @@ for _,job in ipairs(get_jobs(opts)) do
     -- retrieve the items attached to the job before we destroy the references
     local items = get_items(job)
 
+    -- skip jobs whose attached items are already owned by the target building
+    -- but are not already part of the building. They are actively being used to
+    -- construct the building and we can't safely change the building's state.
+    for _,item in ipairs(items) do
+        if not item.flags.in_building and
+                bld == dfhack.items.getHolderBuilding(item) then
+            if not opts.quiet then
+                print(
+                    ('skipping building that is actively being constructed at'..
+                     ' (%d, %d, %d)'):format(bld.centerx, bld.centery, bld.z))
+            end
+            goto continue
+        end
+    end
+
     -- clear non-job items from the planned building footprint
     if not clear_footprint(bld, items) then
         dfhack.printerr(
--- a/build-now.lua
+++ b/build-now.lua
@@ -476,6 +476,10 @@ local function build_building(bld)
             bld:updateOccupancy(x, y)
         end
     end
+    -- doors link to adjacent smooth walls
+    if bld:getType() == df.building_type.Door then
+        dig_now.link_adjacent_smooth_walls(bld.centerx, bld.centery, bld.z)
+    end
 end
 
 local function throw(bld, msg)
--- a/build-now.lua
+++ b/build-now.lua
@@ -114,7 +114,7 @@ local function get_jobs(opts)
 
         -- accept building if if any part is within the processing area
         if bld.z < opts.start.z or bld.z > opts['end'].z
-                or bld.x2 < opts.start.x or bld.x1 > opts['end'].z
+                or bld.x2 < opts.start.x or bld.x1 > opts['end'].x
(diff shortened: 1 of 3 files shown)


lethosor:
An initial comment: I think this script should be at the top level, as opposed to under `modtools/`. There's some rationale for this at https://docs.dfhack.org/en/stable/docs/_auto/modtools.html and https://docs.dfhack.org/en/stable/docs/Lua%20API.html#scripts

myk002:
Ack. Will move to top dir. My thought was that this script was too "cheaty" for the top level, but I see from the docs you linked that `build-now` doesn't really meet the criteria for `modtools`.

build-now.lua
```diff
@@ -0,0 +1,555 @@
...
+    if construction_type == df.construction_type.Wall then
+        dig_now.link_adjacent_smooth_walls(pos)
+    end
+
+    -- for walls and ramps with empty space above, adjust the tile above
+    if construction_type == df.construction_type.Wall
+            or df.construction_type[construction_type]:find('Ramp') then
+        adjust_tile_above(xyz2pos(pos.x, pos.y, pos.z+1), item,
+                          construction_type)
+    end
```
lethosor, line 451: I found a weird bug that I think occurs in the section I've highlighted:

1. Run `createitem BAR IRON 3` (or have a small enough quantity of build materials that you can count easily)
2. Designate an up stair on the surface (b-C-u)
3. Run `build-now`
4. Designate a down stair right above the up stair (b-C-d)
5. Run `build-now`

This leaves behind the building item next to the staircase (its position isn't consistent, but it tends to be 1 tile away), which still has the TSK indicator until the game is advanced:

image

From some debugging, I confirmed that:
- `dfhack.items.remove` is returning false
- The position of the item is set to the position of the down stairs at the top of this function (where `item.pos` is retrieved for the first time)
- The position of the item has changed to be _next_ to the down stairs _before_ `remove()` is called. I'm not seeing where this happens yet.

lethosor (approved):
Looks good! Identified a small issue from some testing with constructions, but other things (furniture, depots) worked well, and I figure rolling this out is best.

build-now.lua
lethosor: Ok, evidently this was easier to narrow down than I thought. The build item is moved in the call to `dfhack.buildings.deconstruct()`. This also happens when building the up stair, but that ends up putting the building item on the ground instead of the air, so `dfhack.items.remove()` works.

From a brief look at `deconstruct()`, I think DF's logic in `bld->deconstructItems(false, false)` might be related to the issue, but calling that function alone for a designated down stair doesn't reproduce the issue.

It's possible that a fix might belong in `deconstruct()` instead of build-now.

myk002: I think I see what's going on here. `bld->deconstructItems(false, false);` will attempt to put the item on adjacent ground, and when it does, there is no problem. If there is no adjacent ground, then it places the object in mid-air and adds a `general_ref_projectile` to indicate that it is falling. This looks like completely rational behavior and not a bug. The problem is that `dfhack.items.remove()` fails for items that have a `PROJECTILE` general ref because `detachItem()` doesn't handle them.

Is there any reason why I shouldn't implement projectile handling in `detachItem()`? From inspection all it would take is deleting the link from `world->proj_list` and removing the general ref. I'll code it up and try it out.

lethosor: Thanks for tracking this down!

https://github.com/DFHack/dfhack/commit/b992b04f0bb7ffef8e7fb152967d438f398543d7 was where the projectile check was added. It doesn't go into much detail about that, but previously there were _no_ checks for unhandled general refs, so I think handling projectiles the way you did in dfhack/dfhack#1926 is reasonable.

PR #311, opened 2021-07-11
Added reveal-adv-map.lua

AtomicChicken:
Whenever a new adventure mode game is started, world tiles in the player's vicinity are flagged as `discovered` and those that are not are hidden from view. This script either reveals or hides the entire adv mode world map by toggling this flag accordingly.


lethosor (approved):
Nice!

PR #313, opened 2021-07-15
[workorder] use newly id'd manager_order_condition_item fields

myk002:
@quietust 
fields given names in https://github.com/DFHack/df-structures/commit/253dfb374cca068843809022dc27605b1b8d99e3

### DIFF
--- a/workorder.lua
+++ b/workorder.lua
@@ -464,9 +464,9 @@ local function create_orders(orders)
                     condition.has_tool_use = tmp
                 end
 
-                condition.anon_1 = -1
-                -- condition.anon_2 = ?
-                condition.anon_3 = -1
+                condition.min_dimension = -1
+                -- condition.contains = ?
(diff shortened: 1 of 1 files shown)


workorder.lua
```diff
@@ -464,9 +464,9 @@ local function create_orders(orders)
                     condition.has_tool_use = tmp
                 end
 
-                condition.anon_1 = -1
-                -- condition.anon_2 = ?
-                condition.anon_3 = -1
```
lethosor, line 469: Another option (albeit less explicit) is to set the `init-value` for both of these fields to `-1` in df-structures. That would make `df.manager_order_condition_item:new()` initialize them to -1. Not really necessary to change the script, IMO, but I'll add an init-value for `reaction_id` just to make other scripts a bit safer (`min_dimension` already has it).

PR #315, opened 2021-07-24
Autonick

Moth-Tolias:
hullo! i made this script a while ago and it's been useful to me, so i asked @BenLubar about it and xey said to submit a PR here.

### DIFF
--- a/autonick.lua
+++ b/autonick.lua
@@ -18,15 +18,13 @@ You may wish to use this script with the "repeat" command, e.g:
 ]====]
 
 names = {};
-index = 1;
 
 path = dfhack.getDFPath () .. "/dfhack-config/autonick.txt";
 
 -- grab list, put in array
 for line in io.lines(path) do
     if line ~= "" then
-        names [index] = line;
-        index = (index + 1);
+        table.insert(names, line);
     end;
 end;
 
--- a/autonick.lua
+++ b/autonick.lua
@@ -18,12 +18,12 @@ You may wish to use this script with the "repeat" command, e.g:
 
 ]====]
 
-local taken = {}
+local seen = {}
 --check current nicknames
 for _,unit in ipairs(df.global.world.units.active) do
     if dfhack.units.isCitizen(unit) and
     unit.name.nickname ~= "" then
-        taken[unit.name.nickname] = true
+        seen[unit.name.nickname] = true
     end
 end
 
@@ -34,8 +34,9 @@ for line in io.lines(path) do
     line = line:trim()
     if (line ~= "")
     and (not line:startswith('#'))
-    and (not taken[line]) then
+    and (not seen[line]) then
         table.insert(names, line)
+        seen[line] = true
     end
 end
 
--- a/autonick.lua
+++ b/autonick.lua
@@ -24,7 +24,8 @@ path = dfhack.getDFPath () .. "/dfhack-config/autonick.txt";
 
 -- grab list, put in array
 for line in io.lines(path) do
-    if (line:trim() ~= "") and (line:sub(1, 1) ~= "#") then
+    line = line:trim();
+    if (line ~= "") and (line:sub(1, 1) ~= "#") then
         table.insert(names, line);
     end;
 end;
--- a/autonick.lua
+++ b/autonick.lua
@@ -18,17 +18,17 @@ You may wish to use this script with the "repeat" command, e.g:
 
 ]====]
 
-names = {};
+names = {}
 
 path = dfhack.getDFPath () .. "/dfhack-config/autonick.txt";
 
 -- grab list, put in array
 for line in io.lines(path) do
-    line = line:trim();
+    line = line:trim()
     if (line ~= "") and (not line:startswith('#')) then
-        table.insert(names, line);
-    end;
-end;
+        table.insert(names, line)
+    end
+end
 
 --check current nicknames
 for _,unit in ipairs(df.global.world.units.active) do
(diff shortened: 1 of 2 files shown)


autonick.lua
```diff
@@ -0,0 +1,57 @@
...
+
+You may wish to use this script with the "repeat" command, e.g:
+``repeat -name autonick -time 3 -timeUnits months -command [ autonick ]``
+
+]====]
+
+names = {};
+index = 1;
+
+path = dfhack.getDFPath () .. "/data/init/names.txt";
```
myk002, line 23: This is a user-written configuration file, right? It should probably go in dfhack-config/autonick.txt

autonick.lua
```diff
@@ -0,0 +1,57 @@
+-- automatically nickname dwarves
+--[====[
+
+autonick
+========
+Automatically nickname dwarves according to ``data/init/names.txt``.
+
+One nickname per line, or an empty line. Repeat entries are discarded.
```
myk002, line 8: It would be useful to also accept comments (lines starting with a '#' character)

autonick.lua
Moth-Tolias: quite right! i normally only use the default /data configs, so i totally overlooked dfhack-config.

autonick.lua
```diff
@@ -0,0 +1,57 @@
...
+
+names = {};
+index = 1;
+
+path = dfhack.getDFPath () .. "/data/init/names.txt";
+
+-- grab list, put in array
+for line in io.lines(path) do
+	if line ~= "" then
+		names [index] = line;
```
myk002, line 28: Consider table.insert(names, line) so you don't have to manually track index

autonick.lua
```diff
@@ -0,0 +1,57 @@
...
+]====]
+
+names = {};
+index = 1;
+
+path = dfhack.getDFPath () .. "/data/init/names.txt";
+
+-- grab list, put in array
+for line in io.lines(path) do
+	if line ~= "" then
```
myk002, line 27: line = line:trim() would be useful here, to remove leading and trailing whitespace

autonick.lua
Moth-Tolias: still need to write the default file, but the location has been changed now.

autonick.lua
```diff
@@ -0,0 +1,56 @@
...
+                --remove current nickname from array
+                table.remove(names, i);
+            end;
+        end;
+    end;
+end;
+
+--assign names
+for _,unit in ipairs(df.global.world.units.active) do
+    if #names > 0 then --if there are any names left
```
myk002, line 47: You can reduce complexity here by breaking out of the loop if #names == 0:
if #names == 0 then break end
if dfhack.units.isCitizen(...

autonick.lua
```diff
@@ -0,0 +1,56 @@
...
+end;
+
+--check current nicknames
+for _,unit in ipairs(df.global.world.units.active) do
+    if dfhack.units.isCitizen(unit) and
+    unit.name.nickname ~= "" then
+        for i,name in ipairs(names) do
+            if name == unit.name.nickname then
+                --remove current nickname from array
+                table.remove(names, i);
```
myk002, line 39: Instead of looping through to remove taken nicknames, how about filtering them out when we build the names list?

You could add this before the io.lines loop above (sorry for the formatting.. I'm writing this on my phone):

local taken = {}
for _,unit in ipairs(df.global.world.units.active) do
    if dfhack.units.isCitizen(unit) and
    unit.name.nickname ~= "" then taken[unit.name.nickname] = true end
end

And then when building names, check if taken[name] is truthy

autonick.lua
```diff
@@ -0,0 +1,56 @@
+-- automatically nickname dwarves
+--[====[
+
+autonick
+========
+Automatically nickname dwarves according to ``dfhack-config/autonick.txt``.
```
myk002, line 6: This might be a bit clearer:

Gives dwarves unique nicknames chosen randomly from ``dfhack-config/autonick.txt``.

autonick.lua
```diff
@@ -0,0 +1,56 @@
...
+
+]====]
+
+names = {};
+
+path = dfhack.getDFPath () .. "/dfhack-config/autonick.txt";
+
+-- grab list, put in array
+for line in io.lines(path) do
+    if (line:trim() ~= "") and (line:sub(1, 1) ~= "#") then
```
myk002, line 27: it would be better to assign the trimmed value back to line -- otherwise the space-padded version will get inserted into the names list.

also, line:startswith('#') also works and is shorter

myk002:
and an overall comment - there isn't a strict code style standard, but DFHack generally doesn't use semicolons for lua scripts

autonick.lua
Moth-Tolias: will do this next.

Moth-Tolias:
should be all good now =] i'm not sure where i would submit the example autonick.txt file though - the main df hack once [if] this gets merged?

myk002 (approved):
@lethosor has the final say, but this looks good to me! Thanks for implementing all the suggestions!

myk002:
> should be all good now =] i'm not sure where i would submit the example autonick.txt file though - the main df hack once [if] this gets merged?

Yeah, once this is merged, add the default autonick.txt to https://github.com/DFHack/dfhack/tree/develop/dfhack-config. That gets installed to <df_dir>/dfhack-config/default/ and anything in default gets copied to <df_dir>/dfhack-config/ if the corresponding fine doesn't already exist

autonick.lua
```diff
@@ -0,0 +1,53 @@
+-- automatically nickname dwarves
+--[====[
+
+autonick
+========
+Gives dwarves unique nicknames chosen randomly from ``dfhack-config/autonick.txt``.
+
+One nickname per line.
+Empty lines, lines beginning with ``#`` and repeat entries are discarded.
```
myk002, line 9: I just noticed that repeat entries are *not* discarded. They could be, though, if we add the names read from the config file to the `taken` set as we read them. Then any repeat entries will be properly ignored.

maybe rename `taken` to `seen` as well to reflect its new role.

lethosor (approved):

Moth-Tolias:
thanks for merging! i'm really glad this got in =]

PR #316, opened 2021-07-31
add do-job-now.lua: sets do_now on selection-related job

TymurGubayev:
recommended usage:

1. add `keybinding add Alt-N do-job-now` to `dfhack*.init`
2. select anything in game
3. press `Alt-N`
4. ~~enjoy~~The script will try its best to find a related job and then set it's `do_now` flag.
   * A selected job is handled directly,
   * for an item with `in_job` a job from its `specific_refs` is affected,
   * for a building, a `ConstructBuilding` or `DestroyBuilding` job,
   * for a unit, either their job (for citizens -- not really interesting) or a job involved the unit (f.e. for animals),
   * for a plant, a job at the same coordinates,
   * for a work order, every job from this work order.


myk002:
Ha, I just wrote a related (but complementary) script: http://www.bay12forums.com/smf/index.php?topic=176889.msg8300718#msg8300718

This looks awesome too. I'll reference it in the docs I just wrote for my script.

TymurGubayev:
> Ha, I just wrote a related (but complementary) script: http://www.bay12forums.com/smf/index.php?topic=176889.msg8300718#msg8300718
> 
> This looks awesome too. I'll reference it in the docs I just wrote for my script.

nice :)

do-task-now.lua
```diff
@@ -0,0 +1,226 @@
...
+        print("... by " .. getUnitName(unit) )
+    end
+end
+
+local function doItemJobNow(item)
+    if not item.flags.in_job then
+        qerror(dfhack.items.getDescription(item, 0) .. " must be in a job! (look for 'TSK')")
+    end
+
+    for _, sref in ipairs(item.specific_refs) do
```
myk002, line 52: you can simplify this with a call to `dfhack.items.getSpecificRef(item, type)`

do-task-now.lua
```diff
@@ -0,0 +1,226 @@
...
+    if #building.jobs > 0
+        and    (
+            building.jobs[0].job_type == df.job_type.ConstructBuilding
+         or building.jobs[0].job_type == df.job_type.DestroyBuilding
+        )
+    then
+        doJobNow(building.jobs[0])
+        return
+    end
+    print("Couldn't find any job for " .. utils.getBuildingName(building))
```
myk002, line 73: maybe clarify here that we couldn't find any ConstructBuilding or DestroyBuilding job

do-task-now.lua
```diff
@@ -0,0 +1,226 @@
...
+        and scr.tab == df.viewscreen_workshop_profilest.T_tab.Orders
+    then
+        orders = scr.orders
+        idx = scr.order_idx
+    end
+    if orders then
+        if idx < #orders then
+            return orders[idx]
+        else
+            if not silent then qerror("Invalid work order selected") end
```
myk002, line 160: this is dead code because silent is always true. you could make silent configurable with a commandline param, or just remove the dead code.

if you do make it configurable, you should change the `qerror` to a `dfhack.printerr`. `qerror` should only be used for errors, but `silent` implies only logging verbosity.

do-task-now.lua
```diff
@@ -0,0 +1,226 @@
...
+    -- help
+    ["-?"] = print_help,
+    ["?"] = print_help,
+    ["--help"] = print_help,
+    ["help"] = print_help,
+    -- action
+    ["default"] = doSelectedEntityJobNow,
+}
+
+-- Lua is beautiful.
```
myk002, line 225: hear hear

do-task-now.lua
```diff
@@ -0,0 +1,226 @@
+-- do-task-now makes a job involving current selection high priority.
+
+-- place this file in your /df/hack/scripts folder.
+
+local function print_help()
+    print [====[
+
+do-task-now
+===========
+Select anything in the game and execute ``do-task-now`` to make
```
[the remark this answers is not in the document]
TymurGubayev: like this?

do-task-now.lua
TymurGubayev: well, this was mostly to make this function look the same as dfhack.gui.getSelected*-functions.

do-task-now.lua
TymurGubayev: :)

TymurGubayev:
after this and #317 are merged I'll add a link to `prioritize` as well at the end of the docs.

TymurGubayev:
> after this and #317 are merged I'll add a link to `prioritize` as well at the end of the docs.
(or you can do it yourself if you like)

do-task-now.lua
```diff
@@ -7,10 +7,17 @@ local function print_help()
...
-deconstruct jobs are considered. For a work order, all its jobs
-are affected.
+
+The script will try its best to find a job related to the selected entity
+(which can be a job, dwarf, animal, item, building, plant or work order) and then
+mark this job as high priority. There is no visual indicator, please look
+at the dfhack console for output. If a work order is selected, every job
+currently present job from this work order is affected, but not the future ones.
+
+For best experience add following to your `dfhack*.init`::
```
myk002: "add following" -> "add the following"

do-task-now.lua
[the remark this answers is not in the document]
myk002: looks good! thanks!

do-task-now.lua
myk002: understood, but I'd still prefer removing the dead code (you don't actually have to use the 'silent' param -- you could even remove it altogether and still pass `true` to the call. it will just get ignored)

do-task-now.lua
```diff
@@ -0,0 +1,233 @@
...
+    if #building.jobs > 0
+        and    (
+            building.jobs[0].job_type == df.job_type.ConstructBuilding
+         or building.jobs[0].job_type == df.job_type.DestroyBuilding
+        )
+    then
+        doJobNow(building.jobs[0])
+        return
+    end
+    print("Couldn't find neither construct nor destroy building job for " .. utils.getBuildingName(building))
```
myk002, line 80: double negative -- should be "Couldn't find either ... or" or "Could find neither ... nor"

lethosor:
Looks good!

How would you feel about renaming it to `do-job-now`? Asking partly because searching for "task" in this PR only turns up the script name - we've been using "job" everywhere else. There is potential for confusion with `tweak do-job-now`, however, which adds a UI toggle for the `job.flags.do_now` flag, but I would personally be ok with the shared name (given that most tweaks tend to be UI changes, not one-off tools).

TymurGubayev:
> How would you feel about renaming it to `do-job-now`? Asking partly because searching for "task" in this PR only turns up the script name - we've been using "job" everywhere else. There is potential for confusion with `tweak do-job-now`, however, which adds a UI toggle for the `job.flags.do_now` flag, but I would personally be ok with the shared name (given that most tweaks tend to be UI changes, not one-off tools).

Yeah, I actually started with `do-job-now`, but then noticed the workshop action is called "Do task now!" and decided this would be less confusing, but since everywhere else and on the wiki it's called "job" let's use it.

lethosor (approved):
Yeah, I guess DF isn't particularly consistent there. Thanks!

PR #317, opened 2021-07-31
New script: prioritize - monitors job creation and boosts priority for configurable job types

myk002:
The prioritize script sets the `do_now` flag on all of the specified types of jobs that are currently ready to be picked up by a dwarf. This will force them to complete the jobs as soon as possible. This script can also continue to monitor creation of new jobs and automatically boost the priority of newly created jobs of specific types.

This is most useful for ensuring important (but low-priority -- according to DF) tasks don't get indefinitely ignored in busy forts. The list of monitored job types is cleared whenever you load a new map, so you can add a line like the one below to your `onMapLoad.init` file to ensure important job types are always completed promptly in your forts:
```
    prioritize -a StoreItemInVehicle PullLever DestroyBuilding RemoveConstruction RecoverWounded FillPond DumpItem SlaughterAnimal
```

This implementation uses the `eventful` plugin to hook the creation of jobs so it can automatically modify new jobs. It cleans up after itself and unregisters the callbacks when no tracked job types remain (or when a map is unloaded).

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,6 +14,7 @@ that repo.
 # Future
 
 ## New Scripts
+- `prioritize`: automatically boosts the priority of current and/or future jobs of the selected types
 - `reveal-adv-map`: exposes/hides all world map tiles in adventure mode
 
 ## Fixes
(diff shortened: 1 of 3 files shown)


myk002:
docs are failing because I pre-emptively refer to `do-task-now` from #316

lethosor (approved):

PR #320, opened 2021-08-10
add warn-starving sanity option

Moth-Tolias:
implements/resolves DFHack/dfhack#1809.

unfortunately my computer is busted at present so i haven't been able to test this, but i believe it should all be correct?


lethosor (approved):
Thanks!

PR #321, opened 2021-08-15
Add gui/petitions.lua: shows list of fort's petitions

TymurGubayev:
image


gui/petitions.lua
```diff
@@ -0,0 +1,307 @@
...
+14                       = PlotFrameTreason
+15                       = PlotInduceWar
+]]
+
+if not dfhack.world.isFortressMode() then return end
+
+-- from gui/unit-info-viewer.lua
+do -- for code folding
+--------------------------------------------------
+---------------------- Time ----------------------
```
lethosor, line 45: Is this section copied from somewhere else? Asking partly due to the different indentation. It might be useful to include in a library so that other scripts can use it too.

gui/petitions.lua
```diff
@@ -0,0 +1,307 @@
...
+--------------------------------------------------
+end
+
+local we = df.global.ui.group_id
+
+local function getAgreementDetails(a)
+    local sb = {} -- StringBuilder
+
+    sb[#sb+1] = {{text = "Agreement #" ..a.id, pen = COLOR_RED}}
+    sb[#sb+1] = NEWLINE
```
lethosor, line 122: Recommendation:
```suggestion
    table.insert(sb, NEWLINE)
```
This would avoid a decent amount of repetition throughout this function.

gui/petitions.lua
ab9rf: See comment on line 42: `from gui/unit-info-viewer.lua`

gui/petitions.lua
```diff
@@ -0,0 +1,307 @@
...
+        end
+    end
+end
+
+function petitions:onRenderFrame(p, frame)
+    petitions.super.onRenderFrame(self, p, frame)
+    p:seek(frame.x1 + 2, frame.y1+frame.height-1):key_string('CUSTOM_F', "toggle fulfilled")
+end
+
+function petitions:onInput(keys)
```
myk002, line 294: If you use a Label widget to display the text, you get scrolling for free

TymurGubayev: I've probably copied it from somewhere...

I've refactored it to use Label now.

gui/petitions.lua
```diff
@@ -0,0 +1,301 @@
...
+    self:onResize(dfhack.screen.getWindowSize()) -- applies new frame_width
+end
+
+function petitions:onRenderFrame(painter, frame)
+    petitions.super.onRenderFrame(self, painter, frame)
+    
+    painter:seek(frame.x1+2, frame.y1 + frame.height-1):key_string('CUSTOM_F', "toggle fulfilled")
+    
+    local label = self.subviews.text
+    if label:getTextHeight() > label.frame_body.height then
```
myk002, line 278: This looks like it should be added to the Label class. Would it still look good if rendered on the label widget instead of the frame boundary?

TymurGubayev: I personally prefer the arrows and the text rendered on the frame. Second-best good-looking is arrows on an extra column on the very right.

I think both are possible to implement, although the second option is probably much easier.

Here some images for comparison:

on the border:
image

on an extra right-most column:
image

arrows on the label widget, no inset (2nd image illustrates a problem with that, notice how arrow hides a character):
image
image

label-widget with `frame_inset = 1`
image

myk002: Thank you for visualizing the options!

Visual-wise, I think the arrows are easiest to see when they're inside the frame. We have a long-standing standard of putting hotkey hints on the bottom frame (and they look fine there too), so I vote for arrows inside frame, hotkey text on bottom frame.

Architecture-wise, the arrow logic should be part of the Label widget. Labels are used all over the place, and they don't always have a frame around them, but they're always scrollable. We could potentially have the logic in the Label widget but render it somewhere else. That might cause confusion, though.

Behavior-wise, the arrows should only appear if the content height is larger than the Label body area height, otherwise existing, static labels would get unneeded arrows. If they appear, maybe they should force at least a right margin of 1 to avoid the character hiding problem you highlighted. If the widget already has a frame_inset>=1 (or frame_inset.r>=1) then this is a noop, otherwise, just the right inset can be increased by 1. The arrows should probably align vertically with the top and bottom frame inset, and be in the column just to the right of the end of the text area, but I guess check what looks best if frame inset is like 5 or something.

Thank you for doing this! This will be an awesome addition to the Label widget!

TymurGubayev: I've created a WIP pull request: https://github.com/DFHack/dfhack/pull/2101 (there is a problem I'm not quite sure what to do about)

myk002: https://github.com/DFHack/dfhack/pull/2101 is now merged. Thanks!

myk002 (changes requested):
needs updating to the latest master and use of the new Label additions

myk002 (approved):

PR #322, opened 2021-08-15
add workorder-job-repeater.lua: improve infinite work orders

TymurGubayev:
This script recreates workshop jobs upon completion in case they come from a work order with amount set to infinite. Usually a job for an infinite work order is only created once a day, resulting in dwarfs lazying around (this doesn't apply to non-infinite work orders). This script makes the dwarfes work harder.

For best experience add following to your ``onMapLoad*.init``:

    workorder-job-repeater start

### DIFF
--- a/workorder-job-repeater.lua
+++ b/workorder-job-repeater.lua
@@ -1,0 +1,219 @@
+-- simulates workshop's "repeat" for infinite work orders
+
+local function print_help()
+    print[====[
+
+workorder-job-repeater
+======================
+This script recreates workshop jobs upon completion in case
+they come from a work order with amount set to infinite. Usually
+a job for an infinite work order is only created once a day,
+resulting in dwarfs lazying around (this doesn't apply to non-infinite
+work orders). This script makes the dwarfes work harder.
+
+For best experience add following to your ``onMapLoad*.init``::
+
+    workorder-job-repeater start
+
+]====]
+end
+
+local utils = require 'utils'
+local eventful = require 'plugins.eventful'
+
+local DEBUG = 1
+local INFO = 2
+local WARN = 3
+local ERROR = 4
+local OFF = 6
+
+local verbosity = INFO
+
+local function log(level, ...)
+    if level < verbosity then return end
+    print (...)
+end
+
+local findManagerOrderById = function (id)
+    for _, o in ipairs(df.global.world.manager_orders) do
+        if o.id == id then
+            return o
+        end
+    end
+    return nil
+end
+
+local findBuildingById = function (id)
+    for _, b in ipairs(df.global.world.buildings.all) do
+        if b.id == id then
+            return b
+        end
+    end
+    return nil
+end
+
+local findJobById = function (id)
+    for _link, j in utils.listpairs(df.global.world.jobs.list) do
+        if j.id == id then
+            return j
+        end
+    end
+    return nil
+end
+
+local only_infty = true -- we will only repeat jobs which work order's amount is set to infinity. Or not.
+
+-- from http://www.bay12forums.com/smf/index.php?topic=164123.msg8300009#msg8300009
+--[[ "I don't fully understand how this works, so this might be horribly wrong
+      The DF job assigner needs a job to be in postings for it to automatically be assigned to a unit."(c)
+--]]
+function addJobToPostings(job)
+    local addedIndex = false
+    -- Find the first free empty postings to add a job to
+    for index, posting in ipairs(df.global.world.jobs.postings) do
+        if posting.job == nil then
+            -- It's free real estate!
+            posting.job = job
+            posting.anon_1 = 0
+            posting.flags.dead = false
+            
+            job.posting_index = posting.idx
+            addedIndex = posting.idx
+            break
+        end
+    end
+    if addedIndex then
+        log(DEBUG, "Added job " .. job.id .. " to postings as index " .. addedIndex)
+    else
+        log(DEBUG, "Couldn't find posting slot to add in job " .. job.id)
+    end
+    
+    return addedIndex
+end
+
+local function repeatJob(order, job)
+    if not (job and order) then
+        qerror('! (job and order)')
+        return
+    end
+    
+    if only_infty and 0 ~= order.amount_total then
+        log(DEBUG, "skipping job " .. dfhack.job.getName(job) .. " (id " .. job.id .. "): order.amount_total > 0")
+        return
+    end
+    
+    if not (order.status.validated and order.status.active) then
+        log(DEBUG, "skipping job " .. dfhack.job.getName(job) .. " (id " .. job.id .. ") from order " .. order.id .. " because of order.status")
+        return
+    end
+    
+    -- for debug purposes: pause the game
+    --df.global.pause_state = true
+
+    log(DEBUG, "Repeat job " .. dfhack.job.getName(job) .. " (id " .. job.id .. ") from order " .. order.id)
+    
+    -- we need a clone because original(which itself is a copy) is destroyed or something.
+    -- cloneJobStruct doesn't copy flags over.
+    local by_manager = job.flags.by_manager
+    local do_now = job.flags.do_now
+    job = dfhack.job.cloneJobStruct(job)
+    job.flags.by_manager = by_manager
+    job.flags.do_now = do_now
+    
+    -- step 0: adjust what we can right now
+    job.flags.working = false
+    job.completion_timer = -1
+    job.items:resize(0) -- remove old items we worked upon
+    
+    local workshop = nil
+    local ixs = {}
+    for ix, gref in ipairs(job.general_refs) do
+        if gref:getType() == df.general_ref_type.BUILDING_HOLDER then
+            local building = gref:getBuilding()
+            if building then
+                log(DEBUG, "Building from gref: " .. building.id)
+                workshop = building
+            else
+                -- should never ever happen, I think
+                log(WARN, "Building from gref not found: ", gref.building_id)
+            end
+        else
+            ixs[ix] = true
+        end
+    end
+    for ix in pairs(ixs) do
+        log(DEBUG, "Removing from gref: ", job.general_refs[ix])
+        job.general_refs:erase(ix)
+    end
+    log(DEBUG, "#gref: " .. #job.general_refs .. " (should be 1)")
+    
+    -- step 1: associate a posting with it
+    local ok = addJobToPostings(job)
+    if not ok then
+        log(WARN, "Couldn't add job of order.id " .. order.id .. " to a posting")
+        return
+    end
+    
+    -- step 2: add the job to the job list
+    local ok = dfhack.job.linkIntoWorld(job, false) -- we don't need a new id, I think?
+    if not ok then -- fallback just in case
+        log(DEBUG, "Couldn't add the job using job.id " .. job.id .. ". Requesting new id.")
+        dfhack.job.linkIntoWorld(job, true)
+    end
+    log(DEBUG, "Job " .. job.id .. " is linked.")
+
+    -- step 3: if a building is associated with the work order, find it
+    if order.workshop_id < 0 then
+        log(DEBUG, "Ignore work orders not associated with workshops")
+        return
+    end
+    
+    if not workshop then
+        log(WARN, "No workshop despite order.workshop_id " .. order.workshop_id .. " -- restoring")
+        workshop = findBuildingById(order.workshop_id)
+    end
+    log(DEBUG, "Associated workshop: " .. workshop.id)
+    
+    -- step 4: add this job to the workshop's job list.
+    workshop.jobs:insert('#', job)
+    log(DEBUG, "Added job " .. job.id .. " to workshop.jobs")
+end
+
+local function start()
+    eventful.onJobCompleted.workorder_repeat_job_immediately = function(job)
+        local order_id = job.order_id
+        local order = order_id > 0 and findManagerOrderById(order_id)
+        if order then
+            -- log(DEBUG, "work order", order_id, order)
+            repeatJob(order, job)
+        end
+    end
+    eventful.enableEvent(eventful.eventType.JOB_COMPLETED, 1) -- check every tick
+
+    log(INFO, "job repeater started")
+end
+
+local function stop()
+    eventful.onJobCompleted.workorder_repeat_job_immediately = nil
+    
+    log(INFO, "job repeater stopped")
+end
+
+local default_action = print_help
+local actions = {
+    -- help
+    ["-?"] = print_help,
+    ["?"] = print_help,
+    ["--help"] = print_help,
+    ["help"] = print_help,
+    --
+    ["default"] = print_help,
+    --
+    ["enable"] = start,
+    ["start"] = start,
+    ["disable"] = stop,
+    ["stop"] = stop,
+}
+
+-- Lua is beautiful.
+(actions[ (...) or "default" ] or default_action)(...)


workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+    end
+    
+    if not workshop then
+        log(WARN, "No workshop despite order.workshop_id " .. order.workshop_id .. " -- restoring")
+        workshop = findBuildingById(order.workshop_id)
+    end
+    log(DEBUG, "Associated workshop: " .. workshop.id)
+    
+    -- step 4: add this job to the workshop's job list.
+    workshop.jobs:insert('#', job)
```
myk002, line 178: what happens if the workshop already has 10 jobs?

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+local function start()
+    eventful.onJobCompleted.workorder_repeat_job_immediately = function(job)
+        local order_id = job.order_id
+        local order = order_id > 0 and findManagerOrderById(order_id)
+        if order then
+            -- log(DEBUG, "work order", order_id, order)
+            repeatJob(order, job)
+        end
+    end
+    eventful.enableEvent(eventful.eventType.JOB_COMPLETED, 1) -- check every tick
```
myk002, line 191: does we really need to check every tick? that can get expensive. would 5 or even 20 be sufficient?

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
+-- simulates workshop's "repeat" for infinite work orders
+
+local function print_help()
+    print[====[
+
+workorder-job-repeater
```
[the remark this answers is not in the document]
TymurGubayev: * _difference with check daily_: yes, this is different, because there is no time to lazy around between jobs until manager comes around to do the daily check
* _name_: I think `workorder` should be the prefix, to group workorder-related scripts together. `workorder-faster-infinity`?
* _workflow_: f.e. when you don't use workflow :) (I actually completely forgot about it). The main difference is, since it's a work order, it isn't bound to a workshop (and an infinite work order automatically gets assigned to multiple workshops). I.e. this script acts on a more global level. I'll try to improve the help in this direction.
* _differences from workflow_: you are correct, the job disappears for the day (or a bit longer for some reason) thus only producing a single cancellation per day

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+            
+            job.posting_index = posting.idx
+            addedIndex = posting.idx
+            break
+        end
+    end
+    if addedIndex then
+        log(DEBUG, "Added job " .. job.id .. " to postings as index " .. addedIndex)
+    else
+        log(DEBUG, "Couldn't find posting slot to add in job " .. job.id)
```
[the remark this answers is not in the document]
TymurGubayev: From the view of this function, it's like "I tried really hard but didn't succeed, what do you want to do about it now?" The caller could then theoretically create a posting - but I don't know how, so there is a check and a warning it didn't work.

https://github.com/DFHack/scripts/blob/b85f26d5f1eebd83c63ae5e6b1ac22242310691a/workorder-job-repeater.lua#L150-L155

I'll change the `WARN` to `ERROR`.

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+    end
+    
+    return addedIndex
+end
+
+local function repeatJob(order, job)
+    if not (job and order) then
+        qerror('! (job and order)')
```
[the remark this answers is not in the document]
TymurGubayev: should never ever happen, actually :) (which is why it's a `qerror` and not `log(ERROR, ...)`) But I'll change it just in case.

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+    job = dfhack.job.cloneJobStruct(job)
+    job.flags.by_manager = by_manager
+    job.flags.do_now = do_now
+    
+    -- step 0: adjust what we can right now
+    job.flags.working = false
+    job.completion_timer = -1
+    job.items:resize(0) -- remove old items we worked upon
+    
+    local workshop = nil
```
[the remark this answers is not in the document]
TymurGubayev: I don't know for sure, it worked for some time and didn't crash - which doesn't necessarily prove the ref won't point to freed memory, maybe the jobs are finishing too quickly or some DF mechanisms are doing something.

I'd say it more probable we are leaking memory here and the script in current form should actually crash/misbehave the same way it did until I started to make clones `job = dfhack.job.cloneJobStruct(job)`.

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+    
+    -- step 1: associate a posting with it
+    local ok = addJobToPostings(job)
+    if not ok then
+        log(WARN, "Couldn't add job of order.id " .. order.id .. " to a posting")
+        return
+    end
+    
+    -- step 2: add the job to the job list
+    local ok = dfhack.job.linkIntoWorld(job, false) -- we don't need a new id, I think?
```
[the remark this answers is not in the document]
TymurGubayev: The main reason is to reuse it is because we can. Also, this mimics a job on repeat more closely -- which may or may not influence how the game treats these jobs. The id isn't bound to the original job struct and just stays the same.

Since it's working fine (as far as I can tell at least) I'd say we leave it as is.

workorder-job-repeater.lua
TymurGubayev: Ah, I forgot the `onJobCompleted` is checked on timer and doesn't actually happen immediately. But it looks like the game is fine with it.

image

workorder-job-repeater.lua
TymurGubayev: `I doubt it ever becomes really expensive, but I'll change it to 10 just in case. And add an option, I guess.

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+local findManagerOrderById = function (id)
+    for _, o in ipairs(df.global.world.manager_orders) do
+        if o.id == id then
+            return o
+        end
+    end
+    return nil
+end
+
+local findBuildingById = function (id)
```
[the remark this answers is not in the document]
TymurGubayev: done

workorder-job-repeater.lua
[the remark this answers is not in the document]
TymurGubayev: Now that I think about what to write regarding `workflow` similarities, I notice it will only add confusion.

`workflow` _protects all **repeat jobs** from removal_. But if you are using work orders, you wouldn't notice this issue at all (like I didn't), because work orders themselves don't get cancelled.

workorder-job-repeater.lua
```diff
@@ -0,0 +1,219 @@
...
+end
+
+local function stop()
+    eventful.onJobCompleted.workorder_repeat_job_immediately = nil
+    
+    log(INFO, "job repeater stopped")
+end
+
+local default_action = print_help
+local actions = {
```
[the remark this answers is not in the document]
TymurGubayev: did that

workorder-job-repeater.lua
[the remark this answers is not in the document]
TymurGubayev: did that

workorder-job-repeater.lua
```diff
@@ -93,7 +126,8 @@ end
 
 local function repeatJob(order, job)
     if not (job and order) then
-        qerror('! (job and order)')
+        -- should never ever happen #hope
+        qerror("both job and order are nil")
```
myk002, line 130: This is incorrect. job *or* order is nil, not job *and* order

workorder-job-repeater.lua
```diff
@@ -14,6 +14,23 @@ work orders). This script makes the dwarfes work harder.
...
+
+``workorder-job-repeater start [<frequency>] | stop``
+
+``workorder-job-repeater --frequency <ticks> | --verbose | --help``
+
+:enable, start <frequency>: starts job monitoring, running every ``<frequency>`` ticks (default: 10).
+:disable, stop:             stops the script.
+
+-f, --frequency <ticks>     sets job monitoring frequency to ``<ticks>``
+                            (only if it's lower than the previously set value,
```
myk002, line 30: After DFHack/dfhack#1876 is merged, you will be able to adjust either up or down

myk002 (changes requested):
Could you add a changelog entry?

TymurGubayev:
> Could you add a changelog entry?

did that as well as other things you caught.

workorder-job-repeater.lua
```diff
@@ -0,0 +1,296 @@
...
+``[enable | disable] workorder-job-repeater``
+
+``workorder-job-repeater start [<frequency>] | stop``
+
+``workorder-job-repeater --frequency <ticks> | --verbose | --help``
+
+:enable, start <frequency>: starts job monitoring, running every ``<frequency>`` ticks (default: 10).
+:disable, stop:             stops the script.
+
+-f, --frequency <ticks>     sets job monitoring frequency to ``<ticks>``
```
myk002, line 29: How is a user to decide what tick frequency to use? Does this need to be configurable?

workorder-job-repeater.lua
```diff
@@ -0,0 +1,296 @@
...
+    
+**Usage**:
+
+``[enable | disable] workorder-job-repeater``
+
+``workorder-job-repeater start [<frequency>] | stop``
+
+``workorder-job-repeater --frequency <ticks> | --verbose | --help``
+
+:enable, start <frequency>: starts job monitoring, running every ``<frequency>`` ticks (default: 10).
```
myk002, line 26: I think you should pick a paradigm and stick with it. Enable/disable or start/stop, not both. Enable/disable before or after the command, not both.

workorder-job-repeater.lua
```diff
@@ -0,0 +1,296 @@
...
+-v, --verbose               toggles script's verbosity.
+-h, --help                  this help.
+
+]====]
+end
+
+local utils = require 'utils'
+local eventful = require 'plugins.eventful'
+
+local DEBUG = 1
```
myk002, line 39: A Lua debug logging framework is useful beyond just this script. Could this be factored out into a general API that could be used by all scripts and configured with `debugfilter`?

workorder-job-repeater.lua
```diff
@@ -0,0 +1,296 @@
...
+local colors = {}
+colors[DEBUG] = COLOR_LIGHTMAGENTA
+colors[INFO] = COLOR_CYAN
+colors[WARN] = COLOR_YELLOW
+colors[ERROR] = COLOR_LIGHTRED
+
+-- globals persist across script runs
+verbosity = verbosity or INFO
+
+if not scriptname then
```
myk002, line 54: Consider using `dfhack.current_script_name()` instead

workorder-job-repeater.lua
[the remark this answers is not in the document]
myk002: Yeah, from looking at the `cloneJobStruct` code (and the Lua API docs), it's making a deep copy. You need to free the allocated objects.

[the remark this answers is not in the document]
TymurGubayev: you mean the return value of `dfhack.job.cloneJobStruct(job)`? But then I wouldn't be able to use it as a new job (in `dfhack.job.linkIntoWorld` etc), would I, or am I missing something?

I hope the way the cloned job object is integrated into the game state makes the game dispose of it later - although I have no idea how to prove this.

workorder-job-repeater.lua
TymurGubayev: I can think of a way to make this into a Lua module/script, but I have no idea how to make it configurable with `debugfilter`. I can give it a try after I finish my other pending projects.

myk002: Yeah, definitely out of scope for this PR. We already cross the Lua/cpp interface for printing, but I'm not sure how we could efficiently make use of the cpp message filtering. It might end up being best to filter locally in Lua and just sync filter configs between the cpp and Lua implementations. But yeah, that's a discussion for another PR.

workorder-job-repeater.lua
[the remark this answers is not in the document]
myk002: I think the usage of the cloned job struct is fine. I'm just talking about the general refs that are being discarded without freeing

workorder-job-repeater.lua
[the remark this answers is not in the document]
myk002: 👍

workorder-job-repeater.lua
myk002: 👍

workorder-job-repeater.lua
TymurGubayev: did that

workorder-job-repeater.lua
TymurGubayev: the more frequent the better, but more CPU intensive. I'd like to leave this as an option just in case.

myk002: fair enough. maybe just document the default here so users know the flag is optional.

myk002 (changes requested):

workorder-job-repeater.lua
TymurGubayev: did that

workorder-job-repeater.lua
TymurGubayev: kept only enable/disable after command

workorder-job-repeater.lua
```diff
@@ -0,0 +1,278 @@
...
+    -- step 0: adjust what we can right now
+    job.flags.working = false
+    job.completion_timer = -1
+
+    -- remove old items we worked upon
+    log(DEBUG, "Freeing old items")
+    for _, item in pairs(job.items) do
+        item:delete()
+    end
+    job.items:resize(0)
```
myk002, line 153: just to check, are there ever any items in this vector? https://github.com/DFHack/dfhack/blob/develop/library/modules/Job.cpp#L89 clears the item and specific refs if `keepEverything` is false (which is the default). Is `keepEverything` somehow getting passed as `true`?

workorder-job-repeater.lua
```diff
@@ -0,0 +1,278 @@
...
+        else
+            ixs[ix] = true
+        end
+    end
+    for ix, gref in pairs(ixs) do
+        log(DEBUG, "Removing gref ", gref)
+        gref:delete()
+        job.general_refs:erase(ix)
+    end
+    log(DEBUG, "#gref: " .. #job.general_refs .. " (should be 1)")
```
myk002, line 176: I don't think this will work as expected. `ixs` will contain the indices to remove, but what those indices point to will change as we call `job.general_refs:erase()`. also, the second loop assumes the value in `ixs` is the general ref, but the first loop actually just sets the value to `true`. My guess is that this code has never actually removed any general references before. Is there a way you can get multiple general refs on a job so you can test this?

workorder-job-repeater.lua
```diff
@@ -0,0 +1,278 @@
...
+
+    -- step 4: add this job to the workshop's job list.
+    workshop.jobs:insert('#', job)
+    log(DEBUG, "Added job " .. job.id .. " to workshop.jobs")
+end
+
+frequency = frequency or 10
+local function start(N)
+    frequency = tonumber(N) or frequency
+    if eventful.onJobCompleted.workorder_repeat_job_immediately then
```
myk002, line 213: this should use the name of the script -- `eventful.onJobCompleted.workorder_job_repeater` (or `eventful.onJobCompleted['workorder-job-repeater']`)

myk002 (changes requested):

workorder-job-repeater.lua
myk002: I still have this question

workorder-job-repeater.lua
myk002, line 176:
```suggestion
    local workshop, gref_ix = nil, nil
    for ix,gref in ipairs(job.general_refs) do
        if gref:getType() == df.general_ref_type.BUILDING_HOLDER then
            workshop = gref:getBuilding()
            gref_ix = ix
        else
            gref:delete()
        end
    end
    if gref_ix then
        job.general_refs[0] = job.general_refs[gref_ix]
        job.general_refs:resize(1)
    else
        job.general_refs:resize(0)
    end
    log(DEBUG, "#gref: " .. #job.general_refs .. " (should be 1 if we found a building)")
```

myk002 (changes requested):

PR #325, opened 2021-08-20
Adds json option to devel/query

cppcooper:
Simple as that.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -38,6 +38,7 @@ that repo.
 - `devel/query`: added support for matching multiple patterns
 - `devel/query`: added support for selecting the highlighted job, plant, building, and map block data
 - `devel/query`: added support for selecting a Lua script (e.g. `dorf_tables`)
+- `devel/query`: added support for selecting a Json file (e.g. `dwarf_profiles.json`)
 - `devel/query`: removed options ``-listall``, ``-listfields``, and ``-listkeys`` - these are now simply default behaviour
 - `dorf_tables`: integrated `devel/query` to show the table definitions when requested with ``-list``
 - `geld`: fixed ``-help`` option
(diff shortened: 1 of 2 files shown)


devel/query.lua
```diff
@@ -331,6 +340,16 @@ function getSelectionData()
         selection = findTable(args.table)
         path_info = args.table
         path_info_pattern = path_info
+    elseif args.json then
+        local json = require("json")
+        local json_file = string.format("%s%s", dfhack.getDFPath(), args.json)
```
lethosor, line 345: Couldn't this just use string concatenation?
```suggestion
        local json_file = dfhack.getDFPath() .. args.json
```
This would fail if `args.json` is nil, but the `elseif` has already checked that.

Actually, I might recommend this, just so users don't always need to prefix paths with `/` (it's totally fine for `//` to be in a path, so `-json /hack/...` would still work):
```suggestion
        local json_file = string.format("%s/%s", dfhack.getDFPath(), args.json)
```
or
```suggestion
        local json_file = dfhack.getDFPath() .. '/' .. args.json
```
(take your pick)

cppcooper: lol I just copied it from `assign-profiles.lua`

changelog.txt
```diff
@@ -38,6 +38,7 @@ that repo.
 - `devel/query`: added support for matching multiple patterns
 - `devel/query`: added support for selecting the highlighted job, plant, building, and map block data
 - `devel/query`: added support for selecting a Lua script (e.g. `dorf_tables`)
+- `devel/query`: added support for selecting a Json file (e.g. dwarf_profiles.json)
```
lethosor, line 41: Single backticks make links, double backticks make code blocks. But no backticks are ok too

lethosor (approved):

PR #326, opened 2021-08-22
devel/query + dwarf-op cleanup

lethosor:
This switches these scripts over to use the new `safe_pairs()` library function from https://github.com/DFHack/dfhack/pull/1917, and also improves some other things I spotted while I was at it.

@cppcooper feel free to take a look

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -41,6 +41,7 @@ that repo.
 - `devel/query`: added support for selecting a Lua script (e.g. `dorf_tables`)
 - `devel/query`: added support for selecting a Json file (e.g. dwarf_profiles.json)
 - `devel/query`: removed options ``-listall``, ``-listfields``, and ``-listkeys`` - these are now simply default behaviour
+- `devel/query`: ``-table`` now accepts the same abbreviations (global names, ``unit``, ``screen``, etc.) as `lua` and `gui/gm-editor`
 - `dorf_tables`: integrated `devel/query` to show the table definitions when requested with ``-list``
 - `geld`: fixed ``-help`` option
 - `gui/gm-editor`: made search case-insensitive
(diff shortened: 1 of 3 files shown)


devel/query.lua
```diff
@@ -570,28 +570,6 @@ function is_exceeding_maxlength(index)
 end
 
 --Section: table helpers
-function safe_pairs(t, keys_only)
```
lethosor, line 573: Nothing was using the `keys_only=true` path here, and it looks like it would only work for DF types anyway. It's possible for callers to ignore values with `for k in safe_pairs(t)` in any case.

changelog.txt
```diff
@@ -41,6 +41,7 @@ that repo.
 - `devel/query`: added support for selecting a Lua script (e.g. `dorf_tables`)
 - `devel/query`: added support for selecting a Json file (e.g. dwarf_profiles.json)
 - `devel/query`: removed options ``-listall``, ``-listfields``, and ``-listkeys`` - these are now simply default behaviour
+- `devel/query`: ``--table`` now accepts the same abbreviations (global names, ``unit``, ``screen``, etc.) as `lua` and `gui/gm-editor`
```
cppcooper, line 44: should just be a single dash ``-table``

lethosor: Whoops, fixed

devel/query.lua
cppcooper: It may have had a purpose at some point, years ago.. but I don't recall what it is/was. So, `safe_pairs` is now available without defining it?

myk002: Recently added: https://github.com/DFHack/dfhack/blob/10db894d99daf71d297f4e877b24ee77ffd37b5d/library/lua/dfhack.lua#L176

lethosor: Yep - and this script was one of the callers I had in mind. This way we won't need to worry about multiple implementations floating around.

Also, to clarify, `pairs()` for DF types returns fields in the same order as `_index_table`, which is why it (or `safe_pairs()`) would be able to replicate the old `keys_only=true` behavior.

PR #331, opened 2021-08-29
[assign-preferences] Update  docs.

acattaneo88:
The plugin `stonesense` does not produce the `MatList.csv` file anymore, so I removed the suggestion about activating it in order to get the tokens.


lethosor (approved):
Looks like https://github.com/dfhack/stonesense/commit/84c72deecba3e0807a576f83575ec0df130e4c20 is the relevant commit, if you're interested.

One thing that we _could_ call out in the docs is `createitem inspect` or `changeitem info`. Those will print material tokens for selected items, which could be useful in some cases.

PR #339, opened 2021-09-24
Fix dfhack #1618 (advfort encrust/stud deletes items)

Ozzatron:
Fixes DFHack/dfhack#1618

This pull request adjusts `advfort_items.lua` to always append the available items for a job in their `job_item_idx` order. This fixes `Stud with <metal>` and `Encrust with Gems` jobs so that they won't delete both reagents.


myk002 (approved):
Looks good. Could you write a changelog entry?

edit: n/m I can get it after merge

PR #340, opened 2021-09-27
[quickfort] implement API for applying blueprint data passed in Lua tables

myk002:
single API that exposes all of quickfort's blueprint actions ("run", "orders", "undo")

Question: should the API docs be just in the `quickfort.lua` file or should it be added to the online docs? Right now I just provide a link in the online docs to the source tree where the API is documented.

also fixes a bug that crept into the last release that broke the `--verbose` flag.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -17,9 +17,11 @@ that repo.
 
 ## Fixes
 - `quickfort`: produce a useful error message instead of a code error when a bad query blueprint key sequence leaves the game in a mode that does not have an active cursor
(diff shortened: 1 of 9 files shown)


drummeur:
I'm always in favor of having online docs, not just in-source docs.  But that also adds another place where they have to be maintained :/

myk002:
> I'm always in favor of having online docs, not just in-source docs. But that also adds another place where they have to be maintained :/

fair enough. I can move them instead of copying them so there is only one place to maintain them.

PR #342, opened 2021-09-29
[quickfort] protect masterwork engravings from being designated for digging

myk002:
By default, `quickfort` will now refuse to designate tiles for digging if they have masterwork engravings on them. The new `--clobber-masterwork-engravings` param allows players to get the old behavior of digging them out anyway if that is actually desired.

Note that masterworks could still be destroyed in adjacent z-levels if `quickfort` designates a tile for channeling or ramp. I figure this is outside the scope of the expected masterwork protection.

I expect the `clobber-masterwork-engravings` param to be rarely used, so I didn't bother wiring it up to the gui interface. If we get more esoteric params in the future, I might add a popup to the gui where they can all be configured.

As a by-product of this PR, dig blueprints that interact with engravings (e.g. toggle show engraving) will now go a lot faster on maps that have a lot of engravings. Not that I expect there to be many of those types of blueprints, but hey, a side benefit is still a benefit : )

In testing this PR, I discovered a bug that I fixed in passing: `quickfort` would designate tiles for channeling even if they were underneath buildings. The UI doesn't allow this, so `quickfort` shouldn't allow this either.

### DIFF
--- a/internal/quickfort/command.lua
+++ b/internal/quickfort/command.lua
@@ -25,13 +25,15 @@ local command_switch = {
     undo='do_undo',
 }
 
-function init_ctx(command, blueprint_name, cursor, aliases, dry_run)
+function init_ctx(command, blueprint_name, cursor, aliases, dry_run,
+                  clobber_masterwork_engravings)
     return {
         command=command,
         blueprint_name=blueprint_name,
         cursor=cursor,
         aliases=aliases,
         dry_run=dry_run,
+        clobber_masterwork_engravings=clobber_masterwork_engravings,
         stats={out_of_bounds={label='Tiles outside map boundary', value=0},
                invalid_keys={label='Invalid key sequences', value=0}},
         messages={},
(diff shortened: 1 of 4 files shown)


myk002:
@drummeur PTAL

drummeur:
That looks good to me, overall.  The only thing I can see that wouldn't work 100% is that my script lets you choose the min quality of engraving to ignore.  I'm not sure if that's actually a particularly useful feature, though.  I personally would always only want MW, but I could see it being feasible for somebody else to decide that they only want MW or Exceptional quality engravings, because it takes too much time to get only masterwork.  That's up to you, though, if that seems worth including.  I can still easily include in my script that logic and just use quickfort for the actual designation parts, which are the chunks I've copied and pasted in the first place.  I also made a repo for my script here https://github.com/drummeur/Trackify

myk002:
Making the target quality configurable is an excellent idea. Let me rework this PR a bit to make that possible.

myk002:
@drummeur needs unit tests, but core logic is done. does this look better?

PR #349, opened 2021-11-08
Submitting first version of Enlinker to try to get feedback

UnknowableCoder:
After a few weeks of posting mostly idiotic questions out on the DFHack Discord, I finally managed to implement what I had long since wanted in DF: an alternative logic system that allows the user to set a threshold (and optional delay) for building activation and de-activation, providing endless new exciting possibilities for dwarfputing in general, on one hand, but also massively simplifying relatively simple things such as e. g. opening and closing a fortress "door" made of both bridges and floodgates through a single lever. Given my chronic lack of inspiration for good names, I call this Enlinker, but I'm sure we could figure out something better... 

Along the way, I decided it would be a good idea to store this information such that it would also work in Adventure Mode, so one could build e. g. a trap-filled maze for others to explore or an alternative entrance to a fortress that an adventurer could also use. This led me to develop some sort of system (still mostly experimental) that could load and save information based on the site-id, as suggested by the helpful folks over on the Discord.

The bulk of the logic is implemented in a C++ plugin that I will be also submitting to consideration through the main DFHack repo, but here I present the Lua code that implements the GUI that allows for user interaction.

**Examples in Fortress Mode:**

Building whose state is not managed by Enlinker:
imagem

Building whose state is managed by Enlinker:
imagem

Setting the building's state to depend on itself (this is bound to be useful at some point...):
imagem

Seeing a list of currently linked-to buildings and their weight in the input count:
imagem

**Example in Adventure Mode:**

Alternative view (all functionalities of Fortress Mode except adding new links - current links can be re-weighted, but no new buildings selected).
imagem

### DIFF
--- a/gui/enlinker-gui.lua
+++ b/gui/enlinker-gui.lua
@@ -1,0 +1,435 @@
+-- Interface front-end for enlinker plugin.
+
+local utils = require 'utils'
+local gui = require 'gui'
+local guidm = require 'gui.dwarfmode'
+local dlg = require 'gui.dialogs'
+local script = require 'gui.script'
+
+local plugin = require('plugins.enlink')
+local bselector = df.global.ui_build_selector
+
+--local buildings = df.global.buildings
+
+function get_current_building()
+
+  local selb = dfhack.gui.getSelectedBuilding(true)
+
+  if not selb then
+    local cursor = guidm.getCursorPos()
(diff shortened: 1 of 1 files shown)


myk002:
putting on hold until we can discuss a bit more on the plugin

myk002:
closing due to lack of docs; please reopen if you have time to finish this project

PR #350, opened 2021-11-09
combine. The merged combine-drinks/plants with greater functionality

Vettlingr:
Combine combines both combine-drinks and combine-plants and adds a lot of new features.
It can combine any food stuff in the selected stockpile or all stockpiles on the map.

I need to know who made the original combine-drinks and combine-plants scripts to know where credits is due.

Some parts of the scripts may be able to be merged, though I've already merged and shortened the script from 600 lines down to 300 by merging functions and using dynamic variables.

There may be requests to decouple eggs from fish or cheese from roasts as well as adding extracts/Liquid_misc as a valid itemType to be merged/combined. For now though, it seems that dwarves are very dilligent in combining milk buckets with barrels and alike, but milk/honey bought from merchants can be merged.

The script will not touch anything stored in bags in barrels, such as meal, seeds and dye.
It is possible to include these too, by finding containeditems within containeditems. 

Keep in mind that eggs and fish uses Caste and Race instead of Mattype and Matindex, this is why they have the bool>=10
So far I've only found eggs, raw fish and fish that does this.

### DIFF
--- a/combine.lua
+++ b/combine.lua
@@ -65,6 +65,13 @@ fat.max     =   20
 roasts.max  =   20
 fish.max    =   10
 
+if f.args.drinks then drinks.max = tonumber(f.args.drinks) end
+if f.args.plants then plants.max = tonumber(f.args.plants) end
+if f.args.meats then meats.max = tonumber(f.args.meats) end
+if f.args.fat then fat.max = tonumber(f.args.fat) end
+if f.args.roasts then roasts.max = tonumber(f.args.roasts) end
+if f.args.fish then fish.max = tonumber(f.args.fish) end
+
 --Not sure if these are needed.
 drinks.Tot=0 drinks.xTot=0
 plants.Tot=0 plants.xTot=0
--- a/combine.lua
+++ b/combine.lua
@@ -149,14 +149,13 @@ function Combineitems(building, tabl, food, bool)
     else
         local foodCount = getItems(rootItems, food, 0, bool)
         local removedFood = { } --as:bool[]
-        food.max=max
-        if f.args.max then max = tonumber(f.args.max)
-            if tonumber(f.args.max)== 0 then max = 500
+        if f.args.max then food.max = tonumber(f.args.max)
+            if tonumber(f.args.max)== 0 then food.max = 500
             end
         end
         for i=0,(foodCount-2) do
             local currentFood = food[i] --as:df.item_foodst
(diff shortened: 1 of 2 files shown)


myk002:
Is this PR still a draft of is it ready for review?

combine.lua
```diff
@@ -0,0 +1,306 @@
...
+local utils = require 'utils'
+
+local f = {
+    validArgs = utils.invert({ 'help', 'drinks', 'plants', 'meat', 'fish', 'fat', 'food', 'roasts', 'max', 'all', 'stockpile' });
+    args = utils.processArgs({...}, validArgs);
+help = [====[
+Combine
+=============
+Merge stacks of food in selected Stockpile or across all stockpiles on the map.
+Valid commands:
```
myk002, line 18: There are some patterns from other scripts that we can follow here to make the API more consistent. How about:
combine -> displays help
combine here -> combines everything it can for the stockpile under the cursor
combine all -> combines everything it can for all stockpiles
combine all --types=plants,roasts --max=100 -> combines all plant and roast barrels with a custom max stack size

So the syntax would be:
```
combine [help|here|all] [<options>]

Where 'help' (or nothing) displays help, 'here' directs combine to act on the selected stockpile, and 'all' directs combine to act on all stockpiles.

Options can be:
-t, --types <comma separated list of types>
    Restricts combine to acting on the specified types. Valid types are: ...
-m, --max <number>
    Sets the preferred target stack size. If not specified, defaults to ...
 
```

The `argparse` module will help you support that syntax. The prioritize script is a good model to follow.
https://github.com/DFHack/scripts/blob/master/prioritize.lua#L532

myk002 (changes requested):
I have some thoughts on the API before we get deeper into the code.

From a brief look through the body of the script, though, it looks like the implementation doesn't match the documentation in a number of places, so I'm not sure what your intentions are. For example, it is unclear whether each of the current options Isa flag or expects a numeric argument, the default stack size differs, and the undocumented stockpile option appears to change the behavior of the script.

myk002:
I updated the title of the PR to track the status. Could you remove the `[waiting on author]` tag when you get back to this script?

myk002:
@Vettlingr This script is a really good idea! Do you have time to address the PR comments?

myk002:
ping

myk002:

PR #351, opened 2022-01-03
Makes setfps also update fps_per_gfps

Putnam3145:
I realized that this doesn't update on its own while looking at timing stuff in-game. No idea if it's used anywhere, but it seems prudent to keep it accurate.


myk002 (approved):

myk002:
Needs a changelog entry, but I can add one after merge

PR #353, opened 2022-02-02
Randomized caste for modtools/transform-unit

Bagelboi:
This randomizes the caste if either the -caste argument is omitted or if the given caste isn't found

### DIFF
--- a/modtools/transform-unit.lua
+++ b/modtools/transform-unit.lua
@@ -83,7 +83,7 @@ if args.untransform then
  caste = normalRace[args.unit].caste
  normalRace[args.unit] = nil
 else
- if not args.race or not args.caste then
+ if not args.race then
   error 'Specficy a target form.'
  end
 
@@ -108,7 +108,12 @@ else
  end
 
  if not caste then
-  error 'Invalid caste.'
+  local rand_caste = math.random( 0, math.abs(#race.caste-1) )
+  if not race.caste[rand_caste] then
+    error 'Creature has no castes or non have been found'
+  else
+    caste = rand_caste
+  end
  end
 end
 


modtools/transform-unit.lua
```diff
@@ -108,7 +108,12 @@ else
  end
 
  if not caste then
-  error 'Invalid caste.'
+  local rand_caste = math.random( 0, math.abs(#race.caste-1) )
```
myk002, line 111: Could you write a unit test to show coverage of the case where race.caste has no elements?  See the `test/` dir for sample unit tests.

modtools/transform-unit.lua
```diff
@@ -108,7 +108,12 @@ else
...
+  if not race.caste[rand_caste] then
+    error 'Creature has no castes or non have been found'
```
myk002, line 113: Typo non -> none

modtools/transform-unit.lua
```diff
@@ -108,7 +108,12 @@ else
...
+  else
```
myk002, line 114: No need to have an else clause if the if clause throws an error. Just replace the else with an end and let the assignment happen unconditionally.

myk002 (changes requested):
I think using a random caste is good when none is specified, but if a caste is specified but not found, the script should throw an error.

myk002:
I updated the title of the PR to track the status. Could you remove the `[waiting on author]` tag when you get back to this script?

myk002:
@Bagelboi do you have time to address the review comments?

myk002:
ping on this

myk002:

PR #358, opened 2022-03-26
add gui/workorder-details.lua: adjusts input item, material, traits

TymurGubayev:
Adjust input items, material, or traits for work orders. Actual jobs created for it will inherit the details.

This is the equivalent of `gui/workshop-job` for work orders, with the additional possibility to set input items' traits.

F.e., we can make a work order to "Cut unknown material":
image
image

Which then appears like this in the workshop:
image
image

It has to be run from a work order's detail screen (:kbd:`j-m`, select work order, :kbd:`d`), because the game initializes work order's `.items` array as soon as the screen is opened (and then copies it over to the job).

For best experience add the following to your ``dfhack*.init``::

    keybinding add D@workquota_details gui/workorder-details


myk002 (changes requested):
Overall, looks solid. Thanks!

A couple questions:
- instead of copying code, can you `reqscript('gui/workshop-job')` and call any existing functions/inherit from existing classes directly?
- instead of `workorder-finetune`, how about `workorder-details` or `workorder-edit`?
- could you add some tests in the [test/gui](https://github.com/DFHack/scripts/tree/master/test/gui) directory to validate that the code continues to work when DF or DFHack behavior changes? This will help a lot to identify code that breaks on the upcoming steam version of DF 
- could you add a changelog entry, under new scripts?

myk002:
Also, does it make sense to be able to adjust order quantities or frequency with this script?

TymurGubayev:
> Also, does it make sense to be able to adjust order quantities or frequency with this script?

IMO this would be too deep in the details: from the work order list view `d` for DF details and another `d` for this script; adjusting frequency or quantities would be better on the list view itself.

TymurGubayev:
> * instead of `workorder-finetune`, how about `workorder-details` or `workorder-edit`?

I definitely don't like `-edit` because it suggests a complete editor, which this script isn't, it only manipulates a small part of a work order.

I am not quite sure about `-details`, because there is already one workorder details view - the game's `workquota_details` view where we land after pressing `d` in the `j-m` view. And here we have "additional details" (or "fine details"). Maybe something like `workorder-details-plus`? I'm not sure.

TymurGubayev:
> * instead of copying code, can you `reqscript('gui/workshop-job')` and call any existing functions/inherit from existing classes directly?

did that. Hat to modify (very minorly) the `gui/workshop-job.lua`. Inheritance won't work (I think) because it's `defclass(JobDetails, guidm.MenuOverlay)` vs `defclass(JobDetails, gui.FramedScreen)`, but just assigning functions works and did eliminate a huge chunk of duplicated code.

gui/workorder-finetune
```diff
@@ -0,0 +1,544 @@
...
+
+For best experience add the following to your ``dfhack*.init``::
+
+    keybinding add D@workquota_details gui/workorder-finetune
+
+]====]
+
+--[[
+credit goes to the author of `gui/workshop-job` where the majority
+of the code here comes from.
```
myk002, line 23: now that you're reusing the `gui/workshop-job` code directly, is this still true?

gui/workorder-finetune
```diff
@@ -0,0 +1,544 @@
...
+                end
+                choice.search_key = table.concat(search_key, ' ')
+            elseif choice.text then
+                choice.search_key = string.lower(choice.text)
+            end
+        end
+    end
+end
+
+guimat.ItemTraitDialog = function(args)
```
myk002, line 134: a script should not be injecting a function into the environment of a library module. You could add it to `gui/materials.lua` in a different PR, though.

gui/workorder-finetune
```diff
@@ -0,0 +1,544 @@
...
+    self.subviews.list:setChoices(choices)
+end
+
+JobDetails.canChangeIType = wsj.JobDetails.canChangeIType
+JobDetails.setItemType = wsj.JobDetails.setItemType
+JobDetails.onChangeIType = wsj.JobDetails.onChangeIType
+JobDetails.canChangeMat = wsj.JobDetails.canChangeMat
+JobDetails.setMaterial = wsj.JobDetails.setMaterial
+JobDetails.findUnambiguousItem = wsj.JobDetails.findUnambiguousItem
+JobDetails.onChangeMat = wsj.JobDetails.onChangeMat
```
myk002, line 428: Stealing individual methods from a class, while technically functional, could lead to issues later if `wsj` is updated with undesired behavior. Could you refactor `wsj.JobDetails` so that the common functions are in a superclass? Then that class could be clearly marked as "shared" and both the `JobDetails` here and the `JobDetails` in `wsj` could inherit from that common class.

gui/workorder-finetune
```diff
@@ -0,0 +1,544 @@
...
+function JobDetails:onGetSelectedJob()
+    return self.job
+end
+
+describe_item_type = wsj.describe_item_type
+is_caste_mat = wsj.is_caste_mat
+describe_material = wsj.describe_material
+list_flags = wsj.list_flags
+
+function describe_item_traits(iobj)
```
myk002, line 358: consider declaring these as `local` so they don't pollute the exported environment (also `set_search_keys` above)

gui/workorder-finetune
```diff
@@ -0,0 +1,544 @@
...
+end
+
+local scr = dfhack.gui.getCurViewscreen()
+if df.viewscreen_workquota_detailsst:is_instance(scr) then
+    -- by opening the viewscreen_workquota_detailsst the
+    -- work order's .items array is initialized
+    JobDetails{ job = scr.order }:show()
+else
+    qerror("This script needs to be run from a work order details screen")
+end
```
myk002, line 544: `qerror` breaks the control flow, so you can simplify (slightly) by reordering:
```
local scr = dfhack.gui.getCurViewscreen()
if not df.viewscreen_workquota_detailsst:is_instance(scr) then
    qerror("This script needs to be run from a work order details screen")
end

-- by opening the viewscreen_workquota_detailsst the
-- work order's .items array is initialized
JobDetails{ job = scr.order }:show()
```

myk002 (changes requested):

gui/workorder-finetune
TymurGubayev: did that

gui/workorder-finetune
TymurGubayev: did that too

gui/workorder-finetune
TymurGubayev: changed the wording a bit there

gui/workorder-finetune
TymurGubayev: Is there a convention how such a shared class should look like? I think to call it `JobDetailsBase` and place in `/scripts/gui/`.

TymurGubayev:
> * could you add some tests in the [test/gui](https://github.com/DFHack/scripts/tree/master/test/gui) directory to validate that the code continues to work when DF or DFHack behavior changes? This will help a lot to identify code that breaks on the upcoming steam version of DF

Started on that. There is currently some ugliness involved, f.e. I couldn't figure out how to emulate typing: `gui.simulateInput(dfhack.gui.getCurViewscreen(true), 'CUSTOM_C')` doesn't do anything on the manager new order screen.

Also, the test file-name has underscore instead of hyphen because `test -dhack/scripts/devel/tests -tworkorder-finetune` doesn't work. It looks like this:

> [DFHack]# test -dhack/scripts/devel/tests -tworkorder-finetune
> Loading tests from hack/scripts/devel/tests
> Loading file: /a-b.lua
> Loading file: /dialogs.lua
> Loading file: /workorder-finetune.lua
> Filtering tests by name pattern:
>   workorder-finetune
> Selected tests: 0/3
> Running 0 tests
> 
> Test summary:
> 0/0 tests passed
> 0/0 checks passed
> 0 test files failed to load
> 0 tests in unreachable modes

gui/workorder-finetune
myk002: `JobDetailsBase` sounds good. It can either stay in wsj (commented to indicate that it is shared across scripts) or it can go in `internal/gui/`. It shouldn't go in a new file in `gui/` because then it would show up in `ls` as an independent script

myk002:
You have to use the actual key code to navigate the df menus. Look in `data/init/interface.txt` and see which mapping of 'c' to use

myk002:
Not matching hyphens sounds like a bug in the test harness. I'll get that fixed.

myk002:
> Not matching hyphens sounds like a bug in the test harness. I'll get that fixed.

Actually, the test harness is working as designed. The problem is that the `-t` param takes a pattern, not a literal. You need to escape the hyphens in the script name:
`test -tworkorder%-finetune`

Or just `test -tfinetune` would do it too

myk002 (approved):
I'd like to test in-game a bit before merging, but code looks good. Thanks for the test!

myk002:
Could you add a changelog entry for this?

TymurGubayev:
> Could you add a changelog entry for this?

I've not forgotten about that, there is a dependency graph in my head: find out if the hyphen is a problem for tests (✔), decide on the script name (will probably go with `workorder-details`), then add the changelog entry.

myk002:
👍

TymurGubayev:
> You have to use the actual key code to navigate the df menus. Look in `data/init/interface.txt` and see which mapping of 'c' to use

under "actual key code" do you mean f.e. `D_JOBLIST` instead of `j` to open the job list? I figured that out, what I can't figure out is how to type something so that the game would apply the filter?
image

myk002:
> what I can't figure out is how to type something so that the game would apply the filter?

Yeah, that's tricky. I did something like that here:
https://github.com/DFHack/scripts/blob/a1df858afd4e1863fa17be659a51b4e8f65966ef/test/gui/blueprint.lua#L371
You essentially have to call the `onInput` method directly

TymurGubayev:
> > what I can't figure out is how to type something so that the game would apply the filter?
> 
> Yeah, that's tricky. I did something like that here:
> 
> https://github.com/DFHack/scripts/blob/a1df858afd4e1863fa17be659a51b4e8f65966ef/test/gui/blueprint.lua#L371
> 
> 
> You essentially have to call the `onInput` method directly

I think this only works when `view` is dfhack UI, but I need to send input to the result of `dfhack.gui.getCurViewscreen(true)`, which doesn't have the `onInput` method - unless I'm missing something.

myk002:
> I think this only works when `view` is dfhack UI, but I need to send input to the result of `dfhack.gui.getCurViewscreen(true)`, which doesn't have the `onInput` method - unless I'm missing something.

Sorry about that -- I thought this was one of *our* filters.

Finding the keycode for each key such that the viewscreen will accept it would work, but would be messy and arduous. Quickfort has an API that might be useful here. For example, to enqueue a 'cut slade' job, you could use this:
```
local quickfort = reqscript('quickfort')

function enqueue_job(name)
    local keys = 'jmq' .. name .. '&1&^^'
    local blueprint_data = {[0]={[0]={[0]=keys}}}
    quickfort.apply_blueprint{mode='config', data=blueprint_data}
end

enqueue_job('cut slade')
```

Quickfort could be used to drive the entire UI flow, if you wish.

Specifically for creating orders, you could also use the `orders` plugin to import the order that you need directly (see `test/plugins/orders.lua` for an example -- the code there for saving and restoring existing manager orders might be useful too so you can guarantee that your test won't make lasting changes to an active game)

myk002 (changes requested):
Could you merge latest master so everything is up to date with the latest code?

myk002:
ping on this

TymurGubayev:
about `local quickfort = reqscript('quickfort')` in test: I don't like the idea of requiring another script or plugin just for the test to run. Any chance you could move the functionality to the dfhack-proper?

myk002:
It would be possible to factor some functionality out, but I'm not sure if there's anything shady about using the public API of quickfort to implement a test. It is used by other scripts (e.g. [assign-minecarts](https://github.com/DFHack/scripts/blob/5ae30f01d9d96453d6ef435d018686db4b4a76f7/assign-minecarts.lua#L45)) to perform similar utility functions. Interdependencies among the scripts is starting to become more common.

gui/workorder-finetune
TymurGubayev: so, I started to actually do this and it would be a **huge** refactoring. The problem is one class inherits from `MenuOverlay` while another from `FramedScreen`. We need multi-inheritance or to change both scripts in such a way most of the rendering happens inside a widget, then `JobDetailsBase` can inherit from `Widget`. But this is IMO too much work for little benefit. Should we actually have the class inheritance in place, any change in the base class that's undesired in one of the scripts will still need to be addressed. The only thing we win is obviously marking those functions as `shared` - maybe we could just leave a comment? "Beware: this script shares functions with `worder-details.lua`" or something.

test/gui/workorder-details.lua
```diff
@@ -0,0 +1,70 @@
...
+    material:   any material
+    traits:     allow_artifact
+    ]]
+    expect.eq(-1, item.item_type, "Input item should change to 'any item'")
+    expect.eq(-1, item.mat_type, "Material should change to 'any material'")
+    expect.true_(item.flags2.allow_artifact, "Trait allow_artifact should change to set")
+
+    -- cleanup
+    wait()
+    send_keys('LEAVESCREEN', 'LEAVESCREEN', 'MANAGER_REMOVE')
```
myk002, line 66: This should verify that cleanup was successful
Also, cleanup will fail if the confirm plugin is active

myk002: Option 1 for handling confirm: https://github.com/DFHack/dfhack/blob/fc384fd1a2bee65315fe79831a55ca74c8e6717b/test/quickfort/ecosystem.lua#L411

option 2: https://github.com/DFHack/scripts/blob/4941caccab6dd775d43250e79b8d1a75b8a24b57/assign-minecarts.lua#L43

TymurGubayev: added a check for the cleanup. Also, got the newest version of the `confirm` plugin :)

Is it possible to set the focus string (`dfhack.gui.getCurFocus()`) for the confirm popups? IMHO this would greatly simplify any checks related to it. Or maybe we could get the currently active dfhack view, similar to what `dfhack.gui.getCurViewscreen()` does.

myk002: confirm is "always" on, so it doesn't have an opportunity to set a focus string. It just renders itself over the active screen.

gui/workorder-details.lua
```diff
@@ -0,0 +1,279 @@
...
+    if job_item.has_material_reaction_product == product_materials then
+        job_item.has_material_reaction_product = ''
+    else
+        job_item.has_material_reaction_product = product_materials
+    end
+end
+
+function JobDetails:unsetFlags(obj)
+    local job_item = obj.iobj
+    for flag, ffield in pairs(job_item_flags_map) do
```
myk002, line 218: job_item_flags_map doesn't exist

test/gui/workorder-details.lua
TymurGubayev: well, that's a shame. I added a check if `confirm`-`order-remove` is enabled and send an additional `SELECT` in this case.

gui/workorder-details.lua
TymurGubayev: yeah... I've added a test for this, and will move the default code to the `materials.lua`

changelog.txt
```diff
@@ -15,6 +15,9 @@ that repo.
 
 ## New Scripts
 
+- `gui/workorder-details`: adjusts work orders' input item, material, traits
+- `assign-minecarts`: assign minecarts to hauling routes that don't have one
```
myk002, line 19: the assign-minecarts line doesn't belong here. merge error?

gui/workorder-details.lua
```diff
@@ -0,0 +1,279 @@
...
+
+This is the equivalent of `gui/workshop-job` for work orders,
+with the additional possibility to set input items' traits.
+
+It has to be run from a work order's detail screen
+(:kbd:`j-m`, select work order, :kbd:`d`).
+
+For best experience add the following to your ``dfhack*.init``::
+
+    keybinding add D@workquota_details gui/workorder-details
```
myk002, line 17: We can do this automatically (and it doesn't need to be documented here since it will get picked up when we generate the online docs). could you add the keybinding to dfhack.init-example in the dfhack repo?

gui/workorder-details.lua
myk002: Could you update this PR with the code that depends on DFHack/dfhack#2211? I can pull both PRs and test them together.

test/gui/workorder-details.lua
```diff
@@ -0,0 +1,155 @@
...
+-- handle confirm plugin: we may need to additionally confirm order removal
+local confirm = require 'plugins.confirm'
+local confirmRemove = function() end
+if confirm.isEnabled() then
+    for _, c in pairs(confirm.get_conf_data()) do
+        if c.id == 'order-remove' then
+            if c.enabled then
+                confirmRemove = function()
+                    wait()
+                    -- without delays `confirm` can miss the key event
```
myk002, line 27: are you sure? I have not found this to be the case in the past.

myk002 (approved):

test/gui/workorder-details.lua
TymurGubayev: it happens sporadically, when it does, removing order fails and test doesn't exit to map screen (and test fails at the `Test order should've been removed`-check). With `delay(5)` it happened once the first time I startet DF after PC restart -- I think it has something to do with anti-virus... Without the first `delay(5)` it happens every 3rd time or so.

gui/workorder-details.lua
TymurGubayev: This will be yet another pull-request?

changelog.txt
TymurGubayev: yeah, no idea how it got there.

test/gui/workorder-details.lua
myk002: Let me look into this a little before merging. Otherwise code looks great!

TymurGubayev: Thanks :)

The most important question: could you reproduce the issue?

myk002: yes, reproduced, and I tracked down why. When the SELECT key is sent to the confirm prompt, the intercepted key isn't passed on to the underlying screen until the next rendered frame. Since we escape out of that screen before the next rendered frame, it is never sent -- until the screen is reopened. If you run the test, and it fails, and then you manually open the manager orders screen, you can see the test order there before it quickly disappears. If you pause the confirm dialog and resend MANAGER_REMOVE, it works. I'll put that suggestion in the PR.

test/gui/workorder-details.lua
```diff
@@ -0,0 +1,155 @@
...
+                    delay(5)
+                    send_keys('SELECT')
+                    delay(5)
```
myk002, line 30:
```suggestion
                    send_keys('CUSTOM_P', 'MANAGER_REMOVE')
```

myk002 (approved):

test/gui/workorder-details.lua
TymurGubayev: Is there a way to "wait until the next rendered frame"? It seems possible with `dfhack.timeout`, but probably cumbersome. On a related note, I can't find docs for the `delay` itself...

myk002: Yes, but it's somewhat hacky. You can get how many ticks occur between each frame and wait that many ticks plus one. I'm not sure if this is always consistent, though, since the ticks per graphics frame is variable.

There are no docs for the unit tests, which is a bad thing. You can see the implementation of `delay` here: https://github.com/DFHack/dfhack/blob/0aa7ec877e4201af76b429438037fb5fc8a501c2/ci/test.lua#L82

It gets injected into the global environment for unit tests along with some other utility functions. In fact, you might be able to use `delay_until` to detect when a frame has been rendered.

PR #360, opened 2022-03-28
add options to autonick

Moth-Tolias:
three real changes, whoops.

- first, listens for -h and --help options. help is also displayed if commands are used, as autonick has none.
- second, ~~add a dialogue to confirm that the user actually wants to nickname, suppressable with -f / --force~~ require the use of the command `all` to proceed, eg `autonick all`. together with 1, this resolves https://github.com/DFHack/dfhack/issues/2050.
- third, report the number of dwarves given nicknames and if autonick.txt has ran out of names, suppressable with -q / --quiet.

something that might be nice (but that i haven't done yet) is to raise an error if the user is not in fort mode, since it doesn't really make sense to run it anywhere else. ~~another possible improvement would be to assume -f if no console exists, so existing configs don't break.~~

### DIFF
--- a/autonick.lua
+++ b/autonick.lua
@@ -44,57 +44,58 @@ if #commands ~= 0 then options.help = true end
 
 if options.help == true then
     print(dfhack.script_help())
-else
-    result = options.force;
-    if options.force ~= true then
-        local utils = require("utils")
-        result = utils.prompt_yes_no("nickname unnicked dwarves? (use -f to disable this check)", true)
+    return
+end
+
+result = options.force;
+if options.force ~= true then
+    local utils = require("utils")
+    result = utils.prompt_yes_no("nickname unnicked dwarves? (use -f to disable this check)", true)
+end
+
+if result == true then
+    local seen = {}
+    --check current nicknames
+    for _,unit in ipairs(df.global.world.units.active) do
+        if dfhack.units.isCitizen(unit) and
+        unit.name.nickname ~= "" then
+            seen[unit.name.nickname] = true
+        end
     end
 
-    if result == true then
-        local seen = {}
-        --check current nicknames
-        for _,unit in ipairs(df.global.world.units.active) do
-            if dfhack.units.isCitizen(unit) and
-            unit.name.nickname ~= "" then
-                seen[unit.name.nickname] = true
-            end
+    local names = {}
+    -- grab list, put in array
+    local path = dfhack.getDFPath () .. "/dfhack-config/autonick.txt";
+    for line in io.lines(path) do
+        line = line:trim()
+        if (line ~= "")
+        and (not line:startswith('#'))
+        and (not seen[line]) then
+            table.insert(names, line)
+            seen[line] = true
         end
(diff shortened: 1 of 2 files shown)


autonick.lua
```diff
@@ -14,41 +14,87 @@ If there are fewer available nicknames than dwarves, the remaining
...
-    and (not line:startswith('#'))
-    and (not seen[line]) then
-        table.insert(names, line)
-        seen[line] = true
+-- autonick does not use commands, so if there are any, the script is being used wrong.
+if #commands ~= 0 then options.help = true end
+
+if options.help == true then
+    print(dfhack.script_help())
+else
```
myk002, line 47: You can avoid indenting the main body of the code by returning in the `if` block:
```
if options.help then
    print(dfhack.script_help())
    return
end

-- no else block needed. the remaining code can be unindented
```

autonick.lua
```diff
@@ -14,41 +14,87 @@ If there are fewer available nicknames than dwarves, the remaining
...
+    result = options.force;
+    if options.force ~= true then
+        local utils = require("utils")
+        result = utils.prompt_yes_no("nickname unnicked dwarves? (use -f to disable this check)", true)
```
myk002, line 51: I'm not sure which is preferable, but an alternative is to display help when there are no arguments and to require something (like the word `all`) to autonick all dwarves. [unforbid](https://docs.dfhack.org/en/stable/docs/_auto/base.html?highlight=unforbid#unforbid) takes this approach, for example. displaying a yes/no prompt seems...heavy handed

myk002 (changes requested):

autonick.lua
Moth-Tolias: agreed, that makes much more sense to me.

Moth-Tolias: should be good now.

autonick.lua
```diff
@@ -41,14 +69,24 @@ for line in io.lines(path) do
 end
 
 --assign names
+count = 0
```
myk002, line 72: missing `local`

myk002 (approved):
other than one minor nit, looks good! could you add a changelog entry?

autonick.lua
Moth-Tolias: whoops! good catch

Moth-Tolias:
missed a couple backticks, whoops! hopefully this should be all good now, for real this time

autonick.lua
```diff
@@ -41,14 +69,24 @@ for line in io.lines(path) do
 end
 
 --assign names
+local count = 0
 for _,unit in ipairs(df.global.world.units.active) do
-    if (#names == 0) then break end
+    if (#names == 0) then
+            if options.quiet ~= true then
+                print("no free names left in autonick.txt.")
```
myk002, line 76: Sorry, didn't see this earlier. Few people will know the path to this file. It would be clearer to say "no free names left in dfhack-config/autonick.txt"

myk002 (approved):

PR #364, opened 2022-04-06
pre-commit config

lethosor:
See https://github.com/DFHack/dfhack/issues/2075


.pre-commit-config.yaml
```diff
@@ -0,0 +1,31 @@
+ci:
+  autofix_prs: false
```
myk002, line 2: This means that it will autofix after merge, right? It just won't push changes to the PR (unless asked to with a magic PR comment)?

lethosor: It'll only fix with a magic comment ("pre-commit.ci autofix" per https://pre-commit.ci/#configuration). Otherwise there will be a failing status associated with the PR, which should be addressed before merging.

My rationale is that PR submitters may not have pre-commit set up and may not be expecting changes to be made to their branch by pre-commit.ci right after they submit a PR.

myk002 (approved):

PR #377, opened 2022-04-27
Use new library code in gui scripts

myk002:
Use new sidebar-setting code in `guidm.MenuOverlay`, use the new `guidm.MenuOverlay:renderMapOverlay()` function where convenient, and standardize error message phrasing when the user is on the wrong viewscreen (for those scripts that care).

Updates to:
- devel/block-borders.lua
- devel/unit-path.lua
- gui/blueprint.lua
- gui/guide-path.lua
- gui/liquids.lua
- gui/mass-remove.lua
- gui/mechanisms.lua
- gui/pathable.lua
- gui/power-meter.lua
- gui/room-list.lua
- gui/siege-engine.lua
- gui/stamper.lua
- gui/teleport.lua

### DIFF
--- a/devel/block-borders.lua
+++ b/devel/block-borders.lua
@@ -5,14 +5,13 @@
 devel/block-borders
 ===================
 
-An overlay that draws borders of map blocks. Must be run from the main fortress
-mode screen. See :doc:`/docs/api/Maps` for details on map blocks.
+An overlay that draws borders of map blocks. See :doc:`/docs/api/Maps` for
+details on map blocks.
 
 ]====]
 
 local gui = require "gui"
 local guidm = require "gui.dwarfmode"
-local utils = require "utils"
 
 local ui = df.global.ui
 
@@ -24,10 +23,6 @@ local DRAW_CHARS = {
     se = string.char(218),
     sw = string.char(191),
 }
-local VALID_SIDEBAR_MODES = utils.invert{
-    df.ui_sidebar_mode.Default,
-    df.ui_sidebar_mode.LookAround,
-}
 -- persist across script runs
 color = color or COLOR_LIGHTCYAN
 
@@ -35,6 +30,7 @@ BlockBordersOverlay = defclass(BlockBordersOverlay, guidm.MenuOverlay)
 BlockBordersOverlay.ATTRS{
     block_size = 16,
     draw_borders = true,
+    sidebar_mode=df.ui_sidebar_mode.LookAround,
 }
 
 function BlockBordersOverlay:onInput(keys)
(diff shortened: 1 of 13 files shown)


myk002:
Most of the scripts could also stand to be using the library widgets instead of rendering all text directly to the sidebar, but that would be a much larger modification

gui/blueprint.lua
```diff
@@ -439,36 +449,14 @@ function BlueprintUI:onRenderBody()
...
-            -- don't overwrite the cursor or start_tile so the user can still
-            -- see them
-            if not same_xy(cursor, pos) and not same_xy(start_pos, pos) then
-                local stile = vp:tileToScreen(pos)
-                dc:map(true):seek(stile.x, stile.y):
-                        pen(COLOR_GREEN, COLOR_BLACK):char('X'):map(false)
-            end
-        end
-    end
+    self:renderMapOverlay(get_overlay_char, self:get_bounds())
```
lethosor, line 459: Out of curiosity, did you check if there was any performance impact from switching to this? (In any script, really)

myk002: I tested every script I modified, and I didn't see any change in performance. F.ex. devel/block-borders uses all the new library code and behaves just the same as before.

myk002: Except, of course, that it is now more flexible about what screen is active when the script is run

devel/block-borders.lua
```diff
@@ -24,17 +23,14 @@ local DRAW_CHARS = {
...
-    df.ui_sidebar_mode.LookAround,
-}
 -- persist across script runs
 color = color or COLOR_LIGHTCYAN
 
 BlockBordersOverlay = defclass(BlockBordersOverlay, guidm.MenuOverlay)
 BlockBordersOverlay.ATTRS{
     block_size = 16,
     draw_borders = true,
+    sidebar_mode=df.ui_sidebar_mode.LookAround,
```
lethosor, line 33: There was a subtle behavioral change introduced here: previously, the script could be invoked from either the Default or LookAround modes, and would preserve the mode that was selected before. Now it always enters LookAround (k), and exits back to Default when it's done.

I don't think it really matters, but again, I am curious - is there an existing way to change sidebar modes only if the pre-selected mode isn't one of _multiple_ valid options?

myk002: I'll look into this. The intended behavior is to restore the previous mode if it is supported by `enterSidebarMode()`. It should only be "restoring" Default if the topmost screen is a DFHack window or if the starting mode is unsupported

lethosor: The part I was specifically focusing on was the "always enters LookAround" behavior, not the "always exits back to Default". The latter is actually not correct. Fixed.

PR #378, opened 2022-05-01
respect current cursor mode in block-borders

myk002:
Fixes DFHack/dfhack#2123

now if a cursor is visible when starting `devel/block-borders`, it will start in cursor mode. Otherwise, it will start in cursor-less mode.

This reverts an accidental behavior change introduced in #377


lethosor (approved):
Nice, easier than I expected!

PR #382, opened 2022-05-10
port unforbid from ruby to lua

myk002:
#2081

First script conversion in the long(ish) road to turning down ruby support.


myk002:
Agree. Since it's a `devel/` script, I think we can relax the standards of backwards compatibility and just remove it.

PR #383, opened 2022-05-10
Port multicmd from ruby to lua

myk002:
DFHack/dfhack#2081

The only behavior change is that multiple successive ';' characters are now skipped instead of producing empty commands.

Docs also updated to reflect the fact that you don't need to put extra spaces in around the ';' characters.


myk002:
hrm, I wonder if I should code this to support ignoring semicolons within quotes. for example: `"tiletypes-command f any ; f designated 1 ; p any ; p hidden 0 ; r 23 23 1"; tiletypes-here`

PR #384, opened 2022-05-10
Port deteriorate*.rb to new unified deteriorate.lua script

myk002:
There are three very similar ruby scripts that need to be ported:
- deteriorateclothes.rb
- deterioratecorpses.rb
- deterioratefood.rb

It seems like a good idea to combine their logic. We can do this either by creating three frontends with the existing command names that access a common backend, or just combine them all into a single new `deteriorate` command that takes options. This PR takes the second route because I plan to add more configurability to the commands, and a new command name makes it clear that the API has changed. However, we can change back to having separate scripts for the sake of backwards compatibility without too much trouble.

I tried to keep the lighthearted tone of the help text from the original scripts, but the additional options and syntax does make the script help more formal. This is unfortunate, but maybe acceptable.


myk002:
btw docs are failing until DFHack/dfhack#2135 is merged

myk002:
docs build will fail until paired DFHack/dfhack#2157 is merged

myk002:
ref: DFHack/dfhack#2081

PR #385, opened 2022-05-11
`startdwarf`: migrate from Ruby to Lua

lethosor:
https://github.com/DFHack/dfhack/issues/2081

Requires dfhack/dfhack#2168


test/startdwarf.lua
```diff
@@ -0,0 +1,95 @@
...
+    test_patch_successful(7)
+end
+
+function test.above_7()
+    test_patch_successful(10)
+end
+
+function test.uint8_overflow()
+    test_patch_successful(257)
+end
```
myk002, line 95: what exactly are we testing here? that a number that would overflow an 8-bit int doesn't throw an error? can we even verify that this is written correctly to memory (though I'm sure it is -- we're using a 32 bit int)?

myk002 (changes requested):

startdwarf.lua
```diff
@@ -0,0 +1,32 @@
+-- change number of dwarves on initial embark
+
+--[====[
+
+startdwarf
+==========
+Use at the embark screen to embark with the specified number of dwarves.
```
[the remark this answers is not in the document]
lethosor: It's actually not correct - it'll work on any screen before that screen, including the title screen. Will fix

test/startdwarf.lua
lethosor: I mistakenly used `patchBytes()` earlier, which truncates everything to 8 bits, instead of `patchMemory()`. I think you're correct that the test doesn't really check for that, though, because nothing is calling `patchMemory()` (i.e. [this comment](https://github.com/DFHack/scripts/blob/1331783a18427a9d66d1ef34690b3380f9b7f936/test/startdwarf.lua#L52) is relevant).

myk002 (changes requested):
No changes since the last review? Needs pushing perhaps?

lethosor:
Needs force-pushing, rather, because I rebased :facepalm:

myk002 (approved):

PR #386, opened 2022-05-13
Use new EditField focus semantics

myk002:
DFHack/dfhack#2137
Partner to core library changes in DFHack/dfhack#2147

Usage is updated for `gui/blueprint` and `gui/gm-editor`. Scripts that use an `EditField` without an activation key do not see any changes in behavior.

I made one quality-of-life change in `gui/gm-editor`. Now when typing a filter, hitting enter is passed through to the list so you can interact with the selected item in one keypress instead of two. This matches the behavior of the standard `FilteredList`.


gui/gm-editor.lua
```diff
@@ -452,11 +444,6 @@ function GmEditorUi:onInput(keys)
         return
     end
 
-    if self.subviews.filter_input.active then
-        self.super.onInput(self,keys)
```
lethosor, line 456: One argument against this change: you need to press `s` to activate the search field, so how do you deactivate the search field? Pressing `Esc` deactivates it _and_ resets it to what it was previously

Example 1: pressing `s` `a` `Enter` opens the first child field containing "a", then `Esc` backs out to the parent object. Now how do you clear the search? `s` `Backspace` `Esc` resets the search back to "a" when you hit `Esc`, and `s` `Backspace` `Enter` opens the first child field.

Example 2: how do I use the "`i: find object with this ID`" feature after searching for the field containing the ID I want to look up? Pressing `Enter` opens the selected field, which isn't what I wanted to do, and pressing `Esc` clears the search, which also isn't what I wanted.

In both of those cases, `Enter` `Esc` gets you back to a state where the search field is not selected, but that's an extra keypress relative to the previous implementation, just in a different place.

> This matches the behavior of the standard FilteredList.

I think the key difference with gm-editor is that it has several keybindings that could also be valid text input (`i`, `I`, `f`, `s`, `]`, and even `?`, going by the help screen). Also, the search field is not always active like it is in most FilteredList dialogs. So I think we need a way to "finish" a search without also triggering an action on the selected item.

myk002: Good points. It's not hard to revert the new behavior. I'll fix this up tomorrow.

lethosor: Ah, I see this behavior was just due to the EditField's `on_submit` handler (https://github.com/DFHack/scripts/pull/395/files#diff-6ba1cc209c84dc86e3b98716abf3a9dd5fa24cae9d569a61ca244f09b9611f59L160), so I guess a custom key would require either implementing `on_select2` for EditField or some custom input handler.

Thanks for the change!

PR #390, opened 2022-05-26
Adding Guild Names to output of list-agreements

Hevlikn:
To assist in selecting the correct guild in game, where many- or alike- guilds exist for certain professions/names.

Todo: The same for Deity Names (so as to properly detail Temple Decorations!) but i can't figure out how to parse historical_entity vectors which might yield the info (such as  `df.global.world.entities.all[religion_id].relations.deities[0]` )

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,7 +23,7 @@ that repo.
 
 ## Misc Improvements
 - `deteriorate`: new ``now`` command immediately deteriorates all items of the specified types.
-- `list-agreements`: now displays translated Guild Name 
+- `list-agreements`: now displays translated Guild Name
 
 ## Removed
 - `devel/unforbidall`: please use `unforbid` instead. You can silence the output with ``unforbid all --quiet``
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@ that repo.
 
 ## Misc Improvements
 - `deteriorate`: new ``now`` command immediately deteriorates all items of the specified types.
+- `list-agreements`: now displays translated Guild Name 
 
 ## Removed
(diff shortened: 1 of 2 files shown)


myk002 (approved):
Looks good, though I'd like to actually run it before it gets merged. Could you add a changelog entry so this improvement is advertised when we make the next release? You can add a line to the `changelog.txt` file under the `Future` version (`Misc Improvements` section).

Did you want to address the deity name TODO in this PR or as a separate PR later?

Also, are there similar changes that should be made to the `gui/petitions` script?

Hevlikn:
> Looks good, though I'd like to actually run it before it gets merged. Could you add a changelog entry so this improvement is advertised when we make the next release? You can add a line to the `changelog.txt` file under the `Future` version (`Misc Improvements` section).

I'm not sure what you mean by 'running it'; put in scripts folder, type script name, get output?
 
 
image

 
> Did you want to address the deity name TODO in this PR or as a separate PR later?

Probably post as Issue now or once merged.
 
> Also, are there similar changes that should be made to the `gui/petitions` script?

Seems reasonable, had no idea this existed!

Will add those changes directly

Hevlikn:
> Also, are there similar changes that should be made to the gui/petitions script?

Output is already sufficient imho:

image

Note the translated name of the guildhall is already stated, as the petitioner is the guild. Will have to look at method to see if it's more optimal than the one i used (or if it's the same with more aliasing)

Hevlikn:
Mooted by #393

myk002:
> I'm not sure what you mean by 'running it'

I mean I reviewed the code, but I hadn't actually verified that it runs as expected in a live game.

PR #392, opened 2022-05-29
[gui/quickfort] implement interactive controls for repetitions and transformations

myk002:
Fixes DFHack/dfhack#2018

Depends on DFHack/dfhack#2147

UI for controlling repetitions and transformations in quickfort blueprints.

There is some UI behavior that is unique to this script that could use some feedback about whether it works. The repetitions counter can be modified by multiple hotkeys or edited directly. Is that obvious from the presentation?

Separate hotkey labels for each transformation type crowded the UI a little too much, so I collapsed them into just a generic "Ctrl+Arrow" label. Does that work?

The transformations get "auto-reduced" so the transformation list doesn't get too long. Longer transformation sequences will get automatically replaced with shorter equivalent sequences. Is that confusing?

Finally, the repetition and transformation settings get saved between script invocations. Should it be? The thinking here is that if you are applying a sequence of blueprints, they are likely to all need to be transformed the same way.

Screenshot:
image

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,7 @@ that repo.
 - `deteriorate`: combines, replaces, and extends previous `deteriorateclothes`, `deterioratecorpses`, and `deterioratefood` scripts.
 - `gui/petitions`: shows list of fort's petitions
 - `gui/quantum`: interactive tool for creating quantum stockpiles
+- `gui/quickfort`: interactive previews and transformations of quickfort blueprints
 - `modtools/fire-rate`: allows modders to adjust the rate of fire for ranged attacks
 
 ## Fixes
--- a/gui/quickfort.lua
+++ b/gui/quickfort.lua
@@ -280,9 +280,15 @@ function QuickfortUI:init()
                                      autoarrange_gap=1}
     main_panel:addviews{
         widgets.Label{text='Quickfort'},
-        widgets.WrappedLabel{view_id='summary',
-            text_pen=COLOR_GREY,
-            text_to_wrap=self:callback('get_summary_label')},
+        widgets.ResizingPanel{subviews={
+            widgets.WrappedLabel{view_id='summary',
+                frame={t=0, l=0},
+                text_pen=COLOR_GREY,
+                text_to_wrap=self:callback('get_summary_label')},
+            widgets.HotkeyLabel{view_id='commit_label',
+                frame={t=1, l=13}, key='SELECT', key_sep=' ', label='to apply.',
+                text_pen=COLOR_GREY, on_activate=self:callback('commit')}
+        }},
         widgets.HotkeyLabel{key='CUSTOM_L', label='Load new blueprint',
             on_activate=self:callback('show_dialog')},
         widgets.ResizingPanel{autoarrange_subviews=true, subviews={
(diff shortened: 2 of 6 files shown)


myk002:
needs unit tests.

lethosor:
My only slight concern about Ctrl-Arrow is that it's bound to workspace switching on macOS by default. People can change that and/or DF, but I'm not sure how best to represent that in DF. I think your way is fine.

lethosor:
I guess an alternative could be to have `t` "activate" the transformation options, and when those are active, the normal arrow keys could change what transformation is selected. Not super important now, but if you ever add more options to the UI, you might run out of keys and need to think about making the UI more modal like this.

lethosor:
A couple more things:

* I think maintaining the transformation makes a lot of sense, especially if people are running multiple blueprints in series that come from the same "set". So I agree with your reasoning.
* Simplifying the transformation list also makes sense to me.
* I saw your point about `R`, then didn't really notice it in the UI until I came back and re-read your post again. Maybe a space separating it from the +- controls would help?
* Replacing `ENTER` with a green `Enter` would help it stand out a bit (unrelated to this change, I know)
* Is there no way to repeat along the x/y axes? I was expecting "up" and "down" to mean along the y-axis, but eventually figured out what was going on. I guess repeating along the z-axis is probably more common.
* `Ctrl-Down` always toggling `hflip` feels counterintuitive to me. However, I also like Ctrl-Left and Right always rotating, so I'm not sure what to suggest. I figured it out eventually, but I was scratching my head for a while trying to figure out why I could only `vflip`.

myk002:
> * I saw your point about `R`, then didn't really notice it in the UI until I came back and re-read your post again. Maybe a space separating it from the +- controls would help?

I'll try a space. I originally had a comma to separate it, but since commas are rendered as ground rubble, it didn't really help.

> * Replacing `ENTER` with a green `Enter` would help it stand out a bit (unrelated to this change, I know)

This is a good idea and applicable to `gui/blueprint` as well, but WrappedLabel needs to be extended to support text attributes. (Option 2, I could go back to my pre-autowrapping version of the help text, which was just a list of Labels with text carefully measured on each line to limit length. Option 3, I could create a Label with the word "Enter" and render it over the wrapped text.)

> * Is there no way to repeat along the x/y axes? I was expecting "up" and "down" to mean along the y-axis, but eventually figured out what was going on. I guess repeating along the z-axis is probably more common.

I kept the nomenclature set by Python quickfort: https://github.com/joelpt/quickfort#transformations-simple-repetition
However, looking at the docs I just linked, Joel introduces the up/down pair as up-z and down-z. That might be clearer on the UI instead of Up/Down. I'll have to make the commandline accept that naming as well. Right now it recognizes `up` or `<` as above and `down` or `>` as below.

There is currently no way to repeat North/South/East/West. I considered implementing it, but I didn't see a strong use case. The "repetition patterns" that python quickfort offers is nifty, but seems to me to be too niche. I can, of course, implement a similar feature if our users request it.

> * `Ctrl-Down` always toggling `hflip` feels counterintuitive to me. However, I also like Ctrl-Left and Right always rotating, so I'm not sure what to suggest. I figured it out eventually, but I was scratching my head for a while trying to figure out why I could only `vflip`.

I can add documentation to gui/quickfort's help text so at least it is explained somewhere.

Also, I could potentially draw arrow characters instead of the word "Arrows" for the hotkey hint. Would that make it clearer that ctrl-down is an option that you can try?

myk002:
Updated. Also, here is how it would look with arrow icons. Is this clearer:
image

lethosor:
Forgot to get back to you on this. I like the arrow keys... wonder why DF doesn't do that. Looks good!

myk002:
screenshot for the record: 
image

PR #393, opened 2022-05-29
list-agreements overhaul

Hevlikn:
+ handling of Deity names (as mentioned in DFHack/scripts#390)

+ age of petitions

+ added status 'expired' for outstanding petitions more than 1 year old

+ race specific guild professions

+ formatting characters to reduce mid-wo
rd wrapping

- consolidated functions for output and name lookup

- minimised repeated work within if...elseif... blocks

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@ that repo.
 
 ## Misc Improvements
 - `deteriorate`: new ``now`` command immediately deteriorates all items of the specified types.
+- `list-agreements`: now displays translated Guild Names, worshiped Deity, age of petition and "man" to race-specific professions
 
 ## Removed
 - `devel/unforbidall`: please use `unforbid` instead. You can silence the output with ``unforbid all --quiet``
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,7 +23,7 @@ that repo.
 
 ## Misc Improvements
 - `deteriorate`: new ``now`` command immediately deteriorates all items of the specified types.
-- `list-agreements`: now displays translated Guild Names, worshiped Deity, age of petition and "man" to race-specific professions
+- `list-agreements`: now displays translated Guild Names, worshiped Deity, age of petition, and "man" to race-specific professions (e.g. "Craftsman" -> "Craftsdwarf")
 
 ## Removed
 - `devel/unforbidall`: please use `unforbid` instead. You can silence the output with ``unforbid all --quiet``
--- a/list-agreements.lua
+++ b/list-agreements.lua
@@ -1,27 +1,34 @@
--- Lists outstanding Guildhall and Temple agreements with Org. Names
+-- list location agreements with guilds or religions
 
 --[====[
 
 list-agreements
 ===============
 
-Lists outstanding Guildhall and Temple agreements in fortress mode.
+Lists Guildhall and Temple agreements in fortress mode.
+In addition:
 
-Additionally:
-
-
-* Translated names of the associated Orders and Guilds
-* Worshiped Deitys and Professions respectively
+* Translates names of the associated Guilds and Orders respectively
+* Displays worshiped Professions and Deities respectively
 * Petition age and status satisfied, denied or expired, or blank for outstanding
 
+Usage::
 
+    list-agreements [options]
 
-Arguments
----------------
+Examples:
 
-    all     list all agreements; past and present
+    ``list-agreements``
+        List outstanding, unfullfilled location agreements.
 
-    help    script help
+    ``list-agreements all``
+        Lists all location agreements, whether satisfied, denied, or expired.
+
+Options:
+
+:``all``     list all agreements; past and present
+
+:``help``    shows this help screen
 
 ]====]
 local playerfortid = df.global.ui.site_id -- Player fortress id
@@ -61,7 +68,7 @@ function get_petition_age(agr)
     local agr_year = agr.details[0].year
     local cur_year_tick = df.global.cur_year_tick
     local cur_year = df.global.cur_year
-    local del_year, del_year_tick, del_month, del_day
+    local del_year, del_year_tick
     --delta, check to prevent off by 1 error, not validated
     if cur_year_tick > agr_year_tick then
         del_year = cur_year - agr_year
@@ -71,14 +78,15 @@ function get_petition_age(agr)
         del_year_tick = agr_year_tick - cur_year_tick
     end
     local julian_day = math.floor(del_year_tick / 1200) + 1
-    del_month = math.floor(julian_day / 28)
-    del_day = julian_day % 28
+    local del_month = math.floor(julian_day / 28)
+    local del_day = julian_day % 28
     return {del_year,del_month,del_day}
 end
 
 function get_guildhall_profession(agr)
     local prof = agr.details[0].data.Location.profession
-    local profname = string.lower(df.profession[prof])
+    local profname = 'profession-less'
+    profname = string.lower(df.profession[prof])
     -- *VERY* important code follows
     if string.find(profname, "man") then
         profname = string.gsub(profname,"man",string.lower(dfhack.units.getRaceNameById(df.global.ui.race_id)))
@@ -89,42 +97,43 @@ end
 function get_agr_party_name(agr)
     --assume party 0 is guild/order, 1 is local government as siteid = playerfortid
     local party_id = agr.parties[0].entity_ids[0]
-    local party_name = dfhack.TranslateName(df.global.world.entities.all[party_id].name, true)
+    local party_name = 'An Unknown Entity or Group'
+    party_name = dfhack.TranslateName(df.global.world.entities.all[party_id].name, true)
     return party_name
 end
 
 function get_deity_name(agr)
     local religion_id = agr.details[0].data.Location.deity_data.Religion
-    for _,deity_id in ipairs(df.global.world.entities.all[religion_id].relations.deities) do
-        return dfhack.TranslateName(df.global.world.history.figures[deity_id].name,true)
-    end
+    local deities = df.global.world.entities.all[religion_id].relations.deities
+    if #deities == 0 then return 'An Unknown Deity' end
+    return dfhack.TranslateName(df.global.world.history.figures[deities[0]].name,true)
 end
 
-
---return resolution status
--- 0 no resolution
--- 1 satisfied
--- 2 denied
--- 3 expired
+--get resolution status, and string
 function is_resolved(agr)
-    local resolution = 0
-    
-    if agr.flags.convicted_accepted then    
-        resolution = 1
+    local res = false
+    local res_str = 'outstanding'
+
+    if agr.flags.convicted_accepted then
+        res = true
+        res_str = 'satisfied'
     elseif agr.flags.petition_not_accepted then
-        resolution = 2
+        res = true
+        res_str = 'denied'
     elseif get_petition_age(agr)[1] ~= 0 then
-        resolution = 3
+        res = true
+        res_str = 'expired'
     end
-    return resolution
+    return res,res_str
 end
 
 --universal handler
-function generate_output(agr,culling,loctype)
+function generate_output(agr,loctype)
     local loc_name = get_location_name(agr.details[0].data.Location.tier,loctype)
     local agr_age = get_petition_age(agr)
     local output_str = 'Establish a '.. loc_name..' for "'..get_agr_party_name(agr)
-    
+    local resolved, resolution_string = is_resolved(agr)
+
     if loctype == df.abstract_building_type.TEMPLE then
         output_str = output_str..'" worshiping "'..get_deity_name(agr)..',"'
     elseif loctype == df.abstract_building_type.GUILDHALL then
@@ -133,18 +142,14 @@ function generate_output(agr,culling,loctype)
         print("Agreement with unknown org")
         return
     end
-    
+
     output_str = output_str..'\n\tas agreed on '..get_petition_date(agr)..'. \t'..agr_age[1]..'y, '..agr_age[2]..'m, '..agr_age[3]..'d ago'
-    if culling and (is_resolved(agr) ~= 0) then
-        return nil
-    elseif is_resolved(agr) == 1 then
-        print( output_str.." (satisfied)")
-    elseif is_resolved(agr) == 2 then
-        print(output_str.." (denied)")
-    elseif is_resolved(agr) == 3 then
-        print(output_str.." (expired)")
-    else
+
+    -- can print '(outstanding)' status here, but imho none is cleaner
+    if not resolved then
         print(output_str)
+    else
+        print(output_str..' ('..resolution_string..')')
     end
 end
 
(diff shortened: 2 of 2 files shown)


Hevlikn:
Example output:

image

list-agreements.lua
```diff
@@ -41,19 +40,50 @@ function get_petition_date(agr)
...
+    return party_name
+end
+
+function get_deity_name(agr)
     religion_id = agr.details[0].data.Location.deity_data.Religion
-    religion_name = dfhack.TranslateName(df.global.world.entities.all[religion_id].name, true)
-    return religion_name
+    for _,deity_id in ipairs(df.global.world.entities.all[religion_id].relations.deities) do
+        return dfhack.TranslateName(df.global.world.history.figures[deity_id].name,true)
+    end
```
myk002, line 86: is this written this way because `deities` can be an empty list? The calling code doesn't appear to handle `nil` as a return value, so perhaps this should be written as:
```
local deities = df.global.world.entities.all[religion_id].relations.deities
if #deities == 0 then return '' end
return dfhack.TranslateName(df.global.world.history.figures[deities[0]].name,true)
```

list-agreements.lua
```diff
@@ -66,26 +96,49 @@ function is_denied(agr)
...
-function generate_output_temple(agr)
-    if is_satisfied(agr) == true then
-        print("Establish a "..get_location_name(agr).." for "..get_religion_name(agr)..", as agreed on "..get_petition_date(agr).." (satisfied)")
-    elseif is_denied(agr) == true then
-        print("Establish a "..get_location_name(agr).." for "..get_religion_name(agr)..", as agreed on "..get_petition_date(agr).." (denied)")
+--universal handler
+function generate_output(agr,loctype)
+    local loc_name = get_location_name(agr.details[0].data.Location.tier,loctype)
+    local agr_age = get_petition_age(agr)
+    output_str = "Establish a ".. loc_name.." for \""..get_agr_party_name(agr)
```
myk002, line 112: It's generally cleaner in Lua to use single quotes for string literals. Then you don't have to escape the double quotes for quoted strings in the output. e.g. `'Establish a '..loc_name..'for "'..get_agr_party_name(agr)`

list-agreements.lua
```diff
@@ -66,26 +96,49 @@ function is_denied(agr)
...
+        print(output_str)
     end
 end
 
+---------------------------------------------------------------------------
+-- Main Script operation
+---------------------------------------------------------------------------
+
+local args = {...}
+local cmd = args[1]
```
myk002, line 138: `cmd` doesn't appear to be used anywhere. however, it would be useful if this script supported the following API:
```
list-agreements
    shows open, unfulfilled agreements
list-agreements all
    shows all agreements, past and present
list-agreements help
    shows script help
```

myk002 (changes requested):

list-agreements.lua
Hevlikn: Yup, that was the intention, but i think i neglected to actually do the handler for the arguments - possibly procrastination because i've not done them before 👀

list-agreements.lua
Hevlikn: This was because i was getting `nil` errors when trying to read the value from `.deities[0]` but handled fine when parsing the array. `.deities` is, to my knowledge, always a single element array for religions, so reading `.deities[0]` *should* be viable.

This was the workaround that generated the desired outcome

list-agreements.lua
```diff
@@ -41,19 +40,50 @@ function get_petition_date(agr)
...
+    agr_year_tick = agr.details[0].year_tick
+    agr_year = agr.details[0].year
+    cur_year_tick = df.global.cur_year_tick
+    cur_year = df.global.cur_year
+    --delta, check to prevent off by 1 error, not validated
+    if cur_year_tick > agr_year_tick then
+        del_year = cur_year - agr_year
+        del_year_tick = cur_year_tick - agr_year_tick
+    else
+        del_year = cur_year - agr_year - 1
```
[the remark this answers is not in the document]
Hevlikn: That's a fair point; i was trying to avoid -ve month/day values in the delta

My understanding is that `_year_tick` resets on year rollover (otherwise it would be more suitably named, such as `world_tick`) and thus if the day-of-year of the current year is before the yearly anniversary of the agreement, the delta would be negative.

However, this is of course mooted if `_year_tick` does not reset

list-agreements.lua
```diff
@@ -41,19 +40,50 @@ function get_petition_date(agr)
     julian_day = math.floor(agr_year_tick / 1200) + 1
     agr_month = math.floor(julian_day / 28) + 1
     agr_day = julian_day % 28
-    return agr_year.."-"..agr_month.."-"..agr_day
+    return string.format("%03d-%02d-%02d",agr_year, agr_month, agr_day)
+end
+
+function get_petition_age(agr)
```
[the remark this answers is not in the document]
Hevlikn: the latter

list-agreements.lua
myk002: The [argparse module](https://docs.dfhack.org/en/stable/docs/Lua%20API.html#argparse) is available if you want to do anything fancy ([example](https://github.com/DFHack/scripts/blob/master/prioritize.lua#L532)), but I think you could get by with an `action_switch` like [this](https://github.com/DFHack/scripts/blob/e4b51365c117f09c3f72fcd16a8ca59ea5c3e5f0/autounsuspend.lua#L29).

changelog.txt
```diff
@@ -23,6 +23,7 @@ that repo.
 
 ## Misc Improvements
 - `deteriorate`: new ``now`` command immediately deteriorates all items of the specified types.
+- `list-agreements`: now displays translated Guild Names, worshiped Deity, age of petition and "man" to race-specific professions
```
myk002, line 26:
```suggestion
- `list-agreements`: now displays translated Guild Names, worshiped Deity, age of petition, and "man" to race-specific professions (e.g. "Craftsman" -> "Craftsdwarf")
```

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
--- Lists all guildhall and temple agreements
+-- Lists outstanding Guildhall and Temple agreements with Org. Names
 
 --[====[
 
 list-agreements
 ===============
 
-Lists all guildhall and temple agreements in fortress mode.
+Lists outstanding Guildhall and Temple agreements in fortress mode.
```
myk002, line 8: you can omit the word "outstanding" here -- it's just one mode of several now.

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
+
+
+
+Arguments
+---------------
+
+    all     list all agreements; past and present
 
+    help    script help
+
```
myk002, line 25: Maybe add a line that explains what happens when run without arguments.

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
--- Lists all guildhall and temple agreements
+-- Lists outstanding Guildhall and Temple agreements with Org. Names
 
 --[====[
 
 list-agreements
```
myk002, line 5: Check out a few other scripts to see how these help blurbs are usually formatted. Here are some relevant examples: https://github.com/DFHack/scripts/blob/master/build-now.lua and https://github.com/DFHack/scripts/blob/master/prioritize.lua

list-agreements.lua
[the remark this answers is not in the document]
myk002: ah, ok, that makes sense. Yes, I was looking at this as if `_year_tick` did not reset, but I think you're right that it does.

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
-    julian_day = math.floor(agr_year_tick / 1200) + 1
-    agr_month = math.floor(julian_day / 28) + 1
-    agr_day = julian_day % 28
-    return agr_year.."-"..agr_month.."-"..agr_day
+function get_petition_age(agr)
+    local agr_year_tick = agr.details[0].year_tick
+    local agr_year = agr.details[0].year
+    local cur_year_tick = df.global.cur_year_tick
+    local cur_year = df.global.cur_year
+    local del_year, del_year_tick, del_month, del_day
```
myk002, line 64: since `del_month` and `del_day` aren't used until after the if blocks, I'd suggest just declaring them `local` below and not here.

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
+    end
     return profname:gsub("_", " ")
 end
 
-function get_religion_name(agr)
-    religion_id = agr.details[0].data.Location.deity_data.Religion
-    religion_name = dfhack.TranslateName(df.global.world.entities.all[religion_id].name, true)
-    return religion_name
+function get_agr_party_name(agr)
+    --assume party 0 is guild/order, 1 is local government as siteid = playerfortid
```
myk002, line 90: is this an assumption or is it documented somewhere?

list-agreements.lua
myk002: That seems.. strange. I'll have to look into the structures involved here.
btw, if the array *is* empty, then the calling code will error out when this function returns nil. maybe add a `return 'Unknown deity'` after the loop?

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
 
-function is_denied(agr)
-    denied = agr.flags.petition_not_accepted
-    return denied
+
+--return resolution status
+-- 0 no resolution
+-- 1 satisfied
+-- 2 denied
+-- 3 expired
```
myk002, line 108: instead of returning an arbitrary numeric code, how about returning a boolean and a descriptive string that can be used directly? That way you can call it as
`if is_resolved(arg) then`
or
`local resolved, resolved_str = is_resolved(arg)`
which might simplify the calling code below

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
+    return resolution
 end
 
-function generate_output_guild(agr)
-    if is_satisfied(agr) == true then
-        print("Establish a "..get_location_name(agr).." for the "..get_guildhall_profession(agr).." guild, as agreed on "..get_petition_date(agr).." (satisfied)")
-    elseif is_denied(agr) == true then
-        print("Establish a "..get_location_name(agr).." for the "..get_guildhall_profession(agr).." guild, as agreed on "..get_petition_date(agr).." (denied)")
+--universal handler
+function generate_output(agr,culling,loctype)
```
myk002, line 123: isn't culling already handled in the main script code at the bottom? does it still need to be passed in here?

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
+-- args handler
+if cmd then
+    if cmd == "all" then
+        cull_resolved = false
+    elseif cmd == "help" then
+        print("list-argeements: \n\t Lists outstanding Guildhall and Temple agreements in fortress mode.")
+        print("")
+        print("Supported arguments:")
+        print("\t all \n\t\t shows all agreements, past and present")
+        print("\t help \n\t\t shows this help")
```
myk002, line 168: You can add all help text to the block at the top and then just call `print(dfhack.script_help())` here

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
+        return
+    else 
+        print("use list-agreements help for supported arguments")
```
myk002, line 171: most scripts just print help on unrecognized commandline params

list-agreements.lua
```diff
@@ -1,120 +1,218 @@
...
 
 for _, agr in pairs(df.agreement.get_vector()) do
     if agr.details[0].data.Location.site == playerfortid then
-        if get_location_type(agr) == df.abstract_building_type.TEMPLE then
-            table.insert(templeagreements, agr)
-        elseif get_location_type(agr) == df.abstract_building_type.GUILDHALL then
-            table.insert(guildhallagreements, agr)
-        end
+		if is_resolved(agr) ~= 0 and cull_resolved then
+		else
```
myk002, line 179:
```suggestion
        if not cull_resolved or not is_resolved(agr) then
```

myk002 (changes requested):

list-agreements.lua
Hevlikn: inference based on documentation and preceding code

relevant structure definition: [df-structures/df.entities L1471](https://github.com/DFHack/df-structures/blob/0fee473dbca37e6d3fd99d2e7447d68418d1158f/df.entities.xml#L1471) 

we've already filtered down to only this sub-type of agreement (location), and filtered out agreements made at other world sites (line 177: `if agr.details[0].data.Location.site == playerfortid then`); so therefore, the 'government' `party[1]` as documented, MUST be the player, and applicant `party[0]`, it being a location agreement will be either a guild or religion (not withstanding toady introducing theatres for performance troupes or other features)

I can see a scenario in which these could appear in arbitrary order, which would be easy enough to handle by testing `for _,party_id in pairs...` `is party_id = playergovernment_id` but i think my logic above is satisfactory as the arrays are strictly typed (unlike, for example JSON in which headers can appear in any order)

list-agreements.lua
Hevlikn: ah is `dfhack.script_help()` a custom function that prints comments between the `[====[`  `]====]` braces?

list-agreements.lua
Hevlikn: [i believe `not 0` returns false, not true as desired](https://www.lua.org/pil/3.3.html)

list-agreements.lua
```diff
@@ -1,120 +1,219 @@
...
-function get_religion_name(agr)
-    religion_id = agr.details[0].data.Location.deity_data.Religion
-    religion_name = dfhack.TranslateName(df.global.world.entities.all[religion_id].name, true)
-    return religion_name
+function get_agr_party_name(agr)
+    --assume party 0 is guild/order, 1 is local government as siteid = playerfortid
+    local party_id = agr.parties[0].entity_ids[0]
+    local party_name = 'An Unknown Entity or Group'
+    party_name = dfhack.TranslateName(df.global.world.entities.all[party_id].name, true)
+    return party_name
```
Hevlikn, line 102: not sure this does what i want it to: return the default if dfhack.TranslateName fails

flip them and `if not party_name then <default>` ?

myk002: a common idiom in Lua is `x = a or b` to mean
```
x = a
if not x then
    x = b
end
```

so in this case you could write `return dfhack.TranslateName() or 'An Unknown Entity or Group'` (assuming `TranslateName` returns nil or false on failure).

myk002 (approved):

list-agreements.lua
```diff
@@ -1,120 +1,222 @@
...
+        output_str = output_str..'" worshiping "'..get_deity_name(agr)..',"'
+    elseif loctype == df.abstract_building_type.GUILDHALL then
+        output_str = output_str..'", a '..get_guildhall_profession(agr)..' guild,'
     else
-        print("Establish a "..get_location_name(agr).." for the "..get_guildhall_profession(agr).." guild, as agreed on "..get_petition_date(agr))
+        print("Agreement with unknown org")
+        return
+    end
+
+    output_str = output_str..'\n\tas agreed on '..get_petition_date(agr)..'. \t'..agr_age[1]..'y, '..agr_age[2]..'m, '..agr_age[3]..'d ago'
```
Hevlikn, line 149: `\t` appears to be inserting ascii-symbol
image

Hevlikn: Resolved in #640

PR #394, opened 2022-06-01
Fix for build-now walls flashing red

wolfboyft:
https://github.com/DFHack/dfhack/issues/2174

### DIFF
--- a/build-now.lua
+++ b/build-now.lua
@@ -454,7 +454,7 @@ local function build_construction(bld)
     local tileattrs = df.tiletype.attrs[tiletype]
     if tileattrs.material == df.tiletype_material.CONSTRUCTION then
         -- modify the construction to the new type
-        local construction, found = utils.binsearch(df.global.world.constructions, pos, "pos", pos_cmp)
+        local construction, found = utils.binsearch(df.global.world.constructions, pos, 'pos', pos_cmp)
         if not found then
             error("Error: could not find construction entry for construction tile at " .. pos.x .. ", " .. pos.y .. ", " .. pos.z)
         end
--- a/build-now.lua
+++ b/build-now.lua
@@ -463,10 +463,10 @@ local function build_construction(bld)
         -- add entries to df.global.world.constructions and adjust tiletypes for
         -- the construction itself
         create_and_link_construction(pos, item, false)
-        set_tiletype(pos, const_to_tile[construction_type])
-        if construction_type == df.construction_type.Wall then
-            dig_now.link_adjacent_smooth_walls(pos)
-        end
+    end
+    set_tiletype(pos, const_to_tile[construction_type])
+    if construction_type == df.construction_type.Wall then
+        dig_now.link_adjacent_smooth_walls(pos)
     end
 
     -- for walls and ramps with empty space above, adjust the tile above
--- a/build-now.lua
+++ b/build-now.lua
@@ -456,7 +456,7 @@ local function build_construction(bld)
         -- modify the construction to the new type
         local construction, found = utils.binsearch(df.global.world.constructions, pos, 'pos', pos_cmp)
         if not found then
-            error("Error: could not find construction entry for construction tile at " .. pos.x .. ", " .. pos.y .. ", " .. pos.z)
+            error('could not find construction entry for construction tile at ' .. pos.x .. ', ' .. pos.y .. ', ' .. pos.z)
         end
         reuse_construction(construction, item)
     else
--- a/build-now.lua
+++ b/build-now.lua
@@ -359,7 +359,12 @@ local function pos_cmp(a, b)
     return utils.compare(a.z, b.z)
 end
 
-local function create_and_link_construction(pos, item, top_of_wall, use_open_space_for_original_tile)
+local function get_original_tile(pos)
+    -- TODO
+    return dfhack.maps.getTileType(pos)
+end
+
+local function create_and_link_construction(pos, item, top_of_wall)
     local construction = df.construction:new()
     utils.assign(construction.pos, pos)
     construction.item_type = item:getType()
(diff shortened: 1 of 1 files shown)


build-now.lua
```diff
@@ -359,16 +359,16 @@ local function pos_cmp(a, b)
...
     local construction = df.construction:new()
     utils.assign(construction.pos, pos)
     construction.item_type = item:getType()
     construction.item_subtype = item:getSubtype()
     construction.mat_type = item:getMaterial()
     construction.mat_index = item:getMaterialIndex()
     construction.flags.top_of_wall = top_of_wall
-    construction.flags.no_build_item = not top_of_wall
-    construction.original_tile = dfhack.maps.getTileType(pos)
+    construction.flags.no_build_item = false
```
lethosor, line 370: We should preserve `no_build_item = true` for most constructions as it improves framerate (see `cleanconst`)

myk002: but only for "real" constructions, not top_of_wall pseudo-constructions

lethosor: Right. So should this line be unchanged?

build-now.lua
```diff
@@ -435,9 +435,18 @@ local function build_construction(bld)
     local construction_type = bld.type
     dfhack.buildings.deconstruct(bld)
 
+    -- check for and remove a construction on the tile already (i.e. building a construction on top of a wall)
+    local use_open_space_for_original_tile = true
+    for i, construction in ipairs(df.global.world.constructions) do
+        if pos.x == construction.pos.x and pos.y == construction.pos.y and pos.z == construction.pos.z then
+            df.global.world.constructions:erase(i)
```
myk002, line 442: instead of erasing the item, we could instead call `utils.binsearch` in `create_and_link_construction` so we can reuse the entry.

build-now.lua
```diff
@@ -435,9 +435,18 @@ local function build_construction(bld)
     local construction_type = bld.type
     dfhack.buildings.deconstruct(bld)
 
+    -- check for and remove a construction on the tile already (i.e. building a construction on top of a wall)
+    local use_open_space_for_original_tile = true
```
myk002, line 439: this should only be true if the original tile is a construction

myk002 (changes requested):

build-now.lua
myk002: I think so. `no_build_item` should always be set to false by `build-now`.

wolfboyft: Wait, but code is for all constructions.

build-now.lua
wolfboyft: We should give the function two modes, then. Or maybe use two functions. If we give it two modes, what would the changed parameters be called?

build-now.lua
wolfboyft: Also, reverting it to construction.flags.no_build_item = not top_of_wall as it was seems to fix the problem wtih not getting build items back when deconstructing a build-now wall...

build-now.lua
wolfboyft: Oops. I meant to initialise that as false and set it to true in the loop.

build-now.lua
myk002: How about if we keep the signature of create_and_link_construction add it is (i.e. don't pass in the new bool) and instead rewrite create_and_link_construction like this:
1) check tile at pos to see if it's a construction
2) if so, use utils.binsearch to get a reference to the construction and overwrite it with our new configuration. The original tile value should already be set correctly.
3) if not, allocate a new construction like we do now.
4) write a new function called get_original_tile that does the check for what original tile should be for newly allocated constructions and call it from create_and_link_construction.

wolfboyft: scatterbrain programming time. this is an optimisation but people may be doing it on huge areas so it might be worth it

build-now.lua
```diff
@@ -435,9 +449,21 @@ local function build_construction(bld)
...
 
-    -- add entries to df.global.world.constructions and adjust tiletypes for
-    -- the construction itself
-    create_and_link_construction(pos, item, false)
+    -- check if we're building on a construction (i.e. building a construction on top of a wall)
+    local tiletype = dfhack.maps.getTileType(pos)
+    local tileattrs = df.tiletype.attrs[tiletype]
+    if tileattrs.material == df.tiletype_material.CONSTRUCTION then
+        -- modify the construction to the new type
+        local construction, found = utils.binsearch(df.global.world.constructions, pos, "pos", pos_cmp)
```
myk002, line 457:
```suggestion
        local construction, found = utils.binsearch(df.global.world.constructions, pos, 'pos', pos_cmp)
```

build-now.lua
```diff
@@ -435,9 +449,21 @@ local function build_construction(bld)
...
+        if not found then
+            error("Error: could not find construction entry for construction tile at " .. pos.x .. ", " .. pos.y .. ", " .. pos.z)
```
myk002, line 459:
```suggestion
            error('could not find construction entry for construction tile at ' .. pos.x .. ', ' .. pos.y .. ', ' .. pos.z)
```

build-now.lua
```diff
@@ -435,9 +449,21 @@ local function build_construction(bld)
...
+        end
+        reuse_construction(construction, item)
+    else
+        -- add entries to df.global.world.constructions and adjust tiletypes for
+        -- the construction itself
```
myk002, line 464: the next line doesn't actually set the tiletype. is this an old comment?

myk002 (approved):

build-now.lua
wolfboyft: Yep oops

myk002 (approved):

lethosor:
Did this make it into a changelog?

myk002:
I have a changelog entry for it on my next pr

PR #398, opened 2022-06-06
Various improvements in petitions.lua

TymurGubayev:
Changes:

* determine names of historical figures in petitions
* hide petitions older than a year by default (AFAIK that's the only way to determine a petition expired)
* correctly build the date diff string for intervals greater than a year
* show type of Residency petition
* make the window bigger

(this addresses most of [#dfhack/2171](https://github.com/DFHack/dfhack/issues/2171) )

### DIFF
--- a/gui/petitions.lua
+++ b/gui/petitions.lua
@@ -81,6 +81,9 @@ end
 function Time:getMonths() -- >>int<< Months as age (not including years)
  return math.floor (self.ticks / TU_PER_MONTH)
 end
+function Time:getYears() -- >>int<<
+ return self.year
+end
 function Time:getMonthStr() -- Month as date
  return MONTHS[self:getMonths()+1] or 'error'
 end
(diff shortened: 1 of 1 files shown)


gui/petitions.lua
```diff
@@ -132,6 +135,17 @@ local function getAgreementDetails(a)
...
         end
+        for _, hf_id in ipairs(p.histfig_ids) do
+            local hf = df.global.world.history.figures[hf_id]
+            local race = df.creature_raw.find(hf.race)
+            local civ = df.historical_entity.find(hf.civ_id)
+            e_descr[#e_descr+1] = table.concat{"The ", race.creature_id, " ", df.profession[hf.profession],
+                                " ", dfhack.TranslateName(hf.name, true),
+                                NEWLINE,
+                                "of ", dfhack.TranslateName( civ.name , true )
+                            }
```
myk002, line 146:
```suggestion
            e_descr[#e_descr+1] = ('The %s %s %s\nof %s'):format(race.creature_id,
                            df.profession[hf.profession],
                            dfhack.TranslateName(hf.name, true),
                            dfhack.TranslateName(civ.name, true))
```

myk002 (approved):
overall, looks good. could you add a changelog entry?

gui/petitions.lua
TymurGubayev: I fixed the whitespaces and changed the formatting of this code block somewhat - I'm still not sure what form is more readable, `table.concat` or `string.format`.

There is no reason to change the changelog, I think -- `gui/petitions` is in the `New Scripts` block.

myk002: :+1:

PR #400, opened 2022-06-07
Make gui/dfstatus not count items owned by traders

wolfboyft:
Also made the flags check use a more compact style which comes with a micro-optimisation


myk002 (approved):
could you add a changelog entry?

wolfboyft:
Oops, of course

wolfboyft:
Was that right?

lethosor (approved):
Looks good!

PR #401, opened 2022-06-08
Add pop-control and max-wave

wolfboyft:
With this you would be able to set (and re-set) population caps, migrants per wave and whether to use hermit persistently per fortress.


wolfboyft:
Concerned about whether this will play nice with settings manager. Maybe settings manager itself should get the ones used in the settings on init (and make that publically accessible) and pop-control should get info from there.

myk002:
I'm happy to merge `max-wave`, but I need to think some more about `pop-control`. Could you split this into two PRs?

max-wave.lua
```diff
@@ -0,0 +1,51 @@
+--Dynamically limit the next immigration wave
+--By Loci, modified by Fleeting Frames
+--[====[
+
+max-wave
+========
+Set the population cap to the lesser of current_pop + wave_size or max_pop.
+Use with the `repeat` command to set a rolling immigration limit.
+
```
myk002, line 9: you should make it clear that this script *overrides* the maximum population set in `data/init/d_init.txt`

max-wave.lua
```diff
@@ -0,0 +1,51 @@
...
+   current_pop = current_pop + 1
+  end
+ end
+
+local new_limit = current_pop + wave_size
+
+if max_pop and new_limit > max_pop then new_limit = max_pop end
+
+df.global.d_init.population_cap = new_limit
+print('max-wave: Population cap set to '.. new_limit)
```
myk002, line 51: if the limit hasn't changed, perhaps instead say that the population cap has already been reached.

myk002 (changes requested):

wolfboyft:
OK, I'll make the changes you suggested and make two new branches and all that in good time.

wolfboyft:
Just gonna do the stuff you said on this branch in this PR first for GUI convenience

max-wave.lua
```diff
@@ -0,0 +1,51 @@
...
+  return
+end
+
+--One would think the game would track this value somewhere...
+for k,v in ipairs(df.global.world.units.active) do
+  if dfhack.units.isCitizen(v) or
+  (dfhack.units.isOwnCiv(v) and
+   dfhack.units.isAlive(v) and
+   df.global.world.raws.creatures.all[v.race].caste[v.caste].flags.CAN_LEARN and
+   not (dfhack.units.isMerchant(v) or dfhack.units.isForest(v) or v.flags1.diplomat or v.flags2.visitor)
```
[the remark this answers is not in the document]
wolfboyft: Well I didn't write this but I'll do my best

wolfboyft:
Might as well defer them together tbh, as discussed on discord

changelog.txt
```diff
@@ -15,6 +15,10 @@ that repo.
 
 ## New Scripts
 
+- `max-wave`: dynamically limit the next immigration wave, can be set to repeat
+- `pop-control`: persistent per fortress population cap, `hermit`, and `max-wave` management
+- `assign-minecarts`: assign minecarts to hauling routes that don't have one
```
myk002, line 20: wait, didn't we fix this already? the `assign-minecarts` line is from the previous release.

max-wave.lua
```diff
@@ -0,0 +1,59 @@
+--Dynamically limit the next immigration wave
+--By Loci, modified by Fleeting Frames and Tachytaenius
+--[====[
+
+max-wave
+========
+Limit the number of migrants that can arrive in the next wave by overriding the population cap value in data/init/d_init.txt (not safe with gui/settings-manager)
```
myk002, line 7: Perhaps explain (and correct me if I'm wrong) that the danger is that gui/settings-manager will write this "temporary" population cap to the d_init.txt file and make it permanent. But that will only happen if the user changes the pop cap in the gui/settings-manager UI, right?

max-wave.lua
```diff
@@ -0,0 +1,59 @@
...
+]====]
+
+local args = {...}
+
+local wave_size = tonumber(args[1])
+local max_pop = tonumber(args[2])
+local current_pop = 0
+
+if not wave_size then
+  qerror('max-wave: wave_size required')
```
myk002, line 32: maybe put a `print(dfhack.script_help())` before this line so users know what wave_size is.

pop-control.lua
```diff
@@ -0,0 +1,124 @@
+-- Script to control the various population caps as well as use of max-wave and hermit persistently per fortress
```
myk002, line 1: This is too long for `ls`. How about just "control the various population caps"?

pop-control.lua
```diff
@@ -0,0 +1,124 @@
...
+===========
+Controls the various population caps as well as use of max-wave and hermit persistently per fortress
+Intended to be placed within ``onMapLoad.init`` as ``pop-control on-load``
+Available arguments:
+
+- ``on-load`` automatically checks for settings for this site and prompts them to be entered if not present
+
+- ``reenter-settings`` lets you revise settings for this site
+
+- ``view-settings`` shows you the current settings for this site
```
myk002, line 14: This is quite different from other scripts. It might fit in better if it had this syntax:
no params - show current settings
`set <num> <num> <num>` - set settings
`start` - start the script
`stop` - stop the script (though I don't know if you can really stop this script)

myk002 (changes requested):

max-wave.lua
wolfboyft: Yes.

pop-control.lua
wolfboyft: I don't think there is any sensible definition of stop, but migrating away from UI makes sense.

changelog.txt
wolfboyft: I might not be merging as well as I thought.

pop-control.lua
wolfboyft: Except, how do we get a prompt on new fort?

wolfboyft: Resolving as "fine for now" as per discord

pop-control.lua
```diff
@@ -0,0 +1,127 @@
...
+-- by Tachytaenius
+--[====[
+pop-control
+===========
+Controls hermit and the various population caps per-fortress.
+Not safe with gui/settings-manager as it will commit temporary
+population cap changes permanently to file.
+Intended to be placed within ``onMapLoad.init`` as ``pop-control on-load``.
+Available arguments:
+
```
myk002, line 11: shouldn't there be a way to initialize this script with the settings that you want?

myk002 (changes requested):

pop-control.lua
wolfboyft: Like, apply defaults and skip the GUI? OK, makes sense.

max-wave.lua
```diff
@@ -0,0 +1,63 @@
+--Dynamically limit the next immigration wave
+--By Loci, modified by Fleeting Frames and Tachytaenius
+--[====[
+
+max-wave
+========
+Limit the number of migrants that can arrive in the next wave by
+overriding the population cap value in data/init/d_init.txt.
+Not safe with gui/settings-manager as it will commit temporary
+population cap changes permanently to file.
```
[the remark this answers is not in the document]
wolfboyft: How do we get from disk though? Lua file IO?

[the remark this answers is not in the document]
myk002: Yeah, it's reading the file anyway when it updates the settings. It can pre-read to find discrepancies. This can actually work for all settings, not just population cap

max-wave.lua
```diff
@@ -0,0 +1,63 @@
+--Dynamically limit the next immigration wave
+--By Loci, modified by Fleeting Frames and Tachytaenius
+--[====[
```
myk002, line 3: some things have changed since this PR started. Docs now go in the `docs/` folder. You can copy an existing file in there to get the current formats, or read the new documentation guidance here: https://docs.dfhack.org/en/latest/docs/Documentation.html

max-wave.lua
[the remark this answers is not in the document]
myk002: after discussing on discord, I think we're wrong here about the danger. `gui/settings-manager` will display the value in the `d_init.txt` file. it will overwrite the pop cap that this script sets, not the other way around. We can just add to this tool's documentation that if you edit the population cap with `gui/settings-manager` after running max-wave, your population cap will be reset and you may get more migrants than you expect. We should similarly add a line to `gui/settings-manager`'s docs (in this PR) that editing the population cap will override any modifications made by scripts such as `max-wave`.

max-wave.lua
```diff
@@ -0,0 +1,63 @@
...
+  qerror('max-wave: wave_size required')
+end
+
+local function isCitizen(unit)
+  return dfhack.units.isCitizen(unit) or
+  (dfhack.units.isOwnCiv(unit) and
+   dfhack.units.isAlive(unit) and
+   df.global.world.raws.creatures.all[unit.race].caste[unit.caste].flags.CAN_LEARN and
+   not (dfhack.units.isMerchant(unit) or dfhack.units.isForest(unit) or unit.flags1.diplomat or unit.flags2.visitor))
+ end
```
myk002, line 45: I think nowadays we have `dfhack.units.isCitizen()`, but check the implementation to make sure it's exactly what you want.

pop-control.lua
```diff
@@ -0,0 +1,127 @@
+-- Script to control the various population caps as well as use of max-wave and hermit persistently per fortress
```
myk002, line 1: going back and reading this script, I realize now how much influence your implementation here had on my thoughts for the config-less DFHack Usability Overhaul track. The repository of canned configurations for other tools, the popup at fort start...it's all here.

https://github.com/DFHack/dfhack/issues/2225

There are issues that my subsequent ruminations turned up, like how does one automatically re-enable a script that has been "persistently" enabled? we don't currently have a mechanism for that. Even this script needs something else to call `on-load`.

But anyway, those issues will get ironed out in the design doc for the next DFHack version, and it doesn't really reflect on what should be in this PR.

max-wave.lua
wolfboyft: Let me know if the docs I pushed are OK.

myk002 (approved):

PR #403, opened 2022-06-10
[gui/quickfort] Don't call chopping air an error

myk002:
Fixes DFHack/dfhack#2183

Don't display chopping designations as an error if there is nothing to chop. It's just a noop. Gathering too.


myk002:
no changelog entry since gui/quickfort is in the "new scripts" section (just because I forgot to add it to the previous release's changelog, but still)

PR #404, opened 2022-06-10
Implemented new 'unload' command in caravan script.

CarabusX:
Helps in cases of endless unloading at the depot.

### DIFF
--- a/caravan.lua
+++ b/caravan.lua
@@ -20,6 +20,8 @@ specified, then the commands apply to all caravans on the map.
   annoying merchants, etc.). Also causes caravans to return to the depot if
   applicable.
 - ``leave [IDS]``: makes caravans pack up and leave immediately.
+- ``rejoin_pack_animals``: reconnects merchant pack animals that were
+  disconnected from their owners (fixes endless unloading at the depot).
 
 ]====]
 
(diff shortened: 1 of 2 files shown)


caravan.lua
```diff
@@ -95,6 +97,45 @@ function commands.leave(...)
...
+end
+
+local function getPrintableUnitName(unit)
+    local visible_name = dfhack.units.getVisibleName(unit)
+    local profession_name = dfhack.units.getProfessionName(unit)
+    if visible_name.has_name then
+        return dfhack.TranslateName( visible_name ) .. ' (' .. profession_name .. ')'
+    else
+        return profession_name  -- for unnamed animals
+    end
```
myk002, line 117:
```suggestion
    if visible_name.has_name then
        return ('%s (%s)'):format(dfhack.TranslateName(visible_name), profession_name)
    end
    return profession_name  -- for unnamed animals
```

myk002 (changes requested):
Code looks good! "rejoin_pack_animals" is kind of a long name, but it is specific and understandable. Calling it just "fix" is an option, and is "forward compatible" in case there are more things we can check for and fix in the future. @lethosor, what do you think?

Could you add a changelog entry in `changelog.txt` that describes this change? It should go in the "Misc Improvements" section of the "Future" release.

myk002:
Also, please consider adding your name to the https://github.com/DFHack/dfhack/blob/develop/docs/Authors.rst file in the DFHack/dfhack repo.

CarabusX:
> Code looks good! "rejoin_pack_animals" is kind of a long name, but it is specific and understandable. Calling it just "fix" is an option, and is "forward compatible" in case there are more things we can check for and fix in the future. @lethosor, what do you think?

Yeah name is kinda long, thats why I called it "rejoin" and not "reconnect".
In case there are more things to fix in the future, it may be good to have separate commands for that, so maybe better if current command is specific?

> Could you add a changelog entry in `changelog.txt` that describes this change? It should go in the "Misc Improvements" section of the "Future" release.

Sure! I will add it there.

CarabusX:
> Also, please consider adding your name to the https://github.com/DFHack/dfhack/blob/develop/docs/Authors.rst file in the DFHack/dfhack repo.

Sure! I will add myself there later.

caravan.lua
CarabusX: I don't like this change as it would print " (Horse)", while I want to print units as they appear in unit list.

CarabusX: I will test it and commit.

myk002 (approved):

lethosor (commented):
I agree with @myk002 that the name is kind of long. I think it's also more specific than it needs to be. From what I understand, it took you some time to figure out that your caravan was stuck because the pack animals were disconnected. I don't think we can expect users to figure that out. From their perspective, the caravan is just stuck, and they don't know why, so I think `unstick` or something similar would be better (and shorter).

Now, if we want to maintain the ability to apply specific, targeted fixes in the future if `unstick` does more than one thing, I would be in favor of making `unstick()` call out to `rejoin_pack_animals()` and whatever else is necessary to get a caravan unstuck.

CarabusX:
> I agree with @myk002 that the name is kind of long. I think it's also more specific than it needs to be. From what I understand, it took you some time to figure out that your caravan was stuck because the pack animals were disconnected. I don't think we can expect users to figure that out. From their perspective, the caravan is just stuck, and they don't know why, so I think `unstick` or something similar would be better (and shorter).
> 
> Now, if we want to maintain the ability to apply specific, targeted fixes in the future if `unstick` does more than one thing, I would be in favor of making `unstick()` call out to `rejoin_pack_animals()` and whatever else is necessary to get a caravan unstuck.

I will call it "unload" then, because it is only about caravan not unloading.
"unstick" seems too general - is the caravan stuck moving, or entering map, or leaving map, or unloading goods?

I will make `unload()` call out to `rejoin_pack_animals()` internally.

myk002:
btw, your master branch will be out of sync after this is merged. you'll want to
```
git fetch upstream
git checkout master
git reset --hard upstream/master
git push --force
```
to get back in sync. Creating branches for PRs will avoid this issue in the future.

PR #405, opened 2022-06-13
Add gui/kitchen-info script - adds more info to the Kitchen screen

CarabusX:


gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
+-- Adds more info to the Kitchen screen.
+--@ enable = true
+--[====[
+
+gui/kitchen-info
+===================
```
myk002, line 6:
```suggestion
================
```

myk002, line 6:
```suggestion
================
```

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+    enable|disable gui/kitchen-info
+
+]====]
+
+gui = require 'gui'
+dialogs = require 'gui.dialogs'
+gps = df.global.gps
+world = df.global.world
+
+local kitchen_overlay = defclass(kitchen_overlay, gui.Screen)
```
myk002, line 21: in order for class identity to be preserved across script invocations, classes cannot be declared `local`

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+kitchen_overlay.focus_path = 'kitchen_overlay'
```
myk002, line 22:
```suggestion
kitchen_overlay.ATTRS {
    focus_path='kitchen_overlay',
}
```

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+
+local COLUMNS_BY_LABEL = {}
+
+local function ArrayToSet (array)
+    local set = {}
+    for _, value in ipairs (array) do
+        set[value] = true
+    end
+    return set
+end
```
myk002, line 83: this can be replaced with a call to `utils.invert()`

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+
+for _, column in ipairs(COLUMNS) do
+    column.activeOnPages = ArrayToSet(column.activeOnPages)
+    column.minLength = math.max(column.label:len(), column.minValueLeftPadding + column.maxValueLength)
```
myk002, line 87: prefer `#column.label` over `column.label:len()`

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+            if (matches) then
+                return x, x + length - 1
+            end
+        end
+    end
+    return -1, -1
+end
+
+local function drawNumber(x, kitchen, firstVisibleIndex, numDisplayedItems)
+    dfhack.screen.paintString({ fg = COLOR_WHITE }, x, 4, "Number")
```
myk002, line 202: "Number" is a bit ambiguous. This is the number of items in stock, right? How about calling it "Stock" or "Count"?

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+    keys_p:seek(keys_x, 1):key('CUSTOM_P'):string(show_processing and ": Hide processing info" or ": Show processing info")
+end
+
+function kitchen_overlay:onInput(keys)
+    if keys.CUSTOM_P then
+        show_processing = not show_processing
+    end
+
+    local parent = self._native.parent
+    gui.simulateInput(parent, keys)
```
myk002, line 377:
```suggestion
    self:sendInputToParent(keys)
```

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+
+    if dfhack.screen.isDismissed(parent) then
+        self:dismiss()
+    end
+end
+
+focus_stack = focus_stack or {'', ''}
```
myk002, line 384: what exactly is this protecting against? does the onStateChange callback fire more than once for the `kitchenpref` screen?

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+
+dfhack.onStateChange['gui/kitchen-info'] = function(eventCode)
+    if (eventCode == SC_VIEWSCREEN_CHANGED) then
+        table.insert(focus_stack, dfhack.gui.getCurFocus())
+        table.remove(focus_stack, 1)
+
+        if enabled and focus_stack[2] == 'kitchenpref' and focus_stack[1] ~= 'dfhack/lua/kitchen_overlay' then
```
myk002, line 391: instead of checking enabled here, it would be more performance-friendly to remove the onStateChange callback when gui/kitchen-info is disabled.

gui/kitchen-info.lua
```diff
@@ -0,0 +1,414 @@
...
+    if dfhack.gui.getCurFocus() == 'kitchenpref' then
+        kitchen_overlay():show()
+    end
+elseif args[1] == 'disable' then
+    enabled = false
+else
+    print([[Usage:
+    gui/kitchen-info enable|disable|help
+    enable|disable gui/kitchen-info
+    ]])
```
myk002, line 413:
```suggestion
    print(dfhack.script_help())
```

myk002 (changes requested):

gui/kitchen-info.lua
CarabusX: Fixed

gui/kitchen-info.lua
CarabusX: But it appears like this (before change) in many other gui scripts?

gui/kitchen-info.lua
CarabusX: Done.

gui/kitchen-info.lua
CarabusX: Fixed here and in 2 other places.

gui/kitchen-info.lua
CarabusX: But the actual column in DF is labeled "Number".
I will call it "drawNumberColumn".

gui/kitchen-info.lua
CarabusX: Will need further change as I still need "parent" for dfhack.screen.isDismissed(parent) check.

gui/kitchen-info.lua
CarabusX: It protects agains kitchen_overlay reopening when kitchenpref becomes the top screen again after kitchen_overlay is dismissed.
I took the code from "gui/extended-status.lua" script.

I guess it could be replaced with a check of (enabled and not isDismissed(parent))

gui/kitchen-info.lua
CarabusX: Done.

gui/kitchen-info.lua
ab9rf: There's a lot of legacies in our current codebase that are not necessarily worth perpetuating.

CarabusX: @myk002 also said: "No functional difference as written, but ATTRS is the official API for setting properties on your class."

Ok, changing it.

gui/kitchen-info.lua
```diff
@@ -0,0 +1,404 @@
...
+                    local bag_item_subtype = reaction_product.item_subtype[i]
+                    local bag_mat_type  = reaction_product.material.mat_type[i]
+                    local bag_mat_index = reaction_product.material.mat_index[i]
+                    local bag_matinfo = dfhack.matinfo.decode(bag_mat_type, bag_mat_index)
+                    local bag_mat_plant_raw = bag_matinfo.plant
+                    local bag_mat_plant_growth = bag_mat_plant_raw.growths[bag_item_subtype]
+
+                    if (bag_mat_plant_growth.name_plural:sub(-6) == "leaves") then
+                        return "Leaves"
+                    end
```
ab9rf, line 119: Is this really the best way to implement this test?

gui/kitchen-info.lua
```diff
@@ -0,0 +1,404 @@
...
+        local mill_matinfo = dfhack.matinfo.decode(mill_mat_type, mill_mat_index)
+        local mill_material = mill_matinfo.material
+
+        if (mill_material.flags.IS_DYE) then
+            return "Dye"
+        elseif (mill_material.state_name.Powder:sub(-5) == "flour") then
+            return "Flour"
+        else
+            return "Other"
+        end
```
ab9rf, line 149: What's the significance here of differentiating "flours" from others?

I'm very leery of using substring matches into item names or descriptions; just because it yields the desired results on vanilla items doesn't mean it always will, or that it will do so on modded items as well.

CarabusX: The point is user friendlyness.
I am already differenting Dye by the flag.
For the rest when the user sees Flour he will know that by milling Plant Abc he will get Plant Abc Flour, like any other flour in the game.
When he sees Other, he will have to look up what sort of Powder is created and whats its use.

If that doesn't give desired results for some modded Flour, which despite name is not really a Flour, then it is the problem of user-friendlyness of mod that named it.
Yes it is sort of hardcoding vanilla stuff, but most thing on this screen relly on assumptions from vanilla, as many jobs like "Process to bag" or "Mill to paste" are a certain vanilla reaction with certain vanilla tags on items.

gui/kitchen-info.lua
CarabusX: Again, user experience.
"Leaves" in the name is what the user sees.
LEAVES in growth type, or LEAF material based on LEAF_TEMPLATE material template are all internal tags that can be named whatever and work the same.

(And tbh it is mostly to support vanilla Quarry Bush, which is the only vanilla plant with this reaction. Modded items will just get "Other" so user have to look it up)

gui/kitchen-info.lua
```diff
@@ -0,0 +1,404 @@
...
+    end
+    return false
+end
+
+local function findStringInLine(x1, x2, y, str)
+    local length = #str
+    local firstByte = str:byte(1)
+    local xEnd = x2 - (length - 1)
+    for x = x1, xEnd do
+        if (dfhack.screen.readTile(x, y).ch == firstByte) then
```
myk002, line 179: `dfhack.screen.readTile(x, y)` appears to return nil when ttf mode (F12) is enabled (PRINT_MODE=2D).

Is there a way to handle this?

CarabusX: You mean the script crashes now? I admit I didn't test other print modes or ttf.
No idea how much DFHack supports ttf mode, I will have to check.
At least I will try to make the code safer against this.

CarabusX: I added support for detecting and handling TrueType text. It works correctly now.
Also updated changelog, moving it to future release.

gui/kitchen-info.lua
```diff
@@ -0,0 +1,433 @@
...
+    local p = gui.Painter.new_xy(x, 6, x + string.len("Permissions") - 1, 6 + numDisplayedItems - 1)
+    local possible  = kitchen.possible[kitchen.page]
+    local forbidden = kitchen.forbidden[kitchen.page]
+
+    for i = 0, numDisplayedItems - 1 do
+        local index = firstVisibleIndex + i
+
+        p:seek(1, i)
+        if (possible[index].Cook) then
+            local color = forbidden[index].Cook and COLOR_LIGHTRED or COLOR_LIGHTBLUE
```
myk002, line 251: Would it be possible to detect if `tweak kitchen-prefs-color` is enabled so we could display the corresponding color here?

gui/kitchen-info.lua
```diff
@@ -0,0 +1,433 @@
...
+    -- Hotkey
+    local keys_p = gui.Painter.new_xy(1, gps.dimy - 3, gps.dimx - 2, gps.dimy - 2)
+    local keys_x = math.min(gps.dimx - 3 - string.len("p: Hide processing info"), 62)
+    keys_p:key_pen(COLOR_LIGHTRED):pen(COLOR_WHITE)
+    keys_p:seek(keys_x, 1):key('CUSTOM_P'):string(show_processing and ": Hide processing info" or ": Show processing info")
+end
+
+function kitchen_overlay:onInput(keys)
+    if keys.CUSTOM_P then
+        show_processing = not show_processing
```
myk002, line 394: this does not play nice with the search filter. Every time a 'p' appears in the search box, the gui/kitchen-info info is toggled.

I actually think we can get away with not having this option at all. What does toggling add? If a player wants it gone, they can just run `disable gui/kitchen-info`

myk002 (changes requested):
I think this script is a good candidate for "default on". Could you do a PR in the dfhack repo to add it to dfhack.init-example?

CarabusX:
> I think this script is a good candidate for "default on". Could you do a PR in the dfhack repo to add it to dfhack.init-example?

I will happily make a PR to make it enabled by default once it gets merged :)

gui/kitchen-info.lua
```diff
@@ -221,6 +221,14 @@ local function findStringInLine(x1, x2, y, str)
     return -1, -1
 end
 
+local function pluginCompatibility_search_isInputActive()
+    return false  -- FIXME: Implement when API is available
```
myk002, line 225: Are you planning on implementing this? As it stands, this PR looks good except for the P functionality. I'd suggest removing the P toggle for now so we can merge this PR, then we can add it back if and when the search plugin gains the appropriate API.

myk002 (changes requested):

myk002 (approved):

CarabusX:
@myk002 Thank you for updating it and merging! :)
I got busy with other stuff, and then forgot that it was still open.

myk002:
: ) no worries! It will likely need some updating for v50, if you're interested. I implemented an overlay framework so added information like this won't require its own viewscreen. See: https://docs.dfhack.org/en/latest/docs/dev/overlay-dev-guide.html

CarabusX:
Yeah I would like to have it work in v50.xx as I am playing it at the moment and very much miss the functionality.
But I will probably wait a little until everything settles and stabilizes up a bit after the release.

myk002:
good idea -- I'm still adjusting the DFHack Screen and Gui modules for the new version, and the Lua widget library needs some work as well before it's usable again.

PR #407, opened 2022-06-14
Add warn-stealers

wolfboyft:
Requires the onUnitNewActive fix being worked on.
`warn-stealers start` in onMapLoad.init will warn when creatures that can steal food, drinks, or items become visible.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@ that repo.
 
 ## New Scripts
 
+- `warn-stealers`: warn when creatures that may steal your food, drinks, or items enter the map
 - `assign-minecarts`: assign minecarts to hauling routes that don't have one
 - `deteriorate`: combines, replaces, and extends previous `deteriorateclothes`, `deterioratecorpses`, and `deterioratefood` scripts.
 - `gui/petitions`: shows list of fort's petitions
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,7 +16,6 @@ that repo.
 ## New Scripts
 
 - `warn-stealers`: warn when creatures that may steal your food, drinks, or items become visible
-- `assign-minecarts`: assign minecarts to hauling routes that don't have one
 ## Fixes
 
 ## Misc Improvements
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,6 +16,7 @@ that repo.
 ## New Scripts
 
 - `warn-stealers`: warn when creatures that may steal your food, drinks, or items become visible
+
 ## Fixes
 
 ## Misc Improvements
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,7 +15,7 @@ that repo.
 
 ## New Scripts
 
-- `warn-stealers`: warn when creatures that may steal your food, drinks, or items enter the map
+- `warn-stealers`: warn when creatures that may steal your food, drinks, or items become visible
 - `assign-minecarts`: assign minecarts to hauling routes that don't have one
 - `deteriorate`: combines, replaces, and extends previous `deteriorateclothes`, `deterioratecorpses`, and `deterioratefood` scripts.
 - `gui/petitions`: shows list of fort's petitions
--- a/warn-stealers.lua
+++ b/warn-stealers.lua
@@ -14,13 +14,13 @@ local eventfulKey = "warn-stealers"
 local numTicksBetweenChecks = 100
 
 function gamemodeCheck()
-    if df.global.gamemode ~= df.game_mode.DWARF then
-        cache = nil
-        print("warn-stealers must be run in fort mode")
-        disable()
-        return false
+    if df.global.gamemode == df.game_mode.DWARF then
+        return true
     end
-    return true
+    cache = nil
+    print("warn-stealers must be run in fort mode")
+    disable()
+    return false
 end
 
 cache = cache or {}
--- a/warn-stealers.lua
+++ b/warn-stealers.lua
@@ -13,11 +13,19 @@ local repeatUtil = require("repeat-util")
 local eventfulKey = "warn-stealers"
 local numTicksBetweenChecks = 100
 
-if df.global.gamemode ~= df.game_mode.DWARF then
-    if df.global.gamemode ~= df.game_mode.NONE then
-        -- errors when gamemode is NONE
-        persistTable.GlobalTable.warnStealersCache = nil
+function gamemodeCheck()
+    if df.global.gamemode ~= df.game_mode.DWARF then
+        if df.global.gamemode ~= df.game_mode.NONE then
+            -- errors when gamemode is NONE
+            persistTable.GlobalTable.warnStealersCache = nil
+        end
+        stop()
+        return false
     end
+    return true
+end
+
+if not gamemodeCheck() then
     return
 end
 
@@ -26,19 +34,11 @@ if not persistTable.GlobalTable.warnStealersCache then
 end
 local cache = persistTable.GlobalTable.warnStealersCache
 
-local function isUnitHidden(unit)
-    local block = dfhack.maps.getTileBlock(unit.pos)
-    if not block then
-        return false
-    end
-    return block.designation[unit.pos.x%16][unit.pos.y%16].hidden
-end
-
 local races = df.global.world.raws.creatures.all
 
-local function addToCacheIfStealerAndHidden(unitId)
+function addToCacheIfStealerAndHidden(unitId)
     local unit = df.unit.find(unitId)
-    if not isUnitHidden(unit) then
+    if not dfhack.units.isHidden(unit) then
         return
     end
     local casteFlags = races[unit.race].caste[unit.caste].flags
@@ -47,7 +47,7 @@ local function addToCacheIfStealerAndHidden(unitId)
     end
 end
 
-local function announce(unit)
+function announce(unit)
     local caste = races[unit.race].caste[unit.caste]
     local casteFlags = caste.flags
     local desires = {}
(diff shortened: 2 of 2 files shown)


wolfboyft:
Should we have something for on-new-fortress that can warn you when there are stealers on embark?

myk002:
so in going through all the scripts for the taxonomy, I came across, https://docs.dfhack.org/en/stable/docs/_auto/base.html?highlight=reveal-hidden-units#reveal-hidden-units
which seems highly related. Could we possibly integrate with that script instead of creating a new one?

wolfboyft:
"This script exposes all units on the map who are currently sneaking or waiting in ambush and thus hidden from the player’s view." That seems like totally unrelated behaviour, I don't see how this could be part of that script.

myk002:
> "This script exposes all units on the map who are currently sneaking or waiting in ambush and thus hidden from the player’s view." That seems like totally unrelated behaviour, I don't see how this could be part of that script.

Indeed. I was conflating "stealers" with "ambushers". Please disregard.

changelog.txt
```diff
@@ -15,6 +15,7 @@ that repo.
 
 ## New Scripts
 
+- `warn-stealers`: warn when creatures that may steal your food, drinks, or items become visible
```
myk002, line 18: needs moving to the "Future" release

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+local eventfulKey = "warn-stealers"
+local numTicksBetweenChecks = 100
+
+if df.global.gamemode ~= df.game_mode.DWARF then
+    if df.global.gamemode ~= df.game_mode.NONE then
+        -- errors when gamemode is NONE
+        persistTable.GlobalTable.warnStealersCache = nil
+    end
+    return
+end
```
myk002, line 22: maybe this should be turned into a function and then called from the start and ontick functions. it should cancel the repeat-util callback as well.

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+
+if not persistTable.GlobalTable.warnStealersCache then
+    persistTable.GlobalTable.warnStealersCache = {}
+end
+local cache = persistTable.GlobalTable.warnStealersCache
+
+local function isUnitHidden(unit)
+    local block = dfhack.maps.getTileBlock(unit.pos)
+    if not block then
+        return false
```
myk002, line 32: i'm not sure when a unit would be in an invalid block, but shouldn't this return true to indicate that the unit is hidden?

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+    return
+end
+
+if not persistTable.GlobalTable.warnStealersCache then
+    persistTable.GlobalTable.warnStealersCache = {}
+end
+local cache = persistTable.GlobalTable.warnStealersCache
+
+local function isUnitHidden(unit)
+    local block = dfhack.maps.getTileBlock(unit.pos)
```
myk002, line 30: consider `local flags = dfhack.maps.getTileFlags(unit.pos)`, then you can just check `flags.hidden` instead of doing the math in the block

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+    if casteFlags.CURIOUS_BEAST_EATER then
+        str = str .. "eat food + "
+    end
+    if casteFlags.CURIOUS_BEAST_GUZZLER then
+        str = str .. "guzzle drinks + "
+    end
+    if casteFlags.CURIOUS_BEAST_ITEM then
+        str = str .. "steal items + "
+    end
+    str = str:sub(1, -4)
```
myk002, line 63:
```suggestion
    local desires = {}
    if casteFlags.CURIOUS_BEAST_EATER then
        table.insert(desires, "eat food")
    end
    if casteFlags.CURIOUS_BEAST_GUZZLER then
        table.insert(desires, "guzzle drinks")
    end
    if casteFlags.CURIOUS_BEAST_ITEM then
        table.insert(desires, "steal items")
    end
    local str = table.concat(str, " + ")
```

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
+-- Script to warn when creatures that may steal food become visible
+--[====[
+warn-stealers
+=============
+Will make a zoomable announcement whenever a creature that can eat food, guzzle drinks, or steal items enters the map and moves into a revealed location.
+Takes ``start`` or ``stop`` as parameters.
```
myk002, line 6:
```suggestion
Usage::

    warn-stealers [start|stop]
```

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
+-- Script to warn when creatures that may steal food become visible
```
myk002, line 1: also consider supporting `enable` syntax. See https://github.com/DFHack/scripts/pull/405/files for an example.

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+    end
+    if casteFlags.CURIOUS_BEAST_ITEM then
+        str = str .. "steal items + "
+    end
+    str = str:sub(1, -4)
+    dfhack.gui.showZoomAnnouncement(-1, unit.pos, "A " .. caste.caste_name[0] .. " has appeared, it may " .. str .. ".", COLOR_RED, true)
+end
+
+local function onTick()
+    for _, unitIdStr in ipairs(cache._children) do
```
myk002, line 68: do we need to check if the game mode is still valid (and shut down if not)?

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+        if cache[unitIdStr] then -- For a bug in persist-table
```
myk002, line 69: what bug is this?

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+            elseif not isUnitHidden(unit) then
+                announce(unit)
+                cache[unitIdStr] = nil -- this isn't stopping it from being iterated over.
+            end
+        end
+    end
+end
+
+local function help()
+    print("syntax: warn-stealers [start|stop]")
```
myk002, line 83:
```suggestion
    print(dfhack.script_help())
```

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+    if not block then
+        return false
+    end
+    return block.designation[unit.pos.x%16][unit.pos.y%16].hidden
+end
+
+local races = df.global.world.raws.creatures.all
+
+local function addToCacheIfStealerAndHidden(unitId)
+    local unit = df.unit.find(unitId)
```
myk002, line 40: do we need to check the game mode here and stop if we're not where we expect to be?

myk002 (changes requested):

warn-stealers.lua
lethosor: There is also `dfhack.units.isHidden()` which handles considerably more cases

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+
+if df.global.gamemode ~= df.game_mode.DWARF then
+    if df.global.gamemode ~= df.game_mode.NONE then
+        -- errors when gamemode is NONE
+        persistTable.GlobalTable.warnStealersCache = nil
+    end
+    return
+end
+
+if not persistTable.GlobalTable.warnStealersCache then
```
myk002, line 24: If we're populating the cache on start, and that covers all units, why do we need a persistent cache?

If a stealer unit is already unhidden, it wouldn't get another announcement in either case, right?
I suppose the cache would catch the case where a unit was announced, then moved back into hidden territory, then the user saved and loaded. Is that worth it? Or should the stealer unit be reannounced anyway?

changelog.txt
wolfboyft: Ididitagain

warn-stealers.lua
wolfboyft: I tested units with negative coords and they were visible. And my testing showed me that it was the tile they were on that defined whether they were visible, but OK, since there're more cases, I'll use that.

warn-stealers.lua
wolfboyft: Iteration iterates over keys with values of nill.
But it seems hard to reproduce.

warn-stealers.lua
wolfboyft: Well we are going to migrate to DFHack's own is unit hidden so

warn-stealers.lua
wolfboyft: Resolving 'cause other conversation

changelog.txt
```diff
@@ -15,6 +15,8 @@ that repo.
...
+- `assign-minecarts`: assign minecarts to hauling routes that don't have one
```
myk002, line 19:
```suggestion

```

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+    if df.global.gamemode ~= df.game_mode.DWARF then
+        cache = nil
+        print("warn-stealers must be run in fort mode")
+        disable()
+        return false
+    end
+    return true
+end
+
+cache = {}
```
myk002, line 27:
```suggestion
cache = cache or {}
```

myk002: Otherwise the cache will get reset every time the script is run

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
+-- Script to warn when creatures that may steal food become visible
+--@ enable = true
+--[====[
+warn-stealers
+=============
+Will make a zoomable announcement whenever a creature that can eat food, guzzle drinks, or steal items enters the map and moves into a revealed location.
+Takes ``start`` or ``stop`` as parameters.
+]====]
+
+local persistTable = require("persist-table")
```
myk002, line 10:
```suggestion
```

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+local eventful = require("plugins.eventful")
+local repeatUtil = require("repeat-util")
+
+local eventfulKey = "warn-stealers"
+local numTicksBetweenChecks = 100
+
+function gamemodeCheck()
+    if df.global.gamemode ~= df.game_mode.DWARF then
```
myk002, line 18: Prefer putting the smaller code block inside the conditional. In this case, try returning true if the game mode is dwarf. Then the rest of the function can lose an indent level.

warn-stealers.lua
```diff
@@ -0,0 +1,107 @@
...
+    if casteFlags.CURIOUS_BEAST_EATER then
+        table.insert(desires, "eat food")
+    end
+    if casteFlags.CURIOUS_BEAST_GUZZLER then
+        table.insert(desires, "guzzle drinks")
+    end
+    if casteFlags.CURIOUS_BEAST_ITEM then
+        table.insert(desires, "steal items")
+    end
+    local str = table.concat(desires, " + ")
```
myk002, line 58: Since the rest of the message is a proper sentence, maybe " and " would be a better joiner than " + "

myk002 (approved):

changelog.txt
wolfboyft: I can't believe I keep fucking this up I'm so sorry

PR #412, opened 2022-07-10
copy script docs into separate files

myk002:
DFHack/dfhack#2228

Help text is unchanged, though we'll have to go through them all to standardize formats and organization. The short help will be taken from the first sentence of the help instead of the first commented line of the script, so the help text might have to be slightly rewritten for each script.

Original help text will also remain in the script files until we merge the help branch.


myk002:
pre-commit.ci autofix

PR #414, opened 2022-07-19
Fix emigration logic

WoosterUK:
The comparator in the master code is incorrect.

Dorf stress is negative for happy dorfs and positive for unhappy ones. Line 33 creates a threshold that is, therefore, above 100 for happy dorfs and below 100 for unhappy ones.

Line 90 in master asks for a 0-100 roll to be *below* this threshold for emigration to occur. This means that happy dorfs will *always* emigrate (contra line 14) and unhappy ones are increasingly unlikely to the unhappier they get.

You can observe this behaviour in a new fort: I saw six out of seven dorfs leave almost instantly.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -21,6 +21,7 @@ that repo.
 ## Fixes
 - `gui/unit-info-viewer`: fix logic for displaying undead creature names
 - `gui/workflow`: restore functionality to the add/remove/order hotkeys on the workflow status screen
+- `emigration`: fix emigrant logic so unhappy dorfs leave as designed
 
 ## Misc Improvements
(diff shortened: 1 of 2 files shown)


myk002 (changes requested):
Thanks! Could you add a changelog entry for this in the docs/changelog.txt file?

myk002:
Unclear if this fixes all of DFHack/dfhack#1784, but it's clearly related

myk002 (approved):

myk002:
Welcome to DFHack! As a follow-up PR, could you possibly add your name to the Authors.rst file in the dfhack repo? https://github.com/DFHack/dfhack/blob/develop/docs/Authors.rst

PR #421, opened 2022-08-17
Update docs for dwarf-op

myk002:
and move dorf_tables to internal/ so it doesn't get listed by ls


docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+``--optimize``
+    Performs a job search for unoptimized dwarves. Run
+    ``dwarf-op --list job_distribution`` to see how jobs are distributed.
+``--applyjobs "[" <job> [<job> ...] "]"``
+    Applies the listed jobs to the selected dwarves. Run
+    ``dwarf-op --list jobs`` to see available jobs.
+``--applyprofessions "[" <profession> [<profession> ...] "]"``
+    Applies the listed professions to the selected dwarves. Run
+    ``dwarf-op --list professions`` to see available professions.
+``--applytypes "[" <profession> [<profession> ...] "]"``
```
cppcooper, line 118: This syntax, I personally find very confusing. I don't understand what this is trying to communicate. And afaik you definitely don't need to `"[" "]"`

Bottom line: if we're trying to make the docs more accessible to everyone, I think this is too obfuscated for most people to know how to read correctly

cppcooper: I feel as though the `[ a b c ]` syntax for lists maybe is the root of this evil. So maybe we could use curly braces instead? I dunno how many other scripts use this, but if it's just this one.. ez pz?

docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+    Selects any dwarves with the specified custom professions.
+``waves "[" waves <num> [<num> ...] "]"``
+    Selects dwarves from the specified migration waves. Waves are enumerated
+    starting at 0 and increasing by 1 with each wave. The waves go by season and
+    year and thus should match what you see in `list-waves` or Dwarf Therapist.
+    It is recommended that you ``--show`` the selected dwarves before modifying
+    them.
+
+Options
+-------
```
cppcooper, line 86: If we're grouping all the options together, you missed `--select`

I haven't tested, but I didn't use -<option> for any of the options. So afaik they should all be single dashes. Have you tested this? It doesn't appear that you've modified the arg list.

docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+Data tables
+-----------
+
+The data tables that ``dwarf-op`` uses are described below. They can be
+inspected with ``dwarf-op --list <table name>``.
+
+``job_distributions``
+    Defines thresholds for each column of distributions. The columns must add up
+    to the values in the thresholds row for that column.  Every other row
+    references an entry in the ``jobs`` table.
```
cppcooper, line 135: This is incorrect. The values in the columns **should** add up to (not technically required) the values in the thresholds row. Not very important to make that distinction though.

docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+``attrib_levels``
+    Defines stat distributions, used for both physical and mental attributes.
+    Each level gives a probability of a dwarf randomly being assigned an
+    attribute level, and it provides a mean and standard deviation for the
+    attribute's value.
+
+    The ``p`` values describe the normal distribution of stats (each ``p`` value
+    has a sub-distribution, which makes the bell curve not so bell-shaped).
+    Labours do not follow the same stat system and are more uniformly random,
+    which are compensated for in the description of jobs/professions.
```
cppcooper, line 146:
```suggestion
    Defines stat distributions for both physical and mental attributes.
    Each level has a probability (p-value, or p) which indicates how likely
    a level will be used for a particular stat such as strength or spacial
    awareness. The levels range from incompetent to unbelievable (god-like)
    and are mostly inline with what the game uses already, but adds one
    level more than what the game uses to push the unbelievable even higher
    on average.

    In addition to a bell shaped p-value curve for the levels, there is
    additionally a standard deviation used to generate the value once a
    level has been selected this makes the bell curve not so bell shaped in
    the end. Labours do not follow the same stat system and are more uniformly
    random, which are compensated for in the description of jobs/professions.
```

This is how I would word it, but perhaps it doesn't need to be this precise or informative.

docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+
+``jobs``
+    Defines ``dwarf-op``'s nameable jobs. Each job is comprised of required
+    professions, optional professions, probabilities for each optional
+    profession, a 'max' number of optional professions, and a list of types
+    (from the ``types`` table below) to apply to dwarves in the defined job.
+
+``professions``
+    These are a subset of the professions DF has. All professions listed will
+    match a profession dwarf fortress has built in, however not all the
```
cppcooper, line 156:
```suggestion
    These are a subset of the professions DF has. All professions listed should
    match a profession dwarf fortress has built in, however not all the
```
I would use the word should (and perhaps I messed up your char count in the suggestion).
Because, it should match.. but we don't control the upstream.

docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+    Each profession is defined with a set of job skills which match the skills
+    built into Dwarf Fortress. Each skill is given a value which represents the
+    bonus a dwarf will get for this skill. The skills are added in a random
+    order, with the first few receiving the highest values (excluding the bonus
+    just mentioned). Thus the bonuses are to ensure a minimum threshold is
+    passed for certain skills deemed critical to a profession.
+
+``types``
+    These are a sort of archetype system for applying to dwarves. It primarily
+    includes physical attributes, but can include skills as well.
```
cppcooper, line 168:
```suggestion
    These are a sort of archetype system for applying to dwarves. It primarily
    includes physical and mental attributes, but can include skills as well.
```

docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+
+    Each type has a probability of being applied to a dwarf just by pure luck -
+    this is in addition to types applied by other means. Each type also has a
```
cppcooper, line 171: and in addition to the attrib levels. This can potentially bump attributes up higher, which maybe can be inferred but I'm not sure it can.

cppcooper (commented):
I have only looked at the primary docs, since we're decoupling the online docs from the inline docs I'll leave it to you to resolve any of those discrepancies I just introduced with suggestions

myk002:
> I have only looked at the primary docs, since we're decoupling the online docs from the inline docs I'll leave it to you to resolve any of those discrepancies I just introduced with suggestions

The in-script docs will be removed as soon as we merge the docs branch into develop. In-script docs will only be used for out-of-tree scripts that don't have .rst files.

docs/dwarf-op.rst
```diff
@@ -1,158 +1,180 @@
...
+    them.
+
+Options
+-------
+
+``--reset``
+    Forget which dwarves have been optimized. However, if you reload an existing
+    save, the optimization list will be reloaded.
+``--resetall``
+    Forget which dwarves have been optimized and remove the persisted
```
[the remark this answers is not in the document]
myk002: compromised by changing to "saved"

docs/dwarf-op.rst
myk002: it's the "standard" syntax of `argparse.processArgs`, which unfortunately many scripts use. I've offered an alternative with `argparse.stringList` which allows you to represent a list with a comma-separated string, but the script would have to be modified to use it.

It's extra confusing because in the language of usage documentation, square brackets means "optional element" (which is why I'm quoting those square brackets to indicate that they are to be taken literally instead of having their usual meaning)

I'd rather not rewrite scripts just now, though, since I have to focus on getting the documentation updated across all scripts for this release. If you'd like to make (and test) the change though, I'd be happy to merge it!

myk002: note that the "Examples" section precedes the Options section, and I specifically included examples with these square bracket lists in them to clarify their usage.

docs/dwarf-op.rst
myk002: also `--list`. I'll add those in.

As for single vs. double dashes, `dwarf-op` uses `argparse.processArgs` to read the commandline (well, it uses the old name of `utils.processArgs`, but that's just an alias for the argparse version nowadays). That function treats single and double dashes identically. I'm standardizing on getopt-style options (single dash for single letter args, double dash for long name args) so the documented syntax won't change even if the underlying script moves from `argparse.processArgs` to `argparse.processArgsGetopt`, which scripts might do as they become more advanced and need more flexible argument parsing.

docs/dwarf-op.rst
myk002: what happens if they don't add up right?

cppcooper:
> made was just to get the current tree to compile In-script docs will only be used for out-of-tree scripts that don't have .rst files.

Oh? The rst files are replacing the help output? Or just taking away the dual purpose?

docs/dwarf-op.rst
cppcooper: Oh, very nice

docs/dwarf-op.rst
myk002: reworded a bit to make this more clear.

myk002:
> > made was just to get the current tree to compile In-script docs will only be used for out-of-tree scripts that don't have .rst files.
> 
> Oh? The rst files are replacing the help output? Or just taking away the dual purpose?

Yes, the output of `help dwarf-op` will print the text rendered from the docs/dwarf-op.rst file. This is also the text that will be returned by `dfhack.script_help()`. When I go back and remove all the in-script help, I'll convert any script I see outputting raw markup text to output the string returned from `dfhack.script_help()` instead.

docs/dwarf-op.rst
cppcooper: I'll give it a shot.

docs/dwarf-op.rst
cppcooper: I'll double check, but less than will possibly crash when it tries to pull the first larger than. If it's larger than though, it should keep chugging just fine since it can keep pulling jobs.

docs/dwarf-op.rst
myk002: Ok. Let me merge this PR so you can avoid conflicts later. Feel free to make more doc changes as you change the code!

PR #423, opened 2022-08-22
Implements CSV lists for dwarf-op

cppcooper:
Updated docs, ran tests, added a unit test for validating the data in dorf_tables (self-linkage, and valid built-ins). I think that's everything in this.


docs/dwarf-op.rst
```diff
@@ -73,14 +73,15 @@ instead of ``all``.
 ``drunks``
     Selects any dwarves who have the ``DRUNK`` profession, including those who
     have been zeroed by the ``--clear`` command option.
-``jobs "[" jobs <jobname> [<jobname> ...] "]"``
+``"jobs,<job>[,<job>,...]"``
     Selects any dwarves with the specified custom professions.
-``waves "[" waves <num> [<num> ...] "]"``
+    e.g. ``--select "jobs,Miner,Admin"``
+``"waves,<num>[,<num>,...]"``
```
myk002, line 79:
```suggestion
``waves,<num>[,<num> ...]``
```

docs/dwarf-op.rst
```diff
@@ -53,9 +53,9 @@ instead of ``all``.
     specified.
 ``all``
     Selects all dwarves.
-``<name>``
+``<name>[,<name>,...]``
```
myk002, line 56:
```suggestion
``<name>[,<name> ...]``
```

docs/dwarf-op.rst
```diff
@@ -73,14 +73,15 @@ instead of ``all``.
...
     Selects any dwarves with the specified custom professions.
-``waves "[" waves <num> [<num> ...] "]"``
+    e.g. ``--select "jobs,Miner,Admin"``
+``"waves,<num>[,<num>,...]"``
     Selects dwarves from the specified migration waves. Waves are enumerated
     starting at 0 and increasing by 1 with each wave. The waves go by season and
     year and thus should match what you see in `list-waves` or Dwarf Therapist.
     It is recommended that you ``--show`` the selected dwarves before modifying
-    them.
+    them. e.g. ``--select "waves,2,3,5,7,11,13"``
```
myk002, line 84: as above, this should go in the examples section

docs/dwarf-op.rst
```diff
@@ -119,13 +120,13 @@ Command options
 ``--optimize``
     Performs a job search for unoptimized dwarves. Run
     ``dwarf-op --list job_distribution`` to see how jobs are distributed.
-``--applyjobs "[" <job> [<job> ...] "]"``
+``--applyjobs "<job>[,<job>,...]"``
```
myk002, line 123:
```suggestion
``--applyjobs <job>[,<job> ...]``
```

docs/dwarf-op.rst
```diff
@@ -119,13 +120,13 @@ Command options
...
     Applies the listed jobs to the selected dwarves. Run
     ``dwarf-op --list jobs`` to see available jobs.
-``--applyprofessions "[" <profession> [<profession> ...] "]"``
+``--applyprofessions "<profession>[,<profession>,...]"``
```
myk002, line 126:
```suggestion
``--applyprofessions <profession>[,<profession> ...]``
```

docs/dwarf-op.rst
```diff
@@ -119,13 +120,13 @@ Command options
...
     Applies the listed professions to the selected dwarves. Run
     ``dwarf-op --list professions`` to see available professions.
-``--applytypes "[" <profession> [<profession> ...] "]"``
+``--applytypes "<profession>[,<profession>,...]"``
```
myk002, line 129:
```suggestion
``--applytypes <profession>[,<profession> ...]``
```

dwarf-op.lua
```diff
@@ -1008,32 +1013,46 @@ end
...
+        else
+            error("The select option entered is not a string.")
+        end
+        local N = #list
+
+        -- parse the selection arguments
+        if N == 1 then
+            -- name
             if string.match(name,option) or string.match(nickname,option) then
                 return true
```
myk002, line 1039: I don't understand why this is only being run if `#list == 1` -- what if the user specifies something like your example of `bob,james,urist`?

myk002: although really it might keep things simple if the user could pass only one name. that would make it easier to distinguish when the 'p' prefix should be given its special meaning. Unless the name to be matched is "pall" or something..

myk002 (changes requested):

dwarf-op.lua
cppcooper: `N==1` is the equivalent of what `type(option) == 'string'` was.

> what if the user specifies something like your example..

See line 1088.

And if the user enters only one name, and it's something like pall or has any keyword inside of it.. this would have always broken in that edge case. They should avoid naming their dwarves `drunk` for example. What it would do is, it would select all dwarves with the drunk profession, and the ones named drunk. So it will break in that way if they use keywords to name dwarves. I've never seen a dwarf with `all` as a substring of their name, but if that happens naturally.. and they are protected.. it'll select them anyway regardless of whether they have the `p` prefix.

They should also avoid using numbers in the names, or jobs (the two primary list-able options). With numbers in their names, and trying to select by waves, the same thing will happen. It'll select the waves, and dwarves with it in their name. It'll always select what they want.. but it may select extra if they choose to name their dwarves in non-recommended ways. This is one of the reasons `--show` exists so the user can verify the selected dwarves.

You do bring up a valid point about the p prefix though, since it always checks the first option for a `p` prefix.. providing a list `paul,brent,davy` would enable the protected selection flag.. However since choosing by name always ignores protected status, it is a non-issue. Anything else and they are in a non-recommended edge case where they're integrating things into names they probably shouldn't be.

docs/dwarf-op.rst
cppcooper: I wrestled with whether to omit the last comma but then thought about what I've been taught to do in math which is `1,2,3,...,N`

So I figured using that well known listing methodology would be best for people to understand what to do. Though they probably won't be dumb enough to try `jobs,Miner, Admin, Farmer`. In any case, I'll drop one of the things, but not both (just in case they are). Do you want the quotes, or the final comma?

myk002: Reference syntax docs don't use the final comma, e.g.
image
(that's from `man apropos`) and that's the standard I've been applying to other DFHack docs. Syntax docs also don't add quotes in -- the user adds them if they're needed, just like any single param that contains spaces. Let's go with no final comma and no quotes.

dwarf-op.lua
myk002: > `N==1` is the equivalent of what `type(option) == 'string'` was.

That would be true if this line appeared above `list = argparse.stringList(option)`, but where the test is currently located, if `option` is `paul,brent,davy` then `#list` will be 3, not 1.

myk002: Gotcha, though it seems like we are unnecessarily restricting what players can name their dwarves. This restriction could have been avoided by starting the name list with a keyword like "name".

cppcooper: I can't remember if there was some way of thinking that led me to want this particular behaviour, but I can think on it tomorrow before I test my spectate unpause code.

I'll probably add a keyword, or figure out what I was thinking and agree with it. In the latter I'll give you the opportunity to show me the err of my thoughts :)

docs/dwarf-op.rst
cppcooper: Ok. An established convention for command line is enough. More applicable than math at any rate. Though I note your suggestions put a space, I think I'll omitt that to be in line with - what despite being a single data point - apropos which I am going to assume is representative of all similar commands.

dwarf-op.lua
cppcooper: I think maybe I wanted to embed the names with other selection options.. but I had restricted it to one name, so that kinda rules that out and maybe even implies had actually considered this problem and decided to avoid it.

I'll add the keyword. Though it might be nice to be able to select specific dwarves along side particular jobs or waves, probably better to let the eccentric guy name their dwarves: {{dwarf1,dwarf2,...,dwarfN}|{Miner1,Farmer3,Admin93}} etc.

myk002 (changes requested):

docs/dwarf-op.rst
```diff
@@ -76,14 +76,17 @@ instead of ``all``.
...
     Selects any dwarves who have the ``DRUNK`` profession, including those who
     have been zeroed by the ``--clear`` command option.
-``jobs "[" jobs <jobname> [<jobname> ...] "]"``
+``names,<name>[,<name>...]``
+    Selects any dwarf with <name> anywhere in their name or nickname. This
+    option ignores protection status.
+``"jobs,<job>[,<job>...]"``
     Selects any dwarves with the specified custom professions.
-``waves "[" waves <num> [<num> ...] "]"``
+``"waves,<num>[,<num>...]"``
```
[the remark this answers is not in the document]
cppcooper: Not sure how I missed these.

myk002 (approved):

myk002:
Could you add a changelog entry for the improvement?

cppcooper:
> Could you add a changelog entry for the improvement?

done

changelog.txt
```diff
@@ -23,13 +23,15 @@ that repo.
 - `gui/workflow`: restore functionality to the add/remove/order hotkeys on the workflow status screen
 - `emigration`: fix emigrant logic so unhappy dwarves leave as designed
 - `gui/gm-unit`: allow ``+`` and ``-`` to adjust skill values as intended instead of letting the filter intercept the characters
+- `dwarf-op`: fixed a crash in dwarf-op caused by a typo in the dorf_tables file
```
lethosor, line 26: I try to use "crash" for full crashes/segfaults only:
```suggestion
- `dwarf-op`: fixed an error in dwarf-op caused by a typo in the dorf_tables file
```

docs/dwarf-op.rst
```diff
@@ -17,10 +17,7 @@ random variation so each dwarf is unique.
 you can specify precisely what archtype you want a for given dwarf, and
 ``dwarf-op`` can generate a random dwarf that matches that archetype.
 
-Usage
------
-
-::
+Usage::
```
myk002, line 20: Sorry, I didn't see this earlier. Could you revert this modification? Usage should be a header, not plain text.

cppcooper: Oh, that probably happened during a merge conflict during rebase. Got a bit confused and must have done the wrong thing

myk002: Yeah, the formatting is new. It took us a while to settle on this style. I must have changed it six or seven times.. that reminds me, I should look at Documentation.rst and make sure it reflects the latest standards

changelog.txt
```diff
@@ -23,13 +23,15 @@ that repo.
 - `gui/workflow`: restore functionality to the add/remove/order hotkeys on the workflow status screen
 - `emigration`: fix emigrant logic so unhappy dwarves leave as designed
 - `gui/gm-unit`: allow ``+`` and ``-`` to adjust skill values as intended instead of letting the filter intercept the characters
+- `dwarf-op`: fixed an error in dwarf-op caused by a typo in the dorf_tables file
```
myk002, line 26: could you be more specific here? What works now when it didn't before? Like, which profession/job value?

myk002 (changes requested):

changelog.txt
cppcooper: Fisherman was in as fisher. I can add that to the log, but a user wouldn't have known what was happening without heavy debugging

cppcooper: I suppose maybe something like:
`fixed an error in dwarf-op preventing it from optimizing dwarves, sometimes`
I dunno, that doesn't really seem any better. I could just go extra long:
```
fixed an error in dwarf-op caused by a typo in the dorf_tables file. This typo (FISHER, instead of FISHERMAN) caused the script to error when creating miners
```

myk002: Would something like "fixed application of MINER job attributes" be accurate?

cppcooper: Nearly, it's not the application of it that was messed up. It was data inside of it. It referenced something that didn't exist.

myk002: from a user perspective, what action did it prevent from working? Which feature was broken?

cppcooper: The miner job. So anything involving it. The optimize feature or applyjobs with miner as one of them. And not even all the time, cause the profession doesn't have to be used.. Although the more miners, the more exponentially likely it is to be used.

I think of it as a linker error, except during runtime. With the unit test it turns that kind of issue into a compile time error (sort of.. Build time? Test time? Lol)

myk002: So how about "fixed application of Miner job to dwarves"?

cppcooper: Yea it is accurate, since the Miner job isn't part of the dwarf-op script.. So naming it specifically I think implies a problem with its definition.

cppcooper, line 26:
```suggestion
- `dwarf-op`: fixed error when applying the Miner job to dwarves
```

myk002 (approved):

PR #427, opened 2022-08-30
Porting scripts from Ruby to Lua

Putnam3145:
So far, ported adaptation and ban-cooking to Lua. Untested. Draft for a reason.

(not quite finished--comment about not implementing "show" in ban-cooking quite possibly has an implied "yet")


myk002:
Thank you so much for helping with this! Ruby has become a real pain point!

If you're making any changes to the docs, you should know that we're in the middle of moving documentation to the [docs](https://github.com/DFHack/scripts/tree/master/docs) directory. Doc text inside the scripts will be removed. Out-of-tree scripts will keep their docs in-script as they currently are. This docs change is only for in-tree scripts.

I'm going through and updating the doc text for all the scripts right now. I've done up to scripts that start with "s". Docs after that point will still be in old format. If you're porting any scripts in the latter part of the alphabet, updated documentation standardsare here: https://dfhack--2247.org.readthedocs.build/en/2247/docs/Documentation.html

myk002:
btw, I wouldn't worry about `ban-cooking show`. It wasn't even in the [docs](https://dfhack--2247.org.readthedocs.build/en/2247/docs/tools/ban-cooking.html). If you don't need it for development, I'd say skip it.

ban-cooking.lua
```diff
@@ -0,0 +1,262 @@
...
+--[====[
+
+ban-cooking
+===========
+A more convenient way to ban cooking various categories of foods than the
+kitchen interface.  Usage:  ``ban-cooking <type>``.  Valid types are ``booze``,
+``honey``, ``tallow``, ``oil``, ``seeds`` (non-tree plants with seeds),
+``brew``, ``fruit``, ``mill``, ``thread``, and ``milk``.
+
+]====]
```
myk002, line 12: Same here. Docs now go in `docs/ban-cooking.rst`.

myk002:
Others are working now on porting the ruby scripts. I'll merge this so we don' t lose your work and integrate it with the changes others are making.

myk002:
pre-commit.ci autofix

lethosor (commented):
Worth noting that the CLI arguments are changing here. They aren't changing in the version discussed on discord so far.

PR #430, opened 2022-09-05
Improves devel/query usability

cppcooper:
Increase the default maxlength value and prints a message when queries are truncated.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,7 +19,7 @@ that repo.
 - `warn-stealers`: warn when creatures that may steal your food, drinks, or items become visible
 
 ## Fixes
-- `devel/query`: updates call to deleted function that was causing an error for tile data
+- `devel/query`: fixed error when --tile is specified
 - `gui/unit-info-viewer`: fix logic for displaying undead creature names
 - `gui/workflow`: restore functionality to the add/remove/order hotkeys on the workflow status screen
 - `emigration`: fix emigrant logic so unhappy dwarves leave as designed
--- a/changelog.txt
+++ b/changelog.txt
@@ -25,6 +25,7 @@ that repo.
 - `gui/gm-unit`: allow ``+`` and ``-`` to adjust skill values as intended instead of letting the filter intercept the characters
 
 ## Misc Improvements
+- `devel/query`: improved usability by informing the user when a query has been truncated
 - `gui/gm-unit`: don't clear the list filter when you adjust a skill value
 - `gui/quickfort`: better formatting for the generated manager orders report
 - `gui/quickfort`: display an error message when the blueprints directory cannot be found
--- a/changelog.txt
+++ b/changelog.txt
@@ -26,7 +26,7 @@ that repo.
 - `gui/gm-unit`: allow ``+`` and ``-`` to adjust skill values as intended instead of letting the filter intercept the characters
 
 ## Misc Improvements
-- `devel/query`: improved usability by informing the user when a query has been truncated
+- `devel/query`: inform the user when a query has been truncated due to ``--maxlength`` being hit.
 - `devel/query`: increased default maxlength value from 257 to 2048
 - `gui/gm-unit`: don't clear the list filter when you adjust a skill value
 - `gui/quickfort`: better formatting for the generated manager orders report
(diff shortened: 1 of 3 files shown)


changelog.txt
```diff
@@ -19,12 +19,15 @@ that repo.
...
 ## Fixes
+- `devel/query`: updates call to deleted function that was causing an error for tile data
 - `gui/unit-info-viewer`: fix logic for displaying undead creature names
 - `gui/workflow`: restore functionality to the add/remove/order hotkeys on the workflow status screen
 - `emigration`: fix emigrant logic so unhappy dwarves leave as designed
 - `gui/gm-unit`: allow ``+`` and ``-`` to adjust skill values as intended instead of letting the filter intercept the characters
 
 ## Misc Improvements
+- `devel/query`: improved usability by informing the user when a query has been truncated
+- `devel/query`: increased default maxlength value from 257 to 2048
```
myk002, line 30: ha, I should probably go back and modify all the example `devel/query` commands I put in the help text. I had added explicit `--maxlenth` parameters because the default was so small.

changelog.txt
```diff
@@ -19,12 +19,15 @@ that repo.
...
 
 ## Fixes
+- `devel/query`: updates call to deleted function that was causing an error for tile data
 - `gui/unit-info-viewer`: fix logic for displaying undead creature names
 - `gui/workflow`: restore functionality to the add/remove/order hotkeys on the workflow status screen
 - `emigration`: fix emigrant logic so unhappy dwarves leave as designed
 - `gui/gm-unit`: allow ``+`` and ``-`` to adjust skill values as intended instead of letting the filter intercept the characters
 
 ## Misc Improvements
+- `devel/query`: improved usability by informing the user when a query has been truncated
```
myk002, line 29:
```suggestion
- `devel/query`: inform the user when a query has been truncated due to ``--maxlength`` being hit.
```

changelog.txt
```diff
@@ -19,12 +19,15 @@ that repo.
 - `warn-stealers`: warn when creatures that may steal your food, drinks, or items become visible
 
 ## Fixes
+- `devel/query`: updates call to deleted function that was causing an error for tile data
```
myk002, line 22: I'm not sure this is understandable by readers of the changelog. The fact that the error was caused by calling a deleted function is not useful to tell users. Could you rephrase this as something like "fixed error when [use case]"?

devel/query.lua
```diff
@@ -274,6 +274,7 @@ function foreach(table, name, callback)
     if getmetatable(table) and table._kind and (table._kind == "enum-type" or table._kind == "bitfield-type") then
         for idx, value in ipairs(table) do
             if is_exceeding_maxlength(index) then
+                print("<query on " .. name .. " truncated after " .. index .. ">")
```
myk002, line 277: instead of duplicating this message, could you integrate it into `is_exceeding_maxlength` (passing it the `name` param) or, if not that, make a wrapper around `is_exceeding_maxlength` that prints the message?

myk002 (changes requested):

changelog.txt
myk002: In other words, try to answer the question for the user: "What can I do now that I couldn't do before?"

cppcooper: It is a development script, does that matter here?

cppcooper: As a side note, I don't precisely know all of what this means for using the script. I just found what was causing the error and fixed it, it may have other edge cases than just `--tile`

myk002: I think that's ok if it fixes other things too. Rephrasing the message as "fixed error when --tile is specified" is definitely good enough.

cppcooper:
```suggestion
- `devel/query`: fixed error when --tile is specified
```

devel/query.lua
cppcooper: "into" I don't like, cause it's an `is_*` function but I could wrap it or rename it and put it into the renamed. I'll try and do this tomorrow

changelog.txt
cppcooper: I picked the number, mostly, out of thin air. I wasn't sure what the longest standard tables users would be looking at would be. What I did know is there are some crazy long tables that really slow down searches.
I presume I can resolve this though, or should I make a note in wherever that doc pr is?

devel/query.lua
cppcooper: Actually a wrapper doesn't work either; needs to return when it is exceeding. I just put the print message in a nested-function.

myk002 (approved):

PR #436, opened 2022-09-12
[gui/launcher] reuse the current view to avoid flicker on command run

myk002:
as discussed on discord


lethosor (approved):
Does the trick on my end!

myk002:
Thanks!

PR #437, opened 2022-09-12
[gui/launcher] dismiss launcher if parent screen changes

myk002:
Not just if the new parent screen is a dfhack screen.

As discussed on Discord.


lethosor:
> I didn't see an `addressof()` function.

It's buried in the `utils` module

PR #439, opened 2022-09-14
Fix incorrect warn-stealers documentation and remove deprecated header.

wolfboyft:


docs/warn-stealers.rst
```diff
@@ -14,4 +14,4 @@ Usage
 
 ::
 
-    warn-stealers [start|stop]
+    warn-stealers [enable|disable]
```
myk002, line 17: Since it works with the enable API, could we document this as:
`enable warn-stealers`
?
I'm trying to standardize on "enable"/"disable" as the command instead of the argument (to make scripts and plugins look less distinct)

myk002: I've also just been documenting the "enable" path (`enable warn-stealers`) instead of something like
`enable|disable warn-stealers` in order to make the syntax less confusing. Not sure if this is the right decision, but it's how all the other enableable tools are documented right now.

myk002: the long term plan here with enableable tools is to collect them into an in-game configuration screen where they can be toggled. having them all use the `enable` syntax then will make much more sense.

myk002 (approved):

PR #460, opened 2022-11-03
Add -dead flag to unretire-anyone

Tjudge1:
pre-commit.ci autofix


myk002:
Instead of a new script, could we possibly integrate the new functionality into the existing `unretire-anyone.lua` script? is this something that even needs a commandline parameter, or can it just be the new behavior of the `unretire-anyone` script?

devel/revive-hf.lua
```diff
@@ -0,0 +1,40 @@
...
+Usage::
+
+    devel/revive-hf HISTFIG_ID
+
+Arguments:
+
+``histfig_id``:
+    the ID of the historical figure to target
+
+]====]
```
myk002, line 19: Note that documentation should now go in the `docs/` subdirectory. If we keep this as a separate file, you can (should) remove the doc comment inside the script file.

devel/revive-hf.lua
```diff
@@ -0,0 +1,40 @@
...
+--[====[
+
+devel/revive-hf
+=============
+
+Revivs the specified historical figure, even if off-site.
+
+Usage::
+
+    devel/revive-hf HISTFIG_ID
```
myk002, line 12: `full-heal` is already associated with healing and reviving units. Instead of adding a new script, could we perhaps add a `--hf` parameter to `full-heal` so it can work on historical figures?

myk002 (changes requested):

devel/revive-hf.lua
Tjudge1: I'm not sure. You wouldn't want to add --hf as a parameter to full-heal with this particular code though. 

Full-heal will only target units that are within the current map. And you can't mess with a unit's body data until they're loaded onto the map, as far as I know.

A work around that Rumrusher suggested, that would line up with full-heal's current functionality, is to spawn in a creature matching a dead histfig's description and then to link the unit's nemesis/soul together up with a new creature.

Tjudge1 (commented):
I'll create an rst file over in docs

Tjudge1:
Added unretire-literally-anyone.rst to docs.
Removed comments from the lua file.

devel/unretire-literally-anyone.lua
```diff
@@ -0,0 +1,69 @@
+local dialogs = require 'gui.dialogs'
```
myk002, line 1: The diff between this and the existing `unretire-anyone.lua` script is:
```
55,57c33,41
<       if (histFig.died_year == -1 or histFlags.ghost) and not histFlags.deity and not histFlags.force then
<         local creature = df.creature_raw.find(histFig.race).caste[histFig.caste]
<         local name = creature.caste_name[0]
---
> 
>       local creature = df.creature_raw.find(histFig.race).caste[histFig.caste]
>       local name = creature.caste_name[0]
>         if histFig.died_year >= -1 then
>             histFig.old_year = df.global.cur_year
>             histFig.old_seconds = df.global.cur_year_tick + 1
>             histFig.died_year = -1
>             histFig.died_seconds = -1
>         end
79d62
<     end
```
Why is this a separate script instead of a modification to the existing script?

devel/revive-hf.lua
```diff
@@ -0,0 +1,40 @@
+-- Revives the specified historical figure
+
+--[====[
+
+devel/revive-hf
+=============
+
+Revivs the specified historical figure, even if off-site.
```
myk002, line 8: In the docs for this script (which should be in a .rst file in the docs dir), could you include *why* you'd want to revive a historical figure? Give readers some idea about what situations this would be useful.

devel/revive-hf.lua
```diff
@@ -0,0 +1,40 @@
+-- Revives the specified historical figure
+
+--[====[
+
+devel/revive-hf
```
myk002, line 5: is this script intended for developers instead of players? it seems to me that it could just go in the root scripts dir.

devel/revive-hf.lua
```diff
@@ -0,0 +1,40 @@
...
+
+local target_hf = -1
+
+for _, arg in ipairs({...}) do
+    if tonumber(arg) and target_hf == -1 then
+        target_hf = tonumber(arg)
+    else
+        qerror('unrecognized argument: ' .. arg)
+    end
+end
```
myk002, line 29: Since you only interpret a single argument anyway, this could be simplified down to something like:
```
local args = {...}
local target_hf = tonumber(args[1])
if not target_hf then qerror('please pass the ID of the target historical figure') end
```

myk002 (changes requested):

devel/unretire-literally-anyone.lua
myk002: (when you integrate into the existing script, please feel free to remove in the in-script help text. We're trying to clean those up and get rid of them as we do these PRs)

AtomicChicken:
> Instead of a new script, could we possibly integrate the new functionality into the existing `unretire-anyone.lua` script? is this something that even needs a commandline parameter, or can it just be the new behavior of the `unretire-anyone` script?

As the original author of `unretire-anyone`, I'd prefer this behaviour to be controlled by an optional parameter if it's to be merged. Aside from bloating the figure selection list with dead creatures, reviving dead historical figures in this manner feels incomplete and may cause issues down the line.

If I recall correctly, `old_year` and `old_seconds` specify the date at which mortal figures are scheduled to die of old age. Is there any particular reason why they're being set to the same date the figure has been revived on?

devel/revive-hf.lua
Tjudge1: Yeah, mainly because it doesn't give the hf a new body yet.

Tjudge1:
I don't recall, but I think you're right. I'll update that.

devel/unretire-literally-anyone.lua
Tjudge1: Because I'm new here, and wasn't sure what the protocol was. And was slightly uncomfortable with modifying somebody else's script wholesale without getting prior approval from the author.

Tjudge1 (commented):
kk

Tjudge1 (commented):
kk

Tjudge1 (commented):
kk

Tjudge1:
pre-commit.ci autofix

Tjudge1 (commented):
kk

Tjudge1 (commented):
1

devel/unretire-literally-anyone.lua
myk002: I think that behavior is granted. could you add the functionality to the existing `unretire-anyone` and add a flag to enable it?

devel/revive-hf.lua
myk002: could we work that functionality it so that the script is generally useful? what more do you think would need to be done to make it worthy of the top dir?

devel/revive-hf.lua
myk002: unresolving since this comment still looks relevant

devel/unretire-literally-anyone.lua
myk002: There's a nice simple example of how to process argument flags in autonick: https://github.com/DFHack/scripts/blob/master/autonick.lua#L32-L38

docs/devel/revive-hf.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Any future desired future interactions with an NPC, who died at an unknown spot or inaccessible location,
+can have be revived to a degree using this function.
+
+This function is still in development, and may cause strange bugs.
+This is because the revived historical figures will lack a physical body associated with them.
+
+Usage
+-----
+
```
myk002, line 14: you need a `::` here to indicate that the indented lines are a code block

docs/devel/revive-hf.rst
```diff
@@ -0,0 +1,20 @@
+devel/revive-hf
+===============
+
```
myk002, line 3: remember to add a `dfhack-tool` directive here. See https://docs.dfhack.org/en/stable/docs/Documentation.html#header-format for details. Is this script specifically intended for adventure mode?

devel/unretire-literally-anyone.lua
Tjudge1: Done.

devel/revive-hf.lua
Tjudge1: Resolved.

Tjudge1 (commented):
removed

Tjudge1 (commented):
removed

docs/unretire-anyone.rst
```diff
@@ -22,3 +22,10 @@ Usage
 ::
 
     unretire-anyone
+
+Options:
```
myk002, line 26: Check out https://docs.dfhack.org/en/stable/docs/Documentation.html#options for how to format the options section

unretire-anyone.lua
```diff
@@ -52,9 +37,13 @@ function showNemesisPrompt(advSetUpScreen)
     if nemesis.figure and not nemesis.flags.ADVENTURER then -- these are already available for unretiring
       local histFig = nemesis.figure
       local histFlags = histFig.flags
-      if (histFig.died_year == -1 or histFlags.ghost) and not histFlags.deity and not histFlags.force then
+      if (histFig.died_year == -1 or histFlags.ghost or options.dead == true) and not histFlags.deity and not histFlags.force then
```
myk002, line 40: No need to check for `== true`. Just "or options.dead" is sufficient.

Tjudge1: Still know literally nothing about Lua. Thanks!

Tjudge1 (commented):
yup.

myk002 (approved):

myk002 (approved):

PR #461, opened 2022-11-06
create-unit: rework naming; add nemesis civless save data; add arg checks

Shim-Panze:
Mostly, made naming more accurately reflect the game's behavior. Added option to generate "wild" name (i.e. civless name) by passing `-name` empty. Made nemesis records created without a civId have proper save data. Fixed some argument checks and added more so civ_id has to actually be for a civilization and group_id can't be a civilization (also implies they can't be the same). Made history event generation take current year into account, and add history event for histfig joining group as well, to better emulate game behavior (using void migrants as reference).

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -24,6 +24,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/blueprint`: support new blueprint phases and options
+- `modtools/create-unit`: better unit naming, more argument checks, assign nemesis save data for units without civilization
 
 ## Removed
 - `gui/create-item`: removed ``--restricted`` option. it is now the default behavior
--- a/changelog.txt
+++ b/changelog.txt
@@ -24,7 +24,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/blueprint`: support new blueprint phases and options
-- `modtools/create-unit`: better unit naming, more argument checks, assign nemesis save data for units without civilization
+- `modtools/create-unit`: better unit naming, more argument checks, assign nemesis save data for units without civilization so they can be properly saved when offloaded
 
 ## Removed
 - `gui/create-item`: removed ``--restricted`` option. it is now the default behavior
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -448,13 +448,13 @@ local function allocateNewChunk(hist_entity)
     hist_entity.save_file_id = df.global.unit_chunk_next_id
     df.global.unit_chunk_next_id = df.global.unit_chunk_next_id+1
     hist_entity.next_member_idx = 0
-    print("allocating chunk:",hist_entity.save_file_id)
+    print("allocating chunk for entity "..hist_entity.id..":",hist_entity.save_file_id)
   else
     local chunkInfo = df.global.world.worldgen
     chunkInfo.next_unit_chunk_id = df.global.unit_chunk_next_id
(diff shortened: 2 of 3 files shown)


modtools/create-unit.lua
```diff
@@ -56,8 +56,12 @@ Creates a unit.  Usage::
 
     -name entityRawName
         Set the unit's name to be a random name appropriate for the
-        given entity. \\LOCAL can be specified instead to automatically
-        use the fort group entity in fortress mode only.
+            given entity.
```
myk002, line 59: Since you're editing this file, could you remove this entire string and use `dfhack.script_help()` instead where it is used? Help text has been moved to the docs directory. See https://github.com/DFHack/scripts/blob/12d78b25994521451392d8538c0d2db22499bfe3/quickfort.lua#L54 for an example

myk002 (changes requested):
I'll review more when I have an actual computer and not just my phone

modtools/create-unit.lua
Shim-Panze: Done [8dd1dd5](https://github.com/DFHack/scripts/pull/461/commits/8dd1dd5fffe3e348546faba6ec79caaa506edc6b)

modtools/create-unit.lua
```diff
@@ -595,22 +444,47 @@ function getRaceCasteIDs(raceStr, casteStr)
...
+    hist_entity.save_file_id = df.global.unit_chunk_next_id
+    df.global.unit_chunk_next_id = df.global.unit_chunk_next_id+1
+    hist_entity.next_member_idx = 0
+    print("allocating chunk:",hist_entity.save_file_id)
+  else
+    local chunkInfo = df.global.world.worldgen
+    chunkInfo.next_unit_chunk_id = df.global.unit_chunk_next_id
+    df.global.unit_chunk_next_id = df.global.unit_chunk_next_id+1
+    chunkInfo.next_unit_chunk_offset = 0
+    print("allocating chunk:",chunkInfo.next_unit_chunk_id)
```
myk002, line 457: would it be useful to differentiate the printed statements so the user can tell which branch was taken?

myk002 (changes requested):
nice work! could you add a changelog.txt entry for these improvements?

modtools/create-unit.lua
Shim-Panze: It certainly won't hurt

Shim-Panze: That should do it [6b93420](https://github.com/DFHack/scripts/pull/461/commits/6b93420f22cfefc74aeac53f00ba47ee80fb6257)

changelog.txt
```diff
@@ -24,6 +24,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/blueprint`: support new blueprint phases and options
+- `modtools/create-unit`: better unit naming, more argument checks, assign nemesis save data for units without civilization
```
myk002, line 27: Could you add a little info about why adding nemesis data is important?

myk002 (changes requested):
Overall looks good! Could you fix the pre commit and lint errors, though?

myk002 (approved):

myk002:
If you click on the pre-commit failure, you can get the magic words to have pre-commit fix it for you

Shim-Panze:
pre-commit.ci autofix

PR #465, opened 2022-11-18
Updates dwarf-op help doc, and fixes processing bug for names

cppcooper:
Precisely as the title says

### DIFF
--- a/dwarf-op.lua
+++ b/dwarf-op.lua
@@ -40,167 +40,6 @@ local validArgs = utils.invert({
 })
 local args = utils.processArgs({...}, validArgs)
 
-local help = [====[
-
-dwarf-op
-========
-Dwarf optimization is a script designed to provide a robust solution
-to hacking dwarves to be better at work. The primary use case is as follows:
-
- 1) take a dwarf
- 2) delete their ability to do anything, even walk (job skills, phyiscal/mental attributes)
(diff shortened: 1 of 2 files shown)


dwarf-op.lua
```diff
@@ -1076,15 +915,19 @@ function CheckWorker(dwf, option)
             local select_type = list[1]
             if GetChar(select_type,1) == 'p' or select_type == 'name' or select_type == 'names' then
                 includeProtectedDwfs = true
-                select_type = string.sub(select_type,2)
+                if GetChar(select_type,1) == 'p' then
```
myk002, line 918: `GetChar` is fine since it's used in other places, but be aware that dfhack.lua also adds `string:startswith()`, so when getting the first character and comparing it you could write it as `if select_type:startswith('p')` instead of `if GetChar(select_type, 1) == 'p'`

myk002 (approved):
changelog entry?

cppcooper:
Adding a change log entry, but before merging this.. I was thinking that perhaps this should be renamed to `modify-dwarves` to be more "discoverable"

myk002:
> I was thinking that perhaps this should be renamed to `modify-dwarves` to be more "discoverable"

if we do, we can do it in another PR. naming is hard and could spark debate.

PR #479, opened 2022-12-08
DFHack/dfhack#2081: Port adaptation and ban-cooking from Ruby to Lua

EarthPulseAcademy:
DFHack/dfhack#2081


adaptation.lua
```diff
@@ -0,0 +1,128 @@
...
+    dfhack.color(color)
+    dfhack.print(s)
+    dfhack.color(COLOR_RESET)
+end
+
+local function usage(s)
+    if s ~= nil then
+        print(s)
+    end
+    print("Usage: adaptation <show|set> <him|all> [value]")
```
myk002, line 37: Yeah, use `print(dfhack.script_help())` here

adaptation.lua
```diff
@@ -0,0 +1,128 @@
...
+local who = args[2]
+local value = args[3]
+
+if mode == "help" then
+    usage()
+elseif mode ~= "show" and mode ~= "set" then
+    usage("Invalid mode '" .. mode .. "': must be either 'show' or 'set'")
+end
+
+if who == nil then
```
myk002, line 52: Here and elsewhere, you usually don't compare against nil in Lua unless you need to differentiate between nil and other firms of falseness. Just `if not who then` is idiomatic.

adaptation.lua
```diff
@@ -0,0 +1,128 @@
...
+    if tonumber(value) < 0 or tonumber(value) > 800000 then
+        usage("Value must be between 0 and 800000")
+    end
+    value = tonumber(value)
+
+end
+
+local num_set = 0
+
+local function set_adaptation_value(unit, v)
```
myk002, line 74: Try to put functions above executing code

myk002 (commented):
More comments coming later, but let me send what I have so far. Also, put #2071 in the description, not the title, to link the PR to the issue

EarthPulseAcademy:
pre-commit.ci autofix

adaptation.lua
```diff
@@ -0,0 +1,128 @@
+-- Original opening comment before lua adaptation
+-- View or set cavern adaptation levels
+-- based on removebadthoughts.rb
+
+-- Rewritten by TBSTeun using OpenAI GPT from adaptation.rb
+
+local help = [====[
+
```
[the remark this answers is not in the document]
EarthPulseAcademy: done

adaptation.lua
```diff
@@ -0,0 +1,128 @@
...
+The ``value`` must be between 0 and 800,000 (inclusive).
+
+]====]
+
+-- Color constants, values mapped to color_value enum in include/ColorText.h
+local COLOR_RESET = -1
+local COLOR_GREEN = 2
+local COLOR_RED = 4
+local COLOR_YELLOW = 14
```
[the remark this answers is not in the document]
EarthPulseAcademy: done

adaptation.lua
EarthPulseAcademy: done

adaptation.lua
EarthPulseAcademy: done

EarthPulseAcademy:
pre-commit.ci autofix

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+end
+
+local args = {...}
+
+local mode = args[1] or "help"
+local who = args[2]
+local value = args[3]
+
+if mode == "help" then
+    usage()
```
myk002, line 70: you might want to `return` after every call to `usage()` here and below. Otherwise, you might incorrectly run processing code on bad input. For example, a command like `adaptation help him please` will print out help text and then throw an error when 'please' is used as a number.

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+
+local function print_color(color, s)
+    dfhack.color(color)
+    dfhack.print(s)
+    dfhack.color(COLOR_RESET)
+end
+
+local function usage(s)
+    if s then
+        print(s)
```
myk002, line 14: since the message passed to `usage` is always an error message, maybe use `dfhack.printerr()` instead of `print()` here. You can make the swap without any other changes; e.g. you still will not need to manually tack on a newline.

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+local args = {...}
+
+local mode = args[1] or "help"
+local who = args[2]
+local value = args[3]
+
+if mode == "help" then
+    usage()
+elseif mode ~= "show" and mode ~= "set" then
+    usage("Invalid mode '" .. mode .. "': must be either 'show' or 'set'")
```
myk002, line 72: while `..` concatenation does work fine, be aware that it is often easier to read when you use `string:format`. This code could be written as:
```lua
usage(('Invalid mode "%s": must be either "show" or "set"'):format(mode))
```

myk002: note also that while double and single quotes are interchangeable, the Lua code in DFHack prefers using single quotes for strings. It's fine to keep this PR as it is as far as quote style, but it's something to remember for future code.

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+end
+
+if not(who) then
```
myk002, line 75:
```suggestion
if not who then
```

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+    usage("Target not specified")
+elseif who ~= "him" and who ~= "all" then
+    usage("Invalid target '" .. who .. "'")
+end
+
+if mode == "set" then
+    if not(value) then
```
myk002, line 82:
```suggestion
    if not value then
```

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+        usage("Value not specified")
+    elseif not tonumber(value) then
+        usage("Invalid value '" .. value .. "'")
+    end
+
+    if tonumber(value) < 0 or tonumber(value) > 800000 then
+        usage("Value must be between 0 and 800000")
+    end
+    value = tonumber(value)
+
```
myk002, line 92:
```suggestion
    value = tonumber(value)
    if value < 0 or value > 800000 then
        usage("Value must be between 0 and 800000 (inclusive)")
    end
```

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+end
+
+if who == "him" then
+    local u = dfhack.gui.getSelectedUnit(true)
+    local num_set = 0
+    if u then
+        num_set = set_adaptation_value(u, value)
```
myk002, line 99: you don't need to "eat" a returned value if you don't use it. Idiomatic Lua drops unneeded return values and missing/extra params all the time.

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+    else
+        print('Please select a dwarf ingame')
```
myk002, line 101: consider `dfhack.printerr` instead of `print` for "error" messages

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+    end
+elseif who == "all" then
+    for _, uu in ipairs(df.global.world.units.all) do
```
myk002, line 104: need a `local num_set = 0` before this line. The one in the other if block is not in scope here, and this effectively names `num_set` as a global variable.

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+        num_set = set_adaptation_value(uu, value)
```
myk002, line 105: `+=` (which I think you meant to write here) isn't an operator in Lua. You have to write it longhand: `num_set = num_set + set_adaptation_value(uu, value)`

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+
+local function usage(s)
+    if s then
+        print(s)
+    end
+    print(dfhack.script_help())
+end
+
+local function set_adaptation_value(unit, v)
+    local num_set = 0
```
myk002, line 20: since this function can only ever return 0 or 1, you can get rid of this variable and return the numbers directly, I think.

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+
+    if not dfhack.units.isCitizen(unit) or unit.flags2.killed then
+        return num_set
```
myk002, line 23:
```suggestion
        return 0
```

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+    if s then
+        print(s)
+    end
+    print(dfhack.script_help())
+end
+
+local function set_adaptation_value(unit, v)
+    local num_set = 0
+
+    if not dfhack.units.isCitizen(unit) or unit.flags2.killed then
```
myk002, line 22: `not dfhack.units.isAlive(unit)` might be better to check than just the `killed` flag directly.

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+        return num_set
+    end
+    local trait_found = false
```
myk002, line 25: don't need this if you return directly from the loop

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+                        print_color(COLOR_RED, t.value .. "\n")
+                    end
+            elseif mode == "set" then
+                print("Unit " .. unit.id .. " (" .. dfhack.TranslateName(dfhack.units.getVisibleName(unit)).. ") changed from " .. t.value .. " to " .. v)
+                t.value = v
+                num_set = num_set + 1
+            end
+            trait_found = true
+            break
+        end
```
myk002, line 44:
```suggestion
        if t.id ~= df.misc_trait_type.CaveAdapt then goto continue end
        if mode == "show" then
            print_color(COLOR_RESET, ('Unit %d (%s) has an adaptation of '):format(unit.id, dfhack.TranslateName(dfhack.units.getVisibleName(unit)))
            if t.value <= 399999 then
                print_color(COLOR_GREEN, t.value .. "\n")
            elseif t.value <= 599999 then
                print_color(COLOR_YELLOW, t.value .. "\n")
            else
                print_color(COLOR_RED, t.value .. "\n")
            end
            return 0
        end
        print("Unit " .. unit.id .. " (" .. dfhack.TranslateName(dfhack.units.getVisibleName(unit)).. ") changed from " .. t.value .. " to " .. v)
        t.value = v
        return 1
        ::continue::
    end
```

myk002: Lua doesn't have a formal `continue` statement. Instead there is an idiom using `goto` and a "continue" label.

myk002: (btw I left the formatting of the second print statement as an exercise for the reader : )

adaptation.lua
```diff
@@ -0,0 +1,110 @@
...
+            local new_trait = df.unit_misc_trait:new()
+            new_trait.id = df.misc_trait_type.CaveAdapt
+            new_trait.value = v
+            num_set = num_set + 1
+            table.insert(unit.status.misc_traits, new_trait)
+            print("Unit " .. unit.id .. " (" .. dfhack.TranslateName(dfhack.units.getVisibleName(unit)) .. ") changed from 0 to " .. v)
+        end
+    end
+
+    return num_set
```
myk002, line 60: this block can be similarly simplified

myk002 (commented):
reviewed adaptation, will review ban-cooking next

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+    kitchen.mat_indices:insert('#', mat_index)
+    kitchen.item_types:insert('#', type)
+    kitchen.item_subtypes:insert('#', subtype)
+    kitchen.exc_types:insert('#', df.kitchen_exc_type.Cook)
+    already_banned[key] = true
+end
+
+local funcs = {}
+
+funcs["booze"] = function()
```
myk002, line 69: `funcs["booze"]` can be simplified to just `funcs.booze`. you only need to use the first form if the string is not a valid identifier (like "hyphen-word")

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+                    end
+                end
+            end
+        end
+    end
+end
+
+funcs["seeds"] = function()
+    for _, p in ipairs(df.global.world.raws.plants.all) do
+        if p.material_defs.type.seed ~= -1 and p.material_defs.idx.seed ~= -1 and not p.flags.TREE then
```
myk002, line 143: you can invert the condition and use `goto continue` for this one to remove an entire level of indentation from the function

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+                    end
+                end
+            end
+        end
+    end
+end
+
+funcs["brew"] = function()
+    for _, p in ipairs(df.global.world.raws.plants.all) do
+        if p.material_defs.type.drink ~= -1 and p.material_defs.idx.drink ~= -1 then
```
myk002, line 180: same here, invert the condition and use `goto` to remove an unneeded level of indentation

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+                    ban_cooking(p.name .. ' ' .. m.id, matinfo.type, matinfo.index, df.item_type.PLANT, -1)
+                end
+            end
+        end
+    end
+end
+
+funcs["thread"] = function()
+    for _, p in ipairs(df.global.world.raws.plants.all) do
+        if p.material_defs.idx.thread ~= -1 then
```
myk002, line 223: here too. use `goto`

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+                    end
+                end
+            end
+        end
+    end
+end
+
+funcs["show"] = function()
+    -- First put together a dictionary/hash table
+    type_list = {}
```
myk002, line 270: needs `local`

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+
+        type_list[key_seed]["text"] = p.name .. " seed"
+        -- drink materials for plants always appear to use the 'SEEDS' item type tag
+        type_list[key_seed]["type"] = "SEEDS"
+        -- item subtype of 'SEEDS' types appears to always be -1, as there is no growth array entry for the 'SEEDS'
+        type_list[key_seed]["subtype"] = -1
+
+        for r, g in ipairs(p.growths) do
+            matinfo = dfhack.matinfo.decode(g)
+            m = matinfo.material
```
myk002, line 316: need `local` for these two

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+
+            if m.flags["EDIBLE_COOKED"] and m.flags["LEAF_MAT"] then
+                for j, s in ipairs(p.material) do
+                    if m.id == s.id then
+
+                        -- need to create a key to reference values in the types_list dictionary.
+                        key_matinfo_type = matinfo.type .. i
```
myk002, line 323: need `local`

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+                -- item type for milk appears to use the 'GLOB' tag
+                type_list[key_matinfo_type]["type"] = "GLOB"
+                type_list[key_matinfo_type]["subtype"] = -1
+            end
+        end
+    end
+
+    iterator = 0
+
+    output = {}
```
myk002, line 377: needs `local` for these two

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+
+    for i, b in pairs(already_banned) do
```
myk002, line 379: `already_banned` cannot be persisted in global state since it would get out of sync with manual "unbanning" actions taken by the player. therefore this will only work to show items that were *just* banned in the current invocation of the script. This is fine, but I wanted to make sure you were aware of the implications. You'd have to use it like `ban-cooking fruit tallow show`. just running `ban-cooking show` by itself would show nothing.

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+        -- initialize our output string with the array entry position (largely stays the same for each item on successive runs, except when items are added/removed)
+        output[iterator] = ''
+        -- initialize our key for accessing our stored items info
+        key = i[1] .. i[2]
```
myk002, line 383: needs `local`

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+        end
+    end
+
+    iterator = 0
+
+    output = {}
+
+    for i, b in pairs(already_banned) do
+        -- initialize our output string with the array entry position (largely stays the same for each item on successive runs, except when items are added/removed)
+        output[iterator] = ''
```
myk002, line 381: it would be cleaner to declare a `local cur` variable here, populate it, and then add it to the output at the end of the loop with `table.insert(output, cur)`. no need to track the index with `iterator`.

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+
+local function already_banned_has_key(key)
+    for k, v in pairs(already_banned) do
+        if k[1] == key[1] and k[2] == key[2] and k[3] == key[3] and k[4] == key[4] then
+            return true
+        end
+    end
+
+    return false
+end
```
myk002, line 27:
```suggestion
local function make_key(mat_type, mat_index, type, subtype)
    return ('%s:%s:%s:%s'):format(mat_type, mat_index, type, subtype)
end
```

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+
+-- Iterate over the elements of the kitchen.item_types list
+for i = 0, #kitchen.item_types - 1 do
+    -- Check if the kitchen.exc_types[i] element is equal to :Cook
+    if kitchen.exc_types[i] == df.kitchen_exc_type.Cook then
+        -- Add a new element to the already_banned dictionary
+        already_banned_key = {kitchen.mat_types[i], kitchen.mat_indices[i], kitchen.item_types[i], kitchen.item_subtypes[i]}
```
myk002, line 34:
```suggestion
        local key = make_key(kitchen.mat_types[i], kitchen.mat_indices[i], kitchen.item_types[i], kitchen.item_subtypes[i])
```

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+        if not(already_banned_has_key(already_banned_key)) then
+            already_banned[already_banned_key] = true
+        end
```
myk002, line 37:
```suggestion
        already_banned[key] = true
```

ban-cooking.lua
```diff
@@ -0,0 +1,440 @@
...
+        if s.value == str then
+            return true
+        end
+    end
+
+    return false
+end
+
+local function ban_cooking(print_name, mat_type, mat_index, type, subtype)
+    key = mat_type, mat_index, type, subtype
```
myk002, line 52:
```suggestion
    local key = make_key(mat_type, mat_index, type, subtype)
```

myk002 (changes requested):

adaptation.lua
EarthPulseAcademy: I feel like using function calls over operators is what is more confusing because there is more information to read. Consider the '+' operator for example, what feels more confusing? 1+1 or add(1, 1)?
That being said, I feel like this is a larger discussion than for pull request comments. I believe consistency is key in making a codebase approachable for new developers and if using string.format is the standard in our coding practices I'll comply and make the change later today. Taking that into account, I will also make the same change on line 39 even though you said leave it in as a reader exercise to keep the codebase approachable for new developers.

adaptation.lua
EarthPulseAcademy: done

adaptation.lua
EarthPulseAcademy: I actually did this so I could make the availability scope for num_set as small as possible, by only declaring it in the if statement and then filling it with the return value of set_adaptation_value. I am actually using this variable to print the amount of units changed. The ruby way was to set num_set in the root scope of the script and using that variable in multiple places. Keeping variables as local as possible reduces bugs.

adaptation.lua
EarthPulseAcademy: I actually didn't. Since him and all are mutually exclusive options here I don't care about what was in num_set previously and can safely discard its value.

adaptation.lua
EarthPulseAcademy: Assuming you mean string concatenation with your print statement comment, I made the change.

ban-cooking.lua
EarthPulseAcademy: Maybe I already fixed this issue with some of my other changes, but I just tested using ban-cooking all before saving and fully exiting out of df, restarting it, reloading the save, and calling ban-cooking show with the output being equal to before I restarted df. Also I manually banned and unbanned some items in my kitchen interface in df and the changes show up on running ban-cooking show. Am I missing something here?

adaptation.lua
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    return 0
+end
+
+if mode == 'help' then
+    usage()
+    return
+elseif mode ~= 'show' and mode ~= 'set' then
+    usage(('Invalid mode %s: must be either \'show\' or \'set\''):format(mode))
```
myk002, line 68: it's usually easier to use whatever the "other" quotes are in the string to avoid escaping them

adaptation.lua
```diff
@@ -0,0 +1,115 @@
...
+
+    value = tonumber(value)
+    if value < 0 or value > 800000 then
+        usage(('Value must be between 0 and 800000 (inclusive), input value was %s'):format(value))
+    end
+end
+
+local num_set = 0
+
+print("who")
```
myk002, line 97: leftover debug print?

adaptation.lua
```diff
@@ -0,0 +1,115 @@
...
+
+if who == 'him' then
+    local u = dfhack.gui.getSelectedUnit(true)
+    if u then
+        num_set = set_adaptation_value(u, value)
+    else
+        dfhack.printerr('Please select a dwarf ingame')
+    end
+elseif who == 'all' then
+    print("all")
```
myk002, line 107: leftover debug print?

ban-cooking.lua
EarthPulseAcademy: I tried to change the key to a different format, but the table as a key is used later, and I am not sure how I can change this key format without losing needed data in the show function

adaptation.lua
```diff
@@ -0,0 +1,115 @@
...
+end
+
+local num_set = 0
+
+print("who")
+
+if who == 'him' then
+    local u = dfhack.gui.getSelectedUnit(true)
+    if u then
+        num_set = set_adaptation_value(u, value)
```
myk002, line 102: this value is never read after being assigned here. you might as well just not assign the return value from `set_adaptation_value(u, value)` to anything and move the `local num_set = 0` from line 95 into the `elseif who == 'all' then` block below.

docs/ban-cooking.rst
```diff
@@ -12,11 +12,16 @@ Usage
 -----
 
 ::
-    ban-cooking <type>
+    ban-cooking <type|all|show>
```
myk002, line 15: since `all` and `show` are literals and `<type>` can be repeated, this should be written as:
```
ban-cooking all [show]
ban-cooking <type> [<type> ...] [show]
```

`show` is something of an anomaly, though, since it wasn't documented before, was intended for debugging, and doesn't follow the pattern of other kinds of commands that do this kind of thing. We don't need to be faithful to the `show` syntax since it was never documented. I think it should be represented as a flag (`-v`, `--verbose`), or it can just remain undocumented.

adaptation.lua
EarthPulseAcademy: done

adaptation.lua
EarthPulseAcademy: done

adaptation.lua
EarthPulseAcademy: done

adaptation.lua
EarthPulseAcademy: done

docs/ban-cooking.rst
```diff
@@ -12,11 +12,16 @@ Usage
...
 
 Valid types are ``booze``, ``brew``, ``fruit``, ``honey``, ``milk``, ``mill``,
 ``oil``, ``seeds`` (i.e. non-tree plants with seeds), ``tallow``, and
-``thread``.
+``thread``. It is possible to include multiple types in a single ban-cooking
+call: ``ban-cooking oil tallow`` will ban both oil and tallow from cooking.
+
+Using ``ban-cooking all`` will ban all types from being cooked.
+
+Using ``ban-cooking show`` will display all types of food that have been banned.
```
myk002, line 24: I'm pretty sure this will have no output. it needs to be specified along other commands that do something.

ban-cooking.lua
```diff
@@ -49,7 +27,7 @@ local function reaction_product_id_contains(reaction_product, str)
 end
 
 local function ban_cooking(print_name, mat_type, mat_index, type, subtype)
-    key = mat_type, mat_index, type, subtype
+    local key = mat_type
```
myk002, line 30: this confuses me. I said that the previous statement was equivalent to this one, but the implication was that this was wrong. Keys should encompass everything you need to differentiate the tuple, but keys cannot be tables. I'm pretty sure you should be stringifying the four tuple of elements. if you still need to address them individually, assign the table of separate elements as the value instead of `true`. a table is still "truthy" so it can still be easily tested for membership.

myk002 (changes requested):

ban-cooking.lua
```diff
@@ -0,0 +1,433 @@
...
+
+    table.sort(output, function(a, b) return a < b end)
+
+    for k, v in pairs(output) do
+        print(k .. ": |" .. v)
+    end
+end
+
+if ... == "help" then
+    print(dfhack.script_help())
```
myk002, line 416: needs a `return` statement, otherwise the script will error out after printing the help

ban-cooking.lua
```diff
@@ -0,0 +1,433 @@
...
+    -- The item hasn't already been banned, so we do that here by appending its values to the various arrays
+    print(print_name .. ' has been banned!')
+    kitchen.mat_types:insert('#', mat_type)
+    kitchen.mat_indices:insert('#', mat_index)
+    kitchen.item_types:insert('#', type)
+    kitchen.item_subtypes:insert('#', subtype)
+    kitchen.exc_types:insert('#', df.kitchen_exc_type.Cook)
+
+    already_banned[key] = {}
+    already_banned[key].isBanned = true
```
myk002, line 38: the `isBanned` value is never used (and is always `true`) as far as I can see. is there a reason you included it?

myk002 (changes requested):

myk002 (approved):

EarthPulseAcademy:
pre-commit.ci autofix

PR #480, opened 2022-12-13
DFHack/dfhack#2081: port fix/stuckdoors from ruby to lua

jowario:
Also I think this has been broken indefinitely because the item part of the ruby script doesn't seem like it would have worked ever as doors are items and would have shown in the tile being searched.

Part of efforts for https://github.com/DFHack/dfhack/issues/2081


fix/stuckdoors.lua
```diff
@@ -0,0 +1,120 @@
...
+-- A door may stick open if the map occupancy flags are wrong.
+
+--[====[
+
+    fix/stuckdoors
+    ===============
+    Fix doors that are stuck open due to incorrect map occupancy flags, eg due to
+    incorrect use of `teleport`.
+
+]====]
```
myk002, line 12: The docs have moved to `docs/fix/stuckdoors.rst`, but I haven't gone through and cleaned docstrings out of the script files yet. If you made any improvements to the help text, could you move the changes to the .rst file? Then you can remove this docstring entirely. It's not used by anything and will only get out of date.

fix/stuckdoors.lua
```diff
@@ -0,0 +1,120 @@
...
+
+if #{...} > 0 then
+    qerror('This script takes no arguments.')
```
myk002, line 15: instead of throwing an error, consider `print(dfhack.script_help())`. That outputs the rendered help text from the associated .rst file.

fix/stuckdoors.lua
```diff
@@ -0,0 +1,120 @@
...
+end
+
+-- Util function: use to filter down to only an active unit
+function activeUnitFilter(unit)
+    return dfhack.units.isActive(unit)
+end
+
+-- Util function: find out if there are any units on the tile with coordinates x,y,z
+function unitOnTile(x, y, z)
+    local units = dfhack.units.getUnitsInBox(x,y,z,x,y,z,activeUnitFilter)
```
myk002, line 25: I take it directly passing `dfhack.units.isActive` as the filter doesn't work?

fix/stuckdoors.lua
```diff
@@ -0,0 +1,120 @@
...
+    return #(units) > 0
```
myk002, line 26: no need for parentheses

fix/stuckdoors.lua
```diff
@@ -0,0 +1,120 @@
...
+end
+
+-- Util function: find out if there are any items on the tile with coordinates x,y,z
+-- There should be one item on the tile: the door itself.
+-- This method is not especially precise as it's doing a simple count check.
+-- A better one would be somehow linking the door building with the door item
```
myk002, line 32: this is not actually that hard to do. `dfhack.buildings.findAtTile(pos)` will get you the door (you can check the type to make sure), and there's a `contained_items` vector that should get you a reference to the door item.

fix/stuckdoors.lua
```diff
@@ -0,0 +1,120 @@
...
+
+    return unstuck_doors_count
+end
+unstickDoors()
+
+
+-- simple test to be run on map, assumes no corrupted doors currently exist
+-- to get started select a door that is currently closed in the UI and run this script
+-- after uncommenting the call this function below
+function unstickDoorsTest()
```
myk002, line 89: A test! I love it! this should go in `test/fix/stuckdoors.lua`. See other tests in https://github.com/DFHack/scripts/tree/master/test for examples, e.g. https://github.com/DFHack/scripts/blob/master/test/assign-minecarts.lua

you can run just your test by building with `BUILD_TESTS` set to `ON`, loading up DF, and running `test -tstuckdoors -d/path/to/your/dev/scripts/test`

You can mock out any part of the `df` or `dfhack` namespace so you don't actually have to interact with game state. that said, an integration test in a separate file (e.g. `test/fix/stuckdoors_integration.lua`) that *does* interact with game state if a suitable setup is found is also welcome. Integration tests will be a key to faster releases in the future, and I'm keen on getting as many as possible in the tree.

fix/stuckdoors.lua
myk002: contained_items are only for items *inside* the building. this includes the building materials and any "clutter", like with workshops or trade depots. I was just suggesting a way to identify the door item that was used to build the door. Your existing method of finding all items on the tile is a good one.

fix/stuckdoors.lua
myk002: that looks good. Thanks!

fix/stuckdoors.lua
myk002: My apologies, I should have realized that you wouldn't already be running the build just to edit scripts! Instructions for compiling are here: https://docs.dfhack.org/en/latest/docs/dev/Compile.html, but you don't really need to compile anything to run the test harness.

Instead, just clone the dfhack repo and copy the files in manually:
1. `git clone https://github.com/DFHack/dfhack.git`
2. cp dfhack/ci/test.lua yourDFdir/hack/scripts
3. mkdir yourDFdir/hack/scripts/test
4. cp test/fix/stuckdoors.lua yourDFdir/hack/scripts/test

Then run `test` at the DFHack prompt in a running game.

fix/stuckdoors.lua
myk002: There isn't a function like that that I know of. Is performance a problem? how long does the script take to run on a mature fort with lots of items and doors? This fort might be a good test case: https://dffd.bay12games.com/file.php?id=15434

myk002: Yeah, 4-5 seconds is in the danger zone. A little performance optimization would be a good idea.

(Thank you! I've been working on that layout for 10 years : p.  I wrote quickfort so I could have better tools to iterate on the design)

myk002:
If the test is holding you back here, I'd be happy to merge your work without it. The script itself looks good! Or did you want to work on the algorithm for performance?

myk002:
I'm sorry I put you off. Thank you for the work you've done here. I'll get it merged.

janxious:
> I'm sorry I put you off. Thank you for the work you've done here. I'll get it merged.

It wasn't you. You've been nothing but helpful and kind. Cheers, and thank you.

PR #482, opened 2022-12-14
update example code in devel/hello-world.lua

myk002:
depends on DFHack/dfhack#2500

screenshot:
image

### DIFF
--- a/devel/hello-world.lua
+++ b/devel/hello-world.lua
@@ -1,31 +1,61 @@
--- Test lua viewscreens.
---[====[
+-- A basic example to start your own gui script from.
+--@ module = true
 
-devel/hello-world
-=================
-A basic example for testing, or to start your own script from.
+local gui = require('gui')
+local widgets = require('gui.widgets')
 
-]====]
-local gui = require 'gui'
+local HOVER_FRAME = copyall(gui.BOUNDARY_FRAME)
+HOVER_FRAME.signature_pen = false
 
-local text = 'Woohoo, lua viewscreen :)'
+HelloWorld = defclass(HelloWorld, gui.Screen)
 
-local screen = gui.FramedScreen{
-    frame_style = gui.GREY_LINE_FRAME,
-    frame_title = 'Hello World',
-    frame_width = #text,
-    frame_height = 1,
-    frame_inset = 1,
-}
-
-function screen:onRenderBody(dc)
-    dc:string(text, COLOR_LIGHTGREEN)
+function HelloWorld:init()
+    local window = widgets.Window{
+        frame={w=20, h=15},
+        frame_title='Hello World',
+        autoarrange_subviews=true,
+        autoarrange_gap=1,
+    }
+    window:addviews{
+        widgets.Label{text={{text='Hello, world!', pen=COLOR_LIGHTGREEN}}},
+        widgets.Label{frame={l=0, t=0}, text="Hover target:"},
+        widgets.Panel{
+            view_id='hover',
+            frame={w=5, h=5},
+            frame_style=HOVER_FRAME,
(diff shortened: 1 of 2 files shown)


devel/hello-world.lua
```diff
@@ -1,31 +1,61 @@
...
-devel/hello-world
-=================
-A basic example for testing, or to start your own script from.
+local gui = require('gui')
+local widgets = require('gui.widgets')
 
-]====]
-local gui = require 'gui'
+local HOVER_FRAME = copyall(gui.BOUNDARY_FRAME)
+HOVER_FRAME.signature_pen = false
```
lethosor, line 8: If this is hiding the "DFHack" text, I don't think we should encourage this by default

devel/hello-world.lua
```diff
@@ -1,31 +1,61 @@
...
+        autoarrange_gap=1,
+    }
+    window:addviews{
+        widgets.Label{text={{text='Hello, world!', pen=COLOR_LIGHTGREEN}}},
+        widgets.Label{frame={l=0, t=0}, text="Hover target:"},
+        widgets.Panel{
+            view_id='hover',
+            frame={w=5, h=5},
+            frame_style=HOVER_FRAME,
+            on_render=function() self.subviews.hover:getMousePos()
```
lethosor, line 26:
```suggestion
            on_render=function() 
                self.subviews.hover:getMousePos()
```

devel/hello-world.lua
```diff
@@ -1,31 +1,61 @@
...
+                local hover = self.subviews.hover
+                if hover:getMousePos() then
+                    hover.frame_background = dfhack.pen.parse{
+                        ch=32, fg=COLOR_LIGHTGREEN, bg=COLOR_LIGHTGREEN}
```
lethosor, line 30:
```suggestion
                        ch=string.byte(' '), fg=COLOR_LIGHTGREEN, bg=COLOR_LIGHTGREEN}
```

devel/hello-world.lua
```diff
@@ -1,31 +1,61 @@
...
+        frame={w=20, h=15},
+        frame_title='Hello World',
+        autoarrange_subviews=true,
+        autoarrange_gap=1,
+    }
+    window:addviews{
+        widgets.Label{text={{text='Hello, world!', pen=COLOR_LIGHTGREEN}}},
+        widgets.Label{frame={l=0, t=0}, text="Hover target:"},
+        widgets.Panel{
+            view_id='hover',
```
lethosor, line 23: Non-blocking, but we may want to consider moving a complex example like this to a separate script, like `devel/hover-example`. I suspect basic scripts won't need it.

lethosor (changes requested):

devel/hello-world.lua
myk002: haha, that's a good point

devel/hello-world.lua
myk002: basic scripts wouldn't, but this example has always been specifically for gui scripts (at least according to the help text). I added in the hover target because I'd really like to start seeing more mouse feedback for gui scripts, and I wanted to show how easy it can be.

lethosor: That's a fair point. I think an ideal example would be some out-of-the-box widget that supports mouse hover/click (like a Label?) so we don't inadvertently encourage people to reinvent it themselves. But I'm not sure if such a widget exists currently.

lethosor (approved):

devel/hello-world.lua
myk002: Ha, that's another really good idea. How are you so full of those?

I actually added a HotkeyLabel to our widget library a few months back (April 2022) that would exemplify that nicely.

PR #483, opened 2022-12-14
[extra-gamelog.rst] Soundsense link points to the wrong forum thread.

Erquint:
The link labelled as "Soundsense" is pointing to "Stonesense" forum thread.
Here directing it at the correct thread.


lethosor (approved):
Haha, good catch!

PR #487, opened 2022-12-18
Port growcrops to lua

AridTag:


growcrops.lua
```diff
@@ -0,0 +1,169 @@
...
+        end
+
+        :: growskip ::
+    end
+
+    print(string.format("Grew %d %s", count, wanted_mat.name))
+end
+
+local args = {...}
+local material = args[1]
```
myk002, line 146: You might be better off making this a local instead of caching it across invocations. The world can get unloaded and a new world loaded in its place and then the values in this global will be stale.

growcrops.lua
```diff
@@ -0,0 +1,169 @@
...
+raws_crop = raws_crop or {}
+
+-- will contain a count of all planted seeds by mat_index
+inventory = {}
```
myk002, line 150: Instead of a "ephemeral global", you can make this a local returned from getInventory and passed to the other methods.

myk002 (commented):
More comments later, but submitting a few quick ones now. That's for the port!

growcrops.lua
```diff
@@ -0,0 +1,169 @@
+-- Instantly grow crops in farm plots
+--[====[
```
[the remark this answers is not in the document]
AridTag: Awesome, I will remove that. I didn't know that the comment wasn't required but I did check docs/growcrops.rst to ensure documentation was up to date. (I didn't make any changes requiring an update)

growcrops.lua
AridTag: I think you meant for this comment to be on line 147 `raws_crop = raws_crop or {}`

I'll make it local. The original script seemed to be using it as a global cache between invocations but I'm not sure _why_ they were doing that

growcrops.lua
AridTag: It would definitely be a cleaner approach. I'll refactor

AridTag:
I have moved the entry point to the top of the file and handled the comments you left.

growcrops.lua
```diff
@@ -0,0 +1,138 @@
...
+        end
+
+        if string.find(lower_crop_name, lower_mat) then
+            table.insert(possible_matches, {mat_index = mat_index, name = crop.name})
+        end
+    end
+
+    if #possible_matches == 1 then
+        return possible_matches[1]
+    end
```
AridTag, line 101: Is there a better way to do this with the lua api?

The ruby script was using `df.match_rawname` but that doesn't appear to exist for the lua api. So I basically inline implemented that function here

growcrops.lua
```diff
@@ -0,0 +1,138 @@
...
+-- e.g. containsRefType(item.general_refs, df.general_ref_building_holderst)
+function containsRefType(refs, type)
+    for i, ref in ipairs(refs) do
+        if ref._type == type then
+            return true
+        end
+    end
+
+    return false
+end
```
AridTag, line 42: This seems like a common thing that scripts would want to do but my incredibly limited knowledge of lua and completely non-existant knowledge of the df api lead me to just implement this method on my own as I couldn't locate an alternative way to do it.

Is there another way?

AridTag, line 42: This seems like a common thing that scripts would want to do but my incredibly limited knowledge of lua and completely non-existant knowledge of the df api lead me to just implement this method on my own as I couldn't locate an alternative way to do it.

Is there another way?

myk002: The [dfhack.jobs module](https://docs.dfhack.org/en/stable/docs/dev/Lua%20API.html#job-module) has `dfhack.job.getGeneralRef(job, type)`, but the function only appears in its general form here: https://github.com/DFHack/dfhack/blob/1c32783dd2628f22c5355b84e49e5c7357b52cb8/library/Types.cpp#L67 which is only available to c++. it could be exposed to Lua (as `dfhack.types.findRef()`?) with some light additions to LuaApi.cpp, but in truth this doesn't come up that often for the seeds type, so it's probably fine to keep this implementation here.

growcrops.lua
myk002: I'm not aware of any function that does this in lua, though if this comes up elsewhere I'll recommend that this be made into a library function.

growcrops.lua
```diff
@@ -0,0 +1,137 @@
...
+    for _, seed in pairs(df.global.world.items.other.SEEDS) do
+        -- Skip seeds that don't match the material
+        if seed.mat_index ~= wanted_mat.mat_index then
+            goto growskip
+        end
+
+        -- skip if seed is not in a specific building (plot)
+        if not containsRefType(seed.general_refs, df.general_ref_building_holderst) then
+            goto growskip
+        end
```
myk002, line 109: this might need more conditions -- maybe ensuring that the containing building actually *is* a farm plot? seeds could be in built bags or cages (when left over from animal training), for example.

growcrops.lua
myk002: they *may* just have not considered the world reload case. the simplicity of this script does not warrant caching, though.

growcrops.lua
```diff
@@ -0,0 +1,137 @@
...
+
+    print(string.format("Grew %d %s", count, wanted_mat.name))
+end
+
+local args = {...}
+local material = args[1]
+local crop_raws = cacheCropRaws()
+local inventory = collectInventory(crop_raws)
+
+if not material or material == "help" or material == "list" then
```
myk002, line 127: I'd suggest looking for "help" (or any unknown param) separately and calling `print(dfhack.script_help())` in that case.

myk002 (changes requested):
very well written!

growcrops.lua
AridTag: I agree, I just don't know how to do that. I wasn't seeing any existing examples of doing something similar, and the original script doesn't appear to handle that either.

Will look into it more deeply

myk002: advfort does some things with general refs that might give you ideas:
https://github.com/DFHack/scripts/blob/727e4921c00e260d7c8d1112daf77115ce3960ee/gui/advfort.lua#L280

growcrops.lua
AridTag: I will at least pull this implementation out into a function (not really sure why I didn't do that to begin with....) so that it can easily be used by others or easily pulled to somewhere more common in the future

AridTag:
I have handled the comments you left, the script now ensures that the seed is in fact inside of a FarmPlot. Thank you for pointing me to advfort that was very useful.

myk002 (approved):

myk002:
pre-commit.ci autofix

PR #488, opened 2022-12-18
Port deathcause.rb

AridTag:
This port is not fully tested. But I desperately need to go to bed. I will look into completing testing after I sleep.

Specifically, I have not tested the printing of weapon information in `display_histfig_death_event`.

I also took the DEATH_TYPES table from unit-info-viewer as an 'improvement' to the text generated by this script.
I feel like these strings are something that should be provided by the base API, is that the case already and I just don't know it?


deathcause.lua
```diff
@@ -0,0 +1,212 @@
+-- show death cause of a creature
+local utils = require 'utils'
+
+local DEATH_TYPES = {
```
myk002, line 4: instead of copying this, could you make `gui/unit-info-viewer` into a [module](https://docs.dfhack.org/en/stable/docs/dev/Lua%20API.html#importing-scripts), change it's `DEATH_TYPES` table to non-local, and access it via `reqscript`? e.g. `local DEATH_TYPES = reqscript('gui/unit-info-viewer').DEATH_TYPES`

If this comes up elsewhere, we can look into moving it into a library, but no need to move it yet.

deathcause.lua
```diff
@@ -0,0 +1,212 @@
...
+        str = str .. string.format(" %s", dfhack.TranslateName(unit.name))
+    end
+
+    if not unit.flags2.killed and not unit.flags3.ghostly then
+        str = str .. " is not dead yet!"
+        print(str)
+        return
+    end
+
+    str = str .. string.format("%s", getDeathStringFromCause(unit.counters.death_cause))
```
myk002, line 105: what is the format adding here? if you are just protecting against `nil` concatenation, `tostring()` might be cleaner.

as an aside that I didn't want to make a comment for elsewhere, there is an alternate syntax for calling string functions that's a little shorter: `formatstring:format(values)`. note that string literals need to be surrounded by parentheses, e.g. `('%s'):format(getDeathStringFromCause(unit.counters.death_cause))`

you might already know about it, and either format is fine. I just wanted to make sure you were aware of your options.

deathcause.lua
```diff
@@ -0,0 +1,212 @@
...
+        local killer = df.unit.find(incident.criminal)
+        if killer then
+            str = str .. string.format(" killed by the %s", getRaceNameSingular(killer.race))
+            if killer.name.has_name then
+                str = str .. string.format(" %s", dfhack.TranslateName(killer.name))
+            end
+        end
+    end
+
+    str = str:gsub(",%s*$", "")
```
myk002, line 122: when does this come up? I don't see you inserting any commas.

deathcause.lua
```diff
@@ -0,0 +1,212 @@
...
+    print(str)
+end
+
+function display_histfig_death_event(histfig_unit, event)
+    local str = getUnitDeathString(histfig_unit);
```
myk002, line 127: this line looks like it can be removed

deathcause.lua
```diff
@@ -0,0 +1,212 @@
...
+        if killer then
+            str = str .. string.format(" killed by the %s", getRaceNameSingular(killer.race))
+            if killer.name.has_name then
+                str = str .. string.format(" %s", dfhack.TranslateName(killer.name))
+            end
+        end
+    end
+
+    str = str:gsub(",%s*$", "")
+    print(str)
```
myk002, line 123: should you end with a period here like you do for other printed strings?

deathcause.lua
```diff
@@ -0,0 +1,212 @@
...
+
+    if event.weapon then
+        if event.weapon.item_type == df.item_type.WEAPON then
+            str = str .. string.format(" using a %s", df.global.world.raws.itemdefs[event.weapon.item_subtype].name)
+        elseif event.weapon.shooter_item_type == df.item_type.WEAPON then
+            str = str .. string.format(", shot by a %s", df.global.world.raws.itemdefs[event.weapon.shooter_item_subtype].name)
+        end
+    end
+
+    str = str:gsub(",%s*$", "")
```
myk002, line 155: again, I'm not seeing where this can happen

deathcause.lua
AridTag: I think this format was just left in while I was in the process of figuring stuff out and didn't get cleaned up.

I was not aware there were other forms for format though, thank you!

deathcause.lua
AridTag: It might not come up anymore, will take a look

deathcause.lua
AridTag: You are absolutely right. Forgot to remove it in my delirious state

deathcause.lua
AridTag: This format was actually wrong, and it needed a space to be added before the string. Side-effect of refactoring.

deathcause.lua
AridTag: Awesome will definitely do this. I wasn't aware of this ability

deathcause.lua
AridTag: This _could_ happen if the event does not have a slayer and the event does not have a weapon. I don't know enough about DF to say whether that can actually happen in the real world or not (I just started playing with the steam release).

AridTag: I am however going to change it to have the other strings prepend the command instead of that first string appending it

deathcause.lua
myk002: > I was not aware there were other forms for format though, thank you!

not just for format, any string function can use the `obj:classFunction()` syntax. DFHack adds a few extra string functions to the standard library as well: https://docs.dfhack.org/en/stable/docs/dev/Lua%20API.html#string-class-extensions

AridTag:
I have dealth with the comments you left and did a little refactoring to clean it up a bit.

I still have not fully tested this script.

myk002 (commented):
Code itself looks good. Once it's tested, it'll be clear to merge. Thanks!

AridTag:
I made a couple of minor tweaks after testing the script to fix "none" being returned instead of died and an error when accessing df.global.world.itemdefs.

But otherwise the script appears to be working correctly

myk002 (approved):

PR #490, opened 2022-12-23
Update gui/launcher hotkeys for v50

lethosor:
Some keys were removed and causing errors in the full display (see also dfhack/dfhack@6a61c750e)

image

### DIFF
--- a/docs/gui/launcher.rst
+++ b/docs/gui/launcher.rst
@@ -57,7 +57,7 @@ As you type, autocomplete options for DFHack commands appear in the right
 column. If the first word of what you've typed matches a valid command, then the
 autocomplete options will also include commands that have similar functionality
 to the one that you've named. Click on an autocomplete list option to select it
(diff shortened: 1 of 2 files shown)


myk002 (approved):

myk002:
oh, needs doc changes too

lethosor:
Done. To be honest, I don't really like the change here, and was hoping we could find some way longer-term to support Tab/Shift-Tab/Ctrl-Up/etc. properly. But I guess it makes sense to have the docs match reality.

myk002:
> Done. To be honest, I don't really like the change here, and was hoping we could find some way longer-term to support Tab/Shift-Tab/Ctrl-Up/etc. properly. But I guess it makes sense to have the docs match reality.
> 
> My main motivation here was just to make the launcher usable.

I agree. I've got half a brain cell thinking about how the new imgui-based framework might be able to work around issues like this, allowing us to use DF keybindings OR actual keys

PR #491, opened 2022-12-24
Port superdwarf script from Ruby to lua

actioninja:
Port of the superdwarf script from ruby to lua.

Added some feedback messages to make it slightly more usable. If this is considered a compat breaker, let me know and I'll remove them.

Refactored it to be inline with the module system, ~~as well as use argument parsing and allow specifying a unit id rather than only working on selected units.~~


superdwarf-lua.lua
```diff
@@ -0,0 +1,126 @@
...
+local repeatUtil = require('repeat-util')
+local utils = require('utils')
+local validArgs = utils.invert({
+    'add',
+    'del',
+    'clear',
+    'unit',
+    'list',
+    'help',
+})
```
lethosor, line 32: This changes the script to take arguments like `-add`, `-list`, etc. instead of the old `add`, `list`. It would be nice if we could keep compatibility. Also, this allows any number of these arguments to be specified, which is probably not necessary - the old version only allowed one at a time.

You can extract the first argument with something like:

```lua
local commandArg = ({...})[1]
```

superdwarf-lua.lua
```diff
@@ -0,0 +1,126 @@
...
+local args = utils.processArgs({...}, validArgs)
+
+
+local id = "superdwarf"
+
+superdwarfIds = {}
```
lethosor, line 38:
```suggestion
superdwarfIds = superdwarfIds or {}
```

This change ensures that the table is not overwritten every time the script is invoked, which looks like it is necessary for `DeleteSuperdwarf()` and `ListSuperdwarfs()` (among others) to work at all.

superdwarf-lua.lua
```diff
@@ -0,0 +1,126 @@
...
+        print(df.unit.find(v).name)
+    end
+end
+
+function ClearSuperdwarfs()
+    local count = #superdwarfIds
+    for i=0, count do superdwarfIds[i]=nil end
+end
+
+function main(...)
```
lethosor, line 81: You're currently not using this at all, but if you move to a single-argument approach as I described above, you could replace this with
```suggestion
function main(argCommand)
```

and the `main(...)` call below will set `argCommand` properly. This would mean you wouldn't need to extract the argument from `{...}` manually at all.

superdwarf-lua.lua
```diff
@@ -0,0 +1,132 @@
+-- makes units very speedy
+--@ module = true
+local help = [====[
```
lethosor, line 3: Documentation needs to go in `docs/superdwarf.txt`, not here

superdwarf-lua.lua
```diff
@@ -0,0 +1,132 @@
...
+
+superdwarf
+==============
```
lethosor, line 6:
```suggestion
==========
```

superdwarf-lua.lua
actioninja: Sure, didn't realize compatibility was a fairly large ocncern

superdwarf-lua.lua
actioninja: Realized this while testing. Not super familiar with ruby, but realized the `||=` used in the initialization is in fact `var = var || value`

lethosor: yep! unfortunately Lua does not have compound-assignment operators, so anything like `a += b` will always have to translate to `a = a + b`

actioninja:
pre-commit.ci autofix

superdwarf.lua
```diff
@@ -0,0 +1,104 @@
...
+    end
+end
+
+function ClearSuperdwarfs()
+    for k,_ in pairs(superdwarfIds) do superdwarfIds[k] = nil end
+end
+
+function main(argCommand)
+    if argCommand == 'help' or not argCommand then
+        print(help)
```
AridTag, line 70: To be more specific about the removal of help documentation from the scripts: You can remove your help variable entirely and replace this print command with `print(dfhack.script_help())`

actioninja: I see, fixed now

myk002 (approved):

superdwarf-lua.lua
```diff
@@ -64,69 +46,59 @@ function AddSuperdwarf(unit)
...
-        else
-            qerror("Invalid ID Provided")
-        end
-    else
-        unit = dfhack.gui.getSelectedUnit()
-    end
-
+    local unit = dfhack.gui.getSelectedUnit()
     if not unit then
         qerror("Invalid unit selection")
```
lethosor: Minor thing: both of these lines will print an error message if no unit is selected. I would recommend getting rid of the `qerror()` line to reduce it to just one message. You could instead silence the first one by using `getSelectedUnit(true)`, but I think the error from getSelectedUnit is a bit more clear (and standardized).

lethosor:
also, I just saw your edit:

> ~~as well as use argument parsing and allow specifying a unit id rather than only working on selected units.~~

I didn't mean you couldn't add features! I just was recommending that we avoid changing the commands that were already implemented.

One way you could support a `-unit` arg could look like this:
* pass `...` to `main()` (and getting rid of the global `commandArg`)
* make `main()` take `(argCommand, ...)`
* in `main()`, call `processArgs({...}, utils.invert('unit')` or similar. This will give you a table with a `unit` argument, and allows commands such as:
    ```
    superdwarf add
    superdwarf add -unit 3
    ```

Up to you, just wanted to make it clear I wasn't opposed to this.

actioninja:
Hah, sorry if it came off as passive aggressive or anything, I edited it like that just to make the change clearer. I get the argument behind keeping parity to avoid breaking existing people's scripts.

I could re implement specifying the unit, maybe just add it as an optional 2nd position argument.

superdwarf-lua.lua
actioninja: Was unsure if a `dfhack.gui.getSelectedUnit()` call would block if it failed like that, good to know.

lethosor:
No worries at all, I was just worried I'd come across as too strict about parity.

A second argument could certainly work as well. I was about to suggest that, then decided to suggest processArgs() to make it cleaner to add _additional_ optional args in the future (since `-unit` would be optional).

Either way, we can get this in as-is if you're good with it.

lethosor (approved):

actioninja:
Alright, ended up readding the unit argument as an option 2nd parameter to the add and del commands, tested both that it works omitting them and with them.

superdwarf.lua
```diff
@@ -61,9 +61,15 @@ function main(argCommand)
         return
     end
 
-    local unit = dfhack.gui.getSelectedUnit()
+    unit = nil
```
lethosor, line 64: This should still be `local`, otherwise it will end up in the script's global scope.
```suggestion
    local unit = nil
```

oorzkws:
Before I open a PR, was this built for 0.47? In 0.50dev there's errors in the syntax that mean the arguments aren't parsed properly.

myk002:
It was written before DF v50. What is the error?

oorzkws:
the use of `local argCommand = ({...})[1]` puts a table inside a table, then checks for [1]. This means it always falls back to looking for the unit, no matter what args I used.

The copy I'm using locally is [here](https://gist.github.com/oorzkws/ca7de291455210ea73445b7e54f3f894)

myk002:
that looks cleaner. Could you possibly submit that version in a PR?

oorzkws:
> that looks cleaner. Could you possibly submit that version in a PR?

Sure, there's a few bits of cleaning I mean to do before submitting though.

lethosor:
@oorzkws I don't understand the issue. There should be no difference between
```lua
    local args = ...
    local argCommand = args[1]
```
and
```lua
    local argCommand = ({...})[1]
```
and the same goes for unitArg later.

What specific error message(s) are you getting?

lethosor:
Oh, the problem is that the *call to main()* also groups the arguments into a table: https://github.com/DFHack/scripts/commit/592f157e#diff-0882f6889c5837a675dfc46ba8f5bb6932bac961339bdd146edefa18906a46b6R92

I think the cleanest fix is
```diff
diff --git a/superdwarf.lua b/superdwarf.lua
index eb3b3b0..cea8285 100644
--- a/superdwarf.lua
+++ b/superdwarf.lua
@@ -110,5 +110,5 @@ function main(args)
 end
 
 if not dfhack_flags.module then
-    main({...})
+    main(...)
 end
```

as opposed to something like `args = ...`, which is actually equivalent to `args = ({...})[1]` and IMO a bit misleading

lethosor:
Gah, looks like you already fixed this in a rewrite in https://github.com/DFHack/scripts/pull/521 (wasn't cross-linked so I missed it, sorry)

PR #492, opened 2022-12-24
Port Source script from ruby to lua

actioninja:
Ports source script from Ruby to Lua.

Only changes are some miscellaneous feedback messages for performing actions.


lethosor:
I'm not sure you have to interface with a plugin here - the Ruby script doesn't.

myk002:
yeah, the functionality in the `liquids` plugin may not even help you that much here. you *could* call into it over Lua, but it might be more trouble than it's worth.

actioninja:
pre-commit.ci autofix

actioninja:
pre-commit.ci autofix

actioninja:
Took a lot of testing and trial and error, as well as figuring out how exactly the df-structures xmls map to the lua interfaces, but seems like it works perfectly now. Ready for code review.

source.lua
```diff
@@ -0,0 +1,145 @@
...
+            repeatUtil.cancel(sourceId)
+        else
+            for _, v in pairs(liquidSources) do
+                local block = dfhack.maps.getTileBlock(v.pos.x, v.pos.y, v.pos.z)
+                local x = v.pos.x
+                local y = v.pos.y
+                if block and isFlowPassable(v.pos) then
+                    local isMagma = v.liquid == 'magma'
+
+                    local flow = block.designation[x%16][y%16].flow_size
```
AridTag, line 40: I believe you can replace this with the following, then you can simplify lines 48-50 to use tileFlags as well

```suggestion
                    local tileFlags = dfhack.maps.getTileFlags(v.pos)
                    local flow = tileFlags.flow_size
```

source.lua
```diff
@@ -0,0 +1,145 @@
...
+    end
+
+    function findLiquidSourceAtPos(pos)
+        for k,v in pairs(liquidSources) do
+            if v.pos == pos then
+                return k
+            end
+        end
+        return -1
+    end
```
myk002, line 101: this looks like it could just as easily be declared outside of the `main()` function since it isn't making use of any lambda captures that I can see

source.lua
```diff
@@ -0,0 +1,145 @@
...
+
+    if command == 'clear' then
+        ClearLiquidSources()
+        print("Cleared sources")
+        return
+    end
+
+    function findLiquidSourceAtPos(pos)
+        for k,v in pairs(liquidSources) do
+            if v.pos == pos then
```
myk002, line 96: again, beware comparing tables to tables. They will always be unequal.

source.lua
```diff
@@ -0,0 +1,145 @@
+--@ module = true
+-- wip
```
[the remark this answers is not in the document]
actioninja: Old comment

source.lua
```diff
@@ -0,0 +1,145 @@
...
+            z = pos.z,
+        },
+    })
+
+    repeatUtil.scheduleEvery(sourceId, 12, 'ticks', function()
+        if next(liquidSources) == nil then
+            repeatUtil.cancel(sourceId)
+        else
+            for _, v in pairs(liquidSources) do
+                local block = dfhack.maps.getTileBlock(v.pos.x, v.pos.y, v.pos.z)
```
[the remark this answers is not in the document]
actioninja: This wasn't working right when I tried, I *thought* it might have been that the single arg version of these functions required a `pos` pointer but considering the suggestion I'll have to try again, I was likely doing something else wrong.

source.lua
actioninja: This is how every existing script does it, I assume the API is relatively new. I'll change it over.

source.lua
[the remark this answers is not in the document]
actioninja: I realize now I got it backwards, I had a pointer, it was expecting a table. An earlier version was mispassing the argument instead of the iterated value, which was in fact a pointer. Compounding problem yielding a confusing result, the classic.

source.lua
```diff
@@ -0,0 +1,145 @@
...
+        AddLiquidSource(targetPos, liquidArg, amountArg)
+        print("Added " .. liquidArg .. " " .. amountArg .. " at [" .. targetPos.x .. ", " .. targetPos.y .. ", " .. targetPos.z .. "]")
+        return
+    end
+
+    print(dfhack.script_help())
+end
+
+if not dfhack_flags.module then
+    main(...)
```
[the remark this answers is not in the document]
actioninja: I *think* this is what you meant?

source.lua
```diff
@@ -0,0 +1,144 @@
...
+    end
+end
+
+function FindLiquidSourceAtPos(pos)
+    for k,v in pairs(liquidSources) do
+        if same_xyz(v.pos, pos) then
+            return k
+        end
+    end
+    return -1
```
myk002, line 85: It is more common to return `nil` in Lua to mean "not found"

source.lua
```diff
@@ -0,0 +1,144 @@
...
+        local liquidArg = args[2]
+        if not liquidArg then
+            qerror('You must specify a liquid to add a source for')
+        end
+        liquidArg = liquidArg:lower()
+        if not (liquidArg == 'magma' or liquidArg == 'water') then
+            qerror('Liquid must be either "water" or "magma"')
+        end
+        if not IsFlowPassable(targetPos) then
+            qerror("Tile not flow passable: I'm afriad I can't let you do that, Dave.")
```
myk002, line 131: "afriad" -> "afraid"

source.lua
```diff
@@ -0,0 +1,144 @@
...
+        end
+        local amountArg = tonumber(args[3]) or 7
+        AddLiquidSource(targetPos, liquidArg, amountArg)
+        print("Added " .. liquidArg .. " " .. amountArg .. " at [" .. targetPos.x .. ", " .. targetPos.y .. ", " .. targetPos.z .. "]")
```
myk002, line 135: I suggest using `format()` for long concatenated strings like this

myk002 (commented):
Looks great! Good job!

source.lua
```diff
@@ -0,0 +1,144 @@
...
+
+function AddLiquidSource(pos, liquid, amount)
+    table.insert(liquidSources, {
+        liquid = liquid,
+        amount = amount,
+        pos = copyall(pos),
+    })
+
+    repeatUtil.scheduleEvery(sourceId, 12, 'ticks', function()
+        if next(liquidSources) == nil then
```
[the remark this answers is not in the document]
actioninja: I can't take too much credit, got it from stackoverflow. There were alternatives but this one was elegant and `next` reminded me a lot of iterators in other languages.

myk002 (approved):

PR #497, opened 2022-12-29
add graphical tile texture browser

myk002:
image

### DIFF
--- a/devel/tile-browser.lua
+++ b/devel/tile-browser.lua
@@ -1,0 +1,108 @@
+local gui = require('gui')
+local widgets = require('gui.widgets')
+
+TileBrowser = defclass(TileBrowser, gui.Screen)
+
+function TileBrowser:init()
+    local main_panel = widgets.Window{
+        view_id='window',
+        frame={w=36, h=58},
+        drag_anchors={title=true, body=true},
+        resizable=true,
+        resize_min={h=20},
+        frame_title='Tile Browser',
+        on_submit=self:callback('set_start_index'),
+    }
+    main_panel:addviews{
+        widgets.EditField{
+            view_id='start_index',
+            frame={t=0, l=0},
+            key='CUSTOM_CTRL_A',
+            label_text='Start index: ',
+            text='0',
+            on_submit=self:callback('set_start_index'),
+            on_char=function(ch) return ch:match('%d') end},
+        widgets.HotkeyLabel{
+            frame={t=1, l=0},
+            label='Prev',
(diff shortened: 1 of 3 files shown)


lethosor:
Does this depend on a core change? I'm getting a blank table on DFHack `0.47.05-r8-152-ge8e3a616`.

image

myk002:
> Does this depend on a core change? I'm getting a blank table on DFHack `0.47.05-r8-152-ge8e3a616`.

yes, the change I just merged should get this working (it was the call in widgets.Label to `Painter:char()` instead of `Painter:tile()` when pen.tile was defined)

lethosor:
Any idea if there's an easy way to detect out-of-bounds tiles? At least, I'm assuming invalid IDs are behind why the background image shows through on the title screen:
image

Everything else looks great, and I'm not complaining about this, just curious if it's easily fixable.

lethosor:
I guess another oddity is that some special tiles cause the logos to be drawn:

image

but I'm also seeing that in `devel/inspect-tile`

myk002:
I couldn't figure out how to detect out of bounds tiles, so I just left it failing "open" (so to speak). The large textures peeking through the back is a nice oddity, also that I don't have a way to prevent. I did document this "feature", though.

devel/tile-browser.lua
```diff
@@ -0,0 +1,108 @@
...
+        table.insert(report, {tile=texpos})
+        if (texpos+1) % 20 == 0 then
+            table.insert(report, NEWLINE)
+        end
+    end
+
+    self.subviews.report:setText(report)
+    if self.parent_rect then
+        self.subviews.window:updateLayout()
+    end
```
lethosor, line 89: Not exactly a recommendation here, but a more general suggestion: I'm noticing that if I hold shift+arrows for a long time and the keys start repeating _really_ fast, the screen keeps updating for a bit (less than a second) after I release the keys. I'm guessing that's due to computation in here, which is ultimately called by `feed()`. If it becomes an issue (which I don't think it is now), I would consider moving all but the bare minimum to `render()`, and probably behind some check so that it's only recalculated if the state has changed.

lethosor (approved):

lethosor (commented):
Also: new scripts should go in `changelog.txt`

devel/tile-browser.lua
myk002: I had always gone by the mantra: "not in the render thread", but it seems here that making the feed fast is even more important. I'll keep that in mind.

PR #498, opened 2022-12-30
Update gui/launcher

myk002:
new skin, update hotkeys, slightly adjusted proportions in full mode, adjusted position in minimal mode to avoid hiding the text behind "anchor" elements", made window draggable and resizable.

### DIFF
--- a/gui/launcher.lua
+++ b/gui/launcher.lua
@@ -422,11 +422,27 @@ end
 ----------------------------------
 -- LauncherUI
 --
+
 LauncherUI = defclass(LauncherUI, gui.Screen)
 LauncherUI.ATTRS{
     focus_path='launcher',
     minimal=false,
 }
+
+local function get_frame_r()
+    -- scan for anchor elements and do our best to avoid them
+    local gps = df.global.gps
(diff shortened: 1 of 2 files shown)


gui/launcher.lua
```diff
@@ -441,13 +438,15 @@ function LauncherUI:init(args)
     }
 
     local update_frames = function()
-        local new_frame = {l=5, r=5}
+        local new_frame = {}
         if self.minimal then
             new_frame.l = 0
-            new_frame.r = 0
+            new_frame.r = 58
```
lethosor, line 444: Is this just for viewscreen_dwarfmodest? Wondering if we can special-case that easily. It looks a bit out of proportion on other screens to me:

image
image

lethosor: Or could we somehow detect things in the anchored layer somehow? (does that require that the tiles are the same size as the text/UI layer?)

myk002: luckily, since this is on the top row, we can detect anchored layer textures fairly easily, since they only scale down and to the right. I'll see if I can code that up.

PR #509, opened 2023-01-03
update gui/cp437-table to use modern widgets

myk002:
movable window, widgets to manage rendering, state, and text cursor

depends on DFHack/dfhack#2551

### DIFF
--- a/gui/cp437-table.lua
+++ b/gui/cp437-table.lua
@@ -1,118 +1,95 @@
 -- An in-game CP437 table
 
---[====[
+local dialog = require('gui.dialogs')
+local gui = require('gui')
+local widgets = require('gui.widgets')
 
-gui/cp437-table
-===============
-
-An in-game CP437 table. Allows typing or selecting characters with the mouse.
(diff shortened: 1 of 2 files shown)


gui/cp437-table.lua
```diff
@@ -1,118 +1,95 @@
...
-                if ch then
-                    self:insert(ch)
-                end
-            end
+function CPDialog:onInput(keys)
+    local x, y = self.subviews.board:getMousePos()
+    if keys._MOUSE_L_DOWN and x then
+        local ch = x + (32 * y)
+        if ch ~= 0 and dfhack.screen.charToKey(ch) then
+            self.subviews.edit:insert(string.char(ch))
```
lethosor, line 80: Where's this method defined? EditField doesn't have an insert() method on develop

myk002: oops, I may not have made a PR for that yet. One second.

myk002: DFHack/dfhack#2551

lethosor (approved):
Seems to look and behave the same way as before

PR #510, opened 2023-01-03
Convert gui tools to ZScreens

myk002:
let the game run uninterruped while DFHack tools are on the screen and allow DFHack tool windows to swap z-order. this allows, for example, map elements to be interacted with and the map itself to be dragged around while tool windows are up.

Depends on DFHack/dfhack#2556


myk002:
I'll commit this for now so I can maybe get some beta feedback on the behavior. Docs still forthcoming.

PR #516, opened 2023-01-09
new script: gui/autochop

myk002:
Companion to DFHack/dfhack#2575

- implemented as a ZScreen so you can just leave it up and monitor/tweak your logging industry in realtime
- scrollable list of burrows, their tree statistics, and their current autochop configuration
- all autochop settings are settable from the UI
- statistics are colored to draw your attention when trees are becoming scarce (relative to your configured target stock levels)
- 10s auto-refresh so statistics are updated as the game plays


myk002:
Screenshot:
image

PR #519, opened 2023-01-10
Prevent indexing a nil table in unsuspend.lua

oorzkws:
`bld` could be nil, but we were attempting to index it before checking. Filling in an empty table resolves this, and we check for a valid property where relevant.


myk002 (approved):
Thanks!

PR #520, opened 2023-01-10
tag untested scripts with "untested" tag

myk002:
ruby scripts are not moved so as not to disturb porting in progress

"unstable" scripts can still be run by prefixing them with "unstable/" (though the scripts that reqscript other unstable scripts will of course fail since the paths won't match. I don't want to make any code changes in this move). docs are likewise moved and show up properly in helpdb (and therefor in  `help` and `gui/launcher`)

### DIFF
--- a/docs/adaptation.rst
+++ b/docs/adaptation.rst
@@ -3,7 +3,7 @@ adaptation
 
 .. dfhack-tool::
     :summary: Adjust a unit's cave adaptation level.
-    :tags: fort armok units
+    :tags: untested fort armok units
 
 View or set level of cavern adaptation for the selected unit or the whole fort.
 
--- a/docs/add-recipe.rst
+++ b/docs/add-recipe.rst
@@ -3,7 +3,7 @@ add-recipe
 
 .. dfhack-tool::
     :summary: Add crafting recipes to a civ.
-    :tags: adventure fort gameplay
+    :tags: untested adventure fort gameplay
 
 Civilizations pick randomly from a pool of possible recipes, which means not all
 civs get high boots, for instance. This script can help fix that. Only weapons,
--- a/docs/add-thought.rst
+++ b/docs/add-thought.rst
@@ -3,7 +3,7 @@ add-thought
 
 .. dfhack-tool::
     :summary: Adds a thought to the selected unit.
-    :tags: fort armok units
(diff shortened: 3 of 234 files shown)


myk002:
hum. our CI fails since the *name* of the script hasn't changed. it's looking for docs in the "wrong" directory. I don't really want to leave CI broken -- these scripts might take a while before they're out of "unstable". Have to think about this.

lethosor:
Would a "flag" / tag work? a la `--@ unstable = true`

myk002:
oh! I like the idea of a tag. that way they can even be hidden by default in `gui/launcher` like we do for tools tagged with `dev`. it wouldn't even be that hard to keep updated either since the spreadsheet can track whether it's tested or not and we can sync the spreadsheet to the tool docs automatically.

lethosor:
I meant a flag in the script itself, like "--@ enable = true", because we already have machinery to read from those whenever a script is invoked. We don't have as straightforward of a way to read tags from docs from run_script() as far as I know

myk002:
> We don't have as straightforward of a way to read tags from docs from run_script() as far as I know

we actually do, I think -- we could do something similar to how we implement `dfhack.script_help()` : 
```lua
function dfhack.script_help(script_name, extension)
    script_name = script_name or dfhack.current_script_name()
    return require('helpdb').get_entry_long_help(script_name)
end
```

myk002:
warning implemented in https://github.com/DFHack/dfhack/pull/2590/commits/679e91d2e7efb09d29c9e8dbcbccf3c253d2c225

lethosor:
Ok, so that works. I think another reason I was thinking of the in-script flag machinery is that it's a single place to update - if you fix the script, you can update the same file to remove an `--@ untested = true` line.

...I guess it is useful to have an "untested" indicator in documentation as well, though.

myk002:
That is true that `--@ untested = true` could be removed in the same PR.. I guess I saw the tags approach as less toily since the spreadsheet sync is *already* part of my workflow for managing the tags assignment.

lethosor:
That is fair. I'll accept this approach :)
I think having the tag show up in the docs is enough to tip the balance towards this approach - the more places it shows up, the better.

lethosor (approved):

PR #522, opened 2023-01-11
Port stripcaged for DFHack/dfhack#2081

arekatir:
Includes some squashed changes based on feedback received.


stripcaged.lua
```diff
@@ -0,0 +1,228 @@
...
+            type = 'Animal trap'
+        end
+
+        -- If count is empty
+        if not next(count) then
+            empty_cages = empty_cages + 1
+        else
+            local sortedlist = {}
+            for classname, n in pairs(count) do
+                sortedlist[#sortedlist + 1] = {classname = classname, count = n}
```
myk002, line 144: This works, but `table.insert(sortedlist, {classname=classname, count=n})` is arguably cleaner

stripcaged.lua
```diff
@@ -0,0 +1,228 @@
...
+            print(('%s %d: '):format(type, cage.id))
+            for _, t in ipairs(sortedlist) do
+                print(' ' .. t.count .. ' ' .. t.classname)
+            end
+        end
+        for k, v in pairs(count) do count_total[k] = (count_total[k] or 0) + v end
+    end
+
+    if #list > 1 then
+        print '\nTotal:'
```
myk002, line 156:
```suggestion
        print('\nTotal:')
```

myk002: to be consistent with surrounding syntax style

stripcaged.lua
```diff
@@ -0,0 +1,228 @@
...
+
+local args = {...}
+
+local list
+if args[2] == 'here' then
+    print "NOTE: The 'here' option isn't well tested for v50 and only works with the keyboard cursor."
+    local it = dfhack.gui.getSelectedItem(true)
+    list = {it}
+    if not df.item_cagest:is_instance(it) and
+        not df.item_animaltrapst:is_instance(it) then
```
myk002, line 179:
```suggestion
            not df.item_animaltrapst:is_instance(it) then
```

stripcaged.lua
```diff
@@ -0,0 +1,228 @@
...
+    end
+end
+
+-- handle magic script arguments
+
+local args = {...}
+
+local list
+if args[2] == 'here' then
+    print "NOTE: The 'here' option isn't well tested for v50 and only works with the keyboard cursor."
```
myk002, line 175: now that I've actually implemented getSelectedItem for v50, this will probably work just fine when clicking on an item and doesn't need the warning.

stripcaged.lua
```diff
@@ -0,0 +1,228 @@
...
+    local it = dfhack.gui.getSelectedItem(true)
+    list = {it}
+    if not df.item_cagest:is_instance(it) and
+        not df.item_animaltrapst:is_instance(it) then
+        list = {}
+        for _, cage in ipairs(df.global.world.items.other.ANY_CAGE_OR_TRAP) do
+            if same_xyz(df.global.cursor, cage.pos) then
```
myk002, line 182: instead of the cursor, you could use the position of the selected item to find a cage at the same location

arekatir: Good idea since df.global.cursor currently provides garbage values unless the player uses the keyboard cursor.

Unfortunately, my naive fix that changed df.global.cursor to it.pos revealed a couple problems. Now that we're using 'it' (item) beyond checking if it's a valid cage or animal trap and then falling back on the keyboard cursor position if it's not, we need to make sure we select a valid item before using its position. Also we should definitely check to see if the player is selecting a placed cage, which counts as a building rather than an item, among other things.

I'll see if I can't fix this and make it more robust.

lethosor: > we should definitely check to see if the player is selecting a placed cage, which counts as a building rather than an item

Even if a cage has been placed, there should still also be an item present at the same position in the world.

> we need to make sure we select a valid item before using its position

Does `dfhack.gui.getSelectedItem()` not handle this? It should, but that may not be implemented yet.

lethosor: Ok, clarified on Discord that you mean the caller now needs to call `getSelectedItem()` and check if it actually returned something (previously, the cursor alone was used).

myk002 (approved):

PR #530, opened 2023-01-15
No need to lint turned-down ruby

hawkowl:
This removes the Ruby steps from the CI.

Not sure if this should go in after all of them are migrated, but while it was on my mind I opened a PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,10 +40,6 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3
-    - name: Set up Ruby 2.7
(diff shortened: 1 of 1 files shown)


lethosor (approved):
Confirmed that there are no `*.rb` files on develop

lethosor:
I was looking at the DFHack repo - didn't realize this was scripts.

This is definitely something we should do, but as a matter of principle, I think it should wait until there aren't any `*.rb` files left in this repo.

myk002:
this is getting very close to being committed. just a few ruby scripts left

myk002:
It's official! Ruby has been turned down. thanks!

PR #531, opened 2023-01-15
Port masspit script from Ruby to Lua

vallode:
Complete rewrite of the masspit.rb script in Lua. Changed from a CLI to a gui-driven solution. When reviewing bare in mind this is my first time interacting with the gui api, anything that seems off please let me know!


gui/masspit.lua
```diff
@@ -0,0 +1,218 @@
+-- Interface to assign caged units in a stockpile to pond/pit.
+
+local utils = require 'utils'
+local gui = require('gui')
+local widgets = require('gui.widgets')
+
+Masspit = defclass(Masspit, widgets.Window)
+Masspit.ATTRS {
+    frame_title='masspit',
```
myk002, line 9: This is what players actually see on the window, so a more "title-like" phrasing would be more appropriate. How about "Mass-pit caged creatures". Just "Mass-pit" might be ok, but it's kind of a strange terminology that not everyone might understand.

gui/masspit.lua
```diff
@@ -0,0 +1,218 @@
...
+    frame={w=43, h=22},
```
myk002, line 10: Since you'll be having players interacting with the map, I'd suggest putting this window off to the side by default. How about `frame={w=43, h=22, r=2, t=18}`

gui/masspit.lua
```diff
@@ -0,0 +1,218 @@
...
+    resizable=true,
+    resize_min={w=43, h=12},
```
myk002, line 12: you can omit specifying dimensions that are the same as the starting frame. `resize_min={h=12}` is equivalent. this helps when you're tweaking the frame size so you only have to change it in one place.

gui/masspit.lua
```diff
@@ -0,0 +1,218 @@
...
+
+    for _, zone in pairs(df.global.world.buildings.other.STOCKPILE) do
+        if (#zone.settings.animals.enabled > 0) then
+            local zone_name = zone.name.length and dfhack.TranslateName(zone.name) or "Animal stockpile #" .. zone.stockpile_number
+            local zone_position = df.coord:new()
+            zone_position.x = zone.centerx
+            zone_position.y = zone.centery
+            zone_position.z = zone.z
+
+            table.insert(choices, {
```
myk002, line 61: how about checking these stockpiles for caged creatures *before* adding them to the choices list. that way the player can only select valid choices

vallode: Yeah this is a good point, I thought the iteration over stockpile items might cause performance issues in larger forts though so I wasn't sure. What do you think, since we are using `getStockpileContents` and trying to find any caged creature this could cause issues for very large stockpiles?

myk002 (approved):

myk002:
There will be a missing doc anchor because of the removed script, but I'll fix that up on the DFHack/dfhack side (in the Removed.rst file)

PR #533, opened 2023-01-15
Port scanitemother from Ruby to Lua

vallode:
This was a simple script so I just threw it together, does the job.


devel/scanitemother.lua
```diff
@@ -0,0 +1,20 @@
...
+    for _, item in pairs(value) do
+        if item == selected_item then
+            table.insert(indices, index)
+        end
+    end
+end
+
+print("Found in the following indecies:")
+for _, index in pairs(indices) do
+    print(" " .. index)
```
myk002, line 19: could this print out both the numeric index and the *name* of the OTHER list? that would make it a little more handy.

vallode: Do you mean the numeric index the index appears in `df.global.world.items.other` at or just the numeric index of our local table? Currently the output is:

```
Found in the following indecies:
 IN_PLAY
 ANY_GENERIC36
 CLOTH
```

myk002: I'd say that's good. Thanks!

PR #536, opened 2023-01-16
Update exportlegends.lua

Mason11987:
As of version .50, sometimes claim borders have more "x" coordinates than "y", which breaks the below.  Not sure why that would be, so commenting it out.

Otherwise this script seems to work now with exportlegends custom

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -488,13 +488,14 @@ function export_more_legends_xml()
             file:write("\t\t<child>"..link.."</child>\n")
         end
 
-        if #entityV.claims.border.x > 0 then
-            file:write("\t\t<claims>")
-            for xK, xVal in ipairs(entityV.claims.border.x) do
-                file:write(xVal..","..entityV.claims.border.y[xK].."|")
-            end
-            file:write("</claims>\n")
-        end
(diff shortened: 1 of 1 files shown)


myk002 (commented):
would it be possible to support non-square boundaries instead of commenting out the functionality?

Mason11987:
> would it be possible to support non-square boundaries instead of commenting out the functionality?

So that is a list of x coordinates, and then a list of y coordinates.  In prior version those can be interpreted as coordinates of (x[0], y[0]), and (x[1], y[1]) and so on.

And the x and y lists were the same size.

Now, it appears that's not the case, so I don't really know what those coordinates are supposed to represent.  If you have this in x [1, 3, 4], and this in y [2, 6], I don't know what points that's supposed to represent anymore.

Either the data is being read wrong, or our understanding of the purpose of those arrays is wrong, imo.

myk002 (approved):

myk002:
ah, ok. understood. and it looks like the entire `claims` section could be optional on the output, so we don't need to print an empty claims section as a stub. Looks good. Thanks!

myk002:
I'm not marking this as "tested" though since it's not useful without the vanilla legends export to support it.

PR #537, opened 2023-01-16
Port starvingdead from Ruby to Lua

vallode:
Ported this from CLI to CLI, no need for a GUI for this one. I "corrected" a behavior whereby the month length was double the length of the actual in-game month, not sure if that was intentional. I also left out a part of the script that made the death's happen faster when there were above a certain amount of undead.


docs/starvingdead.rst
```diff
@@ -20,3 +23,20 @@ Usage
 
     starvingdead start
     starvingdead stop
```
myk002, line 25: Since this is an "auto" script, there are a few more requirements, if you have the time. These types of scripts should now follow the enable API and persist their state (all copy-pasteable from existing scripts, so it shouldn't be that much more work).

docs are here for context: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#enabling-and-disabling-scripts

and you can totally just copy paste the code from `emigrate.lua`, which has very little business logic to get in the way.

The usage for this script should change to something like:

```
    enable starvingdead
    starvingdead [<options>]
```

where `starvingdead` run without options displays current status and configuration

docs/starvingdead.rst
```diff
@@ -20,3 +23,20 @@ Usage
...
+``starvingdead start --decay-rate 28``
+    Undead will lose strength roughly once a month.
+``starvingdead start --decay-rate 1 --death-threshold 1``
+    Undead will lose strength each day and die after they have spent a month
+    on the map.
+
+Options
+-------
+
+``--decay-rate``
```
myk002, line 39:
```suggestion
``--decay-rate <days>``
```

docs/starvingdead.rst
```diff
@@ -20,3 +23,20 @@ Usage
...
+    Specify how often, in days, undead should lose strength.
+``--death-threshold``
```
myk002, line 41:
```suggestion
``--death-threshold <months>``
```

docs/starvingdead.rst
myk002: you'll know you've done the integration correctly if you run `enable` and `starvingdead` shows up with the correct status

vallode: The docs do not mention it but you also need `--@module = true` in your document alongside `--@enable = true` for it to show up in the list... done though :)

myk002: That's right. I'll have to add that to the docs. Thanks!

starvingdead.lua
```diff
@@ -0,0 +1,128 @@
...
+      end
+    end
+  end
+
+  self.timeout_id = dfhack.timeout(self.decay_rate, 'days', self:callback('checkDecay'))
+end
+
+if not dfhack.isMapLoaded() then
+  qerror('This script requires a fortress map to be loaded')
+end
```
myk002, line 52: this needs to be after the isEnabled() global and onStateChange hook, otherwise those will fail to be found when DFHack starts and state will not get reloaded on map load

starvingdead.lua
```diff
@@ -0,0 +1,128 @@
...
+    starvingDeadInstance = StarvingDead{persisted_data.decay_rate, persisted_data.death_threshold}
+  end
+end
+
+if dfhack_flags.enable then
+  if dfhack_flags.enable_state then
+    if starvingDeadInstance then
+      print("Stopping previous instance of StarvingDead...")
+      dfhack.timeout_active(starvingDeadInstance.timeout_id, nil)
+    end
```
myk002, line 99: is this necessary? most tools just return without action when enable is called on an already-enabled tool

starvingdead.lua
```diff
@@ -0,0 +1,128 @@
...
+
+    starvingDeadInstance = StarvingDead{options.decay_rate, options.death_threshold}
+    persist_state()
+  else
+    if not starvingDeadInstance then
+      qerror("StarvingDead is not running!")
+    end
```
myk002, line 106: similarly here, it can just be a noop

starvingdead.lua
```diff
@@ -0,0 +1,128 @@
...
+
+    dfhack.timeout_active(starvingDeadInstance.timeout_id, nil)
+    starvingDeadInstance = nil
+  end
+else
+  if positionals[1] == "help" or options.help then
+    return print(dfhack.script_help())
```
myk002, line 113: though this technically works, I try not to be misleading about intentions and put the print and return on separate lines

starvingdead.lua
```diff
@@ -0,0 +1,128 @@
...
+end
+
+if dfhack_flags.enable then
+  if dfhack_flags.enable_state then
+    if starvingDeadInstance then
+      print("Stopping previous instance of StarvingDead...")
+      dfhack.timeout_active(starvingDeadInstance.timeout_id, nil)
+    end
+
+    starvingDeadInstance = StarvingDead{options.decay_rate, options.death_threshold}
```
myk002, line 101: note that `enable starvingdead` will not have parameters associated with it, so the options will only be relevant when `dfhack_flags.enable` is *not* set

myk002 (commented):
I suggest a bit of a reordering of elements to make the script easier to understand:

`starvingDeadInstance = starvingDeadInstance or nil`

function isEnabled() ...

dfhack.onStateChange[GLOBAL_KEY] ...

StarvingDead class

```lua
if dfhack_flags.module then
    return
end
```

parameter parsing

enabled handling

myk002 (approved):

PR #539, opened 2023-01-16
startdwarf - reflect v50 effective limits

ab9rf:
in testing the only values that worked were 7, 8, or 9; values less than 7 or greater than 9 caused DF to crash


lethosor (commented):
I agree with what you said on Discord - even 8 or 9 are likely unstable, just less obviously so.

I think we should disable startdwarf for now until it can be fixed.

PR #549, opened 2023-01-21
add verbose option to ban-cooking

Moth-Tolias:
Fixes Dfhack/dfhack#2642

!!untested!!

doesn't touch `show` functionality, as we figured that would be unwanted. 
~~future possibility: a "banned N items" status report after everything has been banned. didn't do this because we'd rather wait till we're not "flying blind", as it were~~


ban-cooking.lua
```diff
@@ -2,6 +2,33 @@
 -- based on ban-cooking.rb by Putnam: https://github.com/DFHack/scripts/pull/427/files
 -- Putnams work completed by TBSTeun
 
+local options = {}
+
+local argparse = require('argparse')
```
myk002, line 7: These two should stay up here, but the rest of the code you added needs to stay at the bottom. otherwise. `funcs` won't exist before it is called.

In general, you want the order of code to be:

1) requires/reqscripts and global vars (even if they're declared local like `options` is here)
2) classes and functions
3) arg processing and "main" logic. this is last since it needs to call into the classes and functions defined above

ban-cooking.lua
```diff
@@ -2,6 +2,33 @@
...
+    {'h', 'help', handler=function() options.help = true end},
+    {'v', 'verbose', handler=function() options.verbose = true end},
+})
+
+if options.help == true then
+    print(dfhack.script_help())
+    return
+end
+
+if commands.contains('all') then
```
myk002, line 18: I don't think there is a `contains`, but you probably also don't need to be that general. just testing `commands[1] == 'all'` should be sufficient. You don't need to support `all` appearing later in the argument list.

ban-cooking.lua
```diff
@@ -2,6 +2,33 @@
...
+    for _, v in pairs(funcs) do
+        if _ ~= 'show' then
```
myk002, line 20: ha, I didn't catch this in the original review. `_` is normally used for variables that you *don't* reference. this should have a name.

myk002 (commented):
I agree that printing a total count of banned types at the end would be useful. a global `local count = 0` that increments for every ban and is printed at the end of the script would suffice.

ban-cooking.lua
Moth-Tolias: we did wonder! =p

Moth-Tolias:
should all be fixed now.

ban-cooking.lua
```diff
@@ -410,23 +416,28 @@ funcs.show = function()
...
-local args = {...}
-
-for _, v in ipairs(args) do
+for _, v in ipairs(commands) do
     if funcs[v] then
         funcs[v]()
     end
 end
+
+print('banned ' .. count .. 'items.')
```
myk002, line 443: needs a space before `items` (inside the quote)

ban-cooking.lua
```diff
@@ -410,23 +416,28 @@ funcs.show = function()
     end
 end
 
-if ... == "help" then
+local commands = argparse.processArgsGetopt({...}, {
+    {'h', 'help', handler=function() options.help = true end},
+    {'v', 'verbose', handler=function() options.verbose = true end},
```
myk002, line 421: needs to be documented in docs/ban-cooking.rst

myk002 (approved):

myk002:
oh, sorry, one more thing -- could you add a changelog entry under Misc Improvements?

Moth-Tolias:
how's this?

myk002 (approved):
perfect. Thanks!

PR #550, opened 2023-01-21
Add workaround script to save and restore nicknames

plule:
The Steam release has a current issue where units nicknames are occasionally erased, in events such as killing a forgotten beast.

fix-nicks is a script running once an in-game day that saves unit names in the persistent data, and restore any removed nickname.

**Warning**: That's my first dfhack lua script

It assumes that the current "correct" way to set a nickname is to directly assign .nickname, see: https://github.com/DFHack/dfhack/issues/2680


myk002 (commented):
The core logic looks good, but this code will need a few changes to fit in with other similar scripts.

1) scripts that fix bugs should go in the fix/ subdirectory. I know that we haven't always followed this rule, but new bugfix scripts should go there. Maybe rename the script `fix/nicks`? `fix/restore-nicks`?
2) docs go in a separate .rst file so they can be rendered by our documentation system. docs for this script should go in `docs/fix/nicks.rst` you can copy `docs/fix/drop-webs.rst` to get a good example to follow. Further information here: https://docs.dfhack.org/en/latest/docs/dev/Documentation.html
3) since this script needs to get re-enabled when a player loads a save, it needs to persist and reload its enabled state. you can copy the structure for this from `emigration.lua`. Docs are here: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#enabling-and-disabling-scripts

You can pull in a lot of other boilerplate from `emigration.lua` as well, like the check to make sure the player is in the correct mode when they run the command.

plule:
Thank you for the review @myk002 

> The core logic looks good, but this code will need a few changes to fit in with other similar scripts.
> 
>     1. scripts that fix bugs should go in the fix/ subdirectory. I know that we haven't always followed this rule, but new bugfix scripts should go there. Maybe rename the script `fix/nicks`? `fix/restore-nicks`?

Sure, I went for `fix/restore-nicks` for the sake of being explicit.

>     2. docs go in a separate .rst file so they can be rendered by our documentation system. docs for this script should go in `docs/fix/nicks.rst` you can copy `docs/fix/drop-webs.rst` to get a good example to follow. Further information here: https://docs.dfhack.org/en/latest/docs/dev/Documentation.html

Oh nice, I did not know it would render in-game. Done.

>     3. since this script needs to get re-enabled when a player loads a save, it needs to persist and reload its enabled state. you can copy the structure for this from `emigration.lua`. Docs are here: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#enabling-and-disabling-scripts
> You can pull in a lot of other boilerplate from `emigration.lua` as well, like the check to make sure the player is in the correct mode when they run the command.

I missed that it was not automatically reenabled, I rewrote the boilerplate similarly to the other scripts, and it now works as expected. I kept the nickname persistence with dfhack.persistent though, it was more convenient to work with individual entries.

docs/fix/restore-nicks.rst
```diff
@@ -0,0 +1,30 @@
+fix/restore-nicks
+=================
+
+.. dfhack-tool::
+    :summary: Workaround for the v50.x bug where Dwarf Fortress occasionally erase Dwarf's nicknames.
```
myk002, line 5: This is too long. How about:
`Restore nicknames when DF loses them`.

docs/fix/restore-nicks.rst
```diff
@@ -0,0 +1,30 @@
...
+    :tags: fort bugfix units
+
+Units occasionally lose their nicknames, in event such as killing a forgotten beast.
```
myk002, line 8:
```suggestion
Due to a bug, units occasionally lose their nicknames, such as when killing a forgotten beast.
```

docs/fix/restore-nicks.rst
```diff
@@ -0,0 +1,30 @@
...
+
+`fix/restore-nicks` will save the nicknames of the units once an in-game day, and restore any removed
+nickname.
+
+.. note::
+    It does not make the difference between a nickname removed by the player or by a bug, so
+    it will also prevent to remove the nickname of a dwarf.
```
myk002, line 15:
```suggestion
    It does not distinguish between a nickname removed by the player or by a bug. If you want
    to remove a dwarf's nickname, please run ``fix/restore-nicks forget`` to reset the tracking.
```

docs/fix/restore-nicks.rst
```diff
@@ -0,0 +1,30 @@
...
+-----
+
+``enable fix/restore-nicks``
+    Enable saving and restoring of the nicknames daily.
+
+``disable fix/restore-nicks``
+    Disable saving and restoring the nicknames daily.
+
+``fix/restore-nicks now``
+    Save and restore the nicknames once.
```
myk002, line 27:
```suggestion
    Immediately restore lost nicknames. Note for this to work, ``fix/restore-nicks`` must have
    already been enabled when the dwarf lost their nickname.
```

docs/fix/restore-nicks.rst
```diff
@@ -0,0 +1,30 @@
...
+
+.. note::
+    It does not make the difference between a nickname removed by the player or by a bug, so
+    it will also prevent to remove the nickname of a dwarf.
+
+Usage
+-----
+
+``enable fix/restore-nicks``
+    Enable saving and restoring of the nicknames daily.
```
myk002, line 21:
```suggestion
    Enable daily restoring and saving of the nicknames.
```

docs/fix/restore-nicks.rst
```diff
@@ -0,0 +1,30 @@
...
+
+``disable fix/restore-nicks``
+    Disable saving and restoring the nicknames daily.
```
myk002, line 24:
```suggestion
    Disable the daily check.
```

fix/restore-nicks.lua
```diff
@@ -0,0 +1,120 @@
+-- Workaround for the v50.x bug where Dwarf Fortress occasionally erase Dwarf's nicknames.
+-- It happen when killing certain figures, such as forgotten beasts.
+--@ enable = true
```
myk002, line 3: you will also need `--@ module = true` so your script can automatically attach the state change handler and reenable itself after a load.

docs/fix/restore-nicks.rst
myk002: I know in the code you actually save first, and it looks like correct behavior, but if you say saving first here it might  imply that the data is overwritten before being read, losing the nickname history. This isn't true, but it could be confusing for players.

fix/restore-nicks.lua
```diff
@@ -0,0 +1,120 @@
...
+    enabled = (persisted_data or {enabled=false})['enabled']
+    event_loop()
+end
+
+if dfhack_flags.module then
+    return
+end
+
+if df.global.gamemode ~= df.game_mode.DWARF or not dfhack.isMapLoaded() then
+    -- Possibly to review with adventure mode
```
myk002, line 90: yeah, I don't know how adventure mode will fit into this. we'll have to see.

fix/restore-nicks.lua
```diff
@@ -0,0 +1,120 @@
...
+if dfhack_flags and dfhack_flags.enable then
+    args = {dfhack_flags.enable_state and 'enable' or 'disable'}
+end
+
+if args[1] == "enable" then
+    enabled = true
+elseif args[1] == "disable" then
+    enabled = false
+elseif args[1] == "now" then
+    print("Saving and restoring the nicknames")
```
myk002, line 105:
```suggestion
    print("Restoring and saving nicknames")
```

docs/fix/restore-nicks.rst
plule: Yep, I can see how it could be confusing

fix/restore-nicks.lua
plule: Oh? It looked like it was correctly reenabled after restarting the save, but I may have missed something, thanks

fix/restore-nicks.lua
plule: Let's hope the bug is resolved by then!

myk002 (approved):
This needs a changelog entry, but I can add that for you

PR #551, opened 2023-01-21
Update make-monarch to use assignment_vector_idx

20k:
This updates the entity link to use `assignment_vector_idx`, which is new in 50.xx


myk002 (approved):
have you tested that this script works now?

20k:
Yep, I tried a bunch title transfers and tested a save/reload, seems to work as expected

transferring

myk002:
cool. I'll mark it on the spreadsheet. No need to update the tags in the .rst file. I flush those from the spreadsheet to the docs periodically.

PR #553, opened 2023-01-22
Fix cage selection bugs and ignore caged vermin

arekatir:


stripcaged.lua
```diff
@@ -1,8 +1,29 @@
+local rawargs = {...}
+local args = require 'utils'.invert(rawargs)
```
myk002, line 2: you are better off using `argparse.processArgsGetopt()`. it returns the non-option params as a list.

stripcaged.lua
```diff
@@ -1,8 +1,29 @@
...
+
 local function plural(nr, name)
     -- '1 cage' / '4 cages'
     return string.format('%s %s%s', nr, name, nr ~= 1 and 's' or '')
 end
 
+local function isexcludedvermin(item)
```
myk002, line 9: consider passing an `opts` table here instead of reading from global args lists

stripcaged.lua
arekatir: Thanks. My latest commits try to address this.

stripcaged.lua
```diff
@@ -252,8 +263,13 @@ else
     end
 end
 
+-- Print help if requested
+if opts.help then
+    print(dfhack.script_help())
```
myk002, line 268:
```suggestion
    print(dfhack.script_help())
    return
```

myk002 (approved):

PR #557, opened 2023-01-22
add autofish.lua

AndrielChaoti:
a script to help potential fps problems because of thousands of items being left laying around. stops your fisherdwarves when you reach a certain configurable amount of fish.

Here's an inital pull request for peer review.


autofish.lua
```diff
@@ -0,0 +1,220 @@
+-- handles automatic fishing jobs to limit the number of fish the fortress keeps on hand
+-- autofish [enable | disable] <max> [min] [--include-raw | -r]
```
myk002, line 2: docs should go in `docs/autofish.rst`
You can copy `docs/stripcaged.rst` to get the skeleton structure. More info here: https://docs.dfhack.org/en/latest/docs/dev/Documentation.html

autofish.lua
```diff
@@ -0,0 +1,220 @@
...
+
+local help = [====[
+autofish
+=============
+Manages your fish stocks by toggling fishing labours as you reach certain stock thresholds.
+
+
+Usage
+-------------
+  autofish [enable | disable | status] | <max> [min] [--include-raw | -r]
```
myk002, line 21: I'd suggest:
```
enable autofish
autofish [status]
autofish <max> [<min>] [<options>]
```

we're trying to standardize on `enable X` instead of `X enable` so we have a consistent interface for the upcoming `gui/control-panel`. It's ok to still support the `X enable` syntax, but we've been avoiding documenting that to promote usage of the `enable` interface.

autofish.lua
```diff
@@ -0,0 +1,220 @@
+-- handles automatic fishing jobs to limit the number of fish the fortress keeps on hand
+-- autofish [enable | disable] <max> [min] [--include-raw | -r]
+
+--@ enable=true
```
myk002, line 4: You'll also need `--@ module=true` here too so the state change hooks can be attached
docs link (though I know you've already read this):
https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#enabling-and-disabling-scripts

autofish.lua
```diff
@@ -0,0 +1,220 @@
...
+-r, --include-raw
+  Allow the stock tracking to also count raw fish
+enable
+  Enable this script
+disable
+  Disable this script
+status
+  Print the current status
+max
+  The maximum number of fish you want to stop fishing at.
```
myk002, line 34: it would be useful to document the defaults for max and min

autofish.lua
```diff
@@ -0,0 +1,220 @@
...
+        end
+    end
+    isFishing = state -- save current state
+end
+
+-- check if an item isn't forbidden/rotten/on fire/etc..
+function isValidItem(item)
+    local flags = item.flags
+    if flags.rotten or flags.trader or flags.hostile or flags.forbid
+        or flags.dump or flags.on_fire or flags.garbage_collect then
```
myk002, line 87: you probably also want to check `removed`, `encased`, `spider_web`, and maybe `owned`

autofish.lua
```diff
@@ -0,0 +1,220 @@
...
+        if v:getType() == df.item_type.FISH and isValidItem(v) then
+            prepared = prepared + v:getStackSize()
+        end
+        if (v:getType() == df.item_type.RAW_FISH and isValidItem(v)) and set_useRaw then
+            raw = raw + v:getStackSize()
+        end
+    end
+
+    -- hande pausing/resuming labour
+    local sumFish = prepared + raw
```
myk002, line 109: this looks like it can go inside the `if set_useRaw then` block

autofish.lua
```diff
@@ -0,0 +1,220 @@
...
+    local work_details = df.global.plotinfo.hauling.work_details
+    for k,v in pairs(work_details) do
+        if v.allowed_labors.FISH then
+            -- set limited to true just in case a custom work detail is being
+            -- changed, to prevent *all* dwarves from fishing.
+            v.work_detail_flags.limited = true
+            v.work_detail_flags.enabled = state
+        end
+    end
+    isFishing = state -- save current state
```
myk002, line 80: if we have changed state, it might be nice to report that with a `print('autofish: threshold crossed; stopping fishing')` so the player can see the record on the console if they're confused why people suddenly started/stopped fishing

autofish.lua
myk002, line 109: if this is written as `numFish = set_useRaw and (prepared + raw) or prepared` then you can consolidate the logic below.

AndrielChaoti:
Taken the time to resolve the things you've brought up, let me know if anything else trips any alarms :)

docs/autofish.rst
```diff
@@ -0,0 +1,42 @@
+autofish
+========
+
+.. dfhack_tool::
+    :summary: Keep a specific quantity of fish in your fortress.
+    :tags: fort auto fps jobs
```
myk002, line 6:
```suggestion
    :tags: fort auto fps labors
```

docs/autofish.rst
```diff
@@ -0,0 +1,42 @@
...
+    ``max`` (default: 100) controls the maximum amount of fish you  want to keep
+    on hand in your fortress. Fishing will be disabled when the amount of fish
+    goes above this value.
+
+    ``min`` (default: 50) controls the minimum fish you want before restarting
+    fishing.
+
+    Use ``--include-raw``(``-r``) (default: off) to toggle letting the script
+    also count your raw fish as part of your quota. Use it a second time to
+    disable this.
```
myk002, line 34: Maybe this should default to on (and be renamed `--toggle-raw`). Isn't it akin to designating trees in autochop? we stop designating there when the expected yield will cause your stock to reach the threshold. Isn't it the same here? otherwise if you have lazy fish cleaners, you'll just keep on amassing raw fish.

docs/autofish.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Examples
+--------
+
+``autofish -r 150``
```
myk002, line 39: I think the first example should just be `enable autofish`, since that is the simplest use case, and is all a user really needs to do.

docs/autofish.rst
myk002: right? it's affecting labor assignment, not the actual job

docs/autofish.rst
```diff
@@ -0,0 +1,42 @@
+autofish
+========
+
+.. dfhack_tool::
```
myk002, line 4:
```suggestion
.. dfhack-tool::
```

docs/autofish.rst
AndrielChaoti: That's right.

docs/autofish.rst
```diff
@@ -0,0 +1,42 @@
...
+    :summary: Keep a specific quantity of fish in your fortress.
```
myk002, line 5:
```suggestion
    :summary: Auto-manage fishing labors to control your stock of fish.
```

docs/autofish.rst
```diff
@@ -1,9 +1,9 @@
 autofish
 ========
 
-.. dfhack_tool::
+.. dfhack-tool::
     :summary: Keep a specific quantity of fish in your fortress.
-    :tags: fort auto fps jobs
+    :tags: fort auto fps labor
```
myk002, line 6:
```suggestion
    :tags: fort auto fps labors
```

myk002 (approved):

myk002:
whoops! sorry, one more thing. Could you add a line to `changelog.txt` under `New Scripts`?

myk002 (approved):

myk002:
Congrats! you made it to the spreadsheet! https://docs.google.com/spreadsheets/d/1hiDlo8M_bB_1jE-5HRs2RrrA_VZ4cRu9VXaTctX_nwk/edit#gid=1774645373

Thank you very much for this. Great job!

PR #562, opened 2023-01-23
Port show-unit-syndromes from Ruby to Lua

vallode:
Fixes DFHack/dfhack#2081

The final ruby port, this one took/is taking a while. Most of the actual nitty gritty is done, there's a few functions to port over (getEffectCreatureName, etc.) but overall it's nearing completion.

Don't bother reviewing yet, myk. I need at least 2-3 more passes over this to bring it to the ready state. I'll let you know when it's ready :)

Major changes/diversions from `show-unit-syndromes.rb`:

- No HTML output at all
- No mapping effects to colors (GREEN = good, RED = bad, etc.)

PR for show-unit-syndromes docs removal: https://github.com/DFHack/dfhack/pull/2733


gui/unit-syndromes.lua
```diff
@@ -0,0 +1,543 @@
...
+            material and ("vs. %s"):format(material.stone_name)
+        )
+    end,
+    -- TODO: Unfinished.
+    [df.creature_interaction_effect_type.BODY_MAT_INTERACTION] = function(effect)
+        return ("%s %s"):format(effect.interaction_name, effect.interaction_id)
+    end,
+    -- TODO: Unfinished.
+    [df.creature_interaction_effect_type.BODY_APPEARANCE_MODIFIER] = function(effect)
+        return ("TODO"):format(effect.interaction_name, effect.interaction_id)
```
myk002, line 213: This line actually does look unfinished. are the other "Unfinished" comments still true?

gui/unit-syndromes.lua
```diff
@@ -0,0 +1,543 @@
...
+                    cursor_pen = COLOR_WHITE,
+                    text_to_wrap = '',
+                }
+            }
+        },
+        widgets.HotkeyLabel{
+            frame={l=0, b=0},
+            label='Back',
+            auto_width=true,
+            key='KEYBOARD_CURSOR_LEFT',
```
myk002, line 425: This is displayed as a number since the keypad is also bound to this action. In other tools I render over the number with an arrow symbol. e.g. https://github.com/DFHack/scripts/blob/master/quickstart-guide.lua#L78-L82

myk002: Note that you can also use `key='LEAVESCREEN'` to use Esc for "back", which other tools like `gui/gm-unit` do. This is perfectly acceptable. Since you deactivate the hotkey when you can no longer go back, the ZScreen code will still be able to dismiss the screen properly when the player hits Esc then.

Note if you use Esc/LEAVESCREEN, though, you should also have an onInput() function that interprets keys._R_MOUSE_DOWN to also call `previous_page` when `is_not_main_page()`

gui/unit-syndromes.lua
vallode: I checked all the other todos, there was one that was no longer true. The rest of the unfinished lines are either directly from the old script (like this one) or parts that I couldn't quite crack yet.

gui/unit-syndromes.lua
vallode: That's a great call, had no idea that zscreen manages the esc logic. Good to know.

changelog.txt
```diff
@@ -30,6 +30,7 @@ that repo.
 - `locate-ore`: now only searches revealed tiles by default
 
 ## Removed
+-- `show-unit-syndromes`: replaced by `gui/unit-syndromes`. exporting html is no longer supported.
```
myk002, line 33: Single dash here

gui/unit-syndromes.lua
```diff
@@ -422,14 +421,20 @@ function UnitSyndromes:init()
...
             on_activate=previous_page,
             active=is_not_main_page,
             enabled=is_not_main_page,
         },
     }
 end
 
+function UnitSyndromes:onInput(keys)
+    if keys._R_MOUSE_DOWN then
+        self:previous_page()
```
myk002, line 434: Make sure you return true if you handled the r click, otherwise ZScreen will still close the window

myk002:
Be sure to make a sister PR in the DFHack/dfhack repo to add `show-unit-syndromes` to `docs/about/Removed.rst`. that will fix the docs build errors

myk002 (approved):

PR #563, opened 2023-01-24
added pregnancy-postpone.lua

TaxiService:
a workaround to this crash https://www.bay12games.com/dwarves/mantisbt/view.php?id=11549 
can also be used for checking for pregnancies on the map

### DIFF
--- a/pregnancy-postpone.lua
+++ b/pregnancy-postpone.lua
@@ -1,0 +1,328 @@
+local help = [====[
+pregnancy-postpone
+=============
+Delays childbirths by a week or an user-definable amount of time (in ticks).
+Used as a workaround for a crash involving werebeasts giving birth. (https://www.bay12games.com/dwarves/mantisbt/view.php?id=11549)
+Can also be used simply to gather information about pregnancies in the world/playable area.
+Can act on: a selected unit, citizens on the map, non-citizens on the map, or worldwide.
+
+Examples
+=============
+"pregnancy-postpone --info-only --citizens --non-citizens"
+    Prints pregnancy information about all creatures in the playable area.
+
+"pregnancy-postpone --world --ticks 1200"
+    Postpones historical figures' childbirths worldwide by one day. (including those on your map)
+
+"pregnancy-postpone -iu"
+	Prints pregnancy information about the selected unit.
+
+"pregnancy-postpone -u"
+	Postpones selected unit's childbirth by a week. (the default amount of time)
+
+Options
+=============
+
+"-i" "--info-only"
+    Print pregnancy information without changing anything.
+
+"-s" "--silent"
+    Do not print pregnancy info.
+
+"-u" "--unit"
+    Acts only on the selected unit.
+
+"-c" "--citizens"
+    Acts on fort-controlled citizens in the playable area. (excludes animals)
+
+"-x" "--non-citizens"
+    Acts on all non-citizen units in the playable area, like animals and guests. (excludes citizens)
+
+"-w" "--world"
+    Act on the whole world. (INCLUDING your map, for now)
+
+"-t <number>" "--ticks <number>""
+    Postpone births by this many ticks.
+    If unspecified, the default value is 8400. (1 week)
+    in-game-time conversion table:
+      1 hour  = 50 ticks
+      1 day   = 1200 ticks
+      1 week  = 8400 ticks
+      1 month = 33600 ticks
+      1 year  = 403200 ticks
+
+Notes
+=============
+Short optional arguments can be chained together! 
+   this command: "postpone-pregnancy --silent --citizens --ticks 2400"
+is identical to: "postpone-pregnancy -sct2400"
+Just make sure that the "-t <number>" argument comes last, if you want to use it.
+
+For best results, be sure that the game is paused while running the script!
+]====]
+--[[ Credits:
+---------------
+made by TaxiService during a secretive mood.
+  super duper turbo thanks to:
+<3 ab9rf - for hunting down the cause of my map's crash with a debugger, setting me off on this coding journey.
+<3 sdegrace - for mentioning how to use the lua console and "~" to print all. Invaluable tools for testing!
+<3 myk002, Lethosor, nuvu (vallode), ab9rf (again) - for always pointing me in the right direction and putting up with me and my questions!
+<3 all the giants from the DFHack community, on whose shoulders i stood on! 
+
+notes for function grafters:
+  Whenever i refer to a "dude", it means a "unit OR historical_figure" 
+  ...otherwise, i tried to go for "readable" over "compact". hopefully it'll be understandable enough!
+--]]
+local argparse = require 'argparse'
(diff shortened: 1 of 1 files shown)


pregnancy-postpone.lua
```diff
@@ -0,0 +1,328 @@
...
+For best results, be sure that the game is paused while running the script!
+]====]
+--[[ Credits:
+---------------
+made by TaxiService during a secretive mood.
+  super duper turbo thanks to:
+<3 ab9rf - for hunting down the cause of my map's crash with a debugger, setting me off on this coding journey.
+<3 sdegrace - for mentioning how to use the lua console and "~" to print all. Invaluable tools for testing!
+<3 myk002, Lethosor, nuvu (vallode), ab9rf (again) - for always pointing me in the right direction and putting up with me and my questions!
+<3 all the giants from the DFHack community, on whose shoulders i stood on!
```
myk002, line 70: hear hear

pregnancy-postpone.lua
```diff
@@ -0,0 +1,328 @@
+local help = [====[
```
myk002, line 1: since this is primarily a crash bugfix tool, I'd suggest moving it to the `fix` subdir. The question is, how will players know to run it? will this name mean something to them and they can say "oh, I can fix my crash with this tool"? is there a specific error message we can tell them to look out for? something in the docs that will come up in a google search?

pregnancy-postpone.lua
```diff
@@ -0,0 +1,328 @@
+local help = [====[
+pregnancy-postpone
+=============
+Delays childbirths by a week or an user-definable amount of time (in ticks).
+Used as a workaround for a crash involving werebeasts giving birth. (https://www.bay12games.com/dwarves/mantisbt/view.php?id=11549)
```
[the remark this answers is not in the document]
TaxiService: oh god, when you put it like that... not really?
we need a report for current version, i guess? i *did* send one to mantishub, but it's been a week or more and haven't heard back from them. :\ (and i dont see it posted on their page either)

i presumed it was that bug because ab9rf found that my map crashed due to missing genes info or something. 
now, i don't know what crash this script really prevents, but has some sort of correlation with pregnancies... right?

pregnancy-postpone.lua
```diff
@@ -0,0 +1,328 @@
...
+<3 sdegrace - for mentioning how to use the lua console and "~" to print all. Invaluable tools for testing!
+<3 myk002, Lethosor, nuvu (vallode), ab9rf (again) - for always pointing me in the right direction and putting up with me and my questions!
+<3 all the giants from the DFHack community, on whose shoulders i stood on!
+
+notes for function grafters:
+  Whenever i refer to a "dude", it means a "unit OR historical_figure"
+  ...otherwise, i tried to go for "readable" over "compact". hopefully it'll be understandable enough!
+--]]
+local argparse = require 'argparse'
+local args = {
```
myk002, line 77: in almost all scripts, this would be called `opts`, not `args` (which is usually used for incoming arguments, e.g. `local args = {...}`)

pregnancy-postpone.lua
```diff
@@ -0,0 +1,328 @@
...
+      1 year  = 403200 ticks
+
+Notes
+=============
+Short optional arguments can be chained together!
+   this command: "postpone-pregnancy --silent --citizens --ticks 2400"
+is identical to: "postpone-pregnancy -sct2400"
+Just make sure that the "-t <number>" argument comes last, if you want to use it.
+
+For best results, be sure that the game is paused while running the script!
```
myk002, line 61: I'm trying to wrap my head around the user experience for this script. How is a player intended to interact with this script? should it be run on repeat in the background as a precaution? is the player supposed to save often and if DF crashes, load the last save and then run this script then? If it prevents a crash, what then? should the player just keep on postponing pregnancies forever?

pregnancy-postpone.lua
TaxiService: ah crap. well, it's an easy change to make, at least!

pregnancy-postpone.lua
TaxiService: it's a one-time action that the user performs.
in my play-flow, i tend to save often, and if not, to rely on somewhat frequent autosaves.
so yeah in my case i just load the last save that is playable, and then postpone pregnancies once by an amount of time.

it's weird because i had two versions of the same save that would crash. 
One took about 5 minutes and the other less than a minute, in real time.
if i ran the command as soon as the game was loaded:
the 1 minute one was fixed by postponing pregnancies by 1 in-game day
the 5 minutes one was not! i had to postpone them for a week, so either the problem entity wasnt pregnant yet, or... idk, somehow it wasnt enough time then?


but- yeah. the idea is that the user would run this when they have a crashing save. 
When they do, they try postponing pregnancies by an amount of time, and if that fixes it: good!
i guess the best course of actions would be to wait for as close as possible to the crashing time before running it

TaxiService:
anyway i don't know, maybe i should cancel the pull request until i've made a documentation entry?
also the script is kind of a blanket solution for a problem that could be fixed in other ways...
like, instead of postponing ALL pregnancies i should probably only postpone those of cursed creatures; or also detect pregnancies without genes and null them out or something. :\

i'm sorry, like, i dont know much about pullrequesting and stuff ;_;

myk002:
haha, it's a good script, I'm just having trouble figuring out how to make sure it gets thought of and used by players when it will fix their problems.

So it seems like we have two primary use cases: a general pregnancy interface and a fix for a specific crash.

for the crash fix, it would be great if it fixed the problem *before* a player's game crashes. I suggest a script that just scans for crash-causing pregnancies and removes them, printing a message to the console when it does something so we know it's working. we can run it on repeat in the background automatically. I'm writing a `gui/control-panel` interface that can be used to control bugfix scripts like this.

a separate general pregnancy interface is also useful. maybe postponing could be a function of that interface, but it probably shouldn't be the primary feature (or the name).

how does that sound?

TaxiService:
sounds great! indeed, a general pregnancy manager would be a much more interesting tool.

the actual functioning part of my script are basically four lines of code or something. Most of the logic is for printing crap in the console... so it wouldnt even be necessary if such a tool was controlled via a gui!

now, like, if i were to undertake this task... it will probably take a long time, i will have to research both how to detect such problematic preggers, and to teach myself how to handle the gui systems...
...and also come up with more useful pregnancy tools, like "create" or "remove", or even just "check for pregnancies" (which my tool already does, but you've gotta use the -i argument... and instead it could be a separate function in an of itself!)

i think i read of someone else also working on pregnancy stuff, in some discord channel? i honestly havent been able to keep up with all the chats, pullrequest discussions and crap..! 
anyway i guess i can just close this PR for now and come back once i've got something more substantial? what happens if i click this...

TaxiService:
ok that's what. makes sense!
so yeah hit me up if anything comes up.
also, if anyone else is getting that crash, in the meantime this script can be used to circumvent crashing!

PR #565, opened 2023-01-24
modify allneeds prefix rename+negative needs

rslgp:


docs/allneeds.rst
```diff
@@ -7,10 +7,39 @@ allneeds
 
 Provides an overview of the needs of the fort in general, output is sorted to
 show most unfullfiled needs.
+show only needs that needs attention now (damaging focus) and garantee it shows on dwarf overview
+(obs: multiple Prays counts as unique Pray need)
+
+NEEDS_CODE is usually capital letters from each need
```
myk002, line 13: Could you explain a bit how nicknaming the dwarves makes it easier to care for them? instead of nicknaming dwarves, would it be useful to show which dwarves have which needs in a visual list? Clicking on the dwarf name could jump the screen to that dwarf so the player can take action. The list could be left up as a status panel so the player can refer to it as they are assigning dwarves to things that will satisfy them.

rslgp: in my game im using this mod in this way, i put repeat allneeds u, and have a custom labor CO for craft object for crafts with wood,stone,bone and add the dwarves with CO need and remove when CO is satisfied, 

and another frequent use is martial training, i have 6 squads with no uniform, that i use dwarf therapist to assign into them if nickname contains MT, and after some time, i filter dwarfs with squad that not contains MT in nickname to remove it from squad/training

and less often i have a custom labor for stone encraving, that i assign is nick contains BC be creative

if nick contains PM pray or meditate i can barrow all inside temple to pray

usually have 50 dwarfs with same needs, be able to manage multiple without clicking one by one is the motivation to use this, if i need to access the name again i send allneeds r

---
about your suggestion, if its is possible to select multiples using gui after listing it (but currently i dont know how to code it, but im interested to learn or pair prog this code if its possible), it will be useful as alternative for naming it or one more option, because renaming make compatible to use with dwarf therapist

allneeds.lua
```diff
@@ -1,20 +1,110 @@
...
+    ["AcquireObject"]              =   "AO"   ,
+    ["EatGoodMeal"]              =   "EM"   ,
+    ["Fight"]                      =     "Fi"  ,
+    ["CauseTrouble"]              =    "CT"  ,
+    ["Argue"]                      =     "A"  ,
+    ["BeExtravagant"]              =     "BE" ,
+    ["Wander"]                     =      "W" ,
+    ["HelpSomebody"]              =     "HS" ,
+    ["ThinkAbstractly"]          =     "TA" ,
+    ["AdmireArt"]                  =     "AA" ,
```
lethosor, line 75: This table is difficult to read with the random extra whitespace. Could you reformat it?

lethosor (commented):
I'm also not sure that building a custom rename feature into this one script is the way to go, but I don't use DT.

If we do go with this, I would really rather have more descriptive arguments. Single-letter arguments aren't self-explanatory. Also, by making them the only arguments to the script, it prevents us from adding arguments in the future to do _anything else_, like custom display/filtering options. Instead, I would prefer something more verbose, like

```
allneeds add-nicknames
allneeds remove-nicknames
```

or alternatively:

```
allneeds nicknames add
allneeds nicknames remove
```

or even

```
allneeds -n|--nicknames add
allneeds -n|--nick remove
```

rslgp:
i think is better it to be a separated method, i was unsure about create a new one, but i think is only create a .lua, .rst and edit changelog, correct me if im wrong

allneeds.lua
rslgp: done

rslgp:
i think is better this be a new command, its restricts to only unmet needs

PR #566, opened 2023-01-25
Make locate-ore work on discovered veins by default

vallode:
Closes https://github.com/DFHack/dfhack/issues/2426

locate-ore now takes a `--all` option, which allows it to look at undiscovered veins/ores.

### DIFF
--- a/locate-ore.lua
+++ b/locate-ore.lua
@@ -1,4 +1,7 @@
--- scan the map for ore veins
+-- Scan the map for ore veins
+
+local argparse = require('argparse')
+
 local tile_attrs = df.tiletype.attrs
 
 local function extractKeys(target_table)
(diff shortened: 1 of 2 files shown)


locate-ore.lua
```diff
@@ -132,25 +139,36 @@ local function getOreDescription(ore)
...
-local target = args[1]
+local options, args = {
+    help = false,
+    list = false,
+    show_undiscovered = false
+}, {...}
+
+local positionals = argparse.processArgsGetopt(args, {
+    {'h', 'help', handler=function() options.help = true end},
+    {'l', 'list', handler=function() options.list = true end},
```
myk002, line 150: unless we want `--list` to be an option restricting the output of a command (e.g. `locate-ore hematite --list`), I think we should stick to `list` being a command instead of an option.

myk002 (commented):
thanks for this! I've never looked at this script before. I see now why it doesn't support other types of rock.

myk002 (approved):

PR #567, opened 2023-01-25
Refactor fix/forbid-unreachable into forbid and add unreachable check for unforbid

vallode:
As per myk's suggestions in https://github.com/DFHack/dfhack/issues/2421 I think these two changes are great in conjunction.

`fix/forbid-unreachable` is now `forbid` and has added functionality: `list`, `all`, `unreachable`.
`unforbid` will not skip items that are unreachable unless you pass the `-u`,`--include-unreachable` flag.


docs/fix/forbid-unreachable.rst
```diff
@@ -0,0 +1,15 @@
+fix/forbid-unreachable
```
myk002, line 1: I suggest actually calling this script just `forbid` and make it a general purpose productivity tool instead of a "bugfix". It could have the following syntax:

`forbid` - print out the number of forbidden items per type (and a total at the end)
`forbid all` - forbids all items (useful if player wants to go selectively unforbid just a few items)
`forbid unreachable` - forbids unreachable items

with tags: `fort`, `productivity`, `items`

docs/unforbid.rst
```diff
@@ -19,5 +19,8 @@ Usage
 Options
 -------
 
+``-a``, ``--allow-unreachable``
```
myk002, line 22: `-a` usually means "add" or "all" (which has different semantics here). How about `-u`, `--include-unreachable`?

fix/forbid-unreachable.lua
```diff
@@ -0,0 +1,18 @@
+-- Fixes job cancellation spam when trying to use unreachable items.
+
+local count = 0
+for _,item in pairs(df.global.world.items.all) do
+    local pos = item.pos
+    local block = dfhack.maps.getTileBlock(pos)
+
+    if block then
+        local walkable = block.walkable[pos.x%16][pos.y%16]
```
myk002, line 9: note that walkability is not quite the same as reachability, since a tile can be walkable (a floor) without being reachable (behind a closed drawbridge). For the purposes of this tool, though, I think the terminology and behavior as you have specified is perfectly fine. Maybe just call out this distinction in the docs.

vallode: Added a note, good point.

docs/fix/forbid-unreachable.rst
vallode: That was fun!

forbid.lua
```diff
@@ -0,0 +1,94 @@
...
+
+    -- TODO: Check item container walkability.
+    if block then
+        local walkable = block.walkable[item.pos.x % 16][item.pos.y % 16]
+
+        if walkable == 0 then
+            return false
+        else
+            return true
+        end
```
myk002, line 16:
```suggestion
        return block.walkable[item.pos.x % 16][item.pos.y % 16] ~= 0
```

forbid.lua
```diff
@@ -0,0 +1,94 @@
...
+
+    local count = 0
+    for _, item in pairs(df.global.world.items.all) do
+        if not isItemReachable(item) then
+            item.flags.forbid = true
+            count = count + 1
+        end
+    end
+
+    print(("Forbid %s items, you can unforbid them by running `unforbid all`"):format(count))
```
myk002, line 93: I believe the past tense of "forbid" is "forbade"

forbid.lua
```diff
@@ -0,0 +1,94 @@
+-- Forbid and list forbidden items on the map.
+
+local argparse = require('argparse')
+
+local function isItemReachable(item)
+    local block = dfhack.maps.getTileBlock(item.pos)
+
+    -- TODO: Check item container walkability.
```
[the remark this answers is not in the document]
vallode: I'm not entirely sure, it was my hunch that the items I saw with positions of -30000 were in containers of some sorts. i.e many items on the map I was testing with had no `block` value.

forbid.lua
vallode: Tomato tomato but I thought "[The tool] just forbid n items", not necessarily past tense here. Forbade is clearer, though.

myk002 (approved):

PR #569, opened 2023-01-26
Add create-corpsepiece.lua script

ppaawwll:
i wrote this script (in the skeleton of create-items) to both test something and because create-items itself was unable to spawn proper corpsepiece items on its own. hopefully this will be useful to other people as well.


myk002:
Instead of a separate script, could you integrate this into the existing gui/create-item script?

ppaawwll:
> Instead of a separate script, could you integrate this into the existing gui/create-item script?

perhaps, perhaps. i'll have to dissect and look into gui/create-item to figure out where and how exactly to implement it, but i'll certainly try, and open a separate pull request once i'm done.

PR #570, opened 2023-01-26
Autofish bugfixes

AndrielChaoti:
Oops. This is what I get for trying to code tired. 😛 

- fixed a bug causing autofish to not count raw fish.
- fixed persistent data not being loaded and updated, causing state desyncs.
- reduced update frequency to 7 days.
- fixed running main code when sending disable command.
- comment cleanup.

### DIFF
--- a/autofish.lua
+++ b/autofish.lua
@@ -7,6 +7,7 @@
 local json = require("json")
 local persist = require("persist-table")
 local argparse = require("argparse")
+local repeatutil = require("repeat-util")
 
 local GLOBAL_KEY = "autofish"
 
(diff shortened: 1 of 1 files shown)


autofish.lua
```diff
@@ -169,14 +169,19 @@ local positionals = argparse.processArgsGetopt(args,
...
 })
 
+load_state()
 if positionals[1] == "enable" then
     enabled = true
-    --print_status()
+
 elseif positionals[1] == "disable" then
     enabled = false
+    persist_state()
```
myk002, line 178: Don't you need to cancel the repeat somewhere around here?

AndrielChaoti: don't timers auto-cancel when a map is unloaded when they aren't running per tick?

myk002: What if the player disables the script with the `disable` command?

myk002 (approved):

PR #573, opened 2023-01-28
solve the inception problem

myk002:
Fixes DFHack/dfhack#2740
launcher commands can't launch themselves or else the viewscreen stack gets very unhappy


ab9rf:
while this avoids the specific issue we should probably still put a safety check in core in case some other script manages to create a circular reference. a depth counter is probably sufficient

myk002:
you're 100% right. I'll solve this on the Screen::Hide side, which is the root cause of this particular issue

PR #574, opened 2023-01-28
gui/liquids: add river sources, setting/adding/removing from liquid levels, and cleaning flags

vallode:
Adds a few of the older features of the liquids plugin back into the GUI.

Wondering about the best way to revert setting a river source, right now the tool doesn't provide anything. Once you set it, it's there until you remove it with something like tiletypes.

About `modtools/spawn-liquids` resetting temperature... I noticed that the old liquids plugin did this everywhere but was wondering if we maybe put a boolean on that? Is there a good scenario in which someone might _not_ want to reset temperatures (they'd be updated pretty quickly regardless?)?


myk002:
Unless we have a solid reason to *not* set the temp, I think we should set it. Otherwise it's just lost fps while the temp stabilizes

docs/gui/liquids.rst
```diff
@@ -5,13 +5,14 @@ gui/liquids
     :summary: Interactively paint liquids onto the map.
     :tags: fort armok map
 
-This tool is a gui for spawning water or magma. Allowing you to click and paint
-liquid onto the map.
+This tool is a gui for spawning water, magma, and river sources on the map.
+Allowing you to click and paint liquid onto the map. It can optionally clean
```
myk002, line 9: This appears to be a sentence fragment. Perhaps combine with the previous sentence?

vallode:
> Unless we have a solid reason to _not_ set the temp, I think we should set it. Otherwise it's just lost fps while the temp stabilizes

I see that makes total sense, in that case I leave it as it is with the temperature being reset always.

gui/liquids.lua
```diff
@@ -34,71 +43,112 @@ function SpawnLiquid:init()
...
+                { label = "Drag ", value = SpawnLiquidPaintMode.DRAG, pen = COLOR_WHITE },
+            },
+            initial_option = 1,
+            on_change = function(new, _) self.paint_mode = new end,
+        },
+        widgets.CycleHotkeyLabel{
+            frame = {l = 18, b = 2},
+            label = 'Mode:',
+            auto_width = true,
+            key = 'CUSTOM_E',
```
myk002, line 94: If you'd like to enable movement key passthrough, avoid wasdec as hotkeys

gui/liquids.lua
```diff
@@ -178,6 +232,7 @@ end
 SpawnLiquidScreen = defclass(SpawnLiquidScreen, gui.ZScreen)
 SpawnLiquidScreen.ATTRS {
     focus_path='spawnliquid',
+    pass_movement_keys=true,
```
myk002, line 235: You probably also want to set `pass_mouse_clicks=false` and `force_pause=true` since you're affecting the map

myk002:
> Wondering about the best way to revert setting a river source, right now the tool doesn't provide anything. Once you set it, it's there until you remove it with something like tiletypes.

funny you should bring up tiletypes. I just added a Lua API for tiletypes that you can use. Docs here: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#tiletypes and example usage here: https://github.com/DFHack/scripts/blob/master/build-now.lua#L476-L479

myk002:
did you want to do the tiletypes integration in this PR, or should that be for the future?

vallode:
> did you want to do the tiletypes integration in this PR, or should that be for the future?

Good question... I can replace the existing river source spawning mechanic with the tiletypes functions but I'm not sure about how we would even handle an undo process in this tool.

myk002:
it's ok to leave undo out, I think. eventually we'll have `gui/tiletypes` to make that kind of cleanup easier

changelog.txt
```diff
@@ -30,6 +30,8 @@ that repo.
 - `gui/quickfort`: don't close the window when applying a blueprint so players can apply the same blueprint multiple times more easily
 - `locate-ore`: now only searches revealed tiles by default
 - `unforbid`: does not unforbid unreachable items by default anymore.
+- `modtools/spawn-liquid`: resets temperature to "normal" levels, when spawning water or magma
```
myk002, line 33:
```suggestion
- `modtools/spawn-liquid`: sets tile temperature to stable levels when spawning water or magma
```

docs/gui/liquids.rst
```diff
@@ -5,13 +5,14 @@ gui/liquids
...
-liquid onto the map.
+This tool is a gui for spawning water, magma, and river sources on the map,
+allowing you to click and paint liquid onto the map. It can optionally clean
+existing liquids on the map from salt or being stagnant.
 
 .. warning::
 
     There is **no undo support**. Be sure you know what you are doing before
-    spawning any liquid!
+    spawning any liquid, especially river sources.
```
myk002, line 15:
```suggestion
    spawning any liquid, especially river sources!
```

myk002: : )

myk002 (approved):

PR #575, opened 2023-01-28
Add underwater check to unforbid all

vallode:
Closes: https://github.com/DFHack/dfhack/issues/2417
Also this can be closed now too: https://github.com/DFHack/dfhack/issues/2421

### DIFF
--- a/docs/forbid.rst
+++ b/docs/forbid.rst
@@ -6,7 +6,7 @@ forbid
     :tags: fort productivity items
 
 List forbidden items, forbid all items on the map, or forbid unreachable items.
-Forbidding unreachable items checks any of your citizens can walk to the item.
+Forbidding unreachable items checks if any of your citizens can walk to the item.
 
 Usage
 -----
--- a/docs/forbid.rst
+++ b/docs/forbid.rst
@@ -6,8 +6,7 @@ forbid
     :tags: fort productivity items
 
 List forbidden items, forbid all items on the map, or forbid unreachable items.
-Forbidding unreachable items checks if the tile the item is on is walkable,
-which covers most cases of unreachable items.
+Forbidding unreachable items checks any of your citizens can walk to the item.
 
 Usage
 -----
--- a/unforbid.lua
+++ b/unforbid.lua
@@ -2,25 +2,24 @@
 
 local argparse = require('argparse')
 
-local function unforbid_all(include_unreachable, include_underwater, quiet)
+local function unforbid_all(include_unreachable, quiet)
     if not quiet then print('Unforbidding all items...') end
 
+    local citizens = dfhack.units.getCitizens()
     local count = 0
-    for _, item in ipairs(df.global.world.items.all) do
+    for _, item in pairs(df.global.world.items.all) do
         if item.flags.forbid then
-            local block = dfhack.maps.getTileBlock(item.pos)
-            local tile = dfhack.maps.getTileFlags(item.pos)
+            if not include_unreachable then
+                local reachable = false
 
-            if block then
-                local walkable = block.walkable[item.pos.x % 16][item.pos.y % 16]
-
-                if not include_unreachable and walkable == 0 then
-                    if not quiet then print(('  unreachable: %s (skipping)'):format(item)) end
-                    goto skipitem
+                for _, unit in pairs(citizens) do
+                    if dfhack.maps.canWalkBetween(item.pos, unit.pos) then
+                        reachable = true
+                    end
                 end
 
-                if not include_underwater and (tile.liquid_type == false and tile.flow_size > 3) then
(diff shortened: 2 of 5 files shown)


myk002 (commented):
so tiles with water levels > 3 are walkable?

(btw, single quotes are preferred)

vallode:
> so tiles with water levels > 3 are walkable?
> 
> (btw, single quotes are preferred)

As far as I understand tiles up to 3 water levels are completely safe for Dwarves to traverse.

> Dwarves can safely walk through water up to a depth of 3 - at depth 4 or higher, they will cancel jobs due to "Dangerous terrain" and begin to gain swimming experience. At depth 7, any dwarf that does not have sufficient Swimming skill will drown.

-[https://dwarffortresswiki.org/index.php/Water](https://dwarffortresswiki.org/index.php/Water)

As to single quotes, gotcha.

myk002:
> As far as I understand tiles up to 3 water levels are completely safe for Dwarves to traverse.

I phrased my question poorly : p Why is the water depth check necessary? Is the walkability flag still true for tiles with deep water on it? i.e. are underwater items still getting unforbidden even with the current walkability check?

vallode:
> > As far as I understand tiles up to 3 water levels are completely safe for Dwarves to traverse.
> 
> I phrased my question poorly : p Why is the water depth check necessary? Is the walkability flag still true for tiles with deep water on it? i.e. are underwater items still getting unforbidden even with the current walkability check?

Ah yes, I should have thought that was what you meant. From my testing water level does not impact the "walkability" of a tile. I tested with entirely flooded pits (3 z-levels of water) and with puddles of water, too.

myk002:
maybe what we need here is a general reachability check, like what is used in `autochop`:
```cpp
static bool is_accessible_item(const df::coord &pos, const vector<df::unit *> &citizens) {
    for (auto &unit : citizens) {
        if (Maps::canWalkBetween(unit->pos, pos))
            return true;
    }
    return false;
}
```
This would cover the water case and also other hazards (like fire).

The annoying part here is to calculate the vector of citizens, which requires a loop through all units. Perhaps we could add some cpp library functions exposed to Lua for this. Then you could integrate the check into your existing flag and not need a special one for water. Would that work?

vallode:
A general check for reachability would be great, I can see that coming in useful in a few scripts. That being said are you saying there is a way to do this without a loop through all units or just having a vector of citizens available without having to do a lua-side loop of units?

myk002:
actually, after playing with the API a bit, there's not much benefit in passing the vector across the lua boundary since the vector would need to be copied anyway. I think all we really need is:

cpp API:
```cpp
bool Units::getCitizens(std::vector<df::unit *> &citizens);
```

lua API:
```lua
dfhack.units.getCitizens()
```

and we can do the loop in Lua or cpp for calling Maps::canWalkBetween

unforbid.lua
```diff
@@ -2,28 +2,34 @@
 
 local argparse = require('argparse')
 
-local function unforbid_all(include_unreachable, quiet)
+local function unforbid_all(include_unreachable, include_underwater, quiet)
     if not quiet then print('Unforbidding all items...') end
 
     local count = 0
```
myk002, line 8:
```suggestion
    local citizens = dfhack.units.getCitizens()
    local count = 0
```

unforbid.lua
```diff
@@ -2,28 +2,34 @@
...
     for _, item in ipairs(df.global.world.items.all) do
         if item.flags.forbid then
             local block = dfhack.maps.getTileBlock(item.pos)
+            local tile = dfhack.maps.getTileFlags(item.pos)
 
             if block then
-                local walkable = block.walkable[item.pos.x%16][item.pos.y%16]
+                local walkable = block.walkable[item.pos.x % 16][item.pos.y % 16]
```
myk002, line 15:
```suggestion
                local reachable = false
                for _,unit in ipairs(citizens) do
                    if dfhack.maps.canWalkBetween(item.pos, unit.pos) then
                        reachable = true
                    end
                end
```

myk002:
``dfhack.units.getCitizens()`` is merged for your convenience.

docs/forbid.rst
```diff
@@ -6,8 +6,7 @@ forbid
     :tags: fort productivity items
 
 List forbidden items, forbid all items on the map, or forbid unreachable items.
-Forbidding unreachable items checks if the tile the item is on is walkable,
-which covers most cases of unreachable items.
+Forbidding unreachable items checks any of your citizens can walk to the item.
```
myk002, line 9:
```suggestion
Forbidding unreachable items checks if any of your citizens can walk to the item.
```

myk002 (approved):

PR #581, opened 2023-01-30
modtools/spawn-liquid position argument fix

vallode:
`processArgsGetopt` only takes one long name. Should maybe take a table?


myk002:
it should take multiple long forms. let me check. I thought I implemented that.

umm apparently I didn't implement that.

myk002 (approved):

PR #583, opened 2023-02-01
add gui support to Autofish

AndrielChaoti:
This just adds basic gui support for autofish.


autofish.lua
```diff
@@ -129,6 +130,16 @@ local function print_status()
...
 end
 
+function setEnabled(val)
+    load_state()
+    enabled = val
+    persist_state()
+
+    if not enabled then return end
+    event_loop()
+end
```
myk002, line 140: this is technically already handled by the `enabled` API and shouldn't be necessary here.

gui/autofish.lua
```diff
@@ -0,0 +1,110 @@
...
+local REFRESH_MS = 10000
+
+---
+-- Autofish
+---
+Autofish = defclass(Autofish, widgets.Window)
+Autofish.ATTRS = {
+    frame_title = "Autofish",
+    frame = {w=45, h=12},
+    resizable = false
```
myk002, line 16: this should probably be true so players can squish your screen down to a minimum size so they can still see how much fish they have, but take up as little screen space as possible. Try setting `resize_min` to the minimum size that the screen is useful at.

gui/autofish.lua
```diff
@@ -0,0 +1,110 @@
+-- config ui for autofish
+
+local gui = require("gui")
+local widgets = require("gui.widgets")
+local script = reqscript("autofish")
```
myk002, line 5: I realize that you've named this according to the "plugin" name in `gui/autochop`, but this name conflicts with the `script` module. you're not loading it here, but it could cause confusion for a future reader. I suggest you use "autofish" instead of "script". That's the standard in other scripts that load scripts.

gui/autofish.lua
```diff
@@ -0,0 +1,110 @@
...
+        self:refresh_data()
+    end
+end
+
+
+---
+-- AutofishScreen
+---
+AutofishScreen = defclass(AutofishScreen, gui.ZScreen)
+AutofishScreen.ATTRS = {focus_path = "autofish"}
```
myk002, line 97:
```suggestion
AutofishScreen.ATTRS{focus_path = "autofish"}
```

gui/autofish.lua
```diff
@@ -0,0 +1,110 @@
+-- config ui for autofish
+
+local gui = require("gui")
+local widgets = require("gui.widgets")
+local script = reqscript("autofish")
+
+local REFRESH_MS = 10000
```
myk002, line 7: in order for autorefresh to be useful, you should display some live statistics on your window, such as how many raw and processed fish there currently are. Similarly, adding a minimal mode would allow the players to use your gui as a small status window that they can leave up on the screen.

myk002:
also, needs a line in changelog.txt under New Scripts

autofish.lua
AndrielChaoti: without this, toggling enable in the gui does not do anything. should I just have it call my script through the equivalent of exec(...)?

gui/autofish.lua
AndrielChaoti: The screen is already as small as it can be and still be useful, without a minimal mode.

autofish.lua
myk002: The appropriate interface would be `dfhack.run_command{'enable', 'autofish'}`, I think

gui/autofish.lua
AndrielChaoti: It does currently show whether or not autofish has disabled/enabled fishing, and this can also be useful if settings were updated outside of the GUI for some reason. I don't see a reason to *not* add a fish counter to it though, to make it more useful.

docs/gui/autofish.rst
```diff
@@ -0,0 +1,18 @@
+gui/autofish
+============
+
+.. dfhack-tool::
+    :summary: Auto-manage fishing labors to control your stock of fish.
+    :tags: fort auto labors
+
+This is the configuration interface for the `autofish` module. You can configure
+the maximum and minimum fish thresholds, and whether or not the script should
```
myk002, line 9:
```suggestion
the number of fish you'd like to keep in stock, and whether or not the script should
```

gui/autofish.lua
```diff
@@ -0,0 +1,141 @@
...
+        widgets.Label{
+            view_id="fish_count",
+            frame={t=5, l=0, h=1},
+            auto_height=false
+        },
+        widgets.Label{
+            view_id="current_mode",
+            frame={t=6, l=0, h=1},
+            auto_height=false,
+            --visible=false
```
myk002, line 73:
```suggestion
            visible=autofish.isEnabled
```

gui/autofish.lua
```diff
@@ -0,0 +1,141 @@
...
+        get_count_text(total, autofish.s_minFish, autofish.s_maxFish),
+        " (",
+        {text=tostring(prep), pen=COLOR_GREY}, " prepared, ",
+        {text=tostring(raw), pen=COLOR_GREY}, " raw)"
+    }
+    self.subviews.fish_count:setText(counts_text)
+
+    self.subviews.current_mode.visible=autofish.isEnabled()
+    if autofish.isFishing then
+        self.subviews.current_mode:setText(string.format("Autofish has %s fishing.", (autofish.isFishing and "enabled" or "disabled")))
```
myk002, line 111: The word "enabled" might be too overloaded here. how about "Autofish is currently [not] fishing"?

gui/autofish.lua
```diff
@@ -0,0 +1,141 @@
...
+    local counts_text = {
+        "Fish: ",
+        get_count_text(total, autofish.s_minFish, autofish.s_maxFish),
+        " (",
+        {text=tostring(prep), pen=COLOR_GREY}, " prepared, ",
+        {text=tostring(raw), pen=COLOR_GREY}, " raw)"
+    }
+    self.subviews.fish_count:setText(counts_text)
+
+    self.subviews.current_mode.visible=autofish.isEnabled()
```
myk002, line 109: the visible property can be a function that returns a bool, so you can just put this right in the widget instantiation

myk002 (commented):
could you clean out the commented code? otherwise, looks good!

myk002 (approved):

PR #587, opened 2023-02-01
add chitin (and other) support

ppaawwll:
Fixes DFHack/dfhack#2770

added proper functionality to spawning chitin, shell, ivory, and scale corpsepieces, instead of spawning as "chops". also made liquids spawn as GLOB items.


gui/create-item.lua
```diff
@@ -255,15 +256,16 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
...
  end
  local itemType = dfhack.items.findType(item_type..":NONE")
  local itemSubtype = dfhack.items.findSubtype(item_type..":NONE")
  local material = "CREATURE_MAT:"..raceName..":"..layerMat
  local materialInfo = dfhack.matinfo.find(material)
  local item_id = dfhack.items.createItem(itemType, itemSubtype, materialInfo['type'], materialInfo.index, creator)
  local item = df.item.find(item_id)
- item:setQuality(quality)
- local isCorpsePiece = false
+ -- item:setQuality(quality)
```
myk002, line 268: is this because corpse pieces don't have a quality? it would be nice if that prompt didn't come up when it wasn't relevant.

ppaawwll: nothing happens if you do set it (i just thought passing it through would be funny), but ye it would.

gui/create-item.lua
```diff
@@ -255,15 +256,16 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
   item_type = "GLOB"
  elseif CORPSE_PIECES[layerName] or HAIR_PIECES[layerName] then -- check if hair
   item_type = "CORPSEPIECE"
+ elseif LIQUID_PIECES[layerName] then -- check if hair
+  item_type = "GLOB" -- prolly need to do more work on this (it doesn't like spawning LIQUID items), but it works for now
```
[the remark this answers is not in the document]
ppaawwll: ooooohhhhhhhhhhh

gui/create-item.lua
```diff
@@ -255,15 +258,24 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
...
  local itemType = dfhack.items.findType(item_type..":NONE")
  local itemSubtype = dfhack.items.findSubtype(item_type..":NONE")
  local material = "CREATURE_MAT:"..raceName..":"..layerMat
  local materialInfo = dfhack.matinfo.find(material)
  local item_id = dfhack.items.createItem(itemType, itemSubtype, materialInfo['type'], materialInfo.index, creator)
  local item = df.item.find(item_id)
- item:setQuality(quality)
- local isCorpsePiece = false
+ if liquid then
+  local bucket = df.item.find(dfhack.items.createItem(18, -1, 22, 191, creator))
```
quietust, line 272: I'm not sure what you intended `22, 191` to mean, but in my test world it resolves to "sliver barb seed" which doesn't seem to make much sense as a bucket material.

ppaawwll: this is 100% intentional. 🐇

ppaawwll: i figured the material of the bucket didn't matter as it was not the item being specified by the user, so, 🐇 

also might help to differentiate those buckets from intentionally spawned buckets.

ppaawwll:
oh well :P

ppaawwll:
i guess the "(and other)" part still applies lol

myk002 (commented):
looks like some debug prints are still hanging around. could you remove those?

ppaawwll:
oop ye

ppaawwll:
dere we go

gui/create-item.lua
```diff
@@ -213,8 +214,9 @@ local function getCreatureRaceAndCaste(caste)
  return df.global.world.raws.creatures.list_creature[caste.index],df.global.world.raws.creatures.list_caste[caste.index]
 end
 
-local CORPSE_PIECES = utils.invert{'BONE', 'SKIN', 'CARTILAGE', 'TOOTH', 'NERVE', 'NAIL', 'HORN', 'HOOF' }
+local CORPSE_PIECES = utils.invert{'BONE', 'SKIN', 'CARTILAGE', 'TOOTH', 'NERVE', 'NAIL', 'HORN', 'HOOF', 'CHITIN', 'SHELL', 'IVORY', 'SCALE' }
 local HAIR_PIECES = utils.invert{'HAIR', 'EYEBROW', 'EYELASH', 'MOUSTACHE', 'CHIN_WHISKERS', 'SIDEBURNS' }
+local LIQUID_PIECES = utils.invert{'BLOOD', 'PUS', 'VENOM', 'SWEAT', 'TEARS', 'SPIT', 'MILK' }
```
myk002, line 219: is spawning a plain old bucket of water possible?

gui/create-item.lua
```diff
@@ -255,15 +259,38 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
   item_type = "GLOB"
  elseif CORPSE_PIECES[layerName] or HAIR_PIECES[layerName] then -- check if hair
   item_type = "CORPSEPIECE"
+ elseif LIQUID_PIECES[layerName] then -- check if hair
```
myk002, line 262:
```suggestion
 elseif LIQUID_PIECES[layerName] then
```

gui/create-item.lua
```diff
@@ -255,15 +259,38 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
...
+  item_type = "LIQUID_MISC" -- prolly need to do more work on this (it doesn't like spawning LIQUID items), but it works for now
```
myk002, line 263: what exactly doesn't work, and what do you think could be done to make it better? it might be worthwhile to create an issue on GitHub with any follow-up plans you have

gui/create-item.lua
```diff
@@ -255,15 +259,38 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
...
+     bucketMat = dfhack.matinfo.find("PLANT_MAT:"..n.id..":WOOD")
+    end
+    if bucketMat then break end
+   end
+  end
+  local bucketType = dfhack.items.findType("BUCKET:NONE")
+  local bucket = df.item.find(dfhack.items.createItem(bucketType, -1, bucketMat.type, bucketMat.index, creator))
+  dfhack.items.moveToContainer(item, bucket)
+  guidm.setCursorPos(creator.pos)
+  dfhack.run_command("spotclean")
```
myk002, line 290: can you restore the cursor to its former position after running `spotclean`?

alternately, if you have the time, you could add a `--cursor` option to `spotclean` like so many other tools have for specifying the position without a cursor

gui/create-item.lua
```diff
@@ -255,15 +259,38 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
...
+ end
+
+ -- item:setQuality(quality)
```
myk002, line 293:
```suggestion
```

gui/create-item.lua
```diff
@@ -290,10 +317,12 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
   elseif layerName == "HORN" or layerName == "HOOF" then -- check if nail
    item.corpse_flags.horn = true
    item.material_amount.Horn = 1
-   isCorpsePiece = true
+  elseif layerName == "SHELL" then -- check if nail
```
myk002, line 320:
```suggestion
  elseif layerName == "SHELL" then
```

gui/create-item.lua
```diff
@@ -230,7 +232,9 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
  local layerName
  local layerMat
  local tissueID
- if not generic then -- if we have a specified body part and layer, figure all the stuff out about that
+ local liquid = false
+ local isCorpse = bodypart < -2 -- in the hackWish function, the bodypart variable is initialized to -1, which isn't changed if the spawned item is a corpse, but it's passed to this function with two subtracted from it, making it -3 instead
```
myk002, line 236: what's happening with the `-1 - 2` logic? what is it actually trying to do with the minus 2? this feels like it's more hacky than we need it to be, but I don't know the context

ppaawwll: some lists in lua start from `0`, and some seem to start from `1`. i was having an issue where the corpsepiece actually spawned was one entry further than i selected (if i pick `right upper leg`, i get `left upper leg` instead), meaning i had to adjust for that offset.

then, adding the `generic` option to the body part list bumped every entry's index forward by one, meaning i had to adjust for an offset of _two_ instead.

ppaawwll: i should probably put a comment of this where the offset stuff actually happens.

gui/create-item.lua
ppaawwll: if there was a creature that was made out of water, yes. (assuming i add that material to this list)

ppaawwll:
that good?

gui/create-item.lua
ppaawwll: could probably add the bucket thing to general liquids as well tbh. (also water isn't even in the list of available materials for the LIQUID item, but that might be able to be changed? probably a task for another pull request)

gui/create-item.lua
```diff
@@ -448,12 +450,12 @@ function hackWish(unit)
     if df.item_type.attrs[itemtype].is_stackable then
      createItem({mattype,matindex},{itemtype,itemsubtype},quality,unit,description,amount)
     else
-     local isCorpsePiece = itemtype == df.item_type.CORPSEPIECE
+     local isCorpsePiece = itemtype ~= df.item_type.CORPSEPIECE or itemtype ~= df.item_type.CORPSE
```
myk002, line 452: Isn't this always true? A neq B or A neq C is always true unless A is B is C

gui/create-item.lua
```diff
@@ -448,12 +450,12 @@ function hackWish(unit)
...
      for i=1,amount do
       if not isCorpsePiece then
        createItem({mattype,matindex},{itemtype,itemsubtype},quality,unit,description,1)
       else
-       createCorpsePiece(unit,bodypart-2,partlayerID-1,mattype,matindex,corpsepieceGeneric,quality)
+       createCorpsePiece(unit,bodypart-2,partlayerID-1,mattype,matindex,corpsepieceGeneric,quality) --the offsets here are cause indexes in lua are fucky (some start at 0, some start at 1), so we adjust for that, as well as the index offset created by inserting the "generic" option at the start of the body part selection prompt
```
myk002, line 457: No cursing in code, please.

Also, it sounds like the subtraction is a mistake if the index is invalid (i.e. equal to -1)

gui/create-item.lua
ppaawwll: ah fuck wrong comparison type

gui/create-item.lua
ppaawwll: whoops.

also -1 is what i use to let the function know that it's a whole corpse (i.e. there's no specific material/tissue to spawn so just spawn it all)

ppaawwll: does sound a bit hacky tho, but i'm trying to avoid adding too many variables that only get used in specific cases.

ppaawwll: just got what you meant, whoops.

ppaawwll:
that should do it.

ppaawwll:
that should do it.

gui/create-item.lua
```diff
@@ -255,15 +259,39 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
   item_type = "GLOB"
  elseif CORPSE_PIECES[layerName] or HAIR_PIECES[layerName] then -- check if hair
   item_type = "CORPSEPIECE"
+ elseif LIQUID_PIECES[layerName] then
+  item_type = "LIQUID_MISC" -- prolly need to do more work on this (it doesn't like spawning LIQUID items), but it works for now
```
myk002, line 263:
```suggestion
  item_type = "LIQUID_MISC"
```

gui/create-item.lua
myk002: I can get it

gui/create-item.lua
```diff
@@ -345,12 +378,17 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
...
      for n in pairs(creatorBody.body_parts[i].layers) do
-      -- search through the target creature's body parts and bring back every one which has the desired material
-      if creatorRaceRaw.tissue[creatorBody.body_parts[i].layers[n].tissue_id].tissue_material_str[1] == layerMat and creatorBody.body_parts[i].token ~= "SKULL" and not creatorBody.body_parts[i].flags.SMALL then
+      if item_type == "CORPSE" then
        item.body.components.body_part_status[i].missing = false
-       item.body.components.body_part_status[i].grime = 2
        item.body.components.layer_status[creatorBody.body_parts[i].layers[n].layer_id].gone = false
-       -- save the index of the bone layer to a variable
+      else
+       -- -- search through the target creature's body parts and bring back every one which has the desired material
```
myk002, line 385:
```suggestion
       -- search through the target creature's body parts and bring back every one which has the desired material
```

gui/create-item.lua
```diff
@@ -345,12 +378,17 @@ function createCorpsePiece(creator, bodypart, partlayer, creatureID, casteID, ge
...
+       if creatorRaceRaw.tissue[creatorBody.body_parts[i].layers[n].tissue_id].tissue_material_str[1] == layerMat and creatorBody.body_parts[i].token ~= "SKULL" and not creatorBody.body_parts[i].flags.SMALL then
+        item.body.components.body_part_status[i].missing = false
+        item.body.components.body_part_status[i].grime = 2
+        item.body.components.layer_status[creatorBody.body_parts[i].layers[n].layer_id].gone = false
+        -- -- save the index of the bone layer to a variable
```
myk002, line 390:
```suggestion
        -- save the index of the bone layer to a variable
```

myk002 (approved):

ppaawwll:
tysm!!!!!!!!!!!!!!!!!!!!!!!!!!!!

PR #593, opened 2023-02-03
New Dig GUI Script

johncosker:
### gui/dig

This is a script that creates a GUI for helping designate digs, channels, ramps, etc... in fun shapes and patters.

Note: This GIF is slightly out-of-date on the specifics of the UI
dig

I initially planned on creating a GUI for the existing `dig` C++ plugin but I ended up realizing I wanted to make it do a lot more than that plugin currently did, and decided it was easier to just keep everything in Lua. This is my first time working with Lua so please forgive any newb stuff.

### Features

**Multiple designation modes**

Dig, Channel, etc... all work (except removing stairs... for some reason see TODO section)

**Comes with a handful of shapes: `Rectangle, Ellipse, Rows, Diagonal.`**

The shapes are all derived from a base `Shape` class and are in their own module `shapes.lua`. I tried to make it easy to go and add more shapes, all you need is a name and a function that determines if a point is inside the shape. There are also options that will dynamically populate the GUI panel.

**Draggable Corner Anchors**

Without Auto-Commit enabled the shape will remain un-designated to allow resizing via dragging the corner spaces and also and changes to the options in the GUI panel will apply.

**Invert shape**

Might come in handy for making cool designs.

### Notes

Added a bunch of TODO's to the top of the main script file. Still think there's more to do to make it even better but I had to stop somewhere for now.

My biggest hope is there's a way to integrate this with the vanilla GUI so you can select mode/priority/etc... in the main UI and have this still work with it.

I've tested pretty thoroughly but I wouldn't be surprised if there are some bugs or edge cases I missed. But I think it's working pretty good at the moment.


gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+-- Will need to update as needed to add more option types
+function Dig:add_shape_options()
+    local prefix = "shape_option_"
+    if self.subviews ~= nil then
+        for i, view in ipairs(self.subviews) do
+            if string.sub(view.view_id, 1, string.len(prefix)) == prefix then
+                self.subviews[i] = nil
+            end
+        end
+    end
```
myk002, line 327:
```suggestion
    for i, view in ipairs(self.subviews or {}) do
        if view.view_id:sub(1, #prefix) == prefix then
            self.subviews[i] = nil
        end
    end
```

myk002: not required, but is more concise

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+    if self.shape then
+        if self.shape.options ~= nil then
```
myk002, line 329: instead of successive nested `if`s, consider inverting the condition and returning from the function if true. e.g.
```lua
if not self.shape or not self.shape.options then return end
```

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+            guidm.setCursorPos(pos)
+        end
+    elseif keys.SELECT then
+        pos = guidm.getCursorPos()
+    end
+
+    if keys._MOUSE_L_DOWN and pos then
+        if self.mark and not self.saved_cursor then
+            self.saved_cursor = pos
+            -- The statement after the or is to allow the 1x1 special case for easy doors
```
myk002, line 504:
```suggestion
            -- The statement after the or is to allow the 1x1 special case for easy doorways
```

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+                    pos.x - bounds.x1,
+                    pos.y - bounds.y1,
+                    mouse_pos
+                )
+            else
+                return nil
+            end
+        end
+
+        guidm.renderMapOverlay(get_overlay_pen, bounds)
```
myk002, line 472: another option is to factor the preview display logic out from gui/quickfort and use that, but this isn't complicated enough yet to go that route, I think. Maybe once you implement 3d shapes then there will be a better reason to share logic.

internal/dig/shapes.lua
```diff
@@ -0,0 +1,289 @@
...
+    VERT_EW = { 4, 1 },
+    POINT = { 4, 3 },
+}
+
+-- return the pen, alter based on if we want to display a corner and a mouse over corner
+function make_pen(direction, is_corner, is_mouse_over)
+    return to_pen {
+        ch = "X",
+        fg = COLOR_GREEN,
+        tile = dfhack.screen.findGraphicsTile(
```
myk002, line 33: note that findGraphicsTile can be slow since it's a linear search across all tile pages. You might get better mileage out of pre-calculating all the texpos values and keeping them in a top-scope table. Note that this table should *not* be persisted across script invocations, though, since the texpos values can change if a world is unloaded and  a different world is loaded.

internal/dig/shapes.lua
```diff
@@ -0,0 +1,289 @@
...
+                self.arr[x][y] = not value
+            end
+        end
+    end
+end
+
+-- Return a pen object based on the position of the x,y point within the shape
+-- Contains logic to determine which corner/side/etc... icon to draw to create an
+-- aesthetically pleasing border for the shape. Unfortunately the code itself
+-- very much not aesthetically pleasing but I'm not sure if that can be helped
```
myk002, line 70: check out the Scrollbar tile calculations in library/lua/gui/widgets.lua sometime...

myk002 (commented):
this is absolutely fantastic work! Great job!

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+
+-- TODOS ====================
+
+-- Must Haves
+-----------------------------
+-- Figure out why pause/unpause doesn't work
+-- Line 'shape' between two points with thickness (currently tricky with how shapes are defined as within bounds rectangle)
+--   A totally vertical line with thickness of e.g. 3 would end up with tiles outside the bounds/view rectangle
+--   Could also allow 'multi-point' line that would produce curves
+-- Need to support classic mode? Might just be double checking non-graphics 'pen' settings
```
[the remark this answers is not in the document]
myk002: looking through the code, it looks like you largely already handled this with the pens you've created

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+        math.abs(pos1.z - pos2.z) + 1
+    return width, height, depth
+end
+
+-- Panel to show the Mouse position/dimensions/etc
+-- Stolen from blueprint or quickfort I forget
+ActionPanel = defclass(ActionPanel, widgets.ResizingPanel)
+ActionPanel.ATTRS {
+    get_mark_fn = DEFAULT_NIL,
+    is_setting_start_pos_fn = DEFAULT_NIL,
```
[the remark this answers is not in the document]
johncosker: Fixed

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+    autoarrange_subviews = true,
+}
+
+function ActionPanel:init()
+    self:addviews { widgets.WrappedLabel {
+        view_id = "action_label",
+        text_to_wrap = self:callback("get_action_text"),
+    },
```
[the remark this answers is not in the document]
johncosker: Fixed

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+        }
+    end
+
+    self:addviews { widgets.WrappedLabel {
+        view_id = "settings_label",
+        text_to_wrap = "General Settings:\n",
+    },
+        widgets.CycleHotkeyLabel {
+            view_id = "shape_name",
+            key = "CUSTOM_E",
```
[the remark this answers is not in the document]
johncosker: Oops., changed it to `z`. I've mostly stuck with the new default hotkeys to try and get used to them except I remapped up/down to <,> because I just couldn't adjust

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+Dig.ATTRS {
+    frame_title = "Dig",
+    frame = {
+        w = 47,
+        h = 40,
+        r = 2,
+        t = 18,
+    },
+    resizable = true,
+    resize_min = { h = 10 },
```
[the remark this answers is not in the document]
johncosker: I didn't really test the window re-size much, should I just increase the minimum height?

gui/dig.lua
johncosker: Fixed

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+            end
+
+            self.saved_cursor = nil
+        end
+
+        return true
+    end
+
+    -- send movement keys through, but otherwise we're a modal dialog
+    return not guidm.getMapKey(keys)
```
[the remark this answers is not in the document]
johncosker: Fixed

internal/dig/shapes.lua
johncosker: Maybe that's why it was slow when I tried to change the render code to display the grid tiles on all non-shape tiles in the main script... if it's doing a linear search for every tile on the screen then that would make sense, I hadn't thought of caching the icons, is this also doing disk IO every time to load the icon from the image file too? Or are all the graphics loaded into memory on startup?

johncosker (commented):
Thanks for the feedback! I will look at the tile pre-caching and make sure classic mode works okay.

johncosker:
Hmmmmm, I Just realized I forgot to test the pausing fix and it's still not pausing, I'll try to fix it in the next commit

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+-----------------------------
+-- Figure out why pause/unpause doesn't work
+-- Line 'shape' between two points with thickness (currently tricky with how shapes are defined as within bounds rectangle)
+--   A totally vertical line with thickness of e.g. 3 would end up with tiles outside the bounds/view rectangle
+--   Could also allow 'multi-point' line that would produce curves
+-- Need to support classic mode? Might just be double checking non-graphics 'pen' settings
+
+-- Should Haves
+-----------------------------
+-- As the number of shapes and designations grow it might be better to have list menus for them instead of cycle
```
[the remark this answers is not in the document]
myk002: that sounds good. thanks!

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+-- 3D shapes, would allow stuff like spiral staircases/minecart tracks and other neat stuff, probably not too hard
```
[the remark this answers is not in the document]
myk002: this might be more in the realm of quickfort and blueprints. For example, `quickfort run library/dreamfort.csv -n /central_stairs --repeat down,10` will give you 20 levels of spiral staircase. If there is a way to combine the functionality of `gui/dig`, `gui/blueprint`, and `gui/quickfort`, then I'm all ears. I recorded a few notes on the subject here: https://github.com/DFHack/dfhack/issues/2446

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+-- Grid view without slowness (can ignore if next TODO is done, since nrmal mining mode has grid view)
+--   Lags when drawing the full screen grid on each frame render
+-- Integrate with default mining mode for designation type, priority, etc... (possible?)
```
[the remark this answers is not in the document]
myk002: that's essentially an overlay. for example, see the widget that comes up whenever the manager orders page is opened (at the bottom): 
image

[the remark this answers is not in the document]
myk002: see: https://docs.dfhack.org/en/latest/docs/dev/overlay-dev-guide.html for docs and https://github.com/DFHack/dfhack/blob/develop/plugins/lua/orders.lua for a code example

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+-- Add warning to stairs if not spanning z levels like vanilla does
+-- Figure out how to remove dug stairs with mode (nothing seems to work, include 'dig ramp')
```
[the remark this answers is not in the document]
myk002: that might be a change in behavior that `quickfort` needs to adapt to. it has a database of rules for when designations can be applied that might need updating.

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+-- Triangle shapes, more shapes in general, could even do dynamic creation of residential layouts as long as they can be mathematically represented
+-- Exploration pattern ladder https://dwarffortresswiki.org/index.php/DF2014:Exploratory_mining#Ladder_Rows
+-- Add logic to set thickness.max for applicable shapes based on dimensions
+-- Investigate using single quickfort API call when commiting  https://github.com/joelpt/quickfort/blob/master/README.md#area-expansion-syntax
+
+-- Stretch Goals
+-----------------------------
+-- Shape preview in panel
+-- Shape designer in preview panel to draw repeatable shapes i'e' 2x3 room with door
+
```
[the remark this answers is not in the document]
myk002: I wrote this when I thought you were using `gui/quickfort` to draw the preview. I meant that something I have in mind for `gui/quickfort` (which could benefit `gui/dig` if we can share logic between the two scripts) is to use the actual tile sprites that represent the intended result of the blueprint. For example, if you are applying a channel pattern, the channel tile icon will appear for the preview. representing buildings will be much harder, but same idea.

gui/dig.lua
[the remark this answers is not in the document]
myk002: ack. I did that as well, actually.

gui/dig.lua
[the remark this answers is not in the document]
myk002: yeah, just increasing the minimum is fine. people will complain if they need more.

internal/dig/shapes.lua
myk002: no, it's all in memory. the linear search is for the "CURSORS" page, not the map tile (thankfully), but it's still linear.

In this case, I think the lag that you were seeing was due to the loop over all visible tiles. Lua just isn't as fast as C++ when doing that. I saw something similar in `gui/quickfort` when the blueprint preview covers a large area, and that doesn't do any `findGraphicsTile` lookups at all.

myk002:
> Hmmmmm, I Just realized I forgot to test the pausing fix and it's still not un-pausing, I'll try to fix it in the next commit

Is ```lua
    if Dig.super.onInput(self, keys) then
        return true
    end
```
returning? there might be a sub-widget that is eating the space

**Comment by johncosker** [comment]
After trying it again today it actually does work, maybe I forgot to reload the plugin after I made the change or something.

File: internal/dig/shapes.lua (hunk 5 above)
Comment (johncosker) [reply internal/dig/shapes.lua:33]: That makes sense. Well, maybe if it's eventually possible to integrate with the vanilla dig layout then we get the grid overlay for free.

File: gui/dig.lua [hunk 18]
```diff
@@ -0,0 +1,658 @@
...
+        w = 47,
+        h = 40,
+        r = 2,
+        t = 18,
+    },
+    resizable = true,
+    resize_min = { h = 10 },
+    autoarrange_subviews = true,
+    autoarrange_gap = 1,
+    presets = DEFAULT_NIL,
```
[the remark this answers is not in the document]
johncosker: FIxed

gui/dig.lua
johncosker: Fixed

gui/dig.lua
johncosker: Fixed

myk002:
that..does not compute, so to speak. whereas lua code attached to plugins would need to be reloaded, this is a pure script, and scripts don't need to be reloaded. They are checked for changes on every execution, so as long as you've saved the file, the new script should be what is running.

The only confusing part about this is that if you *remove* a global (e.g. delete it entirely) it will retain its old value. This is particularly painful when removing methods you don't need from a class. They'll stick around until you explicitly clear the script environment by restarting DF or running `devel/clear-script-env`.

johncosker:
Huh, weird, I don't know then, but regardless it seems to work now so that's good. That's good to know about the scripts though. I guess I've been running `reload gui/dig` for nothing this whole time lol

gui/dig.lua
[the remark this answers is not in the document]
johncosker: Addressed: https://github.com/DFHack/dfhack/pull/2832

gui/dig.lua
```diff
@@ -0,0 +1,658 @@
...
+-- Need to support classic mode? Might just be double checking non-graphics 'pen' settings
+
+-- Should Haves
+-----------------------------
+-- As the number of shapes and designations grow it might be better to have list menus for them instead of cycle
+-- 3D shapes, would allow stuff like spiral staircases/minecart tracks and other neat stuff, probably not too hard
+-- Grid view without slowness (can ignore if next TODO is done, since nrmal mining mode has grid view)
+--   Lags when drawing the full screen grid on each frame render
+-- Integrate with default mining mode for designation type, priority, etc... (possible?)
+-- Add warning to stairs if not spanning z levels like vanilla does
```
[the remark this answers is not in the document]
johncosker: Fix for this coming in next commit

gui/dig.lua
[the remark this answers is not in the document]
johncosker: Fixed

internal/dig/shapes.lua
johncosker: > note that findGraphicsTile can be slow since it's a linear search across all tile pages. You might get better mileage out of pre-calculating all the texpos values and keeping them in a top-scope table. Note that this table should _not_ be persisted across script invocations, though, since the texpos values can change if a world is unloaded and a different world is loaded.

So if I'm understanding you right, I should do something like change the `CURSORS` table I'm creating in `shapes.lua` to contain the results of the calls to all the `findGraphicsTile` I expect to use. The second part I'm less sure about, would doing what I outlined avoid this issue? Or do I need to take extra steps to make sure it's not persisted. I don't know enough about the lifetime of the scripts between different states of the game.

gui/dig.lua
```diff
@@ -43,6 +46,8 @@ local widgets = require("gui.widgets")
 local quickfort = reqscript("quickfort")
 local shapes = reqscript("internal/dig/shapes")
 
+reload('gui.widgets')
```
myk002, line 49: was this a line for debugging?

gui/dig.lua
```diff
@@ -190,9 +198,39 @@ function GenericOptionsPanel:init()
...
+            },
             disabled = false,
             show_tooltip = true,
+            on_change = function(new, old) self.dig_panel:updateLayout() end,
+        },
+        widgets.CycleHotkeyLabel {
+            view_id = "stairs_subtype",
+            key = "CUSTOM_G",
+            key_back = "CUSTOM_SHIFT_G",
+            label = "Stair Type: ",
```
myk002, line 211: I think you might need two of these (both defaulting to Auto) so the player can define what they want both at the top and the bottom

myk002: or maybe a single one with the combinations that make sense? Auto/Auto, Down/Up, UpDown/Up, Down/UpDown, UpDown/UpDown

gui/dig.lua
```diff
@@ -553,20 +606,26 @@ function Dig:commit()
...
-            end
-            if z == 0 then
-                return "u" -- up stair
-            elseif z == math.abs(
-                self:get_bounds().z1 - self:get_bounds().z2
-            ) then
-                return "j" -- down stair
+        if mode == "i" then
+            if stairs_type == "auto" then
+                if math.abs(self:get_bounds().z1 - self:get_bounds().z2) == 0 then
```
myk002, line 615: if you're testing against zero, math.abs is a noop

gui/dig.lua
```diff
@@ -553,20 +606,26 @@ function Dig:commit()
...
-                return "`" -- return nothing, they need to specify more than one z-level
-            end
-            if z == 0 then
-                return "u" -- up stair
-            elseif z == math.abs(
-                self:get_bounds().z1 - self:get_bounds().z2
-            ) then
-                return "j" -- down stair
+        if mode == "i" then
+            if stairs_type == "auto" then
```
myk002, line 614: if the player is extending down from an existing staircase, and the existing bottom layer is an up stair, then the top layer should become an UpDown instead of a Down. Similar for if they're extending an existing staircase upward (though I don't think you can convert a down stair into an up/down stair by digging..)

gui/dig.lua
```diff
@@ -632,6 +691,7 @@ DigScreen = defclass(DigScreen, gui.ZScreen)
 DigScreen.ATTRS {
     focus_path = "dig",
     pass_pause = true,
+    initial_pause = false,
```
myk002, line 694: you should probably leave this unspecified so the player can configure their preference via gui/control-panel

gui/dig.lua
```diff
@@ -632,6 +691,7 @@ DigScreen = defclass(DigScreen, gui.ZScreen)
...
     pass_movement_keys = true,
     presets = DEFAULT_NIL,
```
myk002, line 696: this isn't used, right?

internal/dig/shapes.lua
myk002: as a code example:
```
local function getPens()
    local pens = {}
    for _,n in ipairs{false, true} do for _,s in ipairs{false, true} do for _,e in ipairs{false, true} do for _,w in ipairs{false, true} do
        -- ...
    end end end end
    return pens
end
local PENS = getPens()
```
That will do the lookup once when the script runs when in a top-level script. It will *not* be re-run if in an internal module, so you'll have to calculate this in the top-level script and pass it to the functions called in this internal script

gui/dig.lua
johncosker: This is a really good point. Vanilla does this with stairs so 'Auto' should match that

gui/dig.lua
johncosker: I think I'll just go with two single-direction cycle labels.

internal/dig/shapes.lua
johncosker: So basically move all the pen/tile code to dig.lua instead of shapes.lua?

johncosker:
Updated the graphic tile logic so it will now build a map of tile dynamically as they're drawn. After the initial calls the graphic tiles will be stored in `PENS` and re-used for the life of the main script.

Moved the logic to `dig.lua` probably better there anyway and keep `shapes.lua` to just contain stuff specific to... well, shapes.

internal/dig/shapes.lua
```diff
@@ -0,0 +1,212 @@
...
+
+-- Shape definitions
+-- All should have a string name, and a function has_point_fn(self, x, y) which returns true or false based
+-- on if the x,y point is within the shape
+-- Also options can be defined in a table, see existing shapes for example
+
+Ellipse = defclass(Ellipse, Shape)
+Ellipse.ATTRS = {
+    name = "Ellipse",
+    has_point_fn = function(self, x, y)
```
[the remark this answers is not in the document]
johncosker: Would it make sense to define the function in `init()` as I'm going to be also moving the `options` table there? Or at least make it a member function of the class instead of making it a `local function` outside of the class?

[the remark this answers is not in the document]
myk002: Ah, yes, good idea. maybe not using an attribute at all and instead declaring it a member function and overriding it OOP-style is the best choice here.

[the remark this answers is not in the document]
johncosker: As long as that works then I'll do that, I'm not familiar with lua so I wasn't sure exactly what the best approach would be in terms of inheritance so that's probably why I initially put everything in the ATTRS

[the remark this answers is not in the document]
myk002: It's not so much a Lua thing as a DFHack widget thing. ATTRS are good for fields that should be set by the parent widget, like the positioning frame and callbacks that link widgets together. Functionally that is not owned by the parent widget might as well just be a regular member function.

myk002 (approved):
This looks great! thank you so much for working with me throughout all these changes. I think your code has been fantastic from the start, and I think this script will be a foundation for a new, higher, class of DFHack visual design tools.

johncosker:
@myk002 Thanks for the kind words, I'm really excited to be able to contribute, I've used dfhack for probably a decade now and it feels good to be able to 'give back' a little bit. I'm looking forward to seeing what else I can help with.

It really is a great tool, both for the end user and, I've found, for developers. The amount of effort put in to the documentation and ease of use of the infrastructure and libraries is amazing.

myk002:
Are you on the DFHack Discord? https://dfhack.org/discord

I was wondering if you could put together a demo video that I could feature on the release announcement (later today).
File size must be < 8M

If you can't get to this right away, I can put something together, but your demo for this PR was very impressive.

myk002:
I like your original gif so much, I'll just use that : P I can update the gif later if you'd like to make a new one, though

johncosker:
I'll try make a new one, might be good to add some text describing it i.e. 'drag corners to re-size' 

If I get to it I'll reply with it. Also I'll check out the discord when it get the chance!

myk002:
The release is ready to go out imminently. If it's not something you're working on right now that we should hold for, I'd say don't spend too much time on it. The demo gif in the description is already very impressive. It got a lot of "wow"s on discord

johncosker:
Okay yeah no worries then I probably won't be able to work on it until I finish up my work day, so I'd say re-use it. It is slightly out-of-date in terms of the panel layout but no one's really gonna notice or care I'd imagine.

myk002:
yeah, the eyes will be on the dig patterns being drawn. thank you again for this!

PR #594, opened 2023-02-04
WIP: autojustice

Igualop:
### autojustice

This is a script that automate most of the tedious parts of the justice system.

### Features
- Auto interview visitors.
- Auto interview accused units.
- Auto convict non-citizens confessed crimes.

### Notes
- There are a couple of things listed as TODO, but the core functions seems to be working fine.

- This script was (afk) tested on a running fort for a dozen hours without issues, but since I'm totally new with df/dfhack/lua I can't guarantee that all the modifications made to the crime data are 100% correct.

- "justicetools.lua" file contains helper methods used by the "autojustice.lua". Not sure if I should merge those two scripts or where to even put the "justicetools.lua".

- I'm not sure I've correctly followed all naming conventions, feel free to comment on that.

- There are a couple of options that can be added in case everything is working fine (i.e: auto convict citizens when the punishment is just jail time, add notifications to convictions and confessions).

- Might need to test on worlds with multiple forts as crime reports seems to have a siteId field mapped on them.

### Known Issues
- I had a goblin poet getting convicted even though he did join my fortress. I'm not sure why he is not being considered as a citizen.

### DIFF
--- a/autojustice.lua
+++ b/autojustice.lua
@@ -1,0 +1,297 @@
+-- Auto Justice
+--@enable = true
+
+local usage = [[
+Usage
+-----
+enable autojustice
+disable autojustice
+]]
+
+local GLOBAL_KEY = 'autojustice'
+local UNIT_NEW_ACTIVE_DELAY = 100
+local CRIME_REFRESH_DELAY = 120
+
+local json = require('json')
+local persist = require('persist-table')
+local repeatUtil = require('repeat-util')
+local eventful = require('plugins.eventful')
+local justice = reqscript('autojustice/justicetools')
+
+enabled = enabled or false
+
+local lastCrimeCount = 0
+local openCrimes = nil
+local openCrimesWitnessCount = nil
+local undiscoveredCrimes = nil
+local confessedCrimes = nil
+
+function isEnabled()
+    return enabled
+end
+
+function saveState()
+    persist.GlobalTable[GLOBAL_KEY] = json.encode({
+        enabled = enabled or false
+    })
+end
+
+function loadState ()
+    local data = json.decode(persist.GlobalTable[GLOBAL_KEY] or '{}') 
+    enabled = data.enabled or false
+end
+
+function onStateChange (sc)
+    if sc == SC_MAP_UNLOADED then
+        enabled = false
+        return
+    end
+  
+    if sc ~= SC_MAP_LOADED or df.global.gamemode ~= df.game_mode.DWARF then
+        return
+    end
+  
+    loadState()
+    --TODO: make it run here
+end
+
+function serviceToggle ()
+    if dfhack_flags.enable_state then
+        serviceEnable()
+    else
+        serviceDisable()
+    end
+end
+
+function serviceEnable ()
+    saveState()
+    initValues()
+    registerEvents()
+
+    print('autojustice is running')
+end
+
+function serviceDisable ()
+    clearEvents()
+    clearValues()
+
+    print('autojustice has stopped')
+end
+
+function runScript ()
+    if not dfhack.isMapLoaded() then
+        qerror('This script requires a fortress map to be loaded')
+        return
+    end
+
+    -- TODO: Set possible variables here
+end
+
+function registerEvents ()
+    dfhack.onStateChange[GLOBAL_KEY] = onStateChange
+
+    eventful.enableEvent(eventful.eventType.UNIT_NEW_ACTIVE, UNIT_NEW_ACTIVE_DELAY)
+    eventful.onUnitNewActive[GLOBAL_KEY] = onUnitNewActive
+
+    repeatUtil.scheduleEvery(GLOBAL_KEY, CRIME_REFRESH_DELAY, 'ticks', onRefresh)
+end
+
+function clearEvents ()
+    dfhack.onStateChange[GLOBAL_KEY] = nil
+    eventful.onUnitNewActive[GLOBAL_KEY] = nil
+    repeatUtil.cancel(GLOBAL_KEY)
+end
+
+function initValues ()
+    lastCrimeCount = #df.global.world.crimes.all
+
+    openCrimes = {}
+    openCrimesWitnessCount = {}
+    undiscoveredCrimes = {}
+    confessedCrimes = {}
+
+    for i, crime in ipairs (df.global.world.crimes.all) do
+        addNewCrime(crime)
+    end
+end
+
+function clearValues () 
+    lastCrimeCount = 0 
+    
+    openCrimes = nil
+    openCrimesWitnessCount = nil
+    undiscoveredCrimes = nil
+    confessedCrimes = nil
+end
+
+function getSiteId ()
+    return df.global.world.world_data.active_site[0].id
+end
+
+function getInterviewCrime ()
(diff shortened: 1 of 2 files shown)


autojustice.lua
```diff
@@ -0,0 +1,297 @@
+-- Auto Justice
+--@enable = true
+
+local usage = [[
+Usage
+-----
+enable autojustice
+disable autojustice
+]]
```
myk002, line 9: docs should go in `docs/autojustice.rst`. Copy one of the other files in that directory, like `docs/autofish.rst` to get the structure.

autojustice.lua
```diff
@@ -0,0 +1,297 @@
+-- Auto Justice
+--@enable = true
```
myk002, line 2: you'll also need `@module = true`, otherwise the `isEnabled()` function can't be called and this script won't show up as enableable.

autojustice/justicetools.lua
```diff
@@ -0,0 +1,176 @@
+--@ module = true
+
```
myk002, line 2: if you want to keep this separate, this file should be moved to `internal/autojustice/justicetools.lua` so it doesn't show up as a runnable script. However, together they are about 500 lines, which is about average for a script like this. It might be simpler if you just combine the two.

autojustice.lua
```diff
@@ -0,0 +1,297 @@
...
+        enabled = enabled or false
+    })
+end
+
+function loadState ()
+    local data = json.decode(persist.GlobalTable[GLOBAL_KEY] or '{}') 
+    enabled = data.enabled or false
+end
+
+function onStateChange (sc)
```
myk002, line 44: this should be hooked up to dfhack directly -- `dfhack.onStateChange[GLOBAL_KEY] = function(sc)`
otherwise your reload code won't actually restore the enabled state for loaded forts automatically. You can copy the structure from `emigration.lua` for this.

myk002: note that the eventful config should still be where it is in registerEvents() since that should only be run when autojustice is enabled

autojustice.lua
```diff
@@ -0,0 +1,297 @@
...
+    dfhack.onStateChange[GLOBAL_KEY] = onStateChange
+
+    eventful.enableEvent(eventful.eventType.UNIT_NEW_ACTIVE, UNIT_NEW_ACTIVE_DELAY)
+    eventful.onUnitNewActive[GLOBAL_KEY] = onUnitNewActive
+
+    repeatUtil.scheduleEvery(GLOBAL_KEY, CRIME_REFRESH_DELAY, 'ticks', onRefresh)
+end
+
+function clearEvents ()
+    dfhack.onStateChange[GLOBAL_KEY] = nil
```
myk002, line 100: this should not be removed. You'll need it for the next loaded fort

myk002 (commented):
I'll have to find someone more knowledgeable about the justice system to review the business logic, but overall it looks pretty good!

autojustice/justicetools.lua
Igualop: I also think that it's simpler to combine the two. My only reasoning to keep these files separate is just to facilitate in case others might want get/change some info from the justice system.

autojustice.lua
```diff
@@ -0,0 +1,291 @@
...
+
+    if sc ~= SC_MAP_LOADED or df.global.gamemode ~= df.game_mode.DWARF then
+        return
+    end
+
+    loadState()
+    --TODO: make it run here
+end
+
+main()
```
myk002, line 291: now that you are a module, you need a guard above this line. e.g.:
```
if dfhack_flags.module then
    return
end
```

autojustice.lua
```diff
@@ -66,7 +98,21 @@ function runScript ()
         return
     end
 
-    -- TODO: Set possible variables here
+    loadState()
+
+    argparse.processArgsGetopt(args, {
+        {'j', 'jailcitizen', hasArg=true, handler=function(arg) convictMode.citizen.jail = arg end},
```
myk002, line 104: the `arg` passed to this function is a string that you'll have to convert to a boolean. the string "false" is truthy in Lua

docs/autojustice.rst
```diff
@@ -1,22 +1,50 @@
...
     Enable the script
 ``disable autojustice``
     Disable the script
+``autojustice [<options>]``
+    Set conviction options
+
+Options
+-------
+
+``-j``, ``--jailcitizen``
```
myk002, line 22: in the code, you have these options taking an argument, but I don't see it documented here. This should be:
```
``-j``, ``--jailcitizen true|false``
```
now that you have configuration options, you should also write a simple GUI to configure them, which should also show statistics and other useful information. See `gui/autofish` for an example. That can come in a different PR, though, if you like

myk002 (commented):
note that `gui/control-panel` was just added. Until I can get a proper query interface that allows a tool to report *when* it can be enabled, could you add "autojustice" to the `FORT_SERVICES` table in `gui/control-panel`?

docs/autojustice.rst
```diff
@@ -19,22 +19,22 @@ Usage
 Options
 -------
 
-``-j``, ``--jailcitizen``
+``-j true|false``, ``--jailcitizen true|false``
```
myk002, line 22: the usual style here is to only document the parameters for the long form to avoid duplication

docs/autojustice.rst
Igualop: I will take a look on the GUI part later on. 

Should I still add the "autojustice" entry to the `gui/controrl-panel` even without a GUI?

myk002: Yes, the control panel will automatically pick up the GUI config when it's added. Until then, autojustice will just not have a button for launching the gui

myk002:
> * I had a goblin poet getting convicted even though he did join my fortress. I'm not sure why he is not being considered as a citizen.

btw, this is because permanent residents are not "citizens". you might have to make a special check for this case

myk002:
be sure to merge `justicetools.lua` into the main file (or move it into `internal`) to solve the docs build error

myk002:
Looks good! If this is ready for merge, could you fix the pre-commit.ci errors and remove WIP from the PR title?

drhead:
I would appreciate an option to not have the script handle automatic interrogations for visitors.  I do find that some visitors who turn out to be innocent want to leave after an interrogation, so I don't think it's necessarily a good idea to interrogate absolutely everyone.

drhead:
Testing the script out, I also noticed that goblins during sieges are queued for interrogation as if they were visitors.

myk002:
@Igualop could you address the issues drhead found before merge?

myk002:
Has there been any progress on this PR?

master-spike:
I've been looking at this as I actually started working on a plugin with similar features that does the same thing, before it had been pointed out that this PR existed. @Igualop if you're still around and can remember stuff about this, I do have a question - for those `unk_vector_...`s that you noticed always had 1 element in them, I've not noticed them contain anything at all at any point. I've also noticed that when setting the convicted ids the fortress and victim doesn't get the relevant happy/sad thoughts, like they would hitting the convict button. Is it possible that those vectors are related to this?

PR #598, opened 2023-02-05
Enable confirm

robob27:
Sister PR to [#2721](https://github.com/DFHack/dfhack/pull/2721) to enable confirm

- Rename `gui/confirm-opts` to `gui/confirm`
- Use a Window/ZScreen in `gui/confirm`
- Add `confirm` to System Services in control-panep

### DIFF
--- a/gui/confirm.lua
+++ b/gui/confirm.lua
@@ -18,27 +18,26 @@ function Opts:init()
             text_pen = COLOR_GREY,
             cursor_pen = COLOR_WHITE,
             choices = {},
-            on_submit = self:callback('toggle'),
-        },
-        widgets.HotkeyLabel{
-            frame = {b=1, l=0},
-            label='Toggle all',
-            key='CUSTOM_ALT_E',
-            on_activate=function() self:toggle_all(self.subviews.list:setSelected(i)) self:refresh() end,
+            on_submit = function() self:toggle(self.subviews.list:getSelected()) self:refresh() end,
         },
         widgets.HotkeyLabel{
             frame = {b=2, l=0},
+            label='Toggle all',
+            key='CUSTOM_ALT_E',
+            on_activate=function() self:toggle_all(self.subviews.list:getSelected()) self:refresh() end,
+        },
+        widgets.HotkeyLabel{
+            frame = {b=1, l=0},
             label='Resume paused confirmations',
             key='CUSTOM_P',
             on_activate=function() self:unpause_all() self:refresh() end,
             enabled=function() return self.any_paused end
         },
-        widgets.Label{
-            view_id = 'controls',
-            frame = {b = 0, l = 0},
-            text = {
-                {key = 'SELECT', text = ': Toggle'},
-            },
+        widgets.HotkeyLabel{
+            frame = {b=0, l=0},
+            label='Toggle',
+            key='SELECT',
+            on_activate=function() self:toggle(self.subviews.list:getSelected()) self:refresh() end,
         },
     }
     self:refresh()
@@ -86,7 +85,8 @@ function Opts:choice_pen(index, enabled)
     return (enabled and COLOR_GREEN or COLOR_RED) + (index == self.subviews.list:getSelected() and 8 or 0)
 end
 
-function Opts:toggle(_, choice)
+function Opts:toggle(choice)
+    choice = self.data[choice]
     confirm.set_conf_state(choice.id, not choice.enabled)
     self:refresh()
 end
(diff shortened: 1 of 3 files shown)


gui/confirm.lua
```diff
@@ -1,7 +1,7 @@
 -- confirm plugin options
 --[====[
 
-gui/confirm-opts
+gui/confirm
 ================
 A basic configuration interface for the `confirm` plugin.
```
myk002, line 7: you can remove this old copy of the docs. it's no longer used and is just liable to get out of sync with the real docs in the rst

gui/confirm.lua
```diff
@@ -12,14 +12,11 @@ confirm = require 'plugins.confirm'
 gui = require 'gui'
 widgets = require 'gui.widgets'
 
-Opts = defclass(Opts, gui.FramedScreen)
+Opts = defclass(Opts, widgets.Window)
 Opts.ATTRS = {
     frame_style = gui.GREY_LINE_FRAME,
```
myk002, line 17: you should use the default window style here (i.e. remove this line)

gui/confirm.lua
```diff
@@ -31,47 +28,30 @@ function Opts:init()
...
         widgets.Label{
             view_id = 'controls',
             frame = {b = 0, l = 0},
             text = {
-                {key = 'SELECT', text = ': Toggle, '},
-                {key = 'LEAVESCREEN', text = ': Back', on_activate = self:callback('dismiss')},
-                NEWLINE,
-                {key = 'SEC_SELECT', text = ': Toggle all'},
+                {key = 'SELECT', text = ': Toggle'},
             },
```
myk002, line 41: To allow this hotkey hint to be clickable too, it should be a HotkeyLabel that runs `toggle` on SELECT

myk002:
Ok, so next steps:
1. Convert confirm dialogs to Lua-drawn widget.Windows with standard theming and clickable widgets
2. ignore unfocused ZScreens when determining which screen should be queried for a focus string (I'll take care of this)
3. moar confirm checks (at your leisure)

did I miss anything that we talked about?

gui/confirm.lua
```diff
@@ -0,0 +1,121 @@
...
+            text_pen = COLOR_GREY,
+            cursor_pen = COLOR_WHITE,
+            choices = {},
+            on_submit = self:callback('toggle'),
+        },
+        widgets.HotkeyLabel{
+            frame = {b=1, l=0},
+            label='Toggle all',
+            key='CUSTOM_ALT_E',
+            on_activate=function() self:toggle_all(self.subviews.list:setSelected(i)) self:refresh() end,
```
myk002, line 27: what's `i` set to here?
actually, looking at how this is used, this is supposed to be getSelected(), isn't it?

robob27: Yeahhh, just fixed that up

robob27:
> Ok, so next steps:
> 
>     1. Convert confirm dialogs to Lua-drawn widget.Windows with standard theming and clickable widgets
> 
>     2. ignore unfocused ZScreens when determining which screen should be queried for a focus string (I'll take care of this)
> 
>     3. moar confirm checks (at your leisure)
> 
> 
> did I miss anything that we talked about?

I can't think of anything, other than maybe if we were making changes to what I did with the getSelectedStockpile stuff. There is a bit more cleaning up to do in Gui.cpp as well I think in general but nothing show stopping for now I think 🤞

gui/confirm.lua
```diff
@@ -0,0 +1,121 @@
...
+    end
+    self.subviews.list:setChoices(choices)
+end
+
+function Opts:choice_pen(index, enabled)
+    return (enabled and COLOR_GREEN or COLOR_RED) + (index == self.subviews.list:getSelected() and 8 or 0)
+end
+
+function Opts:toggle(choice)
+    choice = self.data[choice]
```
myk002, line 89:
```suggestion
function Opts:toggle(idx)
    local choice = self.data[idx]
```

myk002: since `choice` is normally reserved for the object table in the list item

myk002 (approved):

myk002:
bah, I forgot to list config persistence. from discord: ```
One thing that's still needed for confirm is config persistence. It can go in a separate PR, though
User config should be stored in dfhack-config/confirm.json and not be associated with the current fort since the choices are global
```

robob27:
So many things to consider for this stuff! Oopsie

PR #600, opened 2023-02-06
Caravan overlay

robob27:
Adds an overlay to the caravan script that adds some functionality/info to the trade screen:

image

(Shift+Scroll is vanilla but a lot of people don't know about it. I see people get especially frustrated with the amount of scrolling on this page)

The ctrl+shift click actions seem to work super well!

Couple limitations/caveats with it as is right now:

- I can't find a way to calculate what the new height of the scrollbar should be after collapsing categories/bins programmatically. This means I have to manually calculate it. Based on the number of categories showing or the number of items contained in the collapsed items. I've tested it quite a bit and it seems to work pretty reliably but it would be nice to find a better solution.
- Because of some of the issues created by the preceding point, I can't make "Collapse categories" and "Collapse bins" be separate actions. If collapsing categories, bins must also be collapsed first to avoid incorrectly calculating `i_height`.
- It can probably be DRY'd up a bit by extracting the `while not end_of_container_reached` stuff into a separate function that takes a callback or something.

TODO: 

- check for docs to update
- address any limitations possible

### DIFF
--- a/caravan.lua
+++ b/caravan.lua
@@ -1,31 +1,286 @@
--- Adjusts properties of caravans
---[====[
-
-caravan
-=======
-
-Adjusts properties of caravans on the map. See also `force` to create caravans.
-
-This script has multiple subcommands. Commands listed with the argument
-``[IDS]`` can take multiple caravan IDs (see ``caravan list``). If no IDs are
-specified, then the commands apply to all caravans on the map.
-
-**Subcommands:**
-
-- ``list``: lists IDs and information about all caravans on the map.
-- ``extend [DAYS] [IDS]``: extends the time that caravans stay at the depot by
-  the specified number of days (defaults to 7 if not specified). Also causes
-  caravans to return to the depot if applicable.
-- ``happy [IDS]``: makes caravans willing to trade again (after seizing goods,
-  annoying merchants, etc.). Also causes caravans to return to the depot if
-  applicable.
-- ``leave [IDS]``: makes caravans pack up and leave immediately.
-- ``unload``: fixes endless unloading at the depot. Run this if merchant pack
-  animals were startled and now refuse to come to the trade depot.
-
-]====]
-
+-- Adjusts properties of caravans and provides overlay for enhanced trading
 --@ module = true
+
+local gui = require('gui')
+local overlay = require('plugins.overlay')
+local widgets = require('gui.widgets')
+
+trader_selected_state = trader_selected_state or {}
+broker_selected_state = broker_selected_state or {}
+handle_ctrl_click_on_render = handle_ctrl_click_on_render or false
+handle_shift_click_on_render = handle_shift_click_on_render or false
+
+dfhack.onStateChange.caravanTradeOverlay = function(code)
+    if code == SC_WORLD_UNLOADED then
+        trader_selected_state = {}
+        broker_selected_state = {}
+        handle_ctrl_click_on_render = false
+        handle_shift_click_on_render = false
+    end
+end
+
+local GOODFLAG = {
+    UNCONTAINED_UNSELECTED = 0,
+    UNCONTAINED_SELECTED = 1,
+    CONTAINED_UNSELECTED = 2,
+    CONTAINED_SELECTED = 3,
+    CONTAINER_COLLAPSED_UNSELECTED = 4,
+    CONTAINER_COLLAPSED_SELECTED = 5,
+}
+
+function select_shift_clicked_container_items(new_state, old_state, list_index)
+    -- if ctrl is also held, collapse the container too
+    local also_collapse = dfhack.internal.getModifiers().ctrl
+    local collapsed_item_count = 0
+    for k, goodflag in ipairs(new_state) do
+        if old_state[k] ~= goodflag then
+            local next_item_flag = new_state[k + 1]
+            local this_item_is_container = df.item_binst:is_instance(df.global.game.main_interface.trade.good[list_index][k])
+            if this_item_is_container then
+                local container_is_selected = goodflag == GOODFLAG.UNCONTAINED_SELECTED or goodflag == GOODFLAG.CONTAINER_COLLAPSED_SELECTED
+                if container_is_selected then
+                    local collapsed_this_container = false
+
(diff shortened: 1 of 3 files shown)


robob27:
pre-commit.ci autofix

caravan.lua
```diff
@@ -27,6 +27,312 @@ specified, then the commands apply to all caravans on the map.
 
 --@ module = true
 
+local gui = require('gui')
```
myk002, line 30: could you remove the docstring above (lines 2-26)? they're obsolete.

caravan.lua
```diff
@@ -27,6 +27,312 @@ specified, then the commands apply to all caravans on the map.
...
+    UNCONTAINED_SELECTED = 1,
+    CONTAINED_UNSELECTED = 2,
+    CONTAINED_SELECTED = 3,
+    CONTAINER_COLLAPSED_UNSELECTED = 4,
+    CONTAINER_COLLAPSED_SELECTED = 5,
+}
+
+function is_trade_overlay_active()
+    return df.global.game.main_interface.trade.open
+end
```
myk002, line 59: this is still required for now, but soon I hope to be using your new focus string logic instead. the `viewscreens` attribute will become `dwarfmode@focus_string` and the overlay framework will do the matching.

caravan.lua
```diff
@@ -27,6 +27,312 @@ specified, then the commands apply to all caravans on the map.
...
+
+CaravanOverlay = defclass(CaravanOverlay, overlay.OverlayWidget)
```
myk002, line 61: every overlay we define can be individually enabled/disabled by the player. Is there any reason this is separate from the CaravanInfoOverlay below? As it is, it looks like they should be treated as a single unit.

robob27: Yeah, they should be a single unit for sure. Just inexperience with overlay stuff there. I'll fix it up.

caravan.lua
```diff
@@ -1,31 +1,311 @@
...
+end
+
+function copyGoodflagState()
+    trader_selected_state = copyall(df.global.game.main_interface.trade.goodflag[0])
+    broker_selected_state = copyall(df.global.game.main_interface.trade.goodflag[1])
+end
+
+function CaravanOverlay:onInput(keys)
+    if not is_trade_overlay_active() then return false end
+    if keys._MOUSE_L then
```
myk002, line 211: This will trigger multiple times in rapid succession as the mouse button is held down throughout the click. If you only want to catch the click itself, this should be `keys._MOUSE_L_DOWN`. `keys._MOUSE_L` should only be used for dragging operations.

caravan.lua
```diff
@@ -1,31 +1,311 @@
...
 
-Adjusts properties of caravans on the map. See also `force` to create caravans.
+function is_trade_overlay_active()
+    return df.global.game.main_interface.trade.open
+end
 
-This script has multiple subcommands. Commands listed with the argument
-``[IDS]`` can take multiple caravan IDs (see ``caravan list``). If no IDs are
-specified, then the commands apply to all caravans on the map.
+CaravanOverlay = defclass(CaravanOverlay, overlay.OverlayWidget)
```
myk002, line 35: Honestly, this class doesn't seem be adding much value as a subview. I think this class's methods should be integrated into CaravanInfoOverlay and this class can just be removed. If you do want to keep it as a separate class, though, it should inherit from something appropriate, like `gui.View`. It's not an `overlay.OverlayWidget` since it's not directly registered with the overlay framework, and therefore the ATTRS attached to this class will not be read by anything.

caravan.lua
```diff
@@ -1,31 +1,311 @@
...
+            handle_shift_click_on_render = true
+        elseif dfhack.internal.getModifiers().ctrl then
+            handle_ctrl_click_on_render = true
+        end
+
+        if handle_ctrl_click_on_render or handle_shift_click_on_render then
+            copyGoodflagState()
+        end
+    end
+    CaravanOverlay.super.onInput(self, keys)
```
myk002, line 222: it might be worthwhile to leave a comment here explaining that you are intentionally not returning a value from this method so everything is passed through to the underlying viewscreen. either that or explicitly return false.

caravan.lua
```diff
@@ -1,31 +1,311 @@
...
+    self:addviews{
+        widgets.Label{
+            frame={t=0, l=0},
+            text = 'Shift+Click checkbox:',
+            text_pen=COLOR_LIGHTGREEN,
+        },
+        widgets.Label{
+            frame={t=1, l=0},
+            text = 'select items inside bin',
+        },
```
myk002, line 261: not like what you have is wrong, but just so you know, Labels are quite flexible with their text. This could also be written like this:
```lua
widgets.Label{
    frame={t=0, l=0},
    text={
        {text='Shift+Click checkbox:', pen=COLOR_LIGHTGREEN},
        NEWLINE,
        'select items inside bin',
    },

File: caravan.lua [hunk 8]
```diff
@@ -1,31 +1,311 @@
...
+    CaravanInfoOverlay.super.render(self, dc)
+end
+
+function CaravanInfoOverlay:onInput(keys)
+    if not is_trade_overlay_active() then return false end
+    CaravanInfoOverlay.super.onInput(self, keys)
+end
+
+OVERLAY_WIDGETS = {
+    overlay=CaravanInfoOverlay,
```
myk002, line 307: this will show as "caravan.overlay" which maybe isn't *quite* descriptive enough. How about `tradeScreenExtension` or *something* that explicitly refers to the trade screen.

caravan.lua
myk002: once DFHack/dfhack#2825 is merged, then this can be expressed as a focus string in the `viewscreens` attribute and the `is_trade_overlay_active` function can be removed.

caravan.lua
robob27: Just got this one left, should be able to finish it up tonight.

myk002 (commented):
Code looks good!

Could you add:
1) changelog entry (under "Misc Improvements")
2) documentation for the overlay, detailing what it can do (see https://docs.dfhack.org/en/latest/docs/tools/orders.html#overlay for an example)

myk002 (approved):

PR #601, opened 2023-02-06
WIP Add timestream to enable, handle social activities and seed growth

robob27:
VERY WIP. I started writing these updates on 0.47.5, not sure if it even works in 50.05, and have learned a lot more about DFhack since then. Will try to get around to finishing it up sometime soon. Anyone else feel free to take it over. You won't hurt my feelings 😅

### DIFF
--- a/timestream.lua
+++ b/timestream.lua
@@ -1,3 +1,5 @@
+--@module = true
+--@enable = true
 -- speeds up the calendar, units, or both
 -- based on https://gist.github.com/IndigoFenix/cf358b8c994caa0f93d5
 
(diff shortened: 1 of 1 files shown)


myk002:
*lots* of people ask about timestream. these updates will be welcomed

myk002 (commented):
the state persistence and state change hooks need to be brought into line with other updated tools (e.g. emigration) so `timestream` looks and behaves like other tools controllable from the control panel.

I'd also suggest changing the default to `--units`, which is what I think most players expect to happen when `timestream` is enabled without any additional configuration.

PR #604, opened 2023-02-06
Revamp protect-nicks to instead set the nickname to the historical figure

plule:
It turns out that the source issue of the nickname being erased is that DF no longer sets the historical figure name.

So simply looping on the active units and resetting the nickname with "dfhack.units.setNickname" fixes the same issue, alongside with other problems such as the nickname missing from many places, and being lost when retiring a fort.


myk002 (approved):

ab9rf:
closes dfhack/dfhack#2874

PR #608, opened 2023-02-09
Reindex pathfinding on liquids spawn

vallode:
This fixes the pathing issues with gui/liquids. Specifically when spawning or removing magma.


myk002 (approved):

lethosor:
This should have a (dev-only?) changelog entry

lethosor:
Done - looked to me like only the `liquids` plugin set `reindex_pathfinding` before, so this could be a fix for an issue present in 0.47 as well, so I added it as a non-dev change.

PR #613, opened 2023-02-12
Add script to fix nicknames on unnamed units

dawnmist:
When the user adds a nickname through the DF name interface, it does not properly set the `has_name` flag. This then results in the nickname _not_ being persisted when the game is saved/reloaded, so user entered nicknames on creatures that don't already have a name get lost.

This script periodically scans fort-controlled units that don't have a name, checking whether the user has set a nickname (and if so, fixing the has_name flag) or cleared a previously fixed nickname (clearing the has_name flag).

Prerequisite for fixing nicknames for autobutcher so that the user can set nicknames to protect animals from being butchered (e.g. because they had good stats).

See:
- https://github.com/DFHack/dfhack/issues/2874
- https://github.com/DFHack/dfhack/issues/2871

### DIFF
--- a/fix/nicknames.lua
+++ b/fix/nicknames.lua
@@ -1,0 +1,164 @@
+-- Workaround for the v50.x bug where Dwarf Fortress doesn't set the has_name flag properly when setting/clearing nicknames.
+--
+--@enable = true
+--@module = true
+
+local eventful = require('plugins.eventful')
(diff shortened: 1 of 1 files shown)


dawnmist:
Closing since we realised that `fix/protect-nicks.lua` already provides this functionality.

PR #617, opened 2023-02-13
Persist confirm config

robob27:
Write confirm config to `dfhack-config/confirm.json` any time a confirm is enabled or disabled. Will be linked to a related dfhack/dfhack PR

Related to https://github.com/DFHack/dfhack/issues/2815

### DIFF
--- a/gui/confirm.lua
+++ b/gui/confirm.lua
@@ -1,8 +1,20 @@
 -- config ui for confirm
+--@ module = true
+local GLOBAL_KEY = 'confirm'
+local CONFIRM_CONFIG_FILE = 'dfhack-config/confirm.json'
+local json = require('json')
+local confirm = require('plugins.confirm')
+local gui = require('gui')
+local widgets = require('gui.widgets')
 
-confirm = require 'plugins.confirm'
-gui = require 'gui'
-widgets = require 'gui.widgets'
+dfhack.onStateChange[GLOBAL_KEY] = function(sc)
+    if sc ~= SC_CORE_INITIALIZED then return end
+    local ok, config = pcall(json.decode_file, CONFIRM_CONFIG_FILE)
+    if not ok then return end
+    for _, c in ipairs(config) do
+        confirm.set_conf_state(c.id, c.enabled)
+    end
+end
 
 Opts = defclass(Opts, widgets.Window)
 Opts.ATTRS = {
@@ -51,6 +63,13 @@ function Opts:init()
     end
 end
 
+function Opts:persist_data()
+    if not safecall(json.encode_file, self.data, CONFIRM_CONFIG_FILE) then
+        dfhack.printerr(('failed to save confirm config file: "%s"')
+                :format(path))
+    end
+end
+
 function Opts:refresh()
     self.data = confirm.get_conf_data()
     self.any_paused = false
@@ -89,6 +108,7 @@ function Opts:toggle(idx)
     local choice = self.data[idx]
     confirm.set_conf_state(choice.id, not choice.enabled)
     self:refresh()
+    self:persist_data()
 end
 
 function Opts:toggle_all(choice)
@@ -96,6 +116,7 @@ function Opts:toggle_all(choice)
         confirm.set_conf_state(c.id, not self.data[choice].enabled)
     end
     self:refresh()
+    self:persist_data()
 end
 
 function Opts:unpause_all()
@@ -116,6 +137,10 @@ end
 
 function OptsScreen:onDismiss()
     view = nil
+end
+
+if dfhack_flags.module then
+    return
 end
 
 view = view and view:raise() or OptsScreen{}:show()
--- a/gui/confirm.lua
+++ b/gui/confirm.lua
@@ -10,7 +10,14 @@ local widgets = require('gui.widgets')
 dfhack.onStateChange[GLOBAL_KEY] = function(sc)
     if sc ~= SC_CORE_INITIALIZED then return end
     local ok, config = pcall(json.decode_file, CONFIRM_CONFIG_FILE)
+    local all_confirms = confirm.get_conf_data()
+    -- enable all confirms by default so new confirms are automatically enabled
+    for _, c in ipairs(all_confirms) do
+        confirm.set_conf_state(c.id, true)
+    end
+
     if not ok then return end
+    -- update confirm state based on config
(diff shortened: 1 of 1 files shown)


robob27:
@myk002 Apologies in advance if this is silly again. I moved the logic over here and for some reason it doesn't seem to be working. When the conf state is updated in the `onStateChange` callback, the change doesn't seem to persist to `Opt:init` now and I'm not sure why

gui/confirm.lua
```diff
@@ -1,8 +1,36 @@
...
+        if ok then
+            for _, c in ipairs(conf_data) do
+                -- for some reason changing the conf state here doesn't seem to change the result
+                -- when the dialog is opened like it did when it was in the other script?
+                confirm.set_conf_state(c.id, c.enabled)
+            end
+        end
+        loaded = true
+    end
+end
```
myk002, line 29:
```suggestion
dfhack.onStateChange[GLOBAL_KEY] = function(sc)
    if sc ~= SC_CORE_INITIALIZED then return end

    local ok, config = pcall(json.decode_file, CONFIRM_CONFIG_FILE)
    if not ok then return end
    for _, c in ipairs(config) do
        confirm.set_conf_state(c.id, c.enabled)
    end
end
```

myk002: what you had here should have worked, I think, I just cleaned it up a bit. Did you restart DF to test? it wouldn't have loaded until you caused a state change event.

robob27: Yeah it wasn't working even with restarting DF. The file was properly written, and with print statements I could see that the `onStateChange` code was successfully being called. But then when I invoked `gui/confirm`, the result of `confirm.get_conf_data()` was not affected.

robob27: Seeing the same thing with this version

 of the code. It totally seems like it should work but..

https://user-images.githubusercontent.com/2104555/218814231-0a5249e8-38dc-4abf-8785-4801a64b944e.mp4

robob27: I experimented with it a bit. It seems like if the `onStateChange` code happens too early, the change to conf state doesn't persist.

If I change the `sc` code to `SC_MAP_LOADED`, the confirmations have the expected state when I load up a map and invoke `gui/confirm`. Similarly, if I remove the condition for `sc` code altogether, I can see the `onStateChange` code fire off twice as I load up the game, and then 3 more times once the game has finished loading (event happen in the order: SC_CORE_INITIALIZED, SC_VIEWSCREEN_CHANGED, SC_VIEWSCREEN_CHANGED, SC_VIEWSCREEN_CHANGED, SC_VIEWSCREEN_CHANGED). The confirmations then also have the expected state when I invoke `gui/confirm`.

robob27: If I put:

```
loaded = loaded or false

if dfhack.onStateChange then
    dfhack.onStateChange[GLOBAL_KEY] = function(sc)
        if loaded then return end
        local ok, config = pcall(json.decode_file, CONFIRM_CONFIG_FILE)
        if not ok then return end
        for _, c in ipairs(config) do
            set_conf_state(c.id, c.enabled)
            print('Set ' .. c.id .. ' to ' .. tostring(c.enabled))
        end
        loaded = true
    end
end
```

Back into `plugins/confirm.lua`, everything works as expected. If I don't do the `if dfhack.onStateChange then` check, then I see this error in the console:

```
Loading script: hack/init/dfhack.tools.init
...p\DFHack Dev\Dwarf Fortress\hack\lua\plugins\confirm.lua:17: attempt to index a nil value (field 'onStateChange')
stack traceback:
        ...p\DFHack Dev\Dwarf Fortress\hack\lua\plugins\confirm.lua:17: in main chunk
        [C]: in function 'require'
```

But then it still works as expected:

image

It definitely seems to be a matter of timing but I can't work out exactly what the right time is.

myk002: let me look into why this isn't working as I expect it to. other scripts don't seem to have trouble attaching handlers when they are first loaded

myk002: it looks like the confirm plugin itself is overwriting all of its state with "enabled" when it is enabled (which happens after core init). That part needs updating to work with the state restoration flow.

gui/confirm.lua
```diff
@@ -1,8 +1,36 @@
 -- config ui for confirm
+--@ module = true
+local GLOBAL_KEY = 'confirm'
+local CONFIRM_CONFIG_FILE = 'dfhack-config/confirm.json'
+local json = require('json')
+confirm = require('plugins.confirm')
+gui = require('gui')
+widgets = require('gui.widgets')
```
myk002, line 8:
```suggestion

local json = require('json')
local confirm = require('plugins.confirm')
local gui = require('gui')
local widgets = require('gui.widgets')

local GLOBAL_KEY = 'confirm'
local CONFIRM_CONFIG_FILE = 'dfhack-config/confirm.json'

```

myk002: `require`d environments should be local

gui/confirm.lua
```diff
@@ -1,8 +1,36 @@
...
+                confirm.set_conf_state(c.id, c.enabled)
+            end
+        end
+        loaded = true
+    end
+end
+
+if dfhack_flags.module then
+    return
+end
```
myk002, line 33: this block should go at the end of the file, just before the last line that instantiates the UI

gui/confirm.lua
```diff
@@ -51,8 +79,19 @@ function Opts:init()
...
+                :format(path))
+    end
+end
+
 function Opts:refresh()
     self.data = confirm.get_conf_data()
+    print('updated conf data:')
+    for _, confd in ipairs(self.data) do
+        printall(confd)
+    end
```
myk002, line 94: commenting on this just so we don't forget to remove it before merge

gui/confirm.lua
robob27: Ahhh, okay yeah I think I see that now. Updating.

robob27: I think I made the required update (to the other PR). I just need to make one more change to this PR now I think to ensure that confirms get enabled by default if config is not successfully loaded, since nothing else will should be enabling them now

robob27: I think my latest commit handles this

myk002 (approved):

PR #619, opened 2023-02-14
Check for tile_type being nil to avoid out of bounds index error, and also respect stair designations

johncosker:
Add check for `nil` `tile_type` to try to avoid whatever is happening in https://github.com/DFHack/dfhack/issues/2893

I still couldn't reproduce the original bug but this should avoid it.

Also updated the `getDesignation` code to auto matically convert downstair designations at the top of the stairs to updown.

### DIFF
--- a/gui/dig.lua
+++ b/gui/dig.lua
@@ -764,7 +764,7 @@ function Dig:getDesignation(x, y, z)
             return stairs_bottom_type == "auto" and "u" or stairs_bottom_type
         elseif z == math.abs(self:get_bounds().z1 - self:get_bounds().z2) then
             local tile_type = dfhack.maps.getTileType(self:get_bounds().x1 + x, self:get_bounds().y1 + y, z)
-            local tile_shape = tile_attrs[tile_type].shape
+            local tile_shape = tile_type ~= nil and tile_attrs[tile_type].shape or nil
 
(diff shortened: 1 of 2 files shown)


gui/dig.lua
```diff
@@ -763,12 +763,20 @@ function Dig:getDesignation(x, y, z)
         if z == 0 then
             return stairs_bottom_type == "auto" and "u" or stairs_bottom_type
         elseif z == math.abs(self:get_bounds().z1 - self:get_bounds().z2) then
-            local tile_type = dfhack.maps.getTileType(self:get_bounds().x1 + x, self:get_bounds().y1 + y, z)
+            local tile_type = dfhack.maps.getTileType(self:get_bounds().x1 + x, self:get_bounds().y1 + y, self:get_bounds().z1 + z)
             local tile_shape = tile_type ~= nil and tile_attrs[tile_type].shape or nil
+            local designation = dfhack.maps.getTileFlags(xyz2pos(self:get_bounds().x1 + x, self:get_bounds().y1 + y, self:get_bounds().z1 + z))
```
myk002, line 768: you could pull the coordinates out into a `local pos = xyz2pos(self.getBounds()...xyz)` and pass that to getTileType and getTileFlags to clean this up a little bit

myk002:
also please add a "dev" changelog entry (that is, starting with `-@` instead of just `-`)

myk002 (approved):

PR #623, opened 2023-02-17
Revises query.lua to use df_shortcut_env

cppcooper:
Table lookups now use `df_shortcut_env` instead of `df_expr_to_ref` This fixes a problem trying to index some vectors (probably all)

### DIFF
--- a/devel/query.lua
+++ b/devel/query.lua
@@ -169,12 +169,19 @@ function main()
     query(selection, path_info, args.search, path_info)
 end
 
+local eval_env = utils.df_shortcut_env()
+function eval(s)
+    local f, err = load("return " .. s, "expression", "t", eval_env)
+    if err then qerror(err) end
+    return f()
+end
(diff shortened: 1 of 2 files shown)


devel/query.lua
```diff
@@ -169,12 +169,19 @@ function main()
     query(selection, path_info, args.search, path_info)
 end
 
+local eval_env = utils.df_shortcut_env()
+function eval(s)
+    local f, err = load("return " .. s, "expression", "t", eval_env)
+    if err then qerror(err) end
+    return f()
+end
```
lethosor, line 177: Made an issue for the to-do we discussed on Discord around refactoring this: https://github.com/DFHack/dfhack/issues/2928

lethosor (approved):

lethosor:
Tested locally and confirmed that function calls in the `-table` expression work now

PR #625, opened 2023-02-21
[gui/seedwatch] initial version

silverflyone:
initial version of seedwatch gui


myk002:
I haven't reviewed yet, but woot!

silverflyone:
pre-commit.ci autofix

silverflyone:
> I haven't reviewed yet, but woot! (And thank you)

My pleasure. Reasonably straight forward, as there were a few examples to use. There aren't any changes needed to the seedwatch plugin.

silverflyone:
@myk002 ready for feedback.

gui/seedwatch.lua
```diff
@@ -0,0 +1,310 @@
...
+local REFRESH_MS = 10000
+local MAX_TARGET = 2147483647
+--
+-- SeedSettings
+--
+SeedSettings = defclass(SeedSettings, widgets.Window)
+SeedSettings.ATTRS{
+    lockable=false,
+    frame={l=5, t=5, w=35, h=9},
+    data={id='', name='', quantity=0, target=0,},
```
myk002, line 17: it is dangerous to put mutable tables as default values in ATTRS since all instances will share the same state. Since this is not actually set by the parent view, you'd be better off initializing the `data` field in `SeedSettings:init()`.

Now, it's not *really* an issue here since SeedSettings is effectively a singleton, but I'd rather this antipattern not appear anywhere in our code for other script writers to copy.

gui/seedwatch.lua
```diff
@@ -0,0 +1,310 @@
...
+    self:addviews{
+        widgets.Label{
+            frame={t=0, l=0},
+            text='Seed: ',
+        },
+        widgets.Label{
+            view_id='id',
+            frame={t=0, l=0},
+            visible=false,
+        },
```
myk002, line 30: was this for debugging? it appears to overlap the previous field and is never made visible

gui/seedwatch.lua
```diff
@@ -0,0 +1,310 @@
...
+end
+
+function SeedSettings:show(choice, on_commit)
+    self.data = choice.data
+    self.on_commit = on_commit
+    local data = self.data
+    self.subviews.id:setText(data.id)
+    self.subviews.name:setText(data.name)
+    self.subviews.quantity:setText(tostring(data.quantity))
+    self.subviews.target:setText(tostring(data.target))
```
myk002, line 69: this could be simplified by working the data into function callbacks in the Labels. For example, instead of:
```lua
        widgets.Label{
            frame={t=1, l=0},
            text='Quantity: ',
        },
        widgets.Label{
            view_id='quantity',
            frame={t=1, l=10},
            text_pen=COLOR_GREEN,
        },
```
you could have
```lua
        widgets.Label{
            frame={t=1, l=0},
            text={
                'Quantity:',
                {gap=1, text=function() return self.data.quantity end, pen=COLOR_GREEN},
            },
        },
```

gui/seedwatch.lua
```diff
@@ -0,0 +1,310 @@
...
+    self.visible = false
+end
+
+function SeedSettings:commit()
+    local target = math.tointeger(self.subviews.target.text)
+    if not target or target == '' then
+        target = 0
+    elseif target > MAX_TARGET then
+        target = MAX_TARGET
+    end
```
myk002, line 86:
```suggestion
    local target = math.tointeger(self.subviews.target.text) or 0
    target = math.min(MAX_TARGET, math.max(0, target))
```

gui/seedwatch.lua
```diff
@@ -0,0 +1,310 @@
...
+
+local PROPERTIES_HEADER = '   Quantity     Target   '
+local REFRESH_MS = 10000
+local MAX_TARGET = 2147483647
+--
+-- SeedSettings
+--
+SeedSettings = defclass(SeedSettings, widgets.Window)
+SeedSettings.ATTRS{
+    lockable=false,
```
myk002, line 15: This attribute only existed for a brief time and no longer exists. I just removed other leftover references to it (gui/autochop and gui/automelt)

gui/seedwatch.lua
```diff
@@ -0,0 +1,310 @@
...
+--
+Seedwatch = defclass(Seedwatch, widgets.Window)
+Seedwatch.ATTRS {
+    frame_title='Seedwatch',
+    frame={w=60, h=27},
+    resizable=true,
+    resize_min={h=25},
+    hide_unmonitored=DEFAULT_NIL,
+    manual_hide_unmonitored_touched=DEFAULT_NIL,
+    data = {sum=0, seeds_qty=0, seeds_watched=0, seeds = {{id='', name='', quantity=0, target=0,}}},
```
myk002, line 112: internal data that is not intended to be set by a parent should be initialized in `init()`

gui/seedwatch.lua
```diff
@@ -0,0 +1,310 @@
...
+        },
+
+    }
+
+    self:refresh_data()
+end
+
+function Seedwatch:getDefaultHide()
+    return false
+end
```
myk002, line 216: unused?

gui/seedwatch.lua
silverflyone: Removed lockable

gui/seedwatch.lua
silverflyone: these attr are set each time in refresh_data, so removed.

gui/seedwatch.lua
silverflyone: I'll try that.

gui/seedwatch.lua
silverflyone: yeah. squashed.

gui/seedwatch.lua
silverflyone: removed.

gui/seedwatch.lua
silverflyone: It was used as a passback variable, but also used for debugging.

gui/seedwatch.lua
silverflyone: done

silverflyone:
@myk002 I've applied some changes. Let me know suggestions.

gui/seedwatch.lua
myk002: it can be defined at both levels, with the text (token) overriding the Label-level `text_pen`. I'm pretty sure that should work as advertised

gui/seedwatch.lua
```diff
@@ -0,0 +1,293 @@
...
+-- Seedwatch
+--
+Seedwatch = defclass(Seedwatch, widgets.Window)
+Seedwatch.ATTRS {
+    frame_title='Seedwatch',
+    frame={w=60, h=27},
+    resizable=true,
+    resize_min={h=25},
+    hide_unmonitored=DEFAULT_NIL,
+    manual_hide_unmonitored_touched=DEFAULT_NIL,
```
myk002, line 99:
```suggestion
```

myk002 (approved):

myk002:
pre-commit.ci autofix

PR #626, opened 2023-02-22
`troubleshoot-item`: fix to job handling for v50, and some updates to what is detected and printed

TheBloke:
There was a bug in `troubleshoot-item` that caused items with in-progress jobs to print an error and not print the job details. This was fixed.

While reviewing the script, I additionally added/changed the following:
* Added detection of hidden items, and artifacts (mentioning if the artifact was from a strange mood or not.)
* Fixed detection of items in a chest, and added detection of items in a cabinet, bookcase, or any general building, and print the building type
* Added detection of items held by a unit, or stored in a storage item (bin, barrel), and print unit name or storage item type.
* Clarify the messages printed when an item is used for making a building, or a construction, to differentiate them from items stored in that building.


myk002 (commented):
looks good! could you add changelog entries for the improvements and fixes?

TheBloke:
Sorry didn't think of that - that's done now

TheBloke:
I'm going to make another commit or two, adding a couple more pieces of info

TheBloke:
I'm done!

myk002 (approved):

changelog.txt
```diff
@@ -25,6 +26,10 @@ that repo.
 -@ `unforbid`: fix detection of unreachable items for items in containers
 
 ## Misc Improvements
+- `troubleshoot-item`: output as bullet point list with indenting, with item description and ID at top. Now reports on more data, regarding items that are:
+    - hidden; artifacts; held by a unit, with unit name printed
+    - in a storage item or container, displaying the type of container
+    - are containers, with the contents of the container summarised with count of each item type
```
lethosor, line 32: Unfortunately nested lists don't work like this in the changelog. Any item, whether immediately after the tool name or on a separate line, is treated as an individual change for that tool, and then grouped later when the HTML is generated. Order is typically maintained, but it looks like this:
image

I think the "Now reports on more data" bit could be its own item, including the three list items currently under it. Does that change work for you?

myk002: fixed in https://github.com/DFHack/scripts/commit/e70393ff2132e2f5d00ffeb154e8d9242f89a284

TheBloke: Thanks, both! A single ine would have been fine but if you're happy with Myk's edit then I am too.

PR #627, opened 2023-02-22
New script to automate job suspension/unsuspension: suspendmanager

plule:
Add `suspendmanager`, addressing most of the issue https://github.com/DFHack/dfhack/issues/2955

It is intended to work as a replacement for `autounsuspend` with the additional functionality of also suspending jobs in way that it prevents construction jobs to block each other. It works by computing a "risk of being stuck" factor for a job, based on how close it is to be enclosed. Jobs that are neighbouring another job more at risk of being stuck will be suspended.

Modified:
- `unsuspend` now support `--skipblocking`, to not unsuspend jobs that risk blocking other

Added:
- `suspend [--onlyblocking]`: Suspending all jobs, or only the one that risk blocking other
- `suspendmanager`: Automation of `suspend`/`unsuspend` intended to remove job suspension micromanagement, reusable module for other suspension management tools

Misc:
- `unsuspend --quiet` is now quieter, previously it was only removing the "0 job" messages, now it removes all messages
- `unsuspend` now uses argparse

Configuration GUI:
image

This PR only lacks the migration from `autounsuspend`, which is currently left untouched. I don't really now how to do this

- [x] `suspend.lua`
- [x] `suspendmanager.lua` -- dumb trigger
- [x] `gui/suspendmanager.lua`
- [x] rework `suspendmanager` triggers
- [x] `unsuspend` skips blocking jobs
- [x] doc/terminology cleanup


smartsuspend.lua
```diff
@@ -0,0 +1,117 @@
...
+
+if dfhack_flags.module then
+    return
+end
+
+for _,job in utils.listpairs(df.global.world.jobs.list) do
+    if jobShouldBeSuspended(job) then
+        suspend(job)
+    else
+        job.flags.suspend = false
```
myk002, line 115: this script shouldn't be unsuspending jobs unless it's taking over the job of autounsuspend (which I recommend that it does -- I think this functionality should be part of `autounsuspend`, on by default, and controllable with a setting)

smartsuspend.lua
```diff
@@ -0,0 +1,117 @@
...
+    if buildingRef then
+        return df.building.find(buildingRef.building_id)
+    end
+    return nil
+end
+
+--- Check if a building will block the path
+--- @param building building|building_constructionst|nil
+--- @return boolean
+local function isBlockingConstruction(building)
```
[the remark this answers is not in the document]
plule: Ha, dark knowledge of dfhack, thanks

smartsuspend.lua
plule: Yep, it makes sense

plule:
@myk002 regarding the best way to call this from: Calling it from the main "unsuspend" loop, where it also suspends jobs works well, but at the condition that the `autounsuspend` frequency is lowered from 1 day to 10 ticks. I don't realize how acceptable it is, since that's a pretty big jump.

Alternatively, I can investigate the event route, to try to hook to the "job" completed event, though that's uncharted territory for me

plule:
Welll the gui work went smoother than I expected, so I directly added the integration in unsuspend and autounsuspend.

For now it's going the route of running it every 10 ticks though, which is maybe a problem

autounsuspend.lua
```diff
@@ -8,19 +8,35 @@ local persist = require('persist-table')
 local GLOBAL_KEY = 'autounsuspend' -- used for state change hooks and persistence
 
 enabled = enabled or false
+smart = smart or nil
+if smart == nil then
+    smart = true
+end
```
myk002, line 14:
```suggestion
smart = smart == nil and true or smart
```

docs/autounsuspend.rst
```diff
@@ -16,3 +16,12 @@ Usage
 ::
 
     enable autounsuspend
+
+Options
+-------
+
+``enable_smart``
+    Automatically suspend jobs that risk blocking other construction jobs. See `smartsuspend`.
```
myk002, line 24: should mention that this is the default

gui/autounsuspend.lua
```diff
@@ -0,0 +1,77 @@
...
+function Autounsuspend:refresh_data()
+    self.subviews.enable_toggle:setOption(autounsuspend.isEnabled())
+    self.subviews.smartMode:setOption(autounsuspend.smart)
+    self.next_refresh_ms = dfhack.getTickCount() + REFRESH_MS
+end
+
+function Autounsuspend:onRenderBody()
+    if self.next_refresh_ms <= dfhack.getTickCount() then
+        self:refresh_data()
+    end
```
myk002, line 56: unless you're reporting on the live number of suspended jobs, there's really no reason to have an autorefresh.

unsuspend.lua
```diff
@@ -184,9 +185,20 @@ if dfhack_flags.module then
     return
 end
 
-local unsuspended_count, flow_count, buildingplan_count = 0, 0, 0
+local opts = utils.invert{...}
+local quiet = opts['-q'] or opts['--quiet']
+local smart = opts['-s'] or opts['--smart']
```
myk002, line 190: this is *ok* for now, but at some point you should just use a real commandline parsing function like `argparse.processArgsGetopt()` and avoid having to do the matching and parsing yourself. docs here: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#argparse

docs/unsuspend.rst
```diff
@@ -19,6 +22,15 @@ Usage
 
     unsuspend
 
+Options
+-------
+
+``-q``, ``--quiet``
+    Don't report the number of unsuspended tiles
```
myk002, line 29: jobs, not tiles, right?

autounsuspend.lua
```diff
@@ -8,19 +8,35 @@ local persist = require('persist-table')
...
 local function event_loop()
     if enabled then
-        dfhack.run_script('unsuspend', '--quiet')
-        dfhack.timeout(1, 'days', event_loop)
+        local args = {'--quiet'}
+        if smart then
+            table.insert(args, '--smart')
+        end
+        dfhack.run_script('unsuspend', table.unpack(args))
+        dfhack.timeout(10, 'ticks', event_loop)
```
myk002, line 39: yeah, this is *really* frequent, even more frequent than `autonestbox`, which has similar requirements for "mark it before someone takes it". I'd be concerned about running even cpp code at this frequency. Registering a job completion handler might be the better approach.

myk002 (commented):
I think I need to discuss with the rest of the team about whether these are the appropriate dependency chains. does `smartsuspend` *need* to be a new, separate script, or can it be a mode of `unsuspend`? or should it just be called `suspend` and maybe take on more features about selecting jobs for suspension? Maybe we just need to rename `autounsuspend` to `suspendmanager` or something to make it clearer that it can now both suspend *and* unsuspend, as appropriate.

myk002:
ok, I wrote a plan at DFHack/dfhack#2955. Thoughts?

plule:
It makes sense to me, just like your other suggestions. I should be able to look at this again on Monday.

autounsuspend.lua
plule: I switched to triggering on job creation and completion, and it works as expected :)

gui/autounsuspend.lua
plule: oops forgot to adress this one

plule: And done this time

gui/control-panel.lua
```diff
@@ -30,6 +30,7 @@ local FORT_SERVICES = {
     'prioritize',
     'seedwatch',
     'starvingdead',
+    'suspendmanager',
```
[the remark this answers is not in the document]
plule: Sure, done (still a bit uneasy about disabling the previous stuff)

docs/suspendmanager.rst
```diff
@@ -0,0 +1,26 @@
...
+dwarves getting scared by wildlife. It also prevents construction jobs from
+blocking each others in situations such as the corner of a wall being constructed
+after the two adjacent walls.
+
+Usage
+-----
+
+::
+
+    enable suspendmanager
```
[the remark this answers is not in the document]
plule: `suspendmanager status` is now removed, it now under `suspendmanager`

suspendmanager.lua
```diff
@@ -0,0 +1,155 @@
...
+
+function set_prevent_blocking(enable)
+    prevent_blocking = enable
+    persist_state()
+end
+
+local function run_now()
+    if prevent_blocking then
+        dfhack.run_script('suspend', '--onlyblocking')
+        dfhack.run_script('unsuspend', '--quiet', '--skipblocking')
```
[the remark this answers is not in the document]
plule: Done, with a single loop. For the incremental mode the fact that `job.pos` behave weirdly gets in the way, maybe I'll try again to do it in a further PR, but I'm a bit afraid to introduce bugs right now.

plule:
@myk002 The latest push addresses all the remaining comments. I did a change from the initial plan, which is to move all the reusable bits in `suspendmanager` instead of having them scattered between `suspend` and `unsuspend`, as it was getting weird (`unsuspend` needed `suspend` to check for blocking jobs, and `suspend` could arguably use `unsuspend` for the underwater and buildingplan checks). It makes a nice pattern where `suspendmanager` is the main entrypoint with all these smart checks there, while `suspend` and `unsuspend` are really basic glue code for some of its features.

I was also able to confirm that it behaves correctly with `buildingplan` with the latest (and awesomest) DFHack release.

docs/unsuspend.rst
```diff
@@ -19,6 +19,15 @@ Usage
 
     unsuspend
 
+Options
```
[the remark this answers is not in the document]
plule: You have sharp eyes! Done

changelog.txt
```diff
@@ -15,6 +15,8 @@ that repo.
 
 ## New Scripts
 - `suspendmanager`: Automatic job suspension management (replaces `autounsuspend`)
+- `gui/suspendmanager`: Graphical configuration interface for `suspendmanager`
+- `suspend`: uspends building construction jobs
```
myk002, line 19:
```suggestion
- `suspend`: suspends building construction jobs
```

myk002 (approved):

PR #628, opened 2023-02-23
Refactor gui/dig a bit, support 'stamping' shapes and lines/curves

johncosker:
[Demo.webm](https://user-images.githubusercontent.com/4482627/220819399-771243e2-503d-44fd-b7ef-e6a26b20bc10.webm)

### Refactors `gui/dig`

**Points Driven**

Previously, shapes would be defined by the bounds of the 'view rectangle' marked by the user's mouse. Now, shapes are defined by points (currently just 2, although this paves the way for n-point shapes), and the view rectangle is defined by the bounds of the shape (and any extra points).

This allows for shapes 'bigger' than the rectangle defined by the initial points, which allows for thicc lines despite the user marking an e.g. 1x10 shape.

**Real Lines have Curves**

New shape type: Line Segment, also introduces the new concept for `gui/dig` "extra points", these points are for modifying the shape that's defined by the 'main' points. Currently this is used to define the 'bezier point' to make a curved line. This has a lot of potential for creating some really cool stuff.

**Center Point Dragging**

Now creates a special 'center' point that can be dragged to move the shape around,

**Also**

- Adjusted the non auto-commit behavior, now when the users Control-C's to place a shape, it leaves the shape so the user can drag and re-designate it, to quickly designate multiple areas of a given shape.

**Notes**

IMO there's still some grey area about what exactly should be in `dig.lua` vs. `shapes.lua`, so I'll keep thinking about that... I'd like `shapes.lua` to be able to be used in other scripts if possible. There's a lot in `dig.lua` that would potentially be useful elsewhere as well, e.g. drawing nice borders around designations, so it's still a little in flux.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -41,6 +41,7 @@ that repo.
 - `gui/gm-editor`: now supports multiple independent data inspection windows
 - `gui/gm-editor`: now prints out contents of coordinate vars instead of just the type
 - `rejuvenate`: now takes an --age parameter to choose a desired age.
+- `gui/dig` : Added 'Line' shape that also can draw curves, added draggable center handle
 
 # 50.05-alpha3.1
 
--- a/gui/dig.lua
+++ b/gui/dig.lua
@@ -8,6 +8,7 @@
 
 -- Should Haves
 -----------------------------
+-- Keyboard support
 -- As the number of shapes and designations grow it might be better to have list menus for them instead of cycle
 -- 3D shapes, would allow stuff like spiral staircases/minecart tracks and other neat stuff, probably not too hard
 -- Grid view without slowness (can ignore if next TODO is done, since nrmal mining mode has grid view)
@@ -46,6 +47,7 @@ local function get_dims(pos1, pos2)
     math.abs(pos1.x - pos2.x) + 1,
         math.abs(pos1.y - pos2.y) + 1,
         math.abs(pos1.z - pos2.z) + 1
+
     return width, height, depth
 end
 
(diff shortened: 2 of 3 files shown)


myk002:
I have the perfect second client for `shapes.lua`: `buildingplan`. If a player wants to dig in shapes, there's reason they might want to build in those shapes too. I'm not entirely sure how to handle the UI, though.

Let me finish this first iteration of `buildingplan` with its can-do "I make hollow squares" attitude and maybe we can look into hooking these two fantastic tools together after that.

You can check out the in-progress `buildingplan` at DFHack/dfhack#2882

johncosker:
Yeah I actually was thinking that too, after getting this wrapped up I was going to create a proof-of-concept `gui/dig` mode to create a buildings with outer walls, and floors filling the interior, and optional 'roof' floors using quickfort/building plan.

The new 'stamp' commit behavior would allow creating multi floor building layouts really fast.

Obviously though that would just be a demo as `/gui/dig` should stick to digging and also I wouldn't go as far as putting in anything for selecting materials.

johncosker:
As it stands though there would need to be a lot of copying from `dig.lua` to any other script wanting to use `shapes.lua`, hopefully we can encapsulate more of that in `shapes.lua` or another script to reduce code duplication.

myk002:
since `gui/dig` uses quickfort to actually apply the blueprint, maybe we're doing it a disservice by calling it `gui/dig`? it just a slight configuration tweak away from being `gui/build`. Maybe we should call it `gui/layout` or `gui/designer` and make it configurable whether you want to dig or build (or place stockpiles or zones)? is it better to have a single-purpose tool or a unified interface for applying shapes to your designs?

myk002 (approved):

johncosker:
It would certainly be easier overall to just rename dig.lua and keep using it for building vs. refactoring it. and creating separate tools.

It depends on what else might want to use shapes, the designation border drawing logic, etc...

myk002:
I can't think of anything else offhand.. The other design tools are [here](https://docs.dfhack.org/en/latest/design-tag-index.html) and none really seem relevant. What you are doing here is essentially the start of what I put as a strategic goal for the DFHack project to provide better design tools. You are filling a sorely needed gap.

johncosker:
So, say we rename `gui/dig` to `gui/designer` do we care about supporting anyone that already has `gui/dig` on a hotkey and might be confused if it goes away if they haven't been following the changelogs?

Maybe we could mv `dig.lua` -> `designer.lua` and then create an alias gui/dig -> gui/designer? I haven't used aliases much but I think they're a thing right?

myk002:
if we can't do it easily already, we can extend aliases so it's easy. we're having a similar conversation right now about renaming `autounsuspend` to `suspendmanager` since it's taking on intelligent suspending in addition to unsuspending.

PR #629, opened 2023-02-24
New script: diplo - Change civ relations conveniently

Gnurro:
Added a script to conveniently change diplomatic relations.
Console only for now.
Also my first lua, so it probably needs improvement.


diplo.lua
```diff
@@ -0,0 +1,88 @@
...
+
+-- get a very rough, but readable name for a civ:
+function get_raw_name(civ)
+    local raw_name = ""
+    for _, name_word in pairs(civ.name.words) do
+        if name_word ~= -1 then
+            raw_name = raw_name .. " " .. df.global.world.raws.language.words[name_word].word
+        end
+    end
+    raw_name = string.sub(raw_name, 2)
```
myk002, line 28: what does this chop off?

diplo.lua
```diff
@@ -0,0 +1,88 @@
...
+    return raw_name
+end
+
+-- if no civ ID is entered, just output list of civs:
+if not args[1] then
+    goto outputlist
```
myk002, line 34: though this works, I'd prefer if you made outputlist a function and called it (with a `return` statement afterwards to halt the script)

diplo.lua
```diff
@@ -0,0 +1,88 @@
...
+end
+
+-- make sure that there is a relation to change to:
+if not args[2] then qerror("Missing relation!") end
```
myk002, line 38: you can output the script help here (in addition to the `qerror`) via `print(dfhack.script_help())`

diplo.lua
```diff
@@ -0,0 +1,88 @@
+--[[
+diplo - Quick "Diplomacy"
+Without arguments:
+    Print list of civs the player civ has diplomatic relations to.
+    This list might not match the ingame civ list, but is what matters. Does not include 'no contact' civs, etc.
+    Also shows if relations match between player civ and other civ - a mismatch can potentially happen 'naturally', but is important when trying to make peace proper, as both civs need to be at peace with the other.
+With arguments:
+    diplo CIV_ID RELATION
+    Changes the relation to the civ identified by CIV_ID to the one specified in RELATION, making sure that this is mutual.
+    Relations: 0 = Peace; 1 = War; 3 = Alliance (seems to only apply to site govs...?)
```
myk002, line 10: consider allowing the player to specify (case-insensitive) words instead of numbers

myk002, line 10: if we don't know the use of Alliance, I'd omit it from the docs (though maybe leave a comment about it in the code)

diplo.lua
```diff
@@ -0,0 +1,88 @@
...
+        matched,
+        get_raw_name(cur_civ)
+        })
+    end
+end
+
+print(([[%4s %12s %8s %20s]]):format("ID", "Relation", "Matched", "Name"))
+for _, civ in pairs(civ_list) do
+    print(([[%4s %12s %8s %20s]]):format(civ[1], civ[2], civ[3], civ[4]))
+end
```
myk002, line 88: if there is nothing in the list, you should print out "No contacted civs!" or something instead of just an isolated header. You can check for emptiness with `if not next(civ_list)`

diplo.lua
```diff
@@ -0,0 +1,88 @@
+--[[
```
myk002, line 1: docs should go in .rst files in the `docs/` subdirectory. You can copy one of the other files to get the format, and read more context on documentation standards here: https://docs.dfhack.org/en/latest/docs/dev/Documentation.html

myk002: This is mostly about formatting. The parameters you have defined are fine.

diplo.lua
```diff
@@ -0,0 +1,88 @@
+--[[
+diplo - Quick "Diplomacy"
+Without arguments:
+    Print list of civs the player civ has diplomatic relations to.
+    This list might not match the ingame civ list, but is what matters. Does not include 'no contact' civs, etc.
+    Also shows if relations match between player civ and other civ - a mismatch can potentially happen 'naturally', but is important when trying to make peace proper, as both civs need to be at peace with the other.
+With arguments:
+    diplo CIV_ID RELATION
```
myk002, line 8: consider allowing the keyword `all` instead of a `<CIV_ID>` so players can choose harmony or king of the hill style gameplay

diplo.lua
```diff
@@ -0,0 +1,88 @@
+--[[
+diplo - Quick "Diplomacy"
```
myk002, line 2: @lethosor is our resident naming expert. I'll let him weigh in on whether this is an appropriate name.

diplo.lua
Gnurro: Leading whitespace on the very rudimentarily built names. Is there a dfhack utility that returns the properly inflected/built names? Currently they come out of this like 'COOPERATE CHANNEL' for 'The Cooperative Channel'.

diplo.lua
Gnurro: Alright, I wasn't sure how viable gotos are in lua, but I can sure wrap that in a function.

diplo.lua
Gnurro: I'll add that. Do I need to somehoe 'import' utils for that to work...?

diplo.lua
Gnurro: Should be easy enough to add in form of another function that maps 'war'/'peace' etc to the numbers. I just didn't add it since it feels inconsistent to have that but no recognition of civ names, for which a proper conversion of the names would be needed.

Gnurro: Well, I'd have to do a lot more in-depth research to figure out how alliances and site govs interact with this, but since I removed the site gov listing/accessability it can go. War/peace is *supposed* to be at civ level, but I'm not sure, specially due to no-civ site gov tributes being possible...

diplo.lua
Gnurro: Good idea, I'll add that just for code sanity. List will never be empty, as at least the home civ will be in there.

diplo.lua
Gnurro: I'll see if I can figure that out.

diplo.lua
Gnurro: I might add that, though I think that'd need some more convoluted code. Probably need a more or less static listing of civ ids to iterate over for 'all'.

diplo.lua
Gnurro: I couldn't decide between 'diplomat', 'diplomacy' or the like, so i just opted for something short. Name doesn't really matter as long as it stays at a convenionet length to type, imo.

diplo.lua
myk002: I'd bring that question to the DFHack discord. More people are familiar with languages handling there

diplo.lua
myk002: they just have the ability to create very confusing code. In DFHack, we try to keep the use of `goto` confined to the `continue` use case. e.g.:
```lua
for k,v in pairs(t) do
    if not condition(v) then goto continue end
    -- ... logic
    ::continue::
end
```

diplo.lua
myk002: it's a global dfhack function. nothing to import

diplo.lua
myk002: I thought it would be a loop similar to the current listing function. is there more to it than thtat?

diplo.lua
Gnurro: Using dfhack.TranslateName now.

diplo.lua
Gnurro: Added it.

diplo.lua
Gnurro: Well, basically it is - but just copying the loop with slight changes feels redundant. Might be better to wrap more things in functions and such. I'll see how it goes when I get back to it.

diplo.lua
myk002: even length doesn't matter as much as it used to now that `gui/launcher` has autocomplete.

diplo.lua
Gnurro: Refactored script to use functions, `goto` removed.

diplo.lua
Gnurro: Relation argument parsing function is now used for this.

Gnurro: Removed mentions of alliance - further research needed to properly handle this relation.

diplo.lua
Gnurro: `all` argument can now be used, looping over the civ_list for civ ids.

diplo.lua
Gnurro: Added empty list handling for code sanity, with a warning that something is majorly wrong if the list is actually empty.

diplo.lua
```diff
@@ -0,0 +1,128 @@
+@@ -0,0 +1,79 @@
```
myk002, line 1: this looks like merge error from diff text

diplo.lua
```diff
@@ -0,0 +1,128 @@
...
+                if entity2.group_id == p_civ_id and entity2.relation == entity.relation then
+                    matched = "Yes"
+                end
+            end
+            table.insert(civ_list, {
+            cur_civ_id,
+            rel_str,
+            matched,
+            dfhack.TranslateName(cur_civ.name, true)
+            })
```
myk002, line 47: it is difficult to remember an arbitrary order of values when you come back and modify this script later. it is much more programmer-friendly to give them names, e.g.:
```lua
table.insert(civ_list, {
    civ_id=cur_civ_id,
    relation=entity.relation,
    matched=matched,
    name=dfhack.TranslateName(cur_civ.name, true),
})

File: diplo.lua [hunk 11]
```diff
@@ -0,0 +1,128 @@
...
+function output_civ_list()
+    local civ_list = get_civ_list()
+    if not next(civ_list) then
+        print("Your civilisation has no diplomatic relations! This means something is going wrong, as it should have at least a relation to itself.")
+    else
+        print(([[%4s %12s %8s %30s]]):format("ID", "Relation", "Matched", "Name"))
+        for _, civ in pairs(civ_list) do
+            print(([[%4s %12s %8s %30s]]):format(civ[1], civ[2], civ[3], civ[4]))
+        end
+    end
```
myk002, line 63:
```suggestion
    if not next(civ_list) then
        print("Your civilisation has no diplomatic relations! This means something is going wrong, as it should have at least a relation to itself.")
        return
    end
    print(([[%4s %12s %8s %30s]]):format("ID", "Relation", "Matched", "Name"))
    for _, civ in pairs(civ_list) do
        print(([[%4s %12s %8s %30s]]):format(civ[1], civ[2], civ[3], civ[4]))
    end
```

myk002: no need for an indented else clause if the if clause can return

diplo.lua
```diff
@@ -0,0 +1,128 @@
...
+-- make sure that there is a relation to change to:
+if not args[2] then
+    print(dfhack.script_help())
+    qerror("Missing relation!")
+end
+
+-- change relation(s) according to args:
+handle_all(args[1], relation_parse(args[2]))
+output_civ_list()
+return
```
myk002, line 128: final return statements are not needed; lua adds an implicit `return nil` to the end of scripts/functions

diplo.lua
lethosor: Sorry for the delay - `diplomacy` is my suggestion from a quick read-through of the script.

myk002:
@Gnurro Have you had a chance to make progress on this? It's a great script that I'd like to get into the next beta release, if possible

Gnurro:
> 

Well, the script itself is done - I don't find the time to figure out the RST doc stuff, though.

diplo.lua
Gnurro: Yep. Pushing things by hand can get messy.

diplo.lua
Gnurro: I dislike differently indented alternatives, but if this is the intended code style for DFHack, go with that, I guess.

lethosor: I don't think we have a specific style for this. I personally tend slightly towards matched indentation for similar branches. I'd lean towards considering the first one here (under `if`) an early exit instead, and adding the early `return`, but ultimately I think it doesn't really need to be changed.

myk002:
I'll see if I can get this PR fixed up for merge.

PR #630, opened 2023-02-25
Update dig to allow dropping an extra point before placing second point.

johncosker:
Demo3

A reddit comment got me thinking and I realized a nice UX feature is being able to use the `v` shortcut to drop an extra point for a curve whil placing the second point of a line. 

It actually works really well and I think is really intuitive.

### DIFF
--- a/gui/dig.lua
+++ b/gui/dig.lua
@@ -247,8 +247,13 @@ function GenericOptionsPanel:init()
             end,
             show_tooltip = true,
             on_activate = function()
-                self.dig_panel.placing_extra.active = true
-                self.dig_panel.placing_extra.index = #self.dig_panel.extra_points + 1
+                if self.dig_panel.mark1 and self.dig_panel.mark2 then
+                    self.dig_panel.placing_extra.active = true
+                    self.dig_panel.placing_extra.index = #self.dig_panel.extra_points + 1
+                elseif self.dig_panel.mark1 and not self.dig_panel.mark2 then
+                    local mouse_pos = dfhack.gui.getMousePos()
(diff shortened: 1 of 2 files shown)


gui/dig.lua
```diff
@@ -247,8 +247,13 @@ function GenericOptionsPanel:init()
             end,
             show_tooltip = true,
             on_activate = function()
-                self.dig_panel.placing_extra.active = true
-                self.dig_panel.placing_extra.index = #self.dig_panel.extra_points + 1
+                if self.dig_panel.mark1 and self.dig_panel.mark2 then
+                    self.dig_panel.placing_extra.active = true
+                    self.dig_panel.placing_extra.index = #self.dig_panel.extra_points + 1
+                elseif self.dig_panel.mark1 and not self.dig_panel.mark2 then
+                    local mouse_pos = dfhack.gui.getMousePos()
```
myk002, line 254: be sure to check mouse_pos for nil if the player hits `v` while the mouse is out of bounds

gui/dig.lua
```diff
@@ -247,8 +247,13 @@ function GenericOptionsPanel:init()
...
+                    self.dig_panel.extra_points[#self.dig_panel.extra_points + 1] = { x = mouse_pos.x, y = mouse_pos.y }
```
myk002, line 255:
```suggestion
                    table.insert(self.dig_panel.extra_points, { x = mouse_pos.x, y = mouse_pos.y })
```

myk002: not a required change, but slightly more concise

myk002 (commented):
changelog entry plz

johncosker:
@myk002 would this be a dev changelog or normal?

myk002:
Dev changelog, since it's changing something that hasn't seen a "full" release yet

myk002 (approved):

PR #631, opened 2023-02-25
gui/create-item: add whole corpsepiece spawning

ppaawwll:
since i took a hiatus for two weeks, i updated my branch to the latest and re-added the changed script, adding whole corpsepiece spawning, lowercase entries for the creature tissue layer and creature material prompts for ease of searching.


changelog.txt
```diff
@@ -34,6 +34,7 @@ that repo.
 - `devel/query`: can now properly index vectors in the --table argument
 -@ `forbid`: fix detection of unreachable items for items in containers
 -@ `unforbid`: fix detection of unreachable items for items in containers
+-@ `gui/create-item`: adding corpse spawning broke spawning generic corpsepieces, but has now been fixed.
```
myk002, line 37: 50-07-alpha2 has already been released these entries should go in the "Future" section above.

gui/create-item.lua
```diff
@@ -470,7 +482,7 @@ function hackWish(unit)
       if itemtype ~= df.item_type.CORPSEPIECE and itemtype ~= df.item_type.CORPSE then
        createItem({mattype,matindex},{itemtype,itemsubtype},quality,unit,description,1)
       else
-       createCorpsePiece(unit,bodypart,partlayerID-1,mattype,matindex,corpsepieceGeneric,quality)
+       createCorpsePiece(unit,bodypart,partlayerID-2,mattype,matindex,corpsepieceGeneric,quality)
```
myk002, line 485: i get nervous when I see a `-2` here from the issues we had with it before. we're sure it's correct now?

gui/create-item.lua
```diff
@@ -36,10 +36,10 @@ function getCreaturePartList(creatureID, casteID)
 end
 
 function getCreaturePartLayerList(creatureID, casteID, partID)
- local crplList={}
+ local crplList={{"whole"}}
  for k,crpl in ipairs(df.global.world.raws.creatures.all[creatureID].caste[casteID].body_info.body_parts[partID].layers) do
   local str = crpl.layer_name
-  table.insert(crplList,{str})
+  table.insert(crplList,{string.lower(str)}) -- lowercase string for easier searching
```
myk002, line 42: let's actually not do this here. we have a pending action to make FilteredList widgets search case insensitive by default. Let's not lose the displayed capitalization in these prompts.

gui/create-item.lua
ppaawwll: there's an extra entry in the part layer prompt now so i have to offset it by two, plus it's always set so initializing it to -1 won't do anything. i _could_ do the same thing i did with generic body parts, and set a flag when the user chooses the entry in the tissue layer selection screen (which is "whole") but i feel like i want to avoid variable bloat.

gui/create-item.lua
ppaawwll: alright

changelog.txt
ppaawwll: oop

gui/create-item.lua
ppaawwll: or have we been adding variables all over the place already?

myk002: no, this is fine. as long as you have tested the various code paths that lead here.

ppaawwll: tested corpsepiece (single layer), corpsepiece (generic), corpsepiece (whole), and corpse, which don't seem to have any problems.

changelog.txt
```diff
@@ -17,8 +17,11 @@ that repo.
 
 ## Fixes
 -@ `gui/dig`: Allow placing an extra point (curve) while still placing thesecond main point
+-@ `gui/create-item`: adding corpse spawning broke spawning generic corpsepieces, but has now been fixed.
```
myk002, line 20:
```suggestion
-@ `gui/create-item`: fix generic corpsepiece spawning
```

changelog.txt
```diff
@@ -17,8 +17,11 @@ that repo.
...
 
 ## Misc Improvements
+- `gui/create-item`: added ability to spawn 'whole' corpsepieces (every layer of a part).
+- `gui/create-item`: changed creature tissue layer and creature material prompts to have lowercase entries for ease of searching.
```
myk002, line 24:
```suggestion
- `gui/create-item`: added ability to spawn 'whole' corpsepieces (every layer of a part)
```

changelog.txt
```diff
@@ -43,7 +46,7 @@ that repo.
 - `devel/visualize-structure`: now automatically inspects the contents of most pointer fields, rather than inspecting the pointers themselves
 - `devel/query`: will now search for jobs at the map coordinate highlighted, if no explicit job is highlighted and there is a map tile highlighted
 - `caravan`: add trade screen overlay that assists with seleting groups of items and collapsing groups in the UI
-- `gui/gm-editor` will now inspect a selected building itself if the building has no current jobs
+- `gui/gm-editor` will now inspect a selected building itself if the building has no current jobs.
```
myk002, line 49:
```suggestion
- `gui/gm-editor` will now inspect a selected building itself if the building has no current jobs
```

myk002 (approved):

PR #632, opened 2023-02-26
Update devel/export-dt-ini for DT 42.0.0

cvuchener:
I have an issue with getting offsets for item_crafted ("Cannot allocate item_crafted"). I had to replace it with a derived type (item_constructed).

@myk002 Thanks for updating `stress` (5bdfd379b) but don't change the first argument, it is not a DFHack name.


ab9rf:
`item_craftedst` is a pure virtual class and therefore should never be instantiated. `item_constructedst` is also a pure virtual class and thus also should never be instantiated, but apparently you can get away with it for some reason

cvuchener:
In df-structures symbols.xml `item_constructedst` has a vtable-address but `item_craftedst` doesn't. It used to have one in older versions.

ab9rf:
> In df-structures symbols.xml `item_constructedst` has a vtable-address but `item_craftedst` doesn't. It used to have one in older versions.

that's because `item_craftedst` doesn't have a vtable. since it's an abstract class with no vmethod implementations that differ from its parent class which can be called in a dynamic context (that is, a context in which the compiler cannot determine at compile time which implementation to use), the compiler has no need to generate a vtable for it. there are actually two `item_craftedst`-specific vmethod implementation (`read_file` and `write_file`), but they can only be accessed using an explicit type specifier which is always resolvable at compile time. as an abstract class, instances of that class cannot be instantiated, and thus no object can legally exist having a vtable pointer for this class. since the vtable was unneeded, the compiler (or possibly the linker) optimized away the vtable.

cvuchener:
https://github.com/DFHack/dfhack/blob/develop/library/include/DataDefs.h#L383 `can_allocate` does check for a vtable

myk002:
> @myk002 Thanks for updating `stress` ([5bdfd37](https://github.com/DFHack/scripts/commit/5bdfd379b2d13f56bab455f94af56ec0ce3e0cb1)) but don't change the first argument, it is not a DFHack name.

Ah, my mistake. Sorry about that!

devel/export-dt-ini.lua
```diff
@@ -129,6 +129,7 @@ address('historical_figures_vector',globals,'world','history','figures')
 address('world_site_type',df.world_site,'type')
 address('active_sites_vector',df.world_data,'active_site')
 address('gview',globals,'gview')
+address('external_flag',globals,'game_extra','external_flag')
```
myk002, line 132: this *may* change again in the future when/if the `gamest` padding gets into the base game.

ab9rf: i'm going to be reluctant to rejoin the two halves of `game`; while I think Putnam has the best of intentions with her padding strategy, all it takes is for Toady to decide that another field is needed for some reason for the whole thing to get thrown out of whack again, and I judge that highly likely to happen at some time in the future.

PR #633, opened 2023-02-28
More gui/dig refactoring. Allow freeform n-point shapes. Flipping/mirroring

johncosker:
freeform_demo

Gif got a little cut off but it's mostly in there lol

### N-point shapes

Got rid of `mark1` and `mark2` and refactored to use a list of marks to support shapes that require more than setting basically just 2 corners.

I added a `basic_shape` field to the `Shape` inheritance structure to make handling those shapes easier and be able to have complex shapes be handled a little differently in terms of how the UI works.

"Basic Shapes" are those like Rectangle, Ellipse, Rows, etc... that are fully bounded by rectangles defined by the first and second points and therefore we want to add stuff like drag points to all corners not just the two that the user placed.

Non-basic shapes like the Line and FreeForm aren't necessarily bounded the same way and we only want to place drag points on actual points.

### Also added flipping/mirroring of shapes:

flipping_demo


gui/dig.lua
```diff
@@ -209,6 +211,47 @@ function GenericOptionsPanel:init()
...
+                        on_activate=self.dig_panel:callback('on_transform', 'flipv'),
+                    },
+                    widgets.HotkeyLabel {
+                        key = 'STRING_A061',
+                        frame = { t = 1, l = 4 }, key_sep = ':',
+                        on_activate=self.dig_panel:callback('on_transform', 'fliph'),
+                    },
+                    widgets.WrappedLabel {
+                        frame = { t = 1, l = 7 },
+                        text_to_wrap = function() return 'Transformation directions' end
```
myk002, line 250: this looks like a stub. perhaps the "transform reduction" code in gui/quickfort would be useful to reuse here?

johncosker: Maybe i could put "`(): Rotate clockwise/counter-clockwise`" and `_=: Flip horizontally/vertically` and split them into two lines.

myk002 (approved):

PR #634, opened 2023-03-08
Dig mirror mode, S-curves, various fixes

johncosker:
### Added option to place second bezier point for curves
s_curve

### Add option to set a 'mirror point' when drawing freeform shapes
Allows setting a point and selecting mirror directions (vertical, horizontal, diagonal). Points will be mirrored across the planes defined by the point. The user can choose to 'save' the mirrored points which makes them into regular draggable points that can be transformed, moved, etc...

This works pretty well, but the interface is maybe a ltitle clunky. Should look in to UI improvements soon.
freeform_mirror

### Symmetry fixes
- `Line` shapes have been updated to draw the line both ways, this will guarantee creating symmetrical lines in cases where the Bresenham's algorithm has to 'choose' tiles arbitrarily when tie breaking.
- Updated bezier algorithm to fix rounding errors that created asymmetric curves

### Also
- Changed 'Transform' hotkey to `Y` (conflicted with thickness)
- Added optional guide line crosshair for mouse pos to help line things up

### DIFF
--- a/gui/dig.lua
+++ b/gui/dig.lua
@@ -5,6 +5,9 @@
 
 -- Must Haves
 -----------------------------
+-- Cubic bezier algo
+-- freeform mirror shape
+-- Guidelines
 
 -- Should Haves
 -----------------------------
@@ -1200,7 +1203,6 @@ function Dig:onRenderFrame(dc, rect)
         return self:get_pen(pos.x, pos.y, mouse_pos)
     end
 
-    -- if self.placing_mark.active then
     if self.placing_mark.active and self.placing_mark.index ~= nil then
         self.marks[self.placing_mark.index] = mouse_pos
     end
@@ -1501,6 +1503,7 @@ function Dig:commit()
         self.extra_points = {}
         self.prev_center = nil
     end
+
     self:updateLayout()
 end
 
--- a/gui/dig.lua
+++ b/gui/dig.lua
@@ -39,6 +39,93 @@ local tile_attrs = df.tiletype.attrs
 
 local to_pen = dfhack.pen.parse
 
+SHOW_DEBUG_WINDOW = true
+
+function tableToString(tbl, indent)
+    indent = indent or ""
+    local result = {}
+    for k, v in pairs(tbl) do
+        local key = type(k) == "number" and "[" .. tostring(k) .. "]" or tostring(k)
+        if type(v) == "table" then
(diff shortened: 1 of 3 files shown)


internal/dig/shapes.lua
```diff
@@ -88,57 +88,43 @@ end
...
+    if mirror_point then
+        min_x, min_y, max_x, max_y = update_minmax(mirror_point, min_x, min_y, max_x, max_y)
+    end
+
     return { x = min_x, y = min_y }, { x = max_x, y = max_y }
 end
 
 function Shape:points_to_string(points)
-    local points = points == nil and self.points or points
+    local points = (not points) and self.points or points
```
myk002, line 127:
```suggestion
    points = points or self.points
```

gui/dig.lua
```diff
@@ -5,18 +5,22 @@
 
 -- Must Haves
 -----------------------------
+-- Better UI, it's starting to get really crowded
```
myk002, line 8: tap TaxiService for the design. I can point you to examples for the implementation.

gui/dig.lua
```diff
@@ -1317,23 +1676,104 @@ function Dig:commit()
...
+-- DigScreen
 --
 
 DigScreen = defclass(DigScreen, gui.ZScreen)
 DigScreen.ATTRS {
     focus_path = "dig",
     pass_pause = true,
     pass_movement_keys = true,
+    dig_window = DEFAULT_NIL,
+    debug_window = DEFAULT_NIL
```
myk002, line 1765: if they're not being set by a parent, these don't need to be class attributes. they can be initialized as self.dig_window and self.debug_window in init()

myk002:
could I possibly get a video file (the original screencap, not the processed gif) of the symmetry demo to put on our steam page?

johncosker:
I recorded it using a tool that just processes straight to gifs (ScreenToGif) so I don't have an original video file.

Tomorrow I can do some recording of that and other gui/dig features. I could try to make a montage out of them, too, if I'm feeling ambitious.

johncosker:
@myk002 do you need these any time soon? Idk if I'll get to them tonight

myk002:
> @myk002 do you need these any time soon? Idk if I'll get to them tonight

no, no hurry. whenever you get it done is fine. we can add it to the storefront any time. the steam page isn't even live yet.

myk002 (approved):

PR #635, opened 2023-03-08
Create fix/stuck-instruments

Stealthii:
fix/stuck-instruments is a fixer script for instruments that are permanently stuck in an activity job, usually caused by [DF bug #9485](https://www.bay12games.com/dwarves/mantisbt/view.php?id=9485) which results in instruments being picked up for a performance, but are instead simulated and never get disassociated with the performance activity or have their in_job flag removed.


ab9rf:
needs documentation

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,65 @@
...
+simulated and are now stuck permanently in a job that no longer exists.
+
+This works around the issue encountered with :bug:`9485`, and should be run
+if you notice any instruments lying on the ground that seem to be stuck in a
+job.
+
+Run ``fix/stuck-instruments -n`` or ``fix/stuck-instruments --dry-run`` to
+list how many instruments would be fixed without performing the action.
+
+]====]
```
myk002, line 18: help text should go in docs/fix/stuck-instruments.rst. Check out other files in that directory for the documentation structure. Also more info here: https://docs.dfhack.org/en/latest/docs/dev/Documentation.html

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,65 @@
...
+    if fixed > 0 or dry_run then
+        print(("%s %d stuck instruments."):format(
+                dry_run and "Found" or "Fixed",
+                fixed
+        ))
+    end
+end
+
+
+if not dfhack_flags.module then
```
myk002, line 63: since this file doesn't have `--@ module = true` at the top, this check doesn't do anything and can be removed.

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,65 @@
...
+
+]====]
+
+
+function fixInstruments(args)
+    local dry_run = false
+    local fixed = 0
+    for _, arg in pairs(args) do
+        if args[1]:match('-h') or args[1]:match('help') then
+            print(help)
```
myk002, line 26:
```suggestion
            print(dfhack.script_help())
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,65 @@
...
+list how many instruments would be fixed without performing the action.
+
+]====]
+
+
+function fixInstruments(args)
+    local dry_run = false
+    local fixed = 0
+    for _, arg in pairs(args) do
+        if args[1]:match('-h') or args[1]:match('help') then
```
myk002, line 25: use the `argparse` module for commandline parsing instead of rolling your own. this helps the tools react consistently to user input. Example here: <https://github.com/DFHack/scripts/blob/master/prioritize.lua#L479> and documentation here: <https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#argparse>

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,65 @@
...
+            print(help)
+            return
+        elseif args[1]:match('-n') or args[1]:match('dry') then
+            dry_run = true
+        end
+    end
+    for _, item in ipairs(df.global.world.items.all) do
+        if item:getType() == df.item_type.INSTRUMENT then
```
myk002, line 33: instead of putting all the logic in an if block, consider using the lua version of `continue`:
```lua
for _,item in ipairs(df.global.world.items.all) do
    if item:getType() ~= df.item_type.INSTRUMENT then
        goto continue
    end
    -- ... other logic
    ::continue::
end

Comment (myk002) [reply fix/stuck-instruments.lua:33]: however, this particular suggestion is moot since you could iterate through `df.global.world.items.other.INSTRUMENT` and get the items you want directly. There is also `df.global.world.items.other.INSTRUMENT_STATIONARY`, but you're not interested in stationary instruments, right?

Comment (myk002) [reply fix/stuck-instruments.lua:33]: oh, you can use the `::continue::` idiom below where you check for the general ref type

File: fix/stuck-instruments.lua [hunk 6]
```diff
@@ -0,0 +1,65 @@
...
+if you notice any instruments lying on the ground that seem to be stuck in a
+job.
+
+Run ``fix/stuck-instruments -n`` or ``fix/stuck-instruments --dry-run`` to
+list how many instruments would be fixed without performing the action.
+
+]====]
+
+
+function fixInstruments(args)
```
myk002, line 21: this looks like this script could be run periodically to keep the fort healthy. Would it be a good idea to add an entry to the `REPEATS` table in `gui/control-panel`? You could add that to this PR if you agree.

myk002:
@Stealthii Have you had a chance to make progress on this? I've seen others complaining of the same problem, and it seems like we should be able to offer this fix

Stealthii:
Hi @myk002 - apologies for the late response to these suggestions. I got halfway through making changes to support these suggestions, and just need to finish option parsing and conditional logic changes. I'll update this PR later today.

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    for _, item in ipairs(df.global.world.items.other.INSTRUMENT) do
+        for i, ref in pairs(item.general_refs) do
+            if ref:getType() == df.general_ref_type.ACTIVITY_EVENT then
+                local activity = df.activity_entry.find(ref.activity_id)
+                if not activity then
+                    if not dry_run then
+                        --remove dead activity reference
+                        item.general_refs:erase(i)
```
myk002, line 40: I believe you have to free the reference as well to prevent a memory leak

myk002 (commented):
could you also add a line in `changelog.txt` announcing this script in `New scripts`?

gui/control-panel.lua
```diff
@@ -97,6 +97,9 @@ local REPEATS = {
     ['combine']={
         desc='Combine partial stacks in stockpiles into full stacks.',
         command={'--time', '7', '--timeUnits', 'days', '--command', '[', 'combine', 'all', '-q', ']'}},
+    ['fixInstruments']={
```
myk002, line 100:
```suggestion
    ['stuck-instruments']={
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,64 @@
...
+                dry_run and "Found" or "Fixed",
+                fixed
+        ))
+    end
+end
+
+
+if not dfhack_flags.module then
+    fixInstruments{...}
+end
```
myk002, line 64:
```suggestion
local opts = {}

local positionals = argparse.processArgsGetopt({...}, {
    { 'h', 'help', handler = function() options.help = true end },
    { 'n', 'dry-run', handler = function() options.dry_run = true end },
})

if positionals[1] == 'help' or opts.help then
    print(dfhack.script_help())
    return
end

fixInstruments(opts)
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,64 @@
...
+This works around the issue encountered with :bug:`9485`, and should be run
+if you notice any instruments lying on the ground that seem to be stuck in a
+job.
+
+Run ``fix/stuck-instruments -n`` or ``fix/stuck-instruments --dry-run`` to
+list how many instruments would be fixed without performing the action.
+
+]====]
+
+
```
myk002, line 20:
```suggestion
local argparse = require('argparse')

```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,64 @@
...
+    local dry_run = false
+    local fixed = 0
+    for _, arg in pairs(args) do
+        if args[1]:match('-h') or args[1]:match('help') then
+            print(dfhack.script_help())
+            return
+        elseif args[1]:match('-n') or args[1]:match('dry') then
+            dry_run = true
+        end
+    end
```
myk002, line 31:
```suggestion
function fixInstruments(opts)
    local fixed = 0
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,64 @@
...
+
+    for _, item in ipairs(df.global.world.items.other.INSTRUMENT) do
+        for i, ref in pairs(item.general_refs) do
+            if ref:getType() == df.general_ref_type.ACTIVITY_EVENT then
+                local activity = df.activity_entry.find(ref.activity_id)
+                if not activity then
+                    if not dry_run then
```
myk002, line 38:
```suggestion
                    if not opts.dry_run then
```

fix/stuck-instruments.lua
myk002, line 40:
```suggestion
                        item.general_refs[i]:delete()
                        item.general_refs:erase(i)
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,64 @@
...
+                        --remove dead activity reference
+                        item.general_refs:erase(i)
+                        if item.flags.in_job then
+                            --remove stuck in_job flag if true
+                            item.flags.in_job = false
+                        end
```
myk002, line 44:
```suggestion
                        item.flags.in_job = false
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,64 @@
...
+                    end
+                    fixed = fixed + 1
+                    break
+                end
+            end
+        end
+    end
+
+    if fixed > 0 or dry_run then
```
myk002, line 53:
```suggestion
    if fixed > 0 or opts.dry_run then
```

myk002 (approved):

myk002 (approved):

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,46 @@
...
+                fixed
+        ))
+    end
+end
+
+local opts = {}
+
+local positionals = argparse.processArgsGetopt({...}, {
+    { 'h', 'help', handler = function() options.help = true end },
+    { 'n', 'dry-run', handler = function() options.dry_run = true end },
```
myk002, line 38:
```suggestion
    { 'h', 'help', handler = function() opts.help = true end },
    { 'n', 'dry-run', handler = function() opts.dry_run = true end },
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,46 @@
...
+                    fixed = fixed + 1
+                    break
+                end
+            end
+        end
+    end
+
+    if fixed > 0 or opts.dry_run then
+        print(("%s %d stuck instruments."):format(
+                dry_run and "Found" or "Fixed",
```
myk002, line 28:
```suggestion
                opts.dry_run and "Found" or "Fixed",
```

fix/stuck-instruments.lua
```diff
@@ -0,0 +1,46 @@
...
+                    end
+                    fixed = fixed + 1
+                    break
+                end
+            end
+        end
+    end
+
+    if fixed > 0 or opts.dry_run then
+        print(("%s %d stuck instruments."):format(
```
myk002, line 27:
```suggestion
        print(("%s %d stuck instrument(s)."):format(
```

PR #637, opened 2023-03-12
add gui/biomes: a way to see where one biome ends and another begins

TymurGubayev:
image

* selected biome is highlighted on the map
* the "tooltip" window under the legend is shown on mouse-hover


TymurGubayev:
This is not quite finished (see tasks), but I'll need help with them.

1. I can't find a way to change the color of the text on the map. Is there one? I want to (subtly) highlight map tiles of the selected biome.
2. Is there a way to get the minimap coordinates? I want to place the legend under it. If not, I'll probably move it into the bottom right corner.

myk002:
long time no see! welcome back!

I have not figured out a way to change the color of text being applied to map tiles. It *might* be possible, but I think you'll be better off using the "on/off" icons that TaxiService provided us. run `gui/pathable` to see what they look like, and see `plugins/pathable.cpp` for how they're retrieved from the Textures module. Also see `gui/control-panel.lua` for an example of getting textures in Lua (though there they're applied to the UI grid and not the map grid). We might be able to provide more colors and shapes, but it might be enough to just have the "on" icon for the currently selected biome and the "off" icon for everything else.

the minimap is always in the upper right corner and is always the same size. You can just use a frame attached to the right edge at a constant offset from the top and you'll be fine. `gui/quickfort`, for example, has a frame like this to appear just below the minimap.

I've found in the past that overlaying the entire visible map from Lua can be slow. If performance is an issue here, feel free to extend pathable.cpp with another API endpoint for this tool. Much of the logic you need is already there.

gui/biomes.lua
```diff
@@ -0,0 +1,213 @@
+-- Show biomes on the map.
+
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local guidm = require('gui.dwarfmode')
+
+local biomeTypeNames = {
```
myk002, line 7: @ab9rf is there any way to get these strings by lookup?

gui/biomes.lua
```diff
@@ -0,0 +1,213 @@
...
+            biomesZY[x] = ix
+        end
+    end
+end
+
+gatherBiomeInfo()
+
+local TITLE = "Biomes"
+
+BiomeVisualizerLegend = nil -- for debugging purposes
```
myk002, line 114: for clearing the script environment for debugging/dev purposes, consider using `devel/clear-script-env`

gui/biomes.lua
```diff
@@ -0,0 +1,213 @@
...
+        local safe_index = safe_index
+        -- 304 = `0`, 353 = `a`
+        local idxBaseTile = 353
+        local biomes = biomesMap
+
+        local N = safe_index(biomes, pos.z, pos.y, pos.x)
+        if not N then return end
+        return COLOR_RED, tostring(N), idxBaseTile + (N - 1)
+    end
+    local overlay_bounds = getMapViewBounds()
```
myk002, line 204: if you're iterating over the entire visible viewport, you can pass in `nil` for `overlay_bounds`

myk002:
also, if you like the behavior of `gui/pathable`, you can copy its structure of having the highlighted biome follow the mouse instead of/in addition to having the list selection control which biome is highlighted.

gui/biomes.lua
ab9rf: not at present. someone could go into `df.map.xml`, add an `enum-attr` to the enum def,  and then add `item-attr` values to each `enum-item` so as to create one that could then be used by downstream tools, but nobody has ever done this

myk002:
@TymurGubayev have you made any progress on this? I'd love to get it into the next beta!

TymurGubayev:
@myk002 right, I need to finish this one. I'll try to do it this weekend.

gui/biomes.lua
TymurGubayev: I probably don't understand this. Right now, if I make a change in the script, or more specifically in the `BiomeVisualizerLegend`-class, I execute `gui/biomes`, and the changes are immediately applied. How would I use `devel/clear-script-env` instead? Execute first the `clear-sript-env` and then `gui/biomes`?.. I don't remember, how can I chain commands in dfhack console?

TymurGubayev:
Updated version is almost done, there are few things to still do though.
1. UI: right now, mousewheel events are captured by the active legends window, preventing user from changing z-level. No idea how can I disable this. Ideally, it should "fall through" to the game if not handled by a widget.
2. I'd like to move and resize the tooltip window with the legend window. I have no idea how to do this.
3. Something very strange is happening with the label-widget inside the tooltip window. Line 300 is my workaround. Without it, only a single-line is visible.
4. Possible **bug**: There is something strange going on with biomes in the air, either I don't understand something or it's a bug (probably in DFHack, but maybe DF): there are biomes up there that don't exist on a ground level, but also there are square blocks of tiles where biome info is missing (at least if retrieved the way I do it).

gui/biomes.lua
myk002: you'd only need to use `devel/clear-script-env` if you had previously set a global in the script and an update removes the global. for example, if you had `function MyClass:onInput(keys)` but then you remove the function, *it will still be defined and active* until you run `devel/clear-script-env` (or reload the game). If you just change the function definition, then there's no problem.

myk002:
> 4. Possible **bug**: There is something strange going on with biomes in the air, either I don't understand something or it's a bug (probably in DFHack, but maybe DF): there are biomes up there that don't exist on a ground level, but also there are square blocks of tiles where biome info is missing (at least if retrieved the way I do it).

I'll respond to the others in the review, but this one I think is expected. At least I've run into it before and I think there's a mantis bug for it. Biomes in the air are the same as the region up and left from the current embark area (or something like that).

Are the blocks where biome info is missing also in the air?

gui/biomes.lua
```diff
@@ -0,0 +1,392 @@
...
+    parent_window = DEFAULT_NIL,
+}
+
+function TooltipWindow:init()
+    self:addviews{
+        widgets.Label{
+            view_id = 'label',
+            frame = {t = 0, h = 1000},
+            auto_height = false,
+            --wtf??? without this the label is always a single line
```
myk002, line 317: this means that `updateLayout()` calls are not getting to this widget. If you call the `TooltipWindow`'s `updateLayout()` when the main panel's `updateLayout()` is called, then they should stay in sync. However, resizing from the line between the two "windows" is not a common UX style, and nothing else in DFHack does it that way. Maybe you could add the tooltip as a widget in `BiomeVisualizerLegend`, like how `gui/control-panel` does it? That would simplify the structure quite a bit. You can still specify an interior frame (`gui.INTERIOR_FRAME`) around the tooltip panel, if you wish.

docs/gui/biomes.rst
```diff
@@ -0,0 +1,17 @@
+gui/biomes
+==========
+
+.. dfhack-tool::
+    :summary: Shows biomes map overlay.
```
myk002, line 5:
```suggestion
    :summary: Visualize and inspect biome regions on the map.
```

docs/gui/biomes.rst
```diff
@@ -0,0 +1,17 @@
...
+    :tags: fort inspection map
+
+This script shows biomes map overlay.
```
myk002, line 8:
```suggestion
This script shows the boundaries of the biome regions on the map.
```

docs/gui/biomes.rst
```diff
@@ -0,0 +1,17 @@
...
+Hover over a biome entry in the list to get detailed info about it.
+
+
+Usage
+-----
+
+::
+
+  gui/biomes
```
myk002, line 17:
```suggestion
    gui/biomes
```

gui/biomes.lua
```diff
@@ -0,0 +1,392 @@
+-- Show biomes on the map.
+
+local RELOAD = false -- set to true to help with debugging
+
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local guidm = require('gui.dwarfmode')
+
+local TILE_HIGHLIGHTED = 2585 -- yellow-ish indicator
```
myk002, line 9: This needs to be looked up dynamically since its tilepos might change.
```lua
local TILE_HIGHLIGHTED = dfhack.textures.getOnOffTexposStart()
if TILE_HIGHLIGHTED < 0 then
    TILE_HIGHLIGHTED = --whatever the number for capital X is maybe
end
```

TymurGubayev:
> Are the blocks where biome info is missing also in the air?

yes they are

gui/biomes.lua
TymurGubayev: I was trying to recreate how hotkeys look, i.e.
image
but also hiding the tooltip if not hovering. I'll give it another try :)

gui/biomes.lua
TymurGubayev: is there a range where the textures are fixed? First 256? Also, what does switching to text mode do to this?

gui/biomes.lua
myk002: Note that the hotkeys panel is not resizable -- nor is the buildingplan panel, which also has that dual-frame look. For a resizable window, I think what you want is a single exterior frame with possibly a thinner interior frame around the tooltip. Otherwise it isn't obvious where the player should go to resize the window. You can still make the extended info disappear and the window shrink when it's not hovered, but I'd actually recommend leaving it. The player can always resize the window or dragging it around if it's in the way.

gui/biomes.lua
myk002: I thought there was, but then https://github.com/DFHack/dfhack/issues/3262 was filed and now I don't trust even the initial "pre-world" textures.

For a little background, run `devel/tile-browser` on the title screen. you can see that only about 2500 textures are defined. Those are the ones that I thought were constant, but maybe not between Steam and Classic. The 2585 one is loaded by DFHack, and we might add more textures at any time, so it definitely cannot be treated as a constant.

myk002: first 256 are *probably* constant, since those are the text glyphs.

myk002:
> > Are the blocks where biome info is missing also in the air?
> 
> yes they are

ok, then it's probably not worth worrying about. this is almost certainly a vanilla bug

myk002:
> image

I just realized -- you found an embark with 8 biomes? impressive.

TymurGubayev:
> I just realized -- you found an embark with 8 biomes? impressive.

:) It only actually has 3, other ones are in the air

gui/biomes.lua
TymurGubayev: I can't figure out how to make the window shrink dynamically :(
I'll try a version just showing the info

myk002:
> > I just realized -- you found an embark with 8 biomes? impressive.
> 
> :) It only actually has 3, other ones are in the air

Are you on Discord? figuring out what to do with those phantom biomes might be a good topic for discussion. Should they be hidden in the list? Optionally hidden, with a message saying that they are likely to be buggy? Showing them as identical to the "real" biomes might be confusing.

gui/biomes.lua
TymurGubayev: I give up on the resizable version.
List is fix 5 rows, info is fix 15 rows.
image

TymurGubayev:
> Are you on Discord? figuring out what to do with those phantom biomes might be a good topic for discussion. Should they be hidden in the list? Optionally hidden, with a message saying that they are likely to be buggy? Showing them as identical to the "real" biomes might be confusing.

I was for a bit on Discord, but I'm not sure I like it there, especially for discussions such as this one.

I think how to show the phantom biomes should depend on what happens if you manage to create a farm up there.
If it works as a farm in any other biome, then IMO there shouldn't be any difference in how they are shown.
If it doesn't, I could show them red, assuming the correct biomes are those at z = 0.

I'm most curious about the missing biomes: will the game explode or will it treat them like ice?

I'm not any good at such science experiments.

gui/biomes.lua
TymurGubayev: well, then I have to think about whether I set a global or not, and since most of the time I do not, I'll definitely forget the one time I do.

gui/biomes.lua
```diff
@@ -0,0 +1,376 @@
...
+
+    local name = biomeTypeNames[df.biome_type[biomeTypeId]] or "DFHACK_Unknown"
+
+    return ([[%s %s]]):format(surrounding, name)
+end
+
+function BiomeVisualizerLegend:init()
+    local list = widgets.List{
+        view_id = 'list',
+        frame = { t = 0, h = DEFAULT_LIST_HEIGHT, w = self.frame.w },
```
myk002, line 186: to allow the window to be resizable, change the frame to `{t=0, b=16}`

gui/biomes.lua
```diff
@@ -0,0 +1,376 @@
...
+        frame_inset = 0,
+        icon_width = 1,
+        text_pen = { fg = COLOR_GREY, bg = COLOR_BLACK }, -- this makes selection stand out more
+        on_select = self:callback('onSelectEntry'),
+    }
+    local tooltip_panel = widgets.Panel{
+        view_id='tooltip_panel',
+        autoarrange_subviews=true,
+        frame = { t = list.frame.h + 1, h = 15 },
```
myk002, line 195: to allow the window to be resizable, change the frame to `{b=0, h=15}`

gui/biomes.lua
```diff
@@ -0,0 +1,376 @@
...
+
+local TITLE = "Biomes"
+
+if RELOAD then BiomeVisualizerLegend = nil end
+local DEFAULT_LIST_HEIGHT = 5
+BiomeVisualizerLegend = defclass(BiomeVisualizerLegend, widgets.Window)
+BiomeVisualizerLegend.ATTRS {
+    frame_title=TITLE,
+    frame_inset=0,
+    resizable=false,
```
myk002, line 153: this can be left as true -- comments below

gui/biomes.lua
myk002: also note that `self.frame.w` is likely to be `nil` at this point in the code

gui/biomes.lua
TymurGubayev: made it resizable again with your suggestions

gui/biomes.lua
```diff
@@ -0,0 +1,379 @@
...
+end
+
+local regionBiomeMap = {}
+local biomesMap = {}
+local biomeList = {}
+local function gatherBiomeInfo(z)
+    local maxX, maxY, maxZ = dfhack.maps.getTileSize()
+    maxX = maxX - 1; maxY = maxY - 1; maxZ = maxZ - 1
+
+    local z = z or df.global.window_z
```
myk002, line 88:
```suggestion
    z = z or df.global.window_z
```

gui/biomes.lua
```diff
@@ -0,0 +1,379 @@
...
+
+    --for z = 0, maxZ do
```
myk002, line 90: no longer used? also the corresponding `--end`

gui/biomes.lua
```diff
@@ -0,0 +1,379 @@
...
+            text = ([[%s: %s]]):format(biomeExt.char, GetBiomeName(biomeExt.biome, biomeExt.typeId)),
+            icon = self:get_icon_pen_callback(#choices+1),
+            biomeTypeId = biomeExt.typeId,
+            biome = biomeExt.biome,
+        })
+    end
+    self.list:setChoices(choices)
+end
+
+do -- implementation of onMouseHoverEntry(idx, option)
```
myk002, line 254: why is this in a `do` block?

gui/biomes.lua
```diff
@@ -0,0 +1,379 @@
...
+    return function ()
+        if self.SelectedIndex == ix then
+            return PEN_ACTIVE_ICON
+        else
+            return PEN_NO_ICON
+        end
+    end
+end
+
+function BiomeVisualizerLegend:onSelectEntry(idx, option)
```
myk002, line 233: in my testing, `option` can be nil. it might be a good idea to check that and return early here if option is nil

gui/biomes.lua
```diff
@@ -0,0 +1,379 @@
...
+    self:ShowTooltip(option or self.SelectedOption)
+end
+
+function BiomeVisualizerLegend:ShowTooltip(option)
+    local text = get_tooltip_text(option)
+
+    local tooltip_panel = self.tooltip_panel
+    local lbl = tooltip_panel.subviews.label
+
+    lbl:setText(text)
```
myk002, line 318:
```suggestion
    lbl:setText(text)
    lbl:updateLayout()
```

gui/biomes.lua
```diff
@@ -0,0 +1,379 @@
...
+}
+
+function BiomeVisualizer:init()
+    local legend = BiomeVisualizerLegend{view_id = 'legend'}
+    self:addviews{legend}
+end
+
+function BiomeVisualizer:onRenderFrame(dc, rect)
+    BiomeVisualizer.super.onRenderFrame(self, dc, rect)
+
```
myk002, line 337:
```suggestion

    if not dfhack.screen.inGraphicsMode() and not gui.blink_visible(500) then
        return
    end

```

myk002: so the player can still see the underlying map in ascii mode. otherwise it looks like this:

image

gui/biomes.lua
TymurGubayev: this causes

```
...steamapps\common\Dwarf Fortress\hack\lua\gui\widgets.lua:51: attempt to index a nil value (local 'parent_rect')
stack traceback:
        ...steamapps\common\Dwarf Fortress\hack\lua\gui\widgets.lua:51: in method 'computeFrame'
        ...Library\steamapps\common\Dwarf Fortress\hack\lua\gui.lua:514: in method 'updateLayout'
        ...amapps\common\Dwarf Fortress/hack/scripts/gui/biomes.lua:323: in method 'ShowTooltip'
```

Also, for me, the info is shown on the initial display.

gui/biomes.lua
TymurGubayev: for fold it in a code editor, because it doesn't quite belong there: I'd actually preferred if the implementation of `onMouseHoverEntry(idx, option)` would be in a core script (probably widgets.lua).

gui/biomes.lua
TymurGubayev: yes, it was too slow. I'd like to leave it there for future reference though.

gui/biomes.lua
TymurGubayev: I've added a check in the `get_tooltip_text(option)`

gui/biomes.lua
TymurGubayev: Ah, so *that*'s how the blinking is done :) Good to know

TymurGubayev:
`dfhack.textures.getOnOffTexposStart()` doesn't exist anymore. What do I replace it with?

ab9rf:
> `dfhack.textures.getOnOffTexposStart()` doesn't exist anymore. What do I replace it with?

dfhack/dfhack commit fe8bd4fa786369f7a95ae7925fc992095cc1fb2d may be informative

myk002:
unsuspend might be a good model to follow for the new textures api:
https://github.com/DFHack/scripts/blob/master/unsuspend.lua#L15

myk002:
could you rebase onto current master and add a changelog entry for the new tool?

myk002:
It looks like the used texture needs to change. Since the switch to SDL2, textures are stretched to cover the entire map tile:
image

`gui/pathable` switched to cursor textures:
image

TymurGubayev:
> It looks like the used texture needs to change. Since the switch to SDL2, textures are stretched to cover the entire map tile: 
> `gui/pathable` switched to cursor textures: 

well, I have an idea about that actually:

image

this happens if I cut the 8x12 "on" indicator and place it in the top left corner on a different 32x32 png image with a transparent background.
This is done easily, the next question is, what to do with all the letters. I see the following possibilities.
1. Leave it as is, as the letters naturally have a lot of transparent pixels in them, so it isn't too ugly; in which case I'll probably switch to the cursor texture because otherwise, the mix of indicators in a corner and letters over whole tiles looks a bit weird.
2. Get or create a texture with letters in the corner of a 32x32 (haven't looked too hard for it), which would be some work involved and we'd end up with a somewhat big texture file.
3. Create required textures on the fly, including the indicator one.

I'd much prefer the 3rd approach, especially since we already have half of the required API: `createTile`/`createTileset`. The missing half is something to get the pixels from a loaded texture, or maybe some method to resize textures, etc.

myk002:
I tried (1) for the `unsuspend` overlay (displaying if a building is suspended or planned), but it just did not look good. I ended up designing new textures (rather, I asked someone more artistic than myself to design them): https://github.com/DFHack/dfhack/blob/develop/data/art/unsuspend.png

Adding another texture map to our art directory wouldn't be onerous. It's 5k for 5 `unsuspend` images, so 26 letters would be about 25k.

If you'd like to finish up the dynamic texture API, that would be cool, but I'm not sure if we should wait for that for this PR -- I'd rather get this released sooner rather than later.

docs/gui/biomes.rst
```diff
@@ -0,0 +1,17 @@
+gui/biomes
+==========
+
+.. dfhack-tool::
+    :summary: Visualize and inspect biome regions on the map.
+    :tags: fort inspection map
+
+This script shows the boundaries of the biome regions on the map.
+Hover over a biome entry in the list to get detailed info about it.
```
myk002, line 9: How about also allowing the player to hover the mouse cursor over the map to select the region in the list? If the keyboard cursors are used and the mouse hasn't moved, then that can take precedence to move the list cursor. Then when the mouse moves, it can be the selector again.

gui/biomes.lua
```diff
@@ -0,0 +1,388 @@
...
+                    and TILE_HIGHLIGHTED
+                    or TILE_STARTING_SYMBOL + (N-1)
+        local color = (N == idxSelected)
+                    and COLOR_CYAN
+                    or COLOR_GREY
+        local ch = string.char(string.byte('a') + (N-1))
+        return color, ch, idxTile
+    end
+
+    guidm.renderMapOverlay(get_overlay_pen, nil) -- nil for bounds means entire viewport
```
myk002, line 372: I'm amazed that is as performant as it is. When I was developing `gui/quickfort`, rendering to the entire viewport was palpably slow, but this seems to run without any delay.

edit: `gui/quickfort` is still kind of slow. I guess it's the algorithm there that needs improvement.

gui/biomes.lua
myk002: I think this could be fixed if you use a WrappedLabel and set `text_to_wrap=function() return get_tooltip_text(self:get_option()) end,`

TymurGubayev:
where do I put art assets for scripts?

I'd like to add the following two.

curses_small_letters - top-left
on-off - top-left

Then it looks like this (mouse over a "b" tile)
image

docs/gui/biomes.rst
TymurGubayev: implemented... Now that I re-read your comment, probably not exactly what you had in mind, but hopefully close enough

gui/biomes.lua
TymurGubayev: could you please check if this is needed at all? I couldn't reproduce the initial issue this should fix: the tooltip is there from the start and is updated on hover

myk002:
> where do I put art assets for scripts?

all assets should go in the dfhack repo in the `data/art/` directory.

myk002 (approved):
love it. Great job!

gui/biomes.lua
myk002: latest code shows this is fixed.

TymurGubayev:
yay!

PR #638, opened 2023-03-13
script concept for embark anywhere on DF50

Rumrusher:
here's an embark anywhere attempt script this should let one embark anywhere... still need gui widget support to handle the new mouse only cursor controls

### DIFF
--- a/EmbarkAnywhere_v50.lua
+++ b/EmbarkAnywhere_v50.lua
@@ -1,0 +1,9 @@
+local MapCure=df.global.gview.view.child
+
+MapCure.warn_mm_startx = MapCure.neighbor_hover_mm_sx
+MapCure.warn_mm_endx = MapCure.neighbor_hover_mm_ex
+MapCure.warn_mm_starty = MapCure.neighbor_hover_mm_sy
+MapCure.warn_mm_endy = MapCure.neighbor_hover_mm_ey
+MapCure.in_embark_salt=true
+-- uhh this is a micro script that functions the basic steps to bypass the embark warnings if anyone wants to update this with gui functions and probably better mosue support feel free. 
+-- not made to replace the old embarkAnywhere command
--- a/EmbarkAnywhere_v50.lua
+++ b/EmbarkAnywhere_v50.lua
@@ -1,9 +1,0 @@
-local MapCure=df.global.gview.view.child
-
-MapCure.warn_mm_startx = MapCure.neighbor_hover_mm_sx
-MapCure.warn_mm_endx = MapCure.neighbor_hover_mm_ex
-MapCure.warn_mm_starty = MapCure.neighbor_hover_mm_sy
-MapCure.warn_mm_endy = MapCure.neighbor_hover_mm_ey
-MapCure.in_embark_salt=true
--- uhh this is a micro script that functions the basic steps to bypass the embark warnings if anyone wants to update this with gui functions and probably better mosue support feel free. 
--- not made to replace the old embarkAnywhere command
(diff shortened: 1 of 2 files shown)


myk002:
What is the flow? Does this sound right?

1. player starts to choose embark location, wants to embark on site where the game doesn't allow
2. player runs `gui/embark-anywhere`
3. window appears (under the existing panel? is there enough room there if the window is small?), shows instructions for how to continue
4. window intercepts click, applies changes so embark is successful
5. window closes automatically

how does that sound?

myk002:
btw, once we have this tool, it looks like we can remove the `embark-tools` plugin. I don't think there's anything else we need from that plugin anymore. the mouse controls and sand indicator is now part of vanilla, and the "stable cursor" functionality is no longer applicable.

lethosor:
What is all the "spawn liquid" stuff doing in this?

myk002:
> What is all the "spawn liquid" stuff doing in this?

I suggested using `gui/liquids` as an example for how to create a window and intercept mouse clicks. I think rumrusher just copied that code in for now until we can figure out how to integrate the new logic.

Rumrusher:
> 

oh yeah I totally just copy the entire file and inserted the original script into it.
other wise the player would have to run the original script through a keybind so that they could free up their mouse to be used as a position tool. the script is a WIP that probably need some major tweaks.

myk002:
any progress on this script?

TheTimtam97:
So I have this working using just a basic lua script (see below)

`function embarkanywhere ()
df.global.gview.view.child.warn_mm_startx = df.global.gview.view.child.neighbor_hover_mm_sx
df.global.gview.view.child.warn_mm_endx = df.global.gview.view.child.neighbor_hover_mm_ex
df.global.gview.view.child.warn_mm_starty = df.global.gview.view.child.neighbor_hover_mm_sy
df.global.gview.view.child.warn_mm_endy = df.global.gview.view.child.neighbor_hover_mm_ey
df.global.gview.view.child.in_embark_salt=true
end

embarkanywhere()`

As soon as you run that script, the "Confirm and abort" window pops up and you're free to move your mouse wherever you want. 

Not sure what a gui would even look like tbh. A hotkey and some text in the bottom left corner explaining the hotkey would be nice but overall unnecessary.

I've tried to force it into the old embark tools plugin, but I'm well out of my depth. There's so much in that plugin and I don't know how to start from scratch

myk002:
No, you have it right -- there is very little to do other than the core code. I think this is all that's required:
1. check to make sure you're on the correct screen and this is an appropriate time to run the script
2. display a window informing the user that embarks are unlocked, or allow the player to cancel the current state and go back to vanilla embark warnings
3. dismiss the window when the player successfully embarks

TheTimtam97:
Regarding requirement No. 2.

I was just going to hijack the vanilla embark warnings that pop up in yellow when you embark in salt or next to a necro tower. It would say something to the effect of "This embark has been enabled by DFHack and is not recommended by inexperienced players" From that point, the player can abort just like any regular embark and go back to vanilla. If they want to embark on a different (previously restricted spot), they just rerun the script.

Though I'm having difficulty figuring out where these warnings are stored. Are they stored as a string in memory? Or is it more complicated than that? I'm fairly sure that it's different to the "in_embark_salt" structures because there's nothing about necromancer towers in that list and yet, the warning for towers pops up.

image

myk002:
Isn't there a live message that appears when you move the mouse cursor around? Could you add a message there (too?)

myk002:
It might not be a good user experience to require players to have the mouse in just the right position before they run the command. How about bringing up a window along the edge of the screen asking them to click on the desired location?

PR #639, opened 2023-03-13
fix autosave with code reverse engineered by ab9rf

myk002:
Fixes DFHack/dfhack#2663

### DIFF
--- a/quicksave.lua
+++ b/quicksave.lua
@@ -1,12 +1,4 @@
 -- Makes the game immediately save the state.
---[====[
-
-quicksave
-=========
-If called in dwarf mode, makes DF immediately saves the game by setting a flag
-normally used in seasonal auto-save.
-
-]====]
 
 local gui = require("gui")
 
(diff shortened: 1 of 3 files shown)


quicksave.lua
```diff
@@ -42,16 +34,25 @@ local function restore_autobackup()
 end
 
 function save()
-    -- Request auto-save
+    -- Request auto-save (preparation steps below discovered from rev eng)
     ui_main.autosave_request = true
+    ui_main.autosave_unk = 5
```
ab9rf, line 39: I'm probably going to push a structures change to rename this field since I have a good idea what it does (although not why it does what it does)

and of course this violates our "rule" about not using `unk` or `anon` fields in scripts

myk002: according to Putnam, `autosave_unk` is `auto_save_timer`

and `unk_v50_6` is `info`, if you'd like to make that change too while you're in there

PR #640, opened 2023-03-15
[gui/launcher] replace tabs with spaces in command output

myk002:
this is a stop gap solution for true tabulation, which would take a lot more work

Should this be considered to fix DFHack/dfhack#3023 and fix DFHack/dfhack#3024?


ab9rf (approved):
this is at least an improvement on the status quo and i don't think it's worth delaying getting this for getting a "better" approach to dealing with tabs

PR #642, opened 2023-03-16
gui/dig building placement using buildingplan filters

johncosker:
### Adds `Building` mode for gui/dig

Adds `Building` mode for designation which provides options for inner and outer tiles, added a `?` button which pops up a help text explaining to the user that they should set the building materials in `buildingplan` prior to using the tool, if they want to.

buildings

help_text

### DIFF
--- a/gui/dig.lua
+++ b/gui/dig.lua
@@ -39,7 +39,6 @@ local guidm = require("gui.dwarfmode")
 local widgets = require("gui.widgets")
 local quickfort = reqscript("quickfort")
 local shapes = reqscript("internal/dig/shapes")
-local filterselection = require('plugins.buildingplan.filterselection')
 
 local tile_attrs = df.tiletype.attrs
 
@@ -1787,9 +1786,7 @@ function Dig:get_designation(x, y, z)
 
         -- If not completed surrounded, then use outer tile
         for i, d in ipairs(darr) do
-            -- print(view_bounds.x + x + d[1], view_bounds.y + y + d[2])
             if not (self.shape:get_point(top_left.x + x + d[1], top_left.y + y + d[2])) then
-
                 return building_outer_tiles
             end
         end
--- a/gui/dig.lua
+++ b/gui/dig.lua
@@ -95,11 +95,11 @@ HelpWindow.ATTRS {
 function HelpWindow:init()
     self:addviews {
         widgets.ResizingPanel { autoarrange_subviews = true,
-            frame = { t = 0, l = 0},
+            frame = { t = 0, l = 0 },
             subviews = {
                 widgets.WrappedLabel {
                     view_id = 'help_text',
-                    frame = { t = 0, l = 0},
+                    frame = { t = 0, l = 0 },
                     text_to_wrap = function() return self.message end,
                 }
             }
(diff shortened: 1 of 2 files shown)


myk002:
The filter dialog is tied to the `buildingplan` workflow, and makes some changes to the UI state that wouldn't apply in this case. I think the interface that you'll want to link to is the (not yet written) `gui/buildingplan`. For now, it might be sufficient to just document that building properties should be set in the `buildingplan` interface that you get when building that type of building.

johncosker:
Okay then that makes things easier for now. I'll change the gear icons to question marks and pop up a little message or something

johncosker:
Also need to figure out how to handle renaming gui/dig to gui/design (or something else) in a way that allows keeping existing user's hotkeys

myk002:
there's an `alias` command that I think should do what we need. and if it doesn't we can extend it until it does.
https://docs.dfhack.org/en/latest/docs/tools/alias.html#alias

we can add a default alias to https://github.com/DFHack/dfhack/blob/develop/data/init/dfhack.tools.init

johncosker:
Okay yeah I knew about alias but I wasn't sure about how to make one by default

johncosker:
Ready for review

gui/dig.lua
```diff
@@ -1760,13 +1945,16 @@ DigScreen = defclass(DigScreen, gui.ZScreen)
 DigScreen.ATTRS {
     focus_path = "dig",
     pass_pause = true,
+    initial_pause = false,
```
myk002, line 1948: Some tools might want to set this to `true`, but no tool should have a need to set it explicitly to `false`. Instead, the player should set the default pause state in `gui/control-panel`

gui/dig.lua
```diff
@@ -683,29 +798,80 @@ function GenericOptionsPanel:init()
...
+                },
+                widgets.CycleHotkeyLabel {
+                    view_id = "building_inner_tiles",
+                    key = "CUSTOM_G",
+                    label = "Inner Tiles: ",
+                    frame = { t = 1, l = 5 },
+                    active = true,
+                    enabled = true,
+                    initial_option = 2,
+                    visible = function() return self.dig_panel.subviews.mode_name:getOptionValue().desig == "b" end,
```
myk002, line 870: in order to avoid copying this function everywhere it's needed, you can declare a local function inside the scope if `init` and refer to it by name. e.g. https://github.com/DFHack/dfhack/blob/develop/plugins/lua/buildingplan/planneroverlay.lua#L156

myk002: similar suggestion for the on_click callbacks above

myk002 (commented):
this is definitely good for now. sometime soonish, I'd like to explore some UI mocks for how a `gui/designer` tool that has the functionality described in https://github.com/DFHack/dfhack/issues/2446 so we can visualize what we're working towards. we probably want to involve some of our more ui-inclined community members on discord.

gui/dig.lua
johncosker: Oh I didn't know that functionality existed, cool

myk002 (approved):

PR #658, opened 2023-03-27
[suspendmanager] Detect and protect dead end constructions

plule:
Fixes DFHack/dfhack#3049 and DFHack/dfhack#3157

Replace the suspendmanager blocking detection. This new approach explore groups of connected construction jobs, and suspend corridors that would prevent accessing a deadend or the cluster entirely.

This approach handles many more situations, the most important being the issues linked above. It also behaves better for filling area, which was mentioned on discord.

It still have some short-coming, such as interior corridors in a big cluster not being suspended, and the general algorithm feeling more complicated than it should be but all my attempt to simplify the general logic lead to many blocking points...

Two situations where the previous approach would mostly fail, but are now well handled:

Two-z walls where dwarves must build the walls one by one:

suspendmanager-1

Area filling:

suspendmanager-2

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,6 +22,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/create-item`: ask for number of items to spawn by default
+- `suspendmanager`: improve detection of potentially blocking jobs
 
 ## Removed
 
--- a/suspend.lua
+++ b/suspend.lua
@@ -16,8 +16,14 @@ if help then
     return
 end
 
+local manager = suspendmanager.SuspendManager.new()
+
+if onlyblocking then
+    manager:computeBlockingJobs()
+end
+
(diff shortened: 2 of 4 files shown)


myk002:
Have you made any progress on this? I'd like to get it into the next beta release if possible

plule:
Unfortunately no, I got a bit discouraged by not managing to find a clean approach... I'll attempt to get back on it this week-end to get it on a mergeable state, maybe off by default with a «experimental» warning in the UI if it feels too risky?

myk002:
That's why we have the beta branch on Steam. We can ask for testing and feedback there before we push a new public stable build.

plule:
closing - will reopen after https://github.com/DFHack/scripts/pull/721 that's just focused on the class refacto

PR #661, opened 2023-03-29
[warn-starving] no reports for non-fort-controlled units

myk002:
this also skips reports for residents, which is debatable. should we report on residents? Moreover, should isFortControlled be returning false for residents?

Fixes DFHack/dfhack#3117


ab9rf (approved):
tested against the symptom fort, works as expected

myk002:
ack. ok, merging as-is. If we get complaints about residents not being monitored, we can re-evaluate.

PR #663, opened 2023-03-30
Update general-strike.lua

ab9rf:
only update seeds in farms

closes dfhack/dfhack#3131


myk002 (approved):
Why is there even a isFarmPlot function in DF? What does it check other than the type?

PR #665, opened 2023-04-01
launcher: add label showing pause status

Cubittus:
This adds a label that displays `pause` or `unpause` depending on game state.  It is enabled or disabled according to the state of the input editfield

### DIFF
--- a/gui/launcher.lua
+++ b/gui/launcher.lua
@@ -216,10 +216,16 @@ function EditPanel:init()
             -- to the commandline
             ignore_keys={'STRING_A096'},
             on_char=function(ch, text)
-                if ch == ' ' then return text:match('%S$') end
+                if ch == ' ' and not text:match('%S$') then
+                    self:update_pause_label(true)
+                    return false
+                end
                 return true
             end,
-            on_change=self.on_change,
+            on_change=function(new,old)
+                self:update_pause_label(false, not new:match('%S$'))
+                self.on_change(new,old)
+            end,
             on_submit=self.on_submit,
             on_submit2=self.on_submit2},
         widgets.HotkeyLabel{
@@ -238,9 +244,13 @@ function EditPanel:init()
             key='CUSTOM_ALT_M',
             label=string.char(31)..string.char(30),
             on_activate=self.on_toggle_minimal},
+        widgets.Label{
+            view_id='pauselabel',
+            frame={l=13, t=3, w=15},
+            text={{key='D_PAUSE', key_sep=': ', text='unpause'}}},
         widgets.EditField{
             view_id='search',
-            frame={l=13, t=3, r=1},
+            frame={l=29, t=3, r=1},
             key='CUSTOM_ALT_S',
             label_text='history search: ',
             on_change=function(text) self:on_search_text(text) end,
@@ -258,6 +268,11 @@ function EditPanel:init()
             on_submit2=function()
                 self.on_submit2(self.subviews.editfield.text) end},
     }
+end
+
+function EditPanel:update_pause_label(changing, canchange)
+    self.subviews.pauselabel.enabled = canchange or false
+    self.subviews.pauselabel:setText({{key='D_PAUSE', key_sep=': ', text=(df.global.pause_state ~= changing) and 'unpause' or 'pause'}})
 end
 
 function EditPanel:reset_history_idx()
--- a/gui/launcher.lua
+++ b/gui/launcher.lua
@@ -217,7 +217,7 @@ function EditPanel:init()
             ignore_keys={'STRING_A096'},
             on_char=function(ch, text)
                 if ch == ' ' and not text:match('%S$') then
-                    self:update_pause_label(true)
+                    self:update_pause_label(true, true)
                     return false
                 end
                 return true
@@ -246,7 +246,7 @@ function EditPanel:init()
             on_activate=self.on_toggle_minimal},
         widgets.Label{
             view_id='pauselabel',
-            frame={l=13, t=3, w=15},
+            frame={l=13, t=3, w=14},
             text={{key='D_PAUSE', key_sep=': ', text='unpause'}}},
         widgets.EditField{
             view_id='search',
@@ -271,7 +271,7 @@ function EditPanel:init()
 end
 
 function EditPanel:update_pause_label(changing, canchange)
-    self.subviews.pauselabel.enabled = canchange or false
+    self.subviews.pauselabel.enabled = canchange
     self.subviews.pauselabel:setText({{key='D_PAUSE', key_sep=': ', text=(df.global.pause_state ~= changing) and 'unpause' or 'pause'}})
 end
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -21,6 +21,7 @@ that repo.
 ## Misc Improvements
 - `gui/gm-editor`: can now jump to material info objects from a mat_type reference with a mat_index using ``i``
 - `gui/gm-editor`: they key column now auto-fits to the widest key
+- `gui/launcher`: add a label to show the game pause state and when it can change by pressing space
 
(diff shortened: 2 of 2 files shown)


myk002:
image

I see a few issues with this that we'd need to figure out:
1) when the launcher window is resized smaller, the history search becomes unusable (see screenshot)
2) unlike all other labels with the green hotkey hint, this label is not clickable to trigger the pause/unpause action
3) the pause/unpause label changes state even when a world is not loaded
4) the hotkey hint says "Space", but this is misleading. it's only space at the beginning of a line or double space anywhere else. this may need more words to explain than just "Space"

What do you think of adding this Label underneath the current run and history search widgets and moving the help panel down a tile? that would give you a bit more room to explain the mechanics of pausing/unpausing.

Cubittus:
Thanks - I'll make the following changes:
- Move the Alt+s:history line below Enter:run.
- Make the Space:un/pause clickable
- Make it disabled when there is no loaded fort
- Add a sentence to the Help text at the end of the first paragraph:
  `You can press space to unpause or pause a loaded fort (press space
   twice while typing a command).`

Cubittus:
This should resolve https://github.com/DFHack/dfhack/issues/3125

gui/launcher.lua
```diff
@@ -258,6 +269,34 @@ function EditPanel:init()
...
+    self:update_pause_label(false,true)
+end
+
+function EditPanel:pause_fort()
+    if dfhack.isMapLoaded() then
+        df.global.pause_state = not df.global.pause_state
+    end
+    self:update_pause_label(false)
+end
+function EditPanel:enable_pause_label()
```
myk002, line 281: instead of changing the colors, how about changing the hotkey hint from "Space" to "Double Space"? That might make the intent a little clearer.

gui/launcher.lua
```diff
@@ -238,9 +244,14 @@ function EditPanel:init()
...
             on_activate=self.on_toggle_minimal},
+        widgets.Label{
+            view_id='pauselabel',
+            frame={l=13, t=3, w=21},
+            text={{key='D_PAUSE',key_sep=': ',text='no fort loaded',pen=COLOR_DARKGREY,key_pen=COLOR_GREEN}},
+            on_click=function() self:pause_fort() end},
         widgets.EditField{
             view_id='search',
-            frame={l=13, t=3, r=1},
+            frame={l=1, t=4, r=1},
```
myk002, line 254: I'm not so sure about this change. @TaxiService what do you think?

image

gui/launcher.lua
```diff
@@ -258,6 +269,34 @@ function EditPanel:init()
...
+function EditPanel:update_pause_label(willchange, canchange)
+    if dfhack.isMapLoaded() then
+        if canchange ~= nil then
+            if canchange then self:enable_pause_label()
+            else self:disable_pause_label() end
+        end
+        self.subviews.pauselabel.text[1].text = (df.global.pause_state ~= willchange) and 'unpause' or 'pause'
+    else
+        self:disable_pause_label()
+        self.subviews.pauselabel.text[1].text = 'no fort loaded'
```
myk002, line 298: You should probably disable the Label in this case so it doesn't react on hover.
image

gui/launcher.lua
TaxiService: ehhhhhh i don't know if "Double Space" would necessarily make it "clearer". It would be descriptive, but doesn't necessarily explain _why_ the user has to press spacebar twice for it to unpause.

the real reason is, like, because after the first space, the second is passed to the game or something? There _has_ to be a better way to convey this

gui/launcher.lua
TaxiService: like, moving those two things on a new line _does makes sense_ for readability, but...

...are there solutions that don't require adding a `Space: unpause` hotkeylabel at all?

TaxiService:
Like, sorry to pop in like this, but consider this:
Untitled-1

since the original problem was, like, if i remember correctly...
> "i can't tell when spacebar will be input in the command line or un/pause the game"

like, _what if_ we had a little message that appears in the corner of the frame, informing the player _whenever spacebar won't unpause?_

this would have to dynamically appear/disappear as soon as spacebar _would_ be able to un/pause, but doing this would let us not add "bloat" to the gui, _and_ train the player to look in the lower-left corner of the UI for shit related to un/pausing.
(since other windows will show a `PAUSE FORCED` message in that area, in some cases)

myk002:
What I like about Cubittus's design is that the indicator is close to where the player's eyes are going to be when typing. If the launcher window is maximized (double click on the title bar) then the lower left corner is a *long* way away. A commandline can involve a lot of spaces, and the indicator can flash back and forth several times. whatever we do, it shouldn't be too distracting.

TaxiService:
>A commandline can involve a lot of spaces, and the indicator can flash back and forth several times. whatever we do, it shouldn't be too distracting.

it shouldn't be *too* distracting, but it should be *a bit* distracting. 🤔 Like, in the sense that by interacting with the command line *and* observing how that label behaves, the user would naturally get an intuition for when spacebar would and wouldn't un/pause.

> My main worry is the crowding of elements in a tool that _needs_ to be very clean.

and indeed, i think that adding that `Space: unpause` thing up there would kind of make that area more "crowded"
the label down there shouldn't even collide with the `PAUSE FORCED` label, because if pause-forced is on, spacebar can't unpause!

> What I like about Cubittus's design is that the indicator is close to where the player's eyes are going to be when typing. If the launcher window is maximized (double click on the title bar) then the lower left corner is a _long_ way away. 

that is true. And this is why it should probably be _noticeable enough_ to be seen, but not _distracting enough_ to be bothersome.

Like, _maybe_ we could have it so that the first time the spacebar is captured, and up until the first other keyboard input, the label is like LCYAN; then after the second input, it becomes DGRAY. 🤔 

or, you know, some other kind of measure to make the thing be noticeable at first, but ignorable once you know what it does.


like-- all of this can be achieved even with the `Spacebar: unpause` thing in the middle of the command line area, eh.
i'm just trying to think if there's anything else we can do 🤔
(also i'm not even sure if anything i'm saying makes any sense, remember to ask other people as well!)

Cubittus:
Thanks for all the feedback!
I've played around with a couple of changes.
The video below shows three possibilities:
- The current (The pause label is hidden when no fort is loaded now)
- Moving the pause label above the commandline
- Changing the pause label to only show the current state, but still toggling the highlight as you type

Regarding discoverability of space-to-pause there is a new line added to the default help page.
Anyone opening the tool for the first time will read this in the first paragraph.

https://user-images.githubusercontent.com/627213/229715317-bb582812-1470-4e27-ae7d-5022571a56be.mp4

TaxiService:
you know... the first of those three proposal is growing on me. 
i mean having `Space: un/pause` next to `Enter: run` on the same line, above `Alt+s: History search:`

That said, i like how you found solutions that do not increase that area's height by putting the label next to the minimize hotkey!
but i'm not sure if i like how they look. 🤔 (the second and third proposals) 
like, i mean: the placement is genius, but visually it makes that area feel a bit messier. If i had to choose between these three options i'd probably pick the first one.

I kinda like the third proposal's lack of an hotkeylabel, but maybe _not having the hotkeylabel_ makes "why is this thing there" a little unclear in this case. 🤔 
(btw, the minimize button thing should probably be reworked eventually; we need to find more a consistent solution for *all* windows that can be minimized)

so in the end my feedback is this:
- the first proposal is alright! 
- i still believe that coming up with a more "reusable" solution like having a label like the `PAUSE FORCED` one, would be better in the long run. 🤔 (if you feel it's too far from the text input field, we could move both the "PAUSE FORCED" and this "SPACEBAR CAPTURED" labels in the _top_-left corner of the window, for all windows!)

in any case, i'd like to hear more people's opinions on this.

PR #670, opened 2023-04-04
gm-editor: Add ``g`` shortcut to locate a target or selected value on the map

Cubittus:
When a fort is loaded jump to the first of:
- the coord if the current target or the selected value is a coord
- the `pos` of the current target
- the `centerx`, `centery`, `z` of the current target
- the `pos` of the selected value
- the `centerx`, `centery`, `z` of the selected value

The last two allow browsing lists like world.units.active or world.buildings.all and pressing g to jump to each item in the list without opening them.
Items in containers/being carried won't be located currently - I can add a (recursive) holder/container ref lookup if there is interest.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,6 +18,7 @@ that repo.
 ## Fixes
 
 ## Misc Improvements
+- `gui/gm-editor`: press ``g`` to move the map to the currently selected item/unit/building
 
 ## Removed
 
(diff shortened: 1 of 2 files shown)


myk002:
Note that units (and as you mentioned, items) can have unintuitive "real" positions. You can call `dfhack.units.getPosition()` and `dfhack.items.getPosition()` if you detect the relevant structures

myk002 (approved):
clean and brilliant. thanks!

PR #671, opened 2023-04-05
gm-editor: add read-only mode with persistent state

Cubittus:
Fixes DFHack/dfhack#3171

- new global `read_only` var
- read only state is saved in `config.data.read_only`
- frame position now saved in `config.data.frame`
- config saving moved to `save_config` fn
- keybinding `Ctrl-D` added to toggle `read_only`
- titles of all views updated on change
- added `dev`/`edit` and `safe`/`read` args to clear/set `read_only` on start
- insert/delete/set fns do nothing when `read_only` set
- `editSelected` allows only navigation when `read_only` set

One outstanding issue is that views opened using the `dialog` argument are not added to the `views` global array, so cannot have their `frame_title` updated by other views. I think to fix this I'd need to open the main view before showing the input dialog, which may mean moving the dialog code up into GmEditorUi somewhere.

By default I set `read_only` to true, so it's a bit safer for new users, but it can easily be changed on line 19.
Either way, pressing `Ctrl-D` once, will go back to full read-write access, which will be remembered for future sessions.

### DIFF
--- a/gui/gm-editor.lua
+++ b/gui/gm-editor.lua
@@ -10,9 +10,7 @@ local utils = require 'utils'
 config = config or json.open('dfhack-config/gm-editor.json')
 
 function save_config(data)
-    for k,v in pairs(data) do
-        config.data[k] = v
-    end
+    utils.assign(config.data, data)
     config:write()
 end
 
--- a/gui/gm-editor.lua
+++ b/gui/gm-editor.lua
@@ -9,6 +9,18 @@ local utils = require 'utils'
 
 config = config or json.open('dfhack-config/gm-editor.json')
 
+function save_config(data)
+    for k,v in pairs(data) do
+        config.data[k] = v
+    end
+    config:write()
+end
+
+read_only = true
+if config.data.read_only ~= nil then
+    read_only = config.data.read_only
+end
+
 find_funcs = find_funcs or (function()
     local t = {}
     for k in pairs(df) do
@@ -20,6 +32,7 @@ find_funcs = find_funcs or (function()
 end)()
 
 local keybindings_raw = {
+    {name='toggle_ro', key="CUSTOM_CTRL_D",desc="Toggle between read-only and read-write"},
     {name='offset', key="CUSTOM_ALT_O",desc="Show current items offset"},
     {name='find', key="CUSTOM_F",desc="Find a value by entering a predicate"},
     {name='find_id', key="CUSTOM_I",desc="Find object with this ID, using ref-target if available"},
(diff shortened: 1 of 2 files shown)


gui/gm-editor.lua
```diff
@@ -9,6 +9,18 @@ local utils = require 'utils'
...
 config = config or json.open('dfhack-config/gm-editor.json')
 
+function save_config(data)
+    for k,v in pairs(data) do
+        config.data[k] = v
+    end
+    config:write()
+end
+
+read_only = true
```
lethosor, line 19: I think an editor shouldn't start in read-only mode by default.
```suggestion
read_only = false
```

gui/gm-editor.lua
```diff
@@ -72,7 +85,7 @@ end
 
 GmEditorUi = defclass(GmEditorUi, widgets.Window)
 GmEditorUi.ATTRS{
-    frame=copyall(config.data),
+    frame=copyall(config.data.frame or {}),
```
lethosor, line 88: Instead of making `read_only` a global, you could add it here, then check it with `self.read_only` in other functions. It looks to me like it's only used from within this class (and when saving config)

gui/gm-editor.lua
```diff
@@ -577,6 +615,14 @@ end
...
 local function get_editor(args)
     if #args~=0 then
+        if args[1]=='dev' or args[1]=='edit' then
+            read_only = false
+            table.remove(args,1)
+        end
+        if args[1]=='safe' or args[1]=='read' then
+            read_only = true
+            table.remove(args,1)
+        end
```
lethosor, line 625: For consistency with other tools, I would prefer to use `utils.processArgsGetopt()` or another option here if possible. That would allow things like
```
gui/gm-editor -r
gui/gm-editor --read-only
```
or, I suppose
```
gui/gm-editor -e
gui/gm-editor --editable
```

lethosor (changes requested):

gui/gm-editor.lua
Cubittus: I thought about rewriting the arg processing, but didn't want to touch the existing functionality.  The `dialog` and (undocumented?) `free` args would need to be changed too, and I'm not sure how the getopt code would deal with passing an expression to be evaluated.

Could a getopt rewrite be done in a separate PR?

myk002 (commented):
one nit, otherwise looks good! (I like how you avoided the argparse problem : )

myk002:
side note: if you write ```
Fixes DFHack/dfhack#3171
```
instead of ```
This is for DFHack/dfhack#3171
```
in the description, it will automatically close the issue when this PR is merged, saving us a step.

gui/gm-editor.lua
lethosor: I didn't realize there were existing options. Disregard.

myk002 (approved):

PR #676, opened 2023-04-11
Combine seeds dyes

silverflyone:
Version for review.
Contains: Ability to combine seeds and powders (dyes, flour). Excludes sand and plaster items. Ability to combine items in containers; so that seeds, powders and others are combined into containers (bags). Updated logging to include container changes.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,16 +16,26 @@ that repo.
 ## New Scripts
 
 ## Fixes
-- `caravan`: fix trade good list sometimes disappearing when you collapse a bin
-- `gui/gm-editor`: no longer nudges last open window when opening a new one
+
+## Misc Improvements
+
+## Removed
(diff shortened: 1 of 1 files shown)


silverflyone:
@myk002 interested in your views. Is it worth asking others to help test?

myk002:
I plan to feature this for testing in the beta release channel after we do our stable release this Friday. We should get lots of testing there : )

silverflyone:
Awesome, the more the better. :)

A side effect of combining containers is that all types are included. It is QOL for seeds, dyes and flour, and I think also for the other types since it only applies to items already in containers. Thoughts? 

I'll have a look at doing bolts next; I'll raise another issue to track that one.

silverflyone:
pre-commit.ci autofix

PR #678, opened 2023-04-12
[prerelease-warning] parse our new interim release candidate naming scheme

myk002:
so now we can name our builds like `r2rc1` and not scare the user with dire warnings about mislabeled releases

### DIFF
--- a/gui/prerelease-warning.lua
+++ b/gui/prerelease-warning.lua
@@ -1,6 +1,3 @@
--- Shows the warning about missing configuration file.
-
-local gui = require 'gui'
 local dlg = require 'gui.dialogs'
 local json = require 'json'
 local utils = require 'utils'
@@ -19,12 +16,11 @@ if dfhack.internal.getAddress('gametype') and df.global.gametype == df.game_type
     return
 end
 
-local state = dfhack.getDFHackRelease():lower():match('[a-z]+')
+local state = dfhack.getDFHackRelease():lower():match('([a-z]+)%d*$')
 if not utils.invert{'alpha', 'beta', 'rc', 'r'}[state] then
     dfhack.printerr('warning: unknown release state: ' .. state)
     state = 'unknown'
 end
-
 
 message = ({
     alpha = {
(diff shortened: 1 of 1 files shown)


gui/prerelease-warning.lua
```diff
@@ -19,13 +16,12 @@ if dfhack.internal.getAddress('gametype') and df.global.gametype == df.game_type
     return
 end
 
-local state = dfhack.getDFHackRelease():lower():match('[a-z]+')
+local state = dfhack.getDFHackRelease():lower():match('([a-z]+)%d*$')
```
lethosor, line 19: Why not 
```suggestion
local state = dfhack.getDFHackRelease():lower():match('([a-z]+)%d+$')
```
? Do we want digits at the end to be optional?

myk002: that would match the previous logic, which did accept strings like "alpha" and "beta" and "rc" without trailing digits.

myk002: I didn't have a strong opinion on whether we wanted to prevent that, so I left it accepting the same set.

myk002:
merging so the rc I plan to push later tonight doesn't give a warning

gui/prerelease-warning.lua
lethosor: Well, actually, the previous pattern didn't match numbers _at all_. I think it was designed to just match the `alpha`, `beta`, or `r` in the release, since those were previously the only alphabetic characters.

Now that we can have `r` and `rc` in the same number, we might want to rethink the pattern here - i.e. is `r2rc1` a "prerelease" that warrants more caution than `r2`?

myk002: Yeah, I essentially changed it from a prefix match to a suffix match. Before, it matched `r|rc|alpha|beta` followed by a non lowercase letter followed by anything. Now it matches anything followed by one of those strings followed by (optional) digits. I did it that way so strings like `r2rc2` would result in `rc` and not `r`. I hadn't known about the different warnings the different prefixes would show, and ab9rf and I figured that the inter-stable "feedback" releases (that we name with the r2rc2 format) best matched the wording of the "rc" warning.

PR #684, opened 2023-04-18
[gui/gm-editor] add --freeze option for inhibiting calls to parent's logic()

myk002:
Fixes DFHack/dfhack#3265


myk002:
verified by Rumrusher

PR #691, opened 2023-04-24
modernize gui/prerelease-warning

myk002:
before:
image

after:
image

but now the widgets are clickable, the window is resizable, etc. The extra bit of space on the right is for the scrollbar for longer text messages.

Fixes DFHack/dfhack#3281

Depends on DFHack/dfhack#3285 (or the PR that I'll eventually factor this bit of code out into) for `ZScreenModal`


silverwing235:
@myk002 Uh, title typo? That, or inattention, IG?

PR #693, opened 2023-04-28
Refactor gui/design and move performance intensive code to C++

johncosker:
### Shape drawing code to C++
[#3324](https://github.com/DFHack/dfhack/pull/3324/) needs to be merged before this

The shape drawing code was slow in lua due to the number of boundary checks for large shapes, the C++ code is much more responsive.

### Added util.lua
Currently just includes definitions for a `Point` class, which allows adding/subtracting/equal checks for points without having to operate on each individual axis value


gui/design.lua
```diff
@@ -1576,8 +1385,8 @@ function Design:onInput(keys)
 
     -- Secret shortcut to kill the panel if it becomes
     -- unresponsive during development, should not release
-    -- if keys.CUSTOM_M then
-    --     self.parent_view:dismiss()
```
myk002, line 1580: there's also `devel/pop-screen` for this

myk002 (approved):

gui/design.lua
```diff
@@ -156,36 +152,30 @@ DISABLED_PEN_LEFT, DISABLED_PEN_CENTER, DISABLED_PEN_RIGHT,
...
-
-local function get_cur_filters()
-    return dfhack.buildings.getFiltersByType({}, uibs.building_type,
-        uibs.building_subtype, uibs.custom_type)
-end
-
 -- Debug window
 
 SHOW_DEBUG_WINDOW = false
+DEBUG_PROFILING = true
```
myk002, line 158: this isn't actually used, is it?

johncosker: Yeah that's leftover

PR #694, opened 2023-04-28
Update create-unit.lua

Thazzeus:
These changes fix the errors when running the script, although it doesn't correctly create a unit when running it.

### DIFF
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -161,25 +161,26 @@ function createUnitInner(race_id, caste_id, caste_id_choices, pos, locationChoic
   local cursor = copyall(df.global.cursor)
 
   local isArena = dfhack.world.isArena()
-  local arenaSpawn = df.global.world.arena_spawn
+  local arenaSpawn = df.global.game.main_interface.arena
+  local arenaALTSpawn = df.global.world.arena
 
   local oldSpawnType
   local oldSpawnFilter
-  oldSpawnType = arenaSpawn.type
-  arenaSpawn.type = 0 -- selects the creature at index 0 when the arena spawn screen is produced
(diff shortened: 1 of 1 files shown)


myk002 (commented):
as you discover the meaning of anon fields, be sure to add them to the structures:
https://github.com/DFHack/df-structures/blob/master/df.world.xml#L1702
https://github.com/DFHack/df-structures/blob/master/df.ui-menus.xml#L2350

myk002:
It looks like there is not enough keyboard control in DF yet to make this feasible. I made a GUI version in #173 (based on some of the changes you made in this PR, so thank you for that), which should make most people happy for now. this CLI might need to be put on hold until DF brings keyboard controls back.

Thazzeus:
Made more updates, create-unit.lua is "working" currently.

modtools/create-unit.lua
```diff
@@ -290,12 +291,12 @@ function createUnitInner(race_id, caste_id, caste_id_choices, pos, locationChoic
       end
     end
 
-    gui.simulateInput(dwarfmodeScreen, 'D_LOOK_ARENA_CREATURE') -- open the arena spawning menu
+    gui.simulateInput(dwarfmodeScreen, 'ARENA_CREATE_CREATURE') -- open the arena spawning menu
     local spawnScreen = dfhack.gui.getCurViewscreen() -- df.viewscreen_layer_arena_creaturest
-    gui.simulateInput(spawnScreen, 'SELECT') -- create the selected creature
+    gui.simulateInput(spawnScreen, 'ARENA_CREATE_CREATURE') -- create the selected creature
```
myk002, line 296: this works? it doesn't work for me when testing manually..

PR #695, opened 2023-05-02
Rename unforbid to claim

Penumbra69:
- unforbid.lua -> claim.lua
 - unforbid.rst -> claim.rst
 - update documentation to use claim / claimed
 - Update "forbid.lua" to hint at "claim all" instead of "unforbid all"
 - added a note to the changelog.txt

### DIFF
--- a/docs/claim.rst
+++ b/docs/claim.rst
@@ -1,11 +1,11 @@
-unforbid
-========
+claim
+=====
 
 .. dfhack-tool::
-    :summary: Unforbid all items.
+    :summary: Claim all items.
     :tags: fort productivity items
 
-This tool quickly and easily unforbids all items. This is especially useful
+This tool quickly and easily claims all items. This is especially useful
 after a siege to allow cleaning up the mess (or dumping of caged prisoner's
 equipment with `stripcaged`).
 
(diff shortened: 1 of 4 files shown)


docs/claim.rst
```diff
@@ -1,11 +1,11 @@
-unforbid
-========
+claim
```
myk002, line 1:
```suggestion
.. _unforbid:

claim
```

myk002: when we retire a script, we usually tombstone it in https://github.com/DFHack/dfhack/blob/develop/docs/about/Removed.rst, but I don't think that is warranted here. We tombstone it so the sphinx anchors don't disappear and the docs continue to build. In this case, though, I think we can just add an additional anchor for the `claim` script.

myk002:
the sphinx anchor has to be above a title. I think the verification script needs to be updated in this case.

The logic in this line would need to detect and skip over anchor definitions (lines starting with `..`): https://github.com/DFHack/dfhack/blob/develop/ci/script-docs.py#L45

Penumbra69:
I see now .. I haven't used Sphinx in a v. long time - I think I hive it sorted out this time.

PR #697, opened 2023-05-03
get exportlegends basically working

myk002:
no visual progress bar yet, so this does not implement DFHack/dfhack#2292

however, it does produce a usable xml file, which might be useful now that vanilla has some export functionality.

the vanilla "Export XML" button does not appear to be triggerable via the usual means (setting mouse gps coordinates and simulating a button click), so all the logic that keeps everything in a single directory is now less useful than it was. eventually, the button may be triggerable via a keybinding again, and we can restore much of the code removed in this PR.

as it is, the exported XML file is now written to the top level DF game dir since that's where the vanilla file is.

A better solution might be to wrap all this in a GUI and control the process CWD dynamically as the script walks the player through the steps, but that's a job for DFHack/dfhack#2292

### DIFF
--- a/open-legends.lua
+++ b/open-legends.lua
@@ -1,82 +1,44 @@
 -- open legends screen when in fortress mode
 --@ module = true
---[====[
 
-open-legends
-============
-Open a legends screen when in fortress mode. Requires a world loaded in fortress
-or adventure mode. Compatible with `exportlegends`.
+local dialogs = require('gui.dialogs')
+local gui = require('gui')
+local utils = require('utils')
 
-Note that this script carries a significant risk of save corruption if the game
-is saved after exiting legends mode. To avoid this:
+Restorer = defclass(Restorer, gui.Screen)
+Restorer.ATTRS{
+    focus_path='open-legends'
+}
 
-1. Pause DF
-2. Run `quicksave` to save the game
-3. Run `open-legends` (this script) and browse legends mode as usual
-4. Immediately after exiting legends mode, run `die` to quit DF without saving
-   (saving at this point instead may corrupt your save)
-
-Note that it should be safe to run "open-legends" itself multiple times in the
-same DF session, as long as DF is killed immediately after the last run.
-Unpausing DF or running other commands risks accidentally autosaving the game,
-which can lead to save corruption.
-
-The optional ``force`` argument will bypass all safety checks, as well as the
-save corruption warning.
-
-]====]
-
-local dialogs = require 'gui.dialogs'
-local gui = require 'gui'
-local utils = require 'utils'
-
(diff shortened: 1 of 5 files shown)


open-legends.lua
```diff
@@ -1,82 +1,44 @@
...
 
-function Wrapper:onInput(keys)
-    if self._native.parent.cur_page == 0 and keys.LEAVESCREEN then --hint:df.viewscreen_legendsst
-        local v = df.global.world.world_data.region_details
-        while (#v > 0) do v:erase(0) end
-        for _,item in pairs(region_details_backup) do
-            v:insert(0, item)
-        end
-        self:dismiss()
-        dfhack.screen.dismiss(self._native.parent)
```
lethosor, line 58: Per a discussion with Rumrusher on Discord: this dismiss() call may really be necessary, because DF's dismiss logic for this screen kicks the user back to the main menu. In previous versions, I recall it crashing instead - that's why the dismiss() call was added here.

Side note: that's why I named the class "Wrapper" instead of "Restorer". It does more than just restore the original screen/state: it also catches and overrides some native behavior to address issues like these.

open-legends.lua
```diff
@@ -1,82 +1,44 @@
...
+    while (#v > 0) do
+        table.insert(self.region_details_backup, 1, v[0])
+        v:erase(0)
+    end
 end
 
-function Wrapper:onHelp()
-    self._native.parent:help()
+function Restorer:onIdle()
+    self:dismiss()
```
lethosor, line 24: Why are we calling dismiss() in onIdle()?

myk002: that's it exactly -- the Restorer is underneath the Legends screen waiting to be exposed, at which point it would run once and dismiss itself. What you said in the other comment about the legends input logic for dismiss makes this not such a great idea, though.

open-legends.lua
myk002: The problem is that LEAVESCREEN is no longer the only way to exit, and the mouse-driven way is not easy to detect. This is why I rewrote the logic to not care how the screen was exited. the script crashed both before and after my changes, though, so I left the tags as "unavailable" and deferred research for a later time.

myk002: note we *may* be able to use a tactic similar to the `confirm` plugin for detecting (and intercepting) mouse clicks over the `Done` button.

lethosor: Yeah, that is tricky. The logic in viewscreen_legendsst that dismisses the screen is probably the same logic that kicks the user back to the main menu, so we probably can't just detect whether the screen's dismiss flag is set after user input is handled...

Adding detection for button clicks is probably more reliable, but tricky to make reliable, e.g. if it depends on pixel coordinates and we are off by one relative to DF.

PR #698, opened 2023-05-03
add modlist manager overlay script

lizreu:
this is a fairly straightforward script for managing mod lists in df

motivated by repeatedly enabling the same ~15 mods on every new world creation when tuning worldgen for too long

i contemplated posting this on workshop instead, but i figured it would make most sense to try and get a pr going here first

~~please let me know if this needs docs etc~~ docs added

screenshot


lizreu:
pre-commit.ci autofix

gui/modlistman.lua
```diff
@@ -0,0 +1,446 @@
...
+            label = "Mod Manager",
+            on_activate = function()
+                ModmanageScreen{}:show()
+            end,
+        },
+    }
+end
+
+function ModmanageOverlay:overlay_onupdate(viewscreen)
+    self.visible = is_modlist_visible()
```
myk002, line 418: Let me add a focus string that covers this case so you don't have to manage this manually.

gui/modlistman.lua
```diff
@@ -0,0 +1,446 @@
...
+local default_applied = false
+
+dfhack.onStateChange[GLOBAL_KEY] = function(sc)
+    if sc == SC_VIEWSCREEN_CHANGED then
+        local vs = get_newregion_viewscreen()
+        if vs and not default_applied then
+            default_applied = true
+            for i, v in ipairs(presets_file.data) do
+                if v.default then
+                    load_preset(i)
```
myk002, line 438: it would be nice to inform the player that their default settings have been successfully applied. popping up a message that has to be dismissed would kind of defeat the point of having a convenient default, though. What about if a message appears on the lower left of the screen when the new_region viewscreen first appears (and this code segment takes effect). the message could be cleared when the viewscreen is exited or a subscreen is entered.

gui/modlistman.lua
```diff
@@ -0,0 +1,446 @@
...
+ModmanageMenu = defclass(ModmanageMenu, widgets.Window)
+ModmanageMenu.ATTRS {
+    view_id = "modman_menu",
+    frame_title = "Modlist Manager",
+    frame_style = gui.WINDOW_FRAME,
+    frame_inset=0,
+
+    frame = { w = 40, t = 10, b = 15 },
+
+    resizable = true,
```
myk002, line 146: you should set `resize_min={h=15}` (or so) so the window can't resize so small that the list scrollbar becomes non-functional.

gui/modlistman.lua
myk002, line 1: @warmist is it ok if we retire the existing `gui/mod-manager` so that this script can take the name?

docs/gui/modlistman.rst
```diff
@@ -0,0 +1,17 @@
...
+Adds an optional overlay to the mod list screen that
+allows you to save and load mod list presets, as well
+as set a default mod list preset for new worlds.
+
+Usage
+-----
+
+::
+
+    gui/modlistman
```
myk002, line 17: right now, there's nothing that happens when the script is "run" (since the entrypoint is currently just the overlay). it would be useful if this script provided a list of what mods are currently enabled when run when a fort is loaded, but that doesn't need to be done as part of this PR.

gui/modlistman.lua
```diff
@@ -0,0 +1,446 @@
...
+        return true
+    end
+end
+
+ModmanageMenu = defclass(ModmanageMenu, widgets.Window)
+ModmanageMenu.ATTRS {
+    view_id = "modman_menu",
+    frame_title = "Modlist Manager",
+    frame_style = gui.WINDOW_FRAME,
+    frame_inset=0,
```
myk002, line 142: it is more common to leave this at the default of `1` and use child widget frames that have `0` offsets. In general, `frame_inset` is only set to 0 if you're doing something fancy with interior frames.

myk002 (commented):
this is absolutely wonderful. excellent job!

warmist:
Yup, you can use mod-manager if you want.

myk002:
> Yup, you can use mod-manager if you want.

Thanks!

docs/gui/modlistman.rst
lizreu: i've looked into grabbing the modlist of an active fort but came up with squat, unfortunately

gui/modlistman.lua
lizreu: bottom-left, stays up for 5 seconds, is a separate overlay
included in latest commit

myk002: perfect. Thanks!

docs/gui/modlistman.rst
myk002: I've found a reliable list in `df.global.world.object_loader`, which is what I use to dynamically generate the script path for mods.

Some code here might be refactorable: https://github.com/DFHack/dfhack/blob/develop/library/lua/script-manager.lua#L135

myk002:
Have you had time to finish this up? This script is going to be very useful for DF players!

lizreu:
> Have you had time to finish this up? This script is going to be very useful for DF players!

@myk002 should be ready now

gui/mod-manager.lua
```diff
@@ -1,17 +1,16 @@
--- Simple modlist manager
+-- Save and restore lists of active mods.
 --@ module = true
 
 local argparse = require('argparse')
```
myk002, line 4:
```suggestion
```

gui/mod-manager.lua
```diff
@@ -1,365 +1,462 @@
...
 end
-function manager:onInput(keys)
 
-    if keys.LEAVESCREEN  then
-        self:dismiss()
-    else
-        self:inputToSubviews(keys)
+ModmanageScreen = defclass(ModmanageScreen, gui.ZScreen)
+ModmanageScreen.ATTRS {
+    focus_path = "modman_screen",
```
myk002, line 366:
```suggestion
    focus_path = "mod-manager",
```

myk002 (approved):

gui/mod-manager.lua
```diff
@@ -1,365 +1,461 @@
...
+
+function ModmanageScreen:init()
+    self:addviews{
+        ModmanageMenu{}
+    }
+end
+
+ModmanageOverlay = defclass(ModmanageOverlay, overlay.OverlayWidget)
+ModmanageOverlay.ATTRS {
+    frame = { w=30, h=1 },
```
myk002, line 377:
```suggestion
    frame = { w=16, h=3 },
    frame_style = gui.MEDIUM_FRAME,
```

myk002: Forgot this one: DFHack hotkeys need to be clearly labeled as coming from DFHack

myk002:
@lizreu if you're interested, would you like to take a crack at https://github.com/DFHack/dfhack/issues/3746 ?

PR #701, opened 2023-05-15
update light-aquifers-only

myk002:
and add it to the gui/control-panel autostart options list

Fixes DFHack/dfhack#3368

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -24,7 +24,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/create-item`: ask for number of items to spawn by default
-- `light-aquifers-only`: now available as an fort autostart option in `gui/control-panel`
+- `light-aquifers-only`: now available as an fort autostart option in `gui/control-panel`. note that it will only appear if "armok" tools are configured to be shown on the Preferences tab.
 
 ## Removed
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,7 +16,7 @@ that repo.
 ## New Scripts
 - `exportlegends`: reinstated: export extended legends information for external browsing
 - `modtools/create-item`: reinstated: commandline and API interface for creating items
-- `light-aquifers-only`: reinstated: convert heavy aquifers to light (available as an fort autostart option in `gui/control-panel`)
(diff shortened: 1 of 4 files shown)


changelog.txt
```diff
@@ -16,6 +16,7 @@ that repo.
 ## New Scripts
 - `exportlegends`: export extended legends information for external browsing
 - `modtools/create-item`: commandline and API interface for creating items
+- `light-aquifers-only`: convert heavy aquifers to light (available as an fort autostart option in `gui/control-panel`)
```
lethosor, line 19: Perhaps this could go under Improvements instead?

Actually, none of these are new, strictly speaking - they are making a return in v50, but it looks like light-aquifers-only in particular probably would have worked before this PR too. I flagged these as `reinstated` in the dfhack/dfhack changelog.txt, which I guess could work here too. Or we could add a new section.

I think the control-panel capability could stand as a separate improvement, either way.

myk002: @ab9rf and I had this discussion a few weeks ago too. We were adding these to "New Scripts" since we thought it fit the intent of this section, as far as our current users are concerned. "reinstated" sounds appropriate too. I can add that in here.

docs/light-aquifers-only.rst
```diff
@@ -17,15 +17,15 @@ Usage
...
+If you don't ever want to have to deal with heavy aquifers, you can enable the
+``light-aquifers-only`` command in the "Autostart" tab of `gui/control-panel`
+so it will be run automatically whenever you start a new fort.
 
 Technical details
 -----------------
 
 When run pre-embark, this script changes the drainage of all world tiles that
-would generate Heavy aquifers into a value that results in Light aquifers
+would generate heavy aquifers into a value that results in Light aquifers
```
lethosor, line 28: Might as well fix the other one too :)
```suggestion
would generate heavy aquifers into a value that results in light aquifers
```

changelog.txt
lethosor: Presenting them to new users like this makes sense to me, but I didn't want to suggest that they were new to everyone. Glad `reinstated` sounds good

lethosor (approved):
looks good to me

PR #706, opened 2023-05-18
Add devel/input-monitor

lethosor:
This is sometimes useful for troubleshooting. Opening a PR to remind me to get it into the main repo.

Draft because I know it needs to be brought up to modern standards, particularly for docs. It could probably also use a slight UI revamp. (I don't plan on making it a ZScreen because I think the uniform screen helps with the mouse tracking feature, but I'll probably add some additional explanatory text.)

### DIFF
--- a/devel/input-monitor.lua
+++ b/devel/input-monitor.lua
@@ -1,0 +1,91 @@
+-- Mouse/keyboard monitor
+
+DFH_MOD_SHIFT = 1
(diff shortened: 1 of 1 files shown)


myk002:
I didn't see this when I wrote my own `devel/input-monitor`:
image

myk002:
Are there elements in this PR that could be integrated into the now-existing `devel/input-monitor`?

PR #707, opened 2023-05-19
[quickfort] accept more tile types for placing machines

myk002:
Strictly, these buildings need a gear assembly nearby to support them, but I think quickfort can assume that you're doing it "right" (otherwise the machine will deconstruct after being built).

Fixes DFHack/dfhack#3350

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@ that repo.
 
 ## Fixes
 - `gui/design`: Fix building and stairs designation
+- `quickfort`: fixed detection of tiles where machines are allowed (e.g. water wheels *can* be built on stairs if there is a machine support nearby)
 
 ## Misc Improvements
 - `gui/create-item`: ask for number of items to spawn by default
--- a/internal/quickfort/build.lua
+++ b/internal/quickfort/build.lua
@@ -92,7 +92,9 @@ local function is_valid_tile_has_space_or_is_ramp(pos)
 end
 
 local function is_valid_tile_machine(pos)
-    return is_valid_tile_has_space_or_is_ramp(pos)
+    local shape = df.tiletype.attrs[dfhack.maps.getTileType(pos)].shape
+    local basic_shape = df.tiletype_shape.attrs[shape].basic_shape
+    return is_valid_tile_has_space_or_is_ramp(pos) or basic_shape == df.tiletype_shape_basic.Stair
 end
 
 -- ramps are ok everywhere except under the anchor point of directional bridges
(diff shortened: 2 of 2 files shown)


fricy:
It looks like they don't even deconstruct. The only restriction seems to be that they require a supporting machinery to be designated first. They don't care if it ends up being built, or cancelled.

myk002:
I think this PR is the best I can do. I can't check for the *pre-existence* of a gear since it might be in the same blueprint and it just hasn't been placed yet. This PR would technically allow you to let a horizontal axel/water wheel cross stairs even without a gear, but you could do this in vanilla too by designating a gear, designating the axel/wheel, then removing the gear.

fricy:
I was trying to test the changes, but I'm not sure these PR-s have been published.

 I'm updated to 50.08-r2rc1, but most of the files in scripts/internal/quickfort are dated march 31. Except command.lua (april 15) and dig.lua timestaped yesterday (14 hours ago). 

Steam says I'm on the beta dfhack channel, and validates all files.

myk002:
This PR didn't make it into 50.08-r2rc1, but we do have a newer testing build up on Steam that you could subscribe to. Enter "launchertest" as a beta branch password and then select the "launcher_test" branch. This PR is included with that build.

fricy:
Thx!
Looks good so far. I suppose not much can be done about the trackstops cancelling a carve designation?

Edit:
Just found this in the docs:

>In #build blueprints, there are a few building types that will generate extra manager orders for related materials:
 Track stops will generate an order for a minecart

I'm not seeing the extra orders, so I assume that part doesn"t work at the moment. But, I'd really like a way to opt out of that. :)

myk002:
> Looks good so far. I suppose not much can be done about the trackstops cancelling a carve designation?

I believe the fix for that also in the testing release. Not just trackstops, but all buildings were fixed. Quickfort can now designate (passable) floor tiles under buildings for smoothing and carving (including tracks)

myk002:
> I'm not seeing the extra orders, so I assume that part doesn"t work at the moment. But, I'd really like a way to opt out of that. :)

Orders will be handed very differently once this is done. They will be generated and managed by `buildingplan` instead of `quickfort`. There will also be a `gui/buildingplan` where you'll be able to configure orders behavior like this.

fricy:
> I believe the fix for that also in the testing release. Not just trackstops, but all buildings were fixed. Quickfort can now designate (passable) floor tiles under buildings for smoothing and carving (including tracks)

Yes, that fix also works, thank you.

However there's a different vanilla issue:

If you start adding features (rollers, trackstops, etc) to a designated, but yet unfinished track the game will cancel the carving order on those tiles if the buildings are finished before the track. With your fix I can at least reapply the track blueprint, and it fixes the track everywhere, Before I had to find each tile and designate by hand. The drawback is that the dwarves will recarve every tile of the track this way, not just the ones missing.

I filed an issue for now https://github.com/DFHack/dfhack/issues/3407

> plans are in

I'll check those out. 
That reminded me to add an extra idea to the quickfort/pressure plates ticket....

PR #715, opened 2023-05-22
[suspendmanager] Protect floor designation from being erased

plule:
Fixes https://github.com/DFHack/dfhack/issues/3407

In vanilla DF, floor designations (track carving, smoothing, engraving) order are erased if a building is built on top of it before the job is done. However, if the designation is completed before the building, then both are correctly done.

This PR makes suspendmanager suspend constructions that are on top of floor designations, so that the dwarf can do this kind of construction unattended.

https://github.com/DFHack/scripts/assets/630159/0085373f-4ea1-4e92-b451-a14069b89cc8

Internally, it reworks suspendmanager to have a class. This helps managing multi-step operations, for example here detecting the designations require to loop on the designation jobs, and store the layered jobs. The same kind of situation will occur when implementing dead-end detection.

### DIFF
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -29,15 +29,8 @@ SuspendManager.ATTRS {
     jobsOnDesignation = {},
 }
 
-function SuspendManager.new(preventBlocking)
-    local manager = SuspendManager {
-        preventBlocking = preventBlocking
-    }
-    return manager
-end
-
 -- SuspendManager instance kept between frames
-Instance = Instance or SuspendManager.new(true)
+Instance = Instance or SuspendManager{preventBlocking=true}
 
 function isEnabled()
     return enabled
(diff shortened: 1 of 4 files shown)


suspendmanager.lua
```diff
@@ -16,23 +16,41 @@ end
...
+    --- Jobs that are on a tile with a designation (smooth, carve track, engrave)
+    jobsOnDesignation = {},
+}
+
+function SuspendManager.new(preventBlocking)
+    local manager = SuspendManager {
+        preventBlocking = preventBlocking
+    }
+    return manager
+end
```
myk002, line 37: if there's no singleton, what is the advantage of hiding instantiation behind a method? Normally, clients would call SuspendManager{} directly.

suspendmanager.lua
```diff
@@ -212,17 +243,40 @@ function shouldStaySuspended(job, accountblocking)
     end
 
     -- Internal reasons to be suspended, determined by suspendmanager
-    return shouldBeSuspended(job, accountblocking)
+    return self:shouldBeSuspended(job)
+end
+
+function SuspendManager:refresh()
+    self.jobsOnDesignation = {}
```
myk002, line 250: note that some smooth/carve designations will still just be designations if they haven't been converted in to jobs yet. then you'd check the `smooth` field of the tile flags for smoothing and engraving and the occupancy.carve_track_* fields for carving tracks.

you can see examples of working with these values here: https://github.com/DFHack/scripts/blob/master/internal/quickfort/dig.lua (or https://github.com/DFHack/dfhack/blob/develop/plugins/dig-now.cpp)

suspendmanager.lua
plule: Yep, I'm just so used to languages with constructors that I did the same here - removed

suspendmanager.lua
plule: Added the check on designation/occupancy too, thanks

plule:
closing - will reopen after https://github.com/DFHack/scripts/pull/721 that's just focused on the class refacto

PR #735, opened 2023-06-08
Update emigration.lua

wsfsbvchr:
Followup to https://github.com/DFHack/dfhack/issues/1784#issuecomment-1566146480

Fix for emigration.lua not working since the citizenship logic was changed.

It works, based on some testing by myself in 0.47.05. It can be tested relatively easy with stress-inducing scripts, such as this one I created while testing this: https://github.com/wsfsbvchr/dfhack-utils/blob/main/stress-tests.lua

Some edits might be needed, dunno.

### DIFF
--- a/emigration.lua
+++ b/emigration.lua
@@ -1,5 +1,4 @@
 --Allow stressed dwarves to emigrate from the fortress
--- Updated for 0.47.05 by wsfsbvchr
 -- For 34.11 by IndigoFenix; update and cleanup by PeridexisErrant
 -- old version:  http://dffd.bay12games.com/file.php?id=8404
 --[====[
--- a/emigration.lua
+++ b/emigration.lua
@@ -1,8 +1,8 @@
 --Allow stressed dwarves to emigrate from the fortress
+-- Updated for 0.47.05 by wsfsbvchr
 -- For 34.11 by IndigoFenix; update and cleanup by PeridexisErrant
 -- old version:  http://dffd.bay12games.com/file.php?id=8404
 --[====[
-
 emigration
 ==========
 Allows dwarves to emigrate from the fortress when stressed,
@@ -16,7 +16,6 @@ A happy dwarf (ie with negative stress) will never emigrate.
 Usage::
 
     emigration enable|disable
-
 ]====]
 
 enabled = enabled or false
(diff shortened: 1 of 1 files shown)


emigration.lua
```diff
@@ -39,45 +38,114 @@ function desireToStay(unit,method,civ_id)
...
+	
+	-- free owned rooms
+	for i = #u.owned_buildings-1, 0, -1 do
+		local temp_bld = df.building.find(u.owned_buildings[i].id)
+		dfhack.buildings.setOwner(temp_bld, nil)
+	end
+
+	-- erase the unit from the fortress entity
+	for k,v in pairs(fort_ent.histfig_ids) do
+		if tonumber(v) == hf_id then
```
myk002, line 71: in what cases will this not already be a number?

emigration.lua
```diff
@@ -39,45 +38,114 @@ function desireToStay(unit,method,civ_id)
...
+	local newsite_id = -1
+	
+	-- free owned rooms
+	for i = #u.owned_buildings-1, 0, -1 do
+		local temp_bld = df.building.find(u.owned_buildings[i].id)
+		dfhack.buildings.setOwner(temp_bld, nil)
+	end
+
+	-- erase the unit from the fortress entity
+	for k,v in pairs(fort_ent.histfig_ids) do
```
myk002, line 70: while `pairs` likely still works, the semantics of `ipairs` are more appropriate. here and in all the loops below

emigration.lua
```diff
@@ -39,45 +38,114 @@ function desireToStay(unit,method,civ_id)
...
+	for k,v in pairs(hf.entity_links) do
+		if v._type == df.histfig_entity_link_memberst and v.entity_id == fort_ent.id then
+			hf.entity_links:erase(k)
+			break
+		end
+	end
+
+	-- try to find a new entity for the unit to join
+	for k,v in pairs(civ_ent.entity_links) do
+		if v.type == 1 and v.target ~= fort_ent.id then
```
myk002, line 101: what is the meaning of `1`? if it's an enum value, what's the enum?

emigration.lua
```diff
@@ -39,45 +38,114 @@ function desireToStay(unit,method,civ_id)
...
 end
 
 function canLeave(unit)
     if not unit.status.current_soul then
         return false
     end
 
-    for _, skill in pairs(unit.status.current_soul.skills) do
-        if skill.rating > 14 then return false end
-    end
```
myk002, line 69: yeah, this probably makes sense to remove. even masters can get unhappy

emigration.lua
```diff
@@ -105,13 +173,15 @@ function checkmigrationnow()
         and not unit.flags1.tame
         then
             if unit.flags1.merchant then table.insert(merchant_civ_ids, unit.civ_id) end
-            if unit.flags1.diplomat then table.insert(diplomat_civ_ids, unit.civ_id) end
+            --if unit.flags1.diplomat then table.insert(diplomat_civ_ids, unit.civ_id) end
```
myk002, line 176: no need to keep commented-out code. can remove

emigration.lua
```diff
@@ -105,13 +173,15 @@ function checkmigrationnow()
...
     end
-
+    
+    if #merchant_civ_ids == 0 and #diplomat_civ_ids == 0 then
+        checkForDeserters('wild', df.global.ui.main.fortress_entity.entity_links[0].target)
+    end
     for _, civ_id in pairs(merchant_civ_ids) do checkForDeserters('merchant', civ_id) end
-    for _, civ_id in pairs(diplomat_civ_ids) do checkForDeserters('diplomat', civ_id) end
-    checkForDeserters('wild', -1)
+    --for _, civ_id in pairs(diplomat_civ_ids) do checkForDeserters('diplomat', civ_id) end
```
myk002, line 184: remove

myk002 (commented):
I reviewed this version of the code since it is clearer than the PR meant for `master`, but I'm not sure if we'll merge it. we don't intend to make another 0.47.05 release unless there is a severe issue with the current release, like a crash bug. I'll have to confer with the other devs whether there are any down sides to merging it into the v0.47.05 branch anyway, just so the improvements are saved.

please convert tabs to spaces so the formatter check passes.

emigration.lua
wsfsbvchr: It's df.entity_entity_link_type.CHILD, the other one is df.entity_site_link_type.Claim

emigration.lua
wsfsbvchr: Probably none, yeah.

wsfsbvchr:
I think I did all the changes you suggested + it also passes the checks.

lethosor:
I don't have any particular issues with making changes to the 0.47.05 branch (we have already done it in the other two repos), but I agree that it is unlikely to ever be released formally.

wsfsbvchr:
> No one's aware that we have a duplicate of this? (#736) What's to be done about it, then?

It's the same fix to the other branch, because the v0.47 script doesn't work in 50 and vice versa.

emigration.lua
```diff
@@ -1,8 +1,8 @@
 --Allow stressed dwarves to emigrate from the fortress
+-- Updated for 0.47.05 by wsfsbvchr
```
myk002, line 2:
```suggestion
```

myk002: removing this line as discussed in the other review. Please do add your name to Authors.rst, though!

myk002 (approved):

myk002:
the build actions in this branch depend on an EOL version of ubuntu, which is why they aren't completing (I think). merging anyway since pre-commit.ci passes and no docs were changed.

PR #736, opened 2023-06-08
Update emigration.lua

wsfsbvchr:
Followup to https://github.com/DFHack/dfhack/issues/1784#issuecomment-1566146480

Fix for emigration.lua not working since the citizenship logic was changed.

It works at least in 0.47.05, based on some testing by myself. It can be tested relatively easy with stress-inducing scripts, such as this one I created while testing this: https://github.com/wsfsbvchr/dfhack-utils/blob/main/stress-tests.lua

The 50.* version seems to have some different things in the plugin part of the script, so at least adding those might be necessary. Also the exact location of unit's stress has apparently changed?

### DIFF
--- a/emigration.lua
+++ b/emigration.lua
@@ -1,29 +1,35 @@
 --Allow stressed dwarves to emigrate from the fortress
--- Updated for 0.47.05 and potentially for 50.08 by wsfsbvchr
 -- For 34.11 by IndigoFenix; update and cleanup by PeridexisErrant
 -- old version:  http://dffd.bay12games.com/file.php?id=8404
---@module = true
---@enable = true
+--[====[
+emigration
+==========
+Allows dwarves to emigrate from the fortress when stressed,
+in proportion to how badly stressed they are and adjusted
+for who they would have to leave with - a dwarven merchant
+being more attractive than leaving alone (or with an elf).
+The check is made monthly.
 
-local json = require('json')
-local persist = require('persist-table')
+A happy dwarf (ie with negative stress) will never emigrate.
 
-local GLOBAL_KEY = 'emigration' -- used for state change hooks and persistence
+Usage::
+
+    emigration enable|disable
+]====]
 
 enabled = enabled or false
 
-function isEnabled()
-    return enabled
-end
-
-local function persist_state()
-    persist.GlobalTable[GLOBAL_KEY] = json.encode({enabled=enabled})
+local args = {...}
+if args[1] == "enable" then
+    enabled = true
+elseif args[1] == "disable" then
+    enabled = false
 end
 
 function desireToStay(unit,method,civ_id)
     -- on a percentage scale
     local value = 100 - unit.status.current_soul.personality.stress / 5000
-    if method == 'merchant' then
+    if method == 'merchant' or method == 'diplomat' then
         if civ_id ~= unit.civ_id then value = value*2 end end
     if method == 'wild' then
         value = value*5 end
@@ -47,7 +53,7 @@ function desert(u,method,civ)
 
     local hf_id = u.hist_figure_id
     local hf = df.historical_figure.find(u.hist_figure_id)
-    local fort_ent = df.global.ui.main.fortress_entity
+    local fort_ent = df.global.plotinfo.main.fortress_entity
     local civ_ent = df.historical_entity.find(hf.civ_id)
 
     local newent_id = -1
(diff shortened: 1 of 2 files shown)


myk002:
I'm having a hard time following the changes in this PR. it looks like you've overwritten all the improvements made since 0.47. could you integrate your changes into the existing code instead?

wsfsbvchr:
Of course there's still the fact that I haven't tested if it works in 50.* as well as in 0.47.*.

emigration.lua
```diff
@@ -36,16 +37,91 @@ function desert(u,method,civ)
...
+    local fort_ent = df.global.ui.main.fortress_entity
+    local civ_ent = df.historical_entity.find(hf.civ_id)
+
+    local newent_id = -1
+    local newsite_id = -1
+
+    -- free owned rooms
+    for i = #u.owned_buildings-1, 0, -1 do
+        local temp_bld = df.building.find(u.owned_buildings[i].id)
+        dfhack.buildings.setOwner(temp_bld, nil)
```
myk002, line 59: side note: I got the `dfhack.buildings.setOwner` function working for v50 just last week. good timing.

myk002:
code looks reasonable (and professional, I might add). could you tell us when you've had a chance to do some testing in v50 to make sure it actually works as advertised? In particular, I'm not confident I understand the long-term implications of assigning a unit to another site.

wsfsbvchr:
I'd imagine they live their lives like any of the other histfigs in those non-player site entities. I got the idea for that while experiencing the bug where the dwarves just immediately return after leaving the map; I figured if they're members in another place, they surely will go there. It's admittedly bit cheaty though.

I'm also not sure what happens without any new entity links. I suppose they might settle somewhere anyway.

Oh, and, I just tried v50 in wine, and apparently in v50 `df.global.ui` doesn't exist, so I'll need to figure out where it stores the current fortress entity.

wsfsbvchr:
Apparently it's df.global.plotinfo. v50 feels so clunky and complicated, but looks like there are no errors now, and two dwarves succesfully left my newly started fortress in search for better life.

emigration.lua
```diff
@@ -1,22 +1,29 @@
 --Allow stressed dwarves to emigrate from the fortress
 -- For 34.11 by IndigoFenix; update and cleanup by PeridexisErrant
 -- old version:  http://dffd.bay12games.com/file.php?id=8404
---@module = true
---@enable = true
```
myk002, line 5: these two lines are required for the enabled API to function. see https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#script-enable-api for details.

emigration.lua
```diff
@@ -1,22 +1,29 @@
...
+--[====[
```
myk002, line 4: docs have moved to separate .rst files. If you made any changes to the docs, they should go in https://github.com/DFHack/scripts/blob/master/docs/emigration.rst
script files shouldn't have duplicate docs since they just get out of date.

emigration.lua
```diff
@@ -1,22 +1,29 @@
...
+enabled = enabled or false
 
-local function persist_state()
-    persist.GlobalTable[GLOBAL_KEY] = json.encode({enabled=enabled})
+local args = {...}
+if args[1] == "enable" then
+    enabled = true
+elseif args[1] == "disable" then
+    enabled = false
 end
```
myk002, line 27: this part looks like it was accidentally copied from an earlier version. Could you fix the merge?

emigration.lua
```diff
@@ -112,42 +187,17 @@ local function event_loop()
...
-if args[1] == "enable" then
-    enabled = true
-elseif args[1] == "disable" then
-    enabled = false
-else
-    return
-end
-
-event_loop()
-persist_state()
```
myk002, line 153: this part also looks like it was overwritten with old code

emigration.lua
wsfsbvchr: Ah, I must've uploaded the wrong file or something. Now the correct file should be there.

emigration.lua
wsfsbvchr: Sorry!

emigration.lua
wsfsbvchr: I don't think anything relevant to the docs has changed.

Though I don't know if the part in bold is technically true: "Dwarves who can leave in friendly company (e.g. a dwarven merchant caravan) will choose to do so, but extremely stressed dwarves can choose to leave alone, **or even in the company of a visiting elven bard!**", but I suppose it hasn't been true in many years.

Perhaps it's not intended literally, but as a description of how despaired some of the emigrating dwarves must be?

wsfsbvchr:
I think it should now have the v50 base with my updated emigration solutions, but I can't focus at the moment so here's to hoping.

myk002 (approved):
looks good. thanks!

PR #740, opened 2023-06-13
[suspendmanager] Run less often

plule:
Low hanging fruits for improving suspendmanager performances: the `eventful.onJobInitiated`/`eventful.onJobCompleted` are fired quite more often than I thought, and it's not necessary to refresh the suspension each time.

Add two filters:
- Ensure it only refreshes at most once per simulation tick (using `cur_year_tick`, maybe there is another?)
- Don't refresh on job types that have no impact on suspendmanager


myk002:
In addition, you might want to make sure it runs no more frequently than once a second (or X seconds). There's code you can copy to do this here: https://github.com/DFHack/dfhack/blob/ff8daa1de86956da964263a8021eb5880fb68dba/plugins/lua/overlay.lua#L444

suspendmanager.lua
```diff
@@ -146,6 +150,17 @@ local ERASABLE_DESIGNATION = {
...
+--- Job types that impact suspendmanager
+local FILTER_JOB_TYPES = {
+    [df.job_type.CarveTrack]=true,
+    [df.job_type.ConstructBuilding]=true,
+    [df.job_type.DestroyBuilding]=true,
+    [df.job_type.DetailFloor]=true,
+    [df.job_type.Dig]=true,
+    [df.job_type.DigChannel]=true,
+    [df.job_type.SmoothFloor]=true,
+}
```
myk002, line 162: this is absolutely fine to write it this way, but just so you're aware, many other tools take advantage of `utils.invert` like this:
```lua
local FILTER_JOB_TYPES = utils.invert{
    df.job_type.CarveTrack,
    df.job_type.ConstructBuilding,
    df.job_type.DestroyBuilding,
    df.job_type.DetailFloor,
    df.job_type.Dig,
    df.job_type.DigChannel,
    df.job_type.SmoothFloor,
}
```

on another note, I'm not sure I understand this list. Are we looking for the list of job types that impact pathability? Looking through the header file, I came up with this:
```c
        Dig,
        CarveUpwardStaircase,
        CarveDownwardStaircase,
        CarveUpDownStaircase,
        CarveRamp,
        DigChannel,
        FellTree,
        RemoveConstruction,
        ConstructBuilding,
        RemoveStairs,
```

plule: ooooh I never quite understood this `utils.invert`, now I get it!

And for the purpose of this list, yes, that would be all the job impacting the access to jobs, essentially to filter out things like Urist finished his nap. I've added those.

plule:
> In addition, you might want to make sure it runs no more frequently than once a second (or X seconds). There's code you can copy to do this here: https://github.com/DFHack/dfhack/blob/ff8daa1de86956da964263a8021eb5880fb68dba/plugins/lua/overlay.lua#L444

I'd like to avoid this actually, in my very initial tests of suspendmanager where I was just using `repeat`, setting a higher delay than ~20 ticks was starting to cause issues of dwarves still managing to block themselves. In other words, if dwarves are indeed completing 10 walls a second, there is a high chance that they will end up being stuck if the suspension is not refreshed, since there are likely a LOT of dwarves working at the same time.

suspendmanager.lua
```diff
@@ -151,14 +151,23 @@ local ERASABLE_DESIGNATION = {
...
+--- access to construction job
+local FILTER_JOB_TYPES = utils.invert{
+    df.job_type.CarveRamp,
+    df.job_type.CarveTrack,
+    df.job_type.CarveUpDownStaircase,
+    df.job_type.CarveUpwardStaircase,
+    df.job_type.CarveDownwardStaircase,
+    df.job_type.ConstructBuilding,
+    df.job_type.DestroyBuilding,
+    df.job_type.DetailFloor,
```
myk002, line 164: Why are DetailFloor and SmoothFloor in this list? How do they affect access?

myk002:
> > In addition, you might want to make sure it runs no more frequently than once a second (or X seconds).
> 
> since there are likely a LOT of dwarves working at the same time.

that's fair

suspendmanager.lua
plule: I've expanded the comment, it's to properly update the ERASE_DESIGNATION suspension reason, a smooth/carving/detailing job will require any job on top of it to be suspended to avoid being erased.

In gif form, this chair is about to be built, but I add at the last second a smoothing job. Completing the chair would erase the smoothing job, so it suspends the chair. If SmoothFloor was not part of the filter, then the suspension would not have happened immediately
suspendonsmooth

myk002 (approved):

PR #741, opened 2023-06-14
remove gui/automelt

myk002:
will be replaced more thoroughly by gui/logistics, but right now it's just unrunnable cruft


myk002:
merging and will fix the docs error in a Removed.rst entry

PR #745, opened 2023-06-18
[gui/gm-unit] use string-based hotkeys instead of action keybindings

myk002:
where we really want static symbols

this allows keybindings to behave correctly in a SDL2 world


lethosor:
To clarify, this is just changing how the keys _display_, right? e.g. `+` instead of `Plus`

myk002:
the interface keys were coming in as strings, not as `KEYBOARD_CURSOR_*_Z`. The reason for this is because shift-symbol keys are not matched by DF. if `+` actually triggered `KEYBOARD_CURSOR_UP_Z` properly, then `gui/gm-unit` wouldn't be having this problem.

PR #759, opened 2023-06-30
[caravan] advanced trade dialogs

myk002:
still some work to do, especially with the UI design, but putting this up for feedback

image

image

image

@TaxiService any advice on layout?

depends on DFHack/dfhack#3504

further ideas:

- [x] filter for elf-safe items
- [ ] subtractive substring matching (in FilteredList, look for search tokens that start with `-`)
- [x] filter for things your nobles might mandate soon (according to their preferences)
- [x] include items already in the depot; deselecting them would be equivalent to "move from depot" -- not sure what this means in the data structures though.
- [x] value filter range slider: [0, 20, 50, 100, 500, 1000, Max]
- [x] differentiate tool artifacts by type (especially quires/books)


plule:
It looks incredibly useful.

ctrl-v for "select all" is a bit weird: ctrl-a?

It's possibly just me but I don't understand what the shift-click "range of item" does just from the text

myk002:
yes, I'm changing that to Ctrl-A as soon as EditField stops hogging that hotkey. Also, EditField is about to *start* hogging the Ctrl-V hotkey to mean "paste"

yeah, shift-click might need a better description.

the way it works is:

1. select some item with the mouse or keyboard
2. shift-click some other item. all items from the previously selected item to the one you shift-clicked on will be changed to the opposite state of the one you shift-clicked on. e.g. if you shift-clicked on an item that is not currently being traded, they'll all become marked for trade.

How can this be expressed clearly and concisely?

plule:
Ok, it seems pretty obvious now... Honestly I have a hard time finding a better wording, it's probably best like this. I think I was just being obtuse, and that's a standard interaction that will be understood in a second by people not understanding/reading the text.

myk002:
merging for alpha testing

PR #769, opened 2023-07-11
New script: fix/empty-wheelbarrows

master-spike:
Added a script fix/empty-wheelbarrows.lua, which checks for wheelbarrows with items inside them on the ground and not in a job and empties them in the same manner as empty-bin. Wheelbarrows are currently bugged in that if a dwarf is interrupted while hauling or hauls a rock to the wrong stockpile, the rock wont be removed from the wheelbarrow unless directly used in a job (non-hauling). This behaviour eventually can lead to accumulation of rocks in wheelbarrows assigned to feeder stockpiles.

This script works by iterating over all the tools in the map, checking if they're a wheelbarrow, on the ground, and NOT in use by a job. Then for each item in the wheelbarrow, moves it to the ground on the same tile.


fix/empty-wheelbarrows.lua
```diff
@@ -0,0 +1,21 @@
...
+
+
+for _,e in ipairs(df.global.world.items.other.TOOL) do
+    -- wheelbarrow must be on ground and not in a job
+    if ((not e.flags.in_job) and e.flags.on_ground) then
+        if e.subtype.id == "ITEM_TOOL_WHEELBARROW" then
+            local items = dfhack.items.getContainedItems(e)
+            print('Emptying wheelbarrow: ' .. dfhack.items.getDescription(e, 0))
+            if #items > 0 then
+                print('Emptying wheelbarrow: ' .. dfhack.items.getDescription(e, 0))
```
ab9rf, line 13: prints on L11 and L13 are the same, is this intended?

also, having a quiet option would be beneficial for people who might want to run this job automatically via repeat, so an option to suppress nonerror output would likely be desirable

fix/empty-wheelbarrows.lua
```diff
@@ -0,0 +1,21 @@
+--checks all wheelbarrows on map for rocks stuck in them. If a wheelbarrow isn't in use for a job (hauling) then there should be no rocks in them
+--rocks will occasionally get stuck in wheelbarrows, and accumulate if the wheelbarrow gets used.
+--this script empties all wheelbarrows which have rocks stuck in them.
+
+
+for _,e in ipairs(df.global.world.items.other.TOOL) do
+    -- wheelbarrow must be on ground and not in a job
+    if ((not e.flags.in_job) and e.flags.on_ground) then
+        if e.subtype.id == "ITEM_TOOL_WHEELBARROW" then
```
ab9rf, line 9: using the `isWheelbarrow()` item vmethod for this test would be more resistant to someone who changes the raws

technically any tool with the `HEAVY_OBJECT_HAULING` tool use flag is a "wheelbarrow"

ab9rf (changes requested):

fix/empty-wheelbarrows.lua
master-spike: Nope, the first print is a mistake on my part. I'll add quiet options too

master-spike: That's a mistake, L11 shouldn't be there. I'll implement quiet options as well as well as a dry run option.

fix/empty-wheelbarrows.lua
master-spike: Will adjust to use `e:isWheelbarrow()`

fix/empty-wheelbarrows.lua
```diff
@@ -2,20 +2,36 @@
...
+            for _,i in ipairs(items) do
+                if (not quiet) then print('  ' .. dfhack.items.getDescription(i, 0)) end
+                if (not dryrun) then dfhack.items.moveToGround(i, e.pos) end
+                i_count = i_count + 1
             end
         end
     end
 end
+
+print("fix/empty-wheelbarrows - removed " .. i_count .. " items from " .. e_count .. " wheelbarrows.")
```
ab9rf, line 37: this print should probably also be gated by the `quiet` flag

ab9rf (approved):

fix/empty-wheelbarrows.lua
master-spike: done this

fix/empty-wheelbarrows.lua
```diff
@@ -18,20 +18,24 @@ end
 local i_count = 0
 local e_count = 0
 
+local function printNotQuiet(str)
+    if (not quiet) then print(str) end
+end
```
ab9rf, line 23: 👍

fix/empty-wheelbarrows.lua
```diff
@@ -0,0 +1,41 @@
...
+local quiet = false
+local dryrun = false
+for i,arg in ipairs(args) do
+    if (arg == "--quiet") then
+        quiet = true
+    end
+    if (arg == "--dryrun") then
+        dryrun = true
+    end
+end
```
lethosor, line 16: I would recommend using `utils.processArgsGetopt()` for this. Among other benefits, it supports "short" options like `-q` (if specified) and produces errors for unrecognized options.

master-spike: So something like this?

```
local cmds = argparse.processArgsGetopt(args, {
    {'q', 'quiet', handler=function() quiet = true end},
    {'n', 'dryrun', handler=function() dryrun = true end},
})
```

myk002:
`-n` and `--dryrun` are perfectly reasonable, but to match precedent set by other DFHack tools, it should be `-d` and `--dry-run`

docs/fix/empty-wheelbarrows.rst
```diff
@@ -0,0 +1,33 @@
+fix/empty-wheelbarrows
+======================
+
+.. dfhack-tool::
+    :summary: Empties stuck items from wheelbarrows
```
myk002, line 5: needs to end in a period to be parsed by the in-game help engine.

docs/fix/empty-wheelbarrows.rst
```diff
@@ -0,0 +1,33 @@
...
+    :tags: fort bugfix items
+
+Empties all wheelbarrows which contain rocks that have become 'stuck' in them.
+
+This works around the issue encountered with :bug:`6074`, and should be run
+if you notice wheelbarrows lying around with rocks in them that aren't
+being used in a task.
```
myk002, line 12: is there any downside to periodically running this tool in the background? If not, then you should add it to the Maintenance list in `gui/control-panel` and to the default-enabled list in https://github.com/DFHack/dfhack/blob/develop/data/dfhack-config/init/onMapLoad.control-panel-repeats.init

docs/fix/empty-wheelbarrows.rst
```diff
@@ -0,0 +1,33 @@
...
+
+
+Usage
+-----
+
+``fix/empty-wheelbarrows [options]``
+
+-q, --quiet    surpress console output
+-d, --dry-run  dry run, don't commit changes
```
myk002, line 21: these should go in an `Options` section after the `Examples` section. See https://github.com/DFHack/scripts/blob/master/docs/build-now.rst for an example Options section

docs/fix/empty-wheelbarrows.rst
```diff
@@ -0,0 +1,33 @@
...
+
+This works around the issue encountered with :bug:`6074`, and should be run
+if you notice wheelbarrows lying around with rocks in them that aren't
+being used in a task.
+
+
+Usage
+-----
+
+``fix/empty-wheelbarrows [options]``
```
myk002, line 18:
```suggestion
::

    fix/empty-wheelbarrows [<options>]
```

docs/fix/empty-wheelbarrows.rst
```diff
@@ -0,0 +1,33 @@
...
+--------
+
+``fix/empty-wheelbarrows``
+    Empties all items, listing all wheelbarrows emptied and their contents
+``fix/empty-wheelbarrows --dry-run``
+    Lists all wheelbarrows that would be emptied and their contents without performing the action.
+``fix/empty-wheelbarrows --quiet``
+    Does the action while surpressing output to console
+``repeat --name empty-wheelbarrows --time 1200 command [ fix/empty-wheelbarrows --quiet ]``
+    Runs empty-wheelbarrows quietly every 1200 game ticks, which is once per in-game day.
```
myk002, line 33: this command can be added to `gui/control-panel` so it can be enabled with a button click (and as I referenced above, it can be set to be enabled by default for new users)

fix/empty-wheelbarrows.lua
```diff
@@ -0,0 +1,40 @@
...
+            for _,i in ipairs(items) do
+                printNotQuiet('  ' .. dfhack.items.getDescription(i, 0))
+                if (not dryrun) then dfhack.items.moveToGround(i, e.pos) end
+                i_count = i_count + 1
+            end
+        end
+    end
+end
+
+printNotQuiet(("fix/empty-wheelbarrows - removed %d items from %d wheelbarrows."):format(i_count, e_count))
```
myk002, line 40: to behave more like similar fix tools (e.g. https://github.com/DFHack/scripts/blob/master/fix/general-strike.lua ) the status line should be printed even if quiet as long as i_count > 0

myk002 (commented):
remember to also add a line to `changelog.txt` announcing this new script!

docs/fix/empty-wheelbarrows.rst
master-spike: Are the angle brackets required? I've had the idea they were mainly used for required arguments as opposed to optional ones

fix/empty-wheelbarrows.lua
master-spike: Aight, done this.

docs/fix/empty-wheelbarrows.rst
master-spike: thanks, done

docs/fix/empty-wheelbarrows.rst
master-spike: The issue it solves is only really an issue if you're like me and use a lot of different stone stockpiles and feeders, if people don't use wheelbarrows but have maybe 3000 jugs or other tools I don't know how performance will be impacted as I haven't measured this. If I wanted to add it there I suppose I'd have to open another PR over on the dfhack develop repo?

docs/fix/empty-wheelbarrows.rst
master-spike: Under the maintenance tab I presume?

docs/fix/empty-wheelbarrows.rst
master-spike: Added period.

docs/fix/empty-wheelbarrows.rst
myk002: angle brackets indicate that the word between the brackets is a description. just `[options]` means "you can optionally (`[]`) add the literal word `options`". `[<options>]` means "you can optionally add various options here".

docs/fix/empty-wheelbarrows.rst
myk002: another PR just for the addition of the default enable in `onMapLoad.control-panel-repeats.init`. `gui/control-panel.lua` is in the scripts repo and can be modified in this PR.

docs/fix/empty-wheelbarrows.rst
myk002: that looks right, though I'd change the `1200` to 
```lua
        command={'--time', '1', '--timeUnits', 'days', '--command', '[', 'fix/stuck-wheelbarrows', '-q', ']'}},
```
for clarity

myk002: also, once it's integrated in `gui/control-panel`, I'd remove the repeat example here and instead add a note to the introduction text that you can run it automatically by toggling the Maintenance task in `gui/control-panel`.

master-spike: Alrighty, done all this. For anyone else reading this, I did type `fix/stuck-wheelbarrows` instead of `fix/empty-wheelbarrows` in this conversation - I have since noticed and that mistake hasn't carried over to the code.

docs/fix/empty-wheelbarrows.rst
master-spike: Aighty, that's good to know. I've updated usage accordingly.

myk002 (approved):

myk002:
Absolute gold star for your patience! Thank you very much for this!

PR #771, opened 2023-07-16
suspendmanager job overlay

plule:
Add a job overlay to `suspendmanager`. When suspendmanager is enabled, it hides the "Suspend/Resume construction" button with a text overlay and include the toggle for enabling and disabling `suspendmanager`. This overlay either explains why the job is suspended or simply indicate that it is not suspended.

The reason for overlaying the button is that once enabled, suspendmanager will not take in account anymore any choice that the player makes here. This overlay makes this explicit, while providing feedback on its choice and a way to act on it.

image

image

image

image

No overlay at all for buildingplan jobs in the waiting list.

Fixes https://github.com/DFHack/dfhack/issues/3413

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -25,6 +25,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/unit-syndromes`: make lists searchable
+- `suspendmanager`: display the suspension reason
 
 ## Removed
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -25,7 +25,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/unit-syndromes`: make lists searchable
-- `suspendmanager`: display the suspension reason
+- `suspendmanager`: display the suspension reason when viewing a suspended building
 
 ## Removed
 
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -581,7 +581,7 @@ end
 function StatusOverlay:get_status_string()
     local job = dfhack.gui.getSelectedJob()
     if job and job.flags.suspend then
-        return "Suspended because: " .. Instance:suspensionDescription(job) or "External interruption" .. "."
+        return "Suspended because: " .. Instance:suspensionDescription(job) .. "."
     end
     return "Not suspended."
 end
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -404,11 +404,6 @@ function SuspendManager:suspensionDescription(job)
     end
     local reason = self.suspensions[job.id]
     return reason and REASON_DESCRIPTION[reason] or "External interruption"
---    if not reason then
---        return "External Interruption"
---    end
---
---    return REASON_DESCRIPTION[reason] or "External Interruption"
 end
 
 --- Recompute the list of suspended jobs
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -397,7 +397,7 @@ function SuspendManager:shouldStaySuspended(job)
 end
 
 --- Return a human readable description of why suspendmanager keeps a job suspended
---- or nil if the job is not kept suspended
+--- or "External interruption" if the job is not kept suspended by suspendmanager
 function SuspendManager:suspensionDescription(job)
     local reason = self.suspensions[job.id]
     return reason and REASON_DESCRIPTION[reason] or "External interruption"
@@ -581,7 +581,7 @@ end
 function StatusOverlay:get_status_string()
     local job = dfhack.gui.getSelectedJob()
     if job and job.flags.suspend then
-        return "Suspended because: " .. Instance:suspensionDescription(job) .. "."
+        return "Suspended because: " .. Instance:suspensionDescription(job) or "External interruption" .. "."
     end
     return "Not suspended."
 end
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -51,7 +51,7 @@ REASON_TEXT = {
 --- suspend jobs
 REASON_DESCRIPTION = {
     [REASON.RISK_BLOCKING] = 'May block another build job',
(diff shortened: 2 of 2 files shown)


changelog.txt
```diff
@@ -25,6 +25,7 @@ that repo.
 
 ## Misc Improvements
 - `gui/unit-syndromes`: make lists searchable
+- `suspendmanager`: display the suspension reason
```
myk002, line 28:
```suggestion
- `suspendmanager`: display the suspension reason when viewing a suspended building
```

suspendmanager.lua
```diff
@@ -43,6 +46,15 @@ REASON_TEXT = {
     [REASON.DEADEND] = 'dead end',
 }
 
+--- Description of suspension
+--- This only cover the reason where suspendmanager actively
+--- suspend jobs
+REASON_DESCRIPTION = {
+    [REASON.RISK_BLOCKING] = 'Risk blocking another',
+    [REASON.ERASE_DESIGNATION] = 'On a tile designation',
+    [REASON.DEADEND] = 'Blocking a dead-end'
```
myk002, line 55: from conversation on discord, suggest:
- May block another build job
- Waiting for carve/smooth/engrave job on same tile
- Blocks another build job

suspendmanager.lua
```diff
@@ -378,6 +390,20 @@ function SuspendManager:shouldStaySuspended(job)
     return self.suspensions[job.id]
 end
 
+--- Return a human readable description of why suspendmanager keeps a job suspended
+--- or nil if the job is not kept suspended
+function SuspendManager:suspensionDescription(job)
```
myk002, line 395: It might be more consistent to always have a message on suspended buildings. If a player clicks on a suspended building to discover the reason, and nothing is displayed, it might be confusing. Instead of returning nil, perhaps a message like "Unknown (not suspended by DFHack suspendmanager)"?

suspendmanager.lua
```diff
@@ -531,3 +557,42 @@ end
...
+    frame={w=30, h=5},
+    frame_style=gui.MEDIUM_FRAME,
+    frame_background=gui.CLEAR_PEN,
+}
+
+function JobOverlay:init()
+    self:addviews{
+        widgets.Label{
+            frame={t=0, l=0},
+            text='Staying suspended:'
```
myk002, line 576: suggest: 'Suspended because:'

suspendmanager.lua
```diff
@@ -531,3 +557,42 @@ end
...
+        },
+        widgets.Label{
```
myk002, line 578: can use a `widgets.WrappedLabel` for automatic wrapping, but if you do, you also need to force it to rewrap when the message changes. This means adding a bit of logic to detect when the message changes in `render`.

suspendmanager.lua
```diff
@@ -531,3 +557,42 @@ end
...
+        }
+    }
+end
+
+function JobOverlay:get_reason_string()
+    return Instance:suspensionDescription(dfhack.gui.getSelectedJob())
+end
+
+function JobOverlay:render(dc)
+    if not isEnabled() or not self:get_reason_string() then
```
myk002, line 590: the suspension reason message should probably still be displayed even when `suspendmanager` is not enabled -- the building could have been suspended by `suspendmanager` and then the player temporarily disabled `suspendmanager`.

if `suspendmanager` is disabled, it might be good to put a HotkeyLabel here that enables `suspendmanager` (as long as the button isn't too intrusive or pushy)

suspendmanager.lua
```diff
@@ -531,3 +557,42 @@ end
...
+        return
+    end
+    JobOverlay.super.render(self, dc)
+end
+
+OVERLAY_WIDGETS = {
+    inspector=JobOverlay
```
myk002, line 597: maybe "status" instead of "inspector" (actually, "status" would have been a better name for the `buildingplan` overlay too..)

suspendmanager.lua
plule: Oh, I like the idea of the button here. It goes both way too, to allow disabling it from here. If a player is annoyed by `suspendmanager` suspending a job they get both a reason why and a way to bypass the behavior, or if they actually want to manually suspend some jobs, this is the first place they will search for too.

As far as always displaying the reason, I don't think it a good idea, when suspendmanager is disabled the reason will become out of date, maybe the player even unsuspended and resuspended the job. Essentially when suspendmanager is disabled, we don't have a clue of the reason of suspension.

suspendmanager.lua
plule: Can you spot the mindless copy/paste? :D

suspendmanager.lua
plule: I cut off the "on same tile", with the current overlay I have a single line of real estate

suspendmanager.lua
plule: Agreed, I went for the more concise "Suspended because: External interruption.". This feels vague and right enough I think?

plule:
The last update moved the overlay on top of the suspend button, and added the toggle. It's much more intrusive, but I really like the flow it gives - let me know if this does not feel too pushy here

myk002 (approved):
looks great! thanks!

suspendmanager.lua
```diff
@@ -531,3 +555,81 @@ end
...
+                {text=self:callback('get_status_string')}
+            }
+        },
+    }
+end
+
+function StatusOverlay:get_status_string()
+    local job = dfhack.gui.getSelectedJob()
+    if job and job.flags.suspend then
+        return "Suspended because: " .. Instance:suspensionDescription(job) or "External interruption" .. "."
```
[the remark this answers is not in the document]
plule: oops that's right, thanks

suspendmanager.lua
plule: not needed anymore (and only one line of real-estate)

myk002 (approved):

PR #782, opened 2023-07-31
[quickfort] don't error if the map block is unallocated

myk002:
no changelog since this is essentially only a cosmetic fix and the issue can only be triggered in very special circumstances (map block completely unallocated)


lethosor (approved):
Looks reasonable from a static review

PR #784, opened 2023-08-03
reuse reusable workflos from the dfhack repo

myk002:
refactors docs and lint actions to use the reusable versions
adds test action


myk002:
new actions don't seem to run in this PR. merging to see if they'll run in a new PR

PR #786, opened 2023-08-05
devel/lsmem: add filtering capabilities

lethosor:
It's now possible to list ranges containing a given address, or ranges matching a given filename.

Also mark the script as available.


myk002 (approved):
I toggled the available flag on the spreadsheet tracker too

lethosor:
I keep forgetting about the spreadsheet! Thanks.

PR #793, opened 2023-08-10
Add devel/scan-vtables

lethosor:
Linux-only for now

Requires https://github.com/DFHack/dfhack/pull/3657


lethosor:
Successfully tested on both of the new 2023-08-08 Linux builds; merging

PR #797, opened 2023-08-12
Make `workorder-recheck` work again

TymurGubayev:
This repairs the `workorder-recheck.lua` script and adds an overlay to the work order conditions screen.

image

### DIFF
--- a/workorder-recheck.lua
+++ b/workorder-recheck.lua
@@ -1,11 +1,4 @@
 -- Resets the selected work order to the `Checking` state
---[====[
-
-workorder-recheck
-=================
-Sets the status to ``Checking`` (from ``Active``) of the selected work order.
-This makes the manager reevaluate its conditions.
-]====]
 
 --@ module = true
 
--- a/workorder-recheck.lua
+++ b/workorder-recheck.lua
@@ -3,18 +3,66 @@
 
 workorder-recheck
 =================
-Sets the status to ``Checking`` (from ``Active``) of the selected work order (in the ``j-m`` or ``u-m`` screens).
+Sets the status to ``Checking`` (from ``Active``) of the selected work order.
 This makes the manager reevaluate its conditions.
 ]====]
-local scr = dfhack.gui.getCurViewscreen()
-if df.viewscreen_jobmanagementst:is_instance(scr) then
-    local orders = df.global.world.manager_orders
-    local idx = scr.sel_idx
-    if idx < #orders then
-        orders[idx].status.active = false
+
+--@ module = true
+
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local overlay = require('plugins.overlay')
+
+local function set_current_inactive()
+    local scrConditions = df.global.game.main_interface.info.work_orders.conditions
+    if scrConditions.open then
+        local order = scrConditions.wq
+        order.status.active = false
     else
-        qerror("Invalid order selected")
+        qerror("Order conditions is not open")
     end
-else
-    qerror('Must be called on the manager screen (j-m or u-m)')
 end
+
+-- -------------------
+-- RecheckOverlay
+--
+
+local focusString = 'dwarfmode/Info/WORK_ORDERS/Conditions'
+
+RecheckOverlay = defclass(RecheckOverlay, overlay.OverlayWidget)
+RecheckOverlay.ATTRS{
+    default_pos={x=6,y=2},
+    default_enabled=true,
+    viewscreens=focusString,
+    frame={w=17, h=3},
+    frame_style=gui.MEDIUM_FRAME,
+    frame_background=gui.CLEAR_PEN,
+}
+
+function RecheckOverlay:init()
+    self:addviews{
+        widgets.HotkeyLabel{
+            frame={t=0, l=0},
+            label='recheck',
+            key='CUSTOM_CTRL_A',
+            on_activate=set_current_inactive,
+        },
+    }
+end
+
+-- -------------------
+
+OVERLAY_WIDGETS = {
+    recheck=RecheckOverlay,
+}
+
+if dfhack_flags.module then
+    return
+end
+
+-- Check if on correct screen and perform the action if so
+if not dfhack.gui.matchFocusString(focusString) then
+    qerror('workorder-recheck must be run from the manager order conditions view')
+end
+
+set_current_inactive()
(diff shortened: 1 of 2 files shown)


workorder-recheck.lua
```diff
@@ -3,18 +3,66 @@
 
 workorder-recheck
 =================
-Sets the status to ``Checking`` (from ``Active``) of the selected work order (in the ``j-m`` or ``u-m`` screens).
+Sets the status to ``Checking`` (from ``Active``) of the selected work order.
 This makes the manager reevaluate its conditions.
 ]====]
```
myk002, line 8: docs have moved to docs/workorder-recheck.rst. could you make the edits there? Also, please remove the "unavailable" tag in the rst file since now the script is available : P

Please also remove the docstring here in this file -- it's left over from the .rst migration and will just get out of sync with the real docs.

workorder-recheck.lua
```diff
@@ -3,18 +3,66 @@
...
+
+--@ module = true
+
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local overlay = require('plugins.overlay')
+
+local function set_current_inactive()
+    local scrConditions = df.global.game.main_interface.info.work_orders.conditions
+    if scrConditions.open then
```
myk002, line 18: this is the condition that would allow the overlay to be shown or the focus string to match below. I suppose it doesn't hurt terribly to check it again, but I don't think it's necessary.

workorder-recheck.lua
```diff
@@ -3,18 +3,66 @@
...
+
+-- -------------------
+-- RecheckOverlay
+--
+
+local focusString = 'dwarfmode/Info/WORK_ORDERS/Conditions'
+
+RecheckOverlay = defclass(RecheckOverlay, overlay.OverlayWidget)
+RecheckOverlay.ATTRS{
+    default_pos={x=6,y=2},
```
myk002, line 34: we normally try to put the overlays on the panel itself or in the empty space to the right, but in this case those positions would be very far from the related vanilla status icon. I think the position you chose is probably the best choice, but we may get feedback from beta testers on this and end up moving it later, depending on how it interacts with vanilla elements on various screen sizes.

workorder-recheck.lua
```diff
@@ -3,18 +3,66 @@
...
+    frame={w=17, h=3},
+    frame_style=gui.MEDIUM_FRAME,
+    frame_background=gui.CLEAR_PEN,
+}
+
+function RecheckOverlay:init()
+    self:addviews{
+        widgets.HotkeyLabel{
+            frame={t=0, l=0},
+            label='recheck',
```
myk002, line 46: Though I understand the need for making this label short, "recheck" might be too short for players to grok. I'd want some input from people on the Discord server about what would be best here.

myk002 (commented):
Welcome back! I've honestly missed your work!

myk002:
another option is the new `widgets.TextButton` that is designed to appear in more cramped spaces. It could fit in the single line directly above the active/inactive status icon.

screenshots of how that might look here: https://github.com/DFHack/dfhack/pull/3655

TymurGubayev:
> another option is the new `widgets.TextButton` that is designed to appear in more cramped spaces. It could fit in the single line directly above the active/inactive status icon.

is there a way to find where elements are on a screen? I discovered the tabs can be 1 or 2 lines, which makes positioning tricky

workorder-recheck.lua
TymurGubayev: I added it in case something went terribly wrong. I guess the only possible scenario is if the overlay would somehow be active when it shouldn't...

TymurGubayev:
with `default_pos={x=6,y=10},` and `TextButton`, it looks like this with two rows of tabs

image

but maximizing the DF window results in this:

image

myk002:
You can try this:
- make the overlay 2 tiles tall
- on render, check the position of the status icon. If different from the last time you checked, adjust the TextButton frame and call updateLayout()

You can use dfhack.screen.readTile() to get what's on the screen at that position. The position you check should be relative to your overlay self.frame_rect (iirc)

TymurGubayev:
that would work if `readTile` worked... Here is what I get for every coordinate:
> [lua]# ~ dfhack.screen.readTile(10, 10, false)
> userdata: 00000208D89089B8
> bold                     = true
> bg                       = 0
> tile_color               = false
> ch                       = 0
> fg                       = 7

And passing `true` returns the map tile.

Am I missing something?

myk002:
Hmm, I'm not sure. What does devel/inspect-screen say for those tiles?

myk002:
ha, oh I see:
image
the image is completely rendered in the lower layer, which readTile doesn't pick up. Let me see if I can extend readTile to work in that case

myk002:
for now, you can test in ASCII mode, which works:
image

myk002:
you know what, maybe test for text (the `ch` property) at coordinates (7,8). That will be a character from a tab label if there are two rows of tabs and will be blank otherwise.

myk002:
https://github.com/DFHack/dfhack/pull/3670 will improve the `readTile` situation, but as I wrote above, I don't think you'll actually need it for this PR.

TymurGubayev:
the textbox is now correctly positioned in all 4 cases ( { graphics, ASCII } x { one row, two rows} )

workorder-recheck.lua
```diff
@@ -1,20 +1,97 @@
...
+end
+
+function RecheckOverlay:init()
+    self:addviews{
+        widgets.TextButton{
+            view_id = 'button',
+            -- frame={t=0, l=0, r=0, h=1}, -- is set in `updateTextButtonFrame()`
+            label='recheck',
+            key='CUSTOM_CTRL_A',
+            on_activate=set_current_inactive,
```
myk002, line 62: This button should probably be disabled if the workorder is already inactive. e.g. add:
```lua
enabled=is_current_active,
```

workorder-recheck.lua
myk002: how about "re-check conditions" for this label? or maybe "request manager review"?

TymurGubayev: "request conditions re-check"?

TymurGubayev: or just "request re-check"?

TymurGubayev: image

I definitely want to keep the `check` there because it's everywhere on that screen

myk002 (approved):
This is awesome. Thanks!

PR #798, opened 2023-08-12
use dynamic texture mechanism

shevernitskiy:
Will work with https://github.com/DFHack/dfhack/pull/3663 only.

### DIFF
--- a/gui/civ-alert.lua
+++ b/gui/civ-alert.lua
@@ -118,12 +118,12 @@ end
 last_tp_start = last_tp_start or 0
 CONFIG_BUTTON_PENS = CONFIG_BUTTON_PENS or {}
 local function get_button_pen(idx)
-    local start = dfhack.textures.getAsset("hack/data/art/control-panel.png", 0)
+    local start = dfhack.textures.getAsset('hack/data/art/control-panel.png', 0)
     if last_tp_start == start then return CONFIG_BUTTON_PENS[idx] end
     last_tp_start = start
 
     local tp = function(offset)
-        local texpos = dfhack.textures.getAsset("hack/data/art/control-panel.png", offset)
+        local texpos = dfhack.textures.getAsset('hack/data/art/control-panel.png', offset)
         if texpos == -1 then return nil end
         return texpos
     end
--- a/gui/civ-alert.lua
+++ b/gui/civ-alert.lua
@@ -118,18 +118,13 @@ end
 last_tp_start = last_tp_start or 0
 CONFIG_BUTTON_PENS = CONFIG_BUTTON_PENS or {}
 local function get_button_pen(idx)
-    local start = dfhack.textures.getAsset('hack/data/art/control-panel.png', 0)
+    local start = gui.tp_control_panel(1)
     if last_tp_start == start then return CONFIG_BUTTON_PENS[idx] end
     last_tp_start = start
 
-    local tp = function(offset)
-        local texpos = dfhack.textures.getAsset('hack/data/art/control-panel.png', offset)
-        return texpos >= 0 and texpos or nil
-    end
-
-    CONFIG_BUTTON_PENS[1] = to_pen{fg=COLOR_CYAN, tile=tp(6), ch=string.byte('[')}
-    CONFIG_BUTTON_PENS[2] = to_pen{tile=tp(9), ch=15} -- gear/masterwork symbol
-    CONFIG_BUTTON_PENS[3] = to_pen{fg=COLOR_CYAN, tile=tp(7), ch=string.byte(']')}
+    CONFIG_BUTTON_PENS[1] = to_pen{fg=COLOR_CYAN, tile=gui.tp_control_panel(7), ch=string.byte('[')}
+    CONFIG_BUTTON_PENS[2] = to_pen{tile=gui.tp_control_panel(10), ch=15} -- gear/masterwork symbol
+    CONFIG_BUTTON_PENS[3] = to_pen{fg=COLOR_CYAN, tile=gui.tp_control_panel(8), ch=string.byte(']')}
 
     return CONFIG_BUTTON_PENS[idx]
 end
--- a/gui/civ-alert.lua
+++ b/gui/civ-alert.lua
@@ -118,13 +118,14 @@ end
 last_tp_start = last_tp_start or 0
 CONFIG_BUTTON_PENS = CONFIG_BUTTON_PENS or {}
 local function get_button_pen(idx)
-    local start = dfhack.textures.getControlPanelTexposStart()
+    local start = dfhack.textures.getAsset("hack/data/art/control-panel.png", 0)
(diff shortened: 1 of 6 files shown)


gui/civ-alert.lua
```diff
@@ -118,13 +118,14 @@ end
...
+    local start = dfhack.textures.getAsset("hack/data/art/control-panel.png", 0)
     if last_tp_start == start then return CONFIG_BUTTON_PENS[idx] end
     last_tp_start = start
 
     local tp = function(offset)
-        if start == -1 then return nil end
-        return start + offset
+        local texpos = dfhack.textures.getAsset("hack/data/art/control-panel.png", offset)
+        if texpos == -1 then return nil end
+        return texpos
```
myk002, line 128:
```suggestion
        return texpos >= 0 and texpos or nil
```

gui/control-panel.lua
```diff
@@ -153,12 +153,10 @@ local function save_file(path, save_fn)
...
-    local start = dfhack.textures.getControlPanelTexposStart()
-    local valid = start > 0
-    start = start + 10
-
-    local function tp(offset)
-        return valid and start + offset or nil
+    local tp = function(offset)
+        local texpos = dfhack.textures.getAsset("hack/data/art/control-panel.png", offset + 10)
+        if texpos == -1 then return nil end
+        return texpos
```
myk002, line 159:
```suggestion
        return texpos >= 0 and texpos or nil
```

gui/cp437-table.lua
```diff
@@ -6,10 +6,10 @@ local widgets = require('gui.widgets')
 
 local to_pen = dfhack.pen.parse
 
-local tb_texpos = dfhack.textures.getThinBordersTexposStart()
 local tp = function(offset)
-    if tb_texpos == -1 then return nil end
-    return tb_texpos + offset
+    local texpos = dfhack.textures.getAsset("hack/data/art/border-thin.png", offset)
+    if texpos == -1 then return nil end
+    return texpos
```
myk002, line 12:
```suggestion
        return texpos >= 0 and texpos or nil
```

gui/design.lua
```diff
@@ -115,31 +115,33 @@ end
 -- Utilities
 
 local function get_icon_pens()
-    local start = dfhack.textures.getControlPanelTexposStart()
-    local valid = start > 0
-    start = start + 10
+    local tp = function(offset)
+        local texpos = dfhack.textures.getAsset("hack/data/art/control-panel.png", offset + 10)
+        if texpos == -1 then return nil end
+        return texpos
```
myk002, line 121:
```suggestion
        return texpos >= 0 and texpos or nil
```

gui/launcher.lua
```diff
@@ -473,21 +473,24 @@ function MainPanel:postUpdateLayout()
...
-local LEFT_SPLIT_PEN = dfhack.pen.parse{tile=gui.WINDOW_FRAME.l_frame_pen.tile,
-        ch=199, fg=COLOR_GREY, bg=COLOR_BLACK}
-local RIGHT_SPLIT_PEN = dfhack.pen.parse{tile=tp(17), ch=180, fg=COLOR_GREY, bg=COLOR_BLACK}
+    return texpos
+end
+
+local tp_window = function(offset)
+    local texpos = dfhack.textures.getAsset("hack/data/art/border-window.png", offset)
+    if texpos == -1 then return nil end
+    return texpos
```
myk002, line 485:
```suggestion
        return texpos >= 0 and texpos or nil
```

gui/launcher.lua
```diff
@@ -473,21 +473,24 @@ function MainPanel:postUpdateLayout()
...
-local H_SPLIT_PEN = dfhack.pen.parse{tile=tp(5), ch=196, fg=COLOR_GREY, bg=COLOR_BLACK}
-local V_SPLIT_PEN = dfhack.pen.parse{tile=tp(4), ch=179, fg=COLOR_GREY, bg=COLOR_BLACK}
-local TOP_SPLIT_PEN = dfhack.pen.parse{tile=gui.WINDOW_FRAME.t_frame_pen.tile,
-        ch=209, fg=COLOR_GREY, bg=COLOR_BLACK}
-local BOTTOM_SPLIT_PEN = dfhack.pen.parse{tile=gui.WINDOW_FRAME.b_frame_pen.tile,
-        ch=207, fg=COLOR_GREY, bg=COLOR_BLACK}
-local LEFT_SPLIT_PEN = dfhack.pen.parse{tile=gui.WINDOW_FRAME.l_frame_pen.tile,
-        ch=199, fg=COLOR_GREY, bg=COLOR_BLACK}
-local RIGHT_SPLIT_PEN = dfhack.pen.parse{tile=tp(17), ch=180, fg=COLOR_GREY, bg=COLOR_BLACK}
+    return texpos
```
myk002, line 479: `        return texpos >= 0 and texpos or nil`

unsuspend.lua
```diff
@@ -135,17 +135,11 @@ function SuspendOverlay:refresh_screen_buildings()
...
-
-    local function tp(offset)
-        return valid and start + offset or nil
-    end
-
-    return tp(3), tp(2), tp(1), tp(0)
+local tp = function(offset)
+    local texpos = dfhack.textures.getAsset("hack/data/art/unsuspend", offset)
+    if texpos == -1 then return nil end
+    return texpos
```
myk002, line 141:
```suggestion
        return texpos >= 0 and texpos or nil
```

myk002 (commented):
I think I prefer this API to what we had before. I'm not sure if it's perfect, but it's absolutely better. Just stylistic notes, but otherwise looks good.

could you use single quotes around the asset paths? we've mostly standardized on single quotes for strings in our Lua code.

myk002:
wait -- after writing that suggestion of `        return texpos >= 0 and texpos or nil` so many times, something struck me -- why don't we just have `getAsset` return nil if the texture isn't found/isn't valid? that would clean up the Lua usage patterns. If it's still useful for it to return -1 in cpp, then there can be a translation at the LuaApi.cpp layer.

shevernitskiy:
> wait -- after writing that suggestion of ` return texpos >= 0 and texpos or nil` so many times, something struck me -- why don't we just have `getAsset` return nil if the texture isn't found/isn't valid? that would clean up the Lua usage patterns. If it's still useful for it to return -1 in cpp, then there can be a translation at the LuaApi.cpp layer.

That's right!
Cpp api should do all centralized work, lua should just use.
Maybe it can be the next iteration of rework textures? Refactor? There i will need some help:)
(if my cpp level is "look, it is compiles", but lua level is "what the heck, what is this? where are braces? where are types?!") :))

shevernitskiy:
As I mentioned in main PR, it will be useful and more clear to move assets loading to lua plugin, where we can declare callbacks to get specific asset more centralized. Other plugins and scripts will just use this callbacks.
And all checks could move to cpp and/or to that asset init part.

myk002 (approved):

PR #799, opened 2023-08-13
Make `gui/workorder-details.lua` work again

TymurGubayev:
image

image

image


TymurGubayev:
issues:

- [ ] `Ctrl+D` opens another DFHack screen as well
- [ ] I haven't found a better way to hide the widget when in the wrong context, other than `self.subviews.button.visible = false` in `onRenderBody`

docs/gui/workorder-details.rst
```diff
@@ -9,8 +9,7 @@ This tool allows you to adjust item types, materials, and/or traits for items
 used in manager workorders. The jobs created from those workorders will inherit
 the details.
```
myk002, line 10: the introductory text could use a few more examples of things you can do with this tool that you wouldn't be able to do in vanilla. what do you think are the most useful situations where this tool helps?

docs/gui/workorder-details.rst
myk002, line 1: remember to remove the 'unavailable' tag and mention the overlay (could you add an "Overlay" section for `workorder-recheck` as well? we missed that in the last review). Example Overlay sections:

- https://github.com/DFHack/scripts/blob/master/docs/gui/civ-alert.rst 
- https://github.com/DFHack/scripts/blob/master/docs/exportlegends.rst

gui/workorder-details.lua
```diff
@@ -1,33 +1,18 @@
...
 
+local overlay = require('plugins.overlay')
+
 local utils = require 'utils'
 local gui = require 'gui'
 local guimat = require 'gui.materials'
 local widgets = require 'gui.widgets'
-local dlg = require 'gui.dialogs'
 
 local wsj = reqscript 'gui/workshop-job'
```
myk002, line 18: This would need to be updated to modern standards with a gui.ZScreen (maybe a gui.ZScreenModal for this) and widgets.Window. gui.FramedScreen is deprecated.

myk002:
> * [ ]  `Ctrl+D` opens another DFHack screen as well

This can likely be solved by changing the `gui/design` keybinding from `@dwarfmode` to `@dwarfmode/Default` in `dfhack.keybindings.init`

> * [ ]  I haven't found a better way to hide the widget when in the wrong context, other than `self.subviews.button.visible = false` in `onRenderBody`

The preferred solution here is to modify `modules/Gui.cpp` to produce the more specific focus string that you need. If that is not sufficient, then yes you can set the button invisible in `onRender[Whatever]`

> * [ ]  also, I haven't tested it beyond ensuring the UI is kind of working

The UI definitely needs some updating and testing. It uses code from `gui/workshop-job`, which itself hasn't been looked at since v50 dropped. It also should use a ZScreen and draggable windows to fit in with the modern UI themes.

docs/gui/workorder-details.rst
TymurGubayev: probably use dyed or dyeable (=undyed I think) cloth, or encrust unimproved items...

I'm not quite sure.

TymurGubayev:
[ ] I'm not quite sure what is the replacement for `dlg.showMessage`

myk002:
> * [ ]  I'm not quite sure what is the replacement for `dlg.showMessage`

That's still fine to use. At some point, I'll swap out the implementation to use a ZScreenModal, but that should be transparent to the callers.

TymurGubayev:
Sphinx is complaining because of `docs/changelogs/news.rst:861:undefined label: gui/workorder-details`

What's the approach with removed scripts here, should I keep the referenced docs as is, or maybe make its only content a reference to the new script?

myk002:
Tool removal is a tad janky since it crosses repos. You have to add an entry to https://github.com/DFHack/dfhack/blob/develop/docs/about/Removed.rst, but CI for both repos will fail until everything is consistent.

gui/job-details.lua
```diff
@@ -0,0 +1,438 @@
...
+            status = { text = dfhack.TranslateName(dfhack.units.getVisibleName(worker)), pen = COLOR_GREEN }
+        end
+    end
+
+    local window = widgets.Window{
+        frame_title='Details',
+        resizable = true,
+        resize_min={w=50, h=20},
+        frame = { l = 10, w = 50 },
+    }
```
myk002, line 51: while this is perfectly valid (and probably the way I did it myself in some earlier scripts), I would rather encourage you to keep the ZScreen and the Window as separate classes. This allows the ZScreen subclass to focus on tool lifecycle and the Window subclass to focus on content and interactive behaviors. In particular, if there is no Window subclass, then future editors might be tempted to add an `onInput` function to the ZScreen subclass, which could break the ZScreen stack.

gui/job-details.lua
```diff
@@ -0,0 +1,438 @@
...
+--
+
+local focusStrings = 'dwarfmode/JobDetails'
+
+DetailsHotkeyOverlay = defclass(DetailsHotkeyOverlay, overlay.OverlayWidget)
+DetailsHotkeyOverlay.ATTRS{
+    default_pos={x=0,y=0},
+    default_enabled=true,
+    viewscreens=focusStrings,
+    frame={w=1+6+2+(7)+1, h=3},
```
myk002, line 377: isn't `h=1` enough?

myk002, line 377: the numbers in the width could use some explanation in a comment

gui/job-details.lua
```diff
@@ -0,0 +1,438 @@
...
+    local scr = ScrJobDetails()
+    return scr.context == df.job_details_context_type.MANAGER_WORK_ORDER
+end
+
+DetailsHotkeyOverlay_ManagerWorkOrder = defclass(DetailsHotkeyOverlay_ManagerWorkOrder, DetailsHotkeyOverlay)
+DetailsHotkeyOverlay_ManagerWorkOrder.ATTRS{
+    default_pos={x=5, y=5}, -- {x=5, y=5} is right above the job title
+}
+function DetailsHotkeyOverlay_ManagerWorkOrder:onRenderBody(dc)
+    if isManagerOrderScreen() then
```
myk002, line 402: instead of checking here, would it be simpler to attach the overlays to more specific focus strings? The job details context type *is* included, e.g. `dwarfmode/JobDetails/MANAGER_WORK_ORDER`

gui/job-details.lua
```diff
@@ -0,0 +1,438 @@
...
+        job = scr.jb
+    elseif scr.context == df.job_details_context_type.MANAGER_WORK_ORDER then
+        job = scr.wq
+    end
+    if job == nil then
+        qerror("Unhandled screen context: ".. df.job_details_context_type[scr.context])
+    end
+
+    local dlg = JobDetails{ job = job, context = scr.context }
+    dlg:show()
```
myk002, line 363: naming conflict with global `dlg`

gui/job-details.lua
```diff
@@ -0,0 +1,438 @@
...
+    viewscreens=focusStrings,
+    frame={w=1+6+2+(7)+1, h=3},
+}
+
+function DetailsHotkeyOverlay:init()
+    self:addviews{
+        widgets.TextButton{
+            view_id = 'button',
+            frame={t=0, l=0, r=0, h=1},
+            label='details',
```
myk002, line 385: Be sure to include the word "DFHack" in here somewhere so it's clear it's not a vanilla widget. Maybe 'DFHack adjust job details'?

gui/job-details.lua
TymurGubayev: but does the conflict matter? It's local to this scope, it doesn't break anything (I actually stored the value in a global variable in earlier iterations of the rework -- when pressing `ESC` wouldn't close the window)

gui/job-details.lua
TymurGubayev: I probably copied this over from somewhere... Changed to 1 now

gui/job-details.lua
TymurGubayev: well, now - https://github.com/DFHack/dfhack/pull/3730 - it is :)

gui/job-details.lua
TymurGubayev: Wouldn't this make the buttons (unnecessary) wide...

I was trying to stay in style with this:
image

And the pop-up window has `DFHack` on it:

image

It's not hard to change, I'm just not quite sure we need to.

myk002: It's a fair question. The red brackets perhaps are distinctive enough.

Instead of `details`, though, maybe `Configure job inputs`? "Details" might be too generic.

gui/job-details.lua
myk002: ping on this

TymurGubayev: I'm on it, busy with work and stuff, but I'll get it done

docs/gui/job-details.rst
```diff
@@ -1,24 +1,18 @@
...
-    :tags: unavailable fort inspection jobs
+    :summary: Adjust the input materials and traits used for a job or manager order.
+    :tags: fort inspection jobs workorders interface
 
-This tool allows you to inspect or change the input reagents for the selected
-workshop job (in :kbd:`q` mode).
-
-.. image:: /docs/images/workshop-job.png
+This tool allows you to inspect or change the input reagents for the selected job:
+open the job's "Details" screen by clicking the magnifying glass, then press :kbd:`Ctrl-D`.
```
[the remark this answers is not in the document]
TymurGubayev: With new focus strings it doesn't work from global jobs list anymore, I'll add it back.
I also want to add it to work order configuration view, because it can be opened for every single order, while the details view only for some.

gui/job-details.lua
```diff
@@ -92,14 +86,13 @@ function JobDetails:init(args)
             frame = { l = 0, b = 0 },
             text = {
                 { key = 'LEAVESCREEN', text = ': Back',
-                  on_activate = self:callback('dismiss')
+                  -- on_activate = self:callback('dismiss')
```
myk002, line 89: yeah, not needed. the base ZScreen superclass takes care of window lifecycles

gui/job-details.lua
```diff
@@ -0,0 +1,425 @@
...
+
+DetailsHotkeyOverlay_ManagerWorkOrder = defclass(DetailsHotkeyOverlay_ManagerWorkOrder, DetailsHotkeyOverlay)
+DetailsHotkeyOverlay_ManagerWorkOrder.ATTRS{
+    default_pos={x=5, y=5}, -- {x=5, y=5} is right above the job title
+    viewscreens='dwarfmode/JobDetails/MANAGER_WORK_ORDER',
+}
+
+DetailsHotkeyOverlay_BuildingTask = defclass(DetailsHotkeyOverlay_BuildingTask, DetailsHotkeyOverlay)
+DetailsHotkeyOverlay_BuildingTask.ATTRS{
+    default_pos={x=-120, y=6}, -- {x=-120, y=6} is right above the job title on all but smallest widths
```
myk002, line 410: I'd move this to the right by 5 so it still looks good at the smallest width

gui/job-details.lua
```diff
@@ -0,0 +1,425 @@
...
+                  enabled = self:callback('canChangeIType'),
+                  on_activate = self:callback('onChangeIType') },
+                { key = 'CUSTOM_M', text = ': Material, ',
+                  enabled = self:callback('canChangeMat'),
+                  on_activate = self:callback('onChangeMat') },
+                { key = 'CUSTOM_T', text = ': Traits',
+                  enabled = self:callback('canChangeTrait'),
+                  on_activate = self:callback('onChangeTrait') }
+            }
+        },
```
myk002, line 78: these aren't clickable. Try making them individual `HotkeyLabel`s

myk002 (commented):
what does the `20` mean?
image

gui/job-details.lua
TymurGubayev: Did that and I think discovered a bug: it highlights every hotkeylabel up to mouse X position (but correct action is executed)

image


image

gui/job-details.lua
TymurGubayev: I "reserved" the whole line using `frame={w=1000, h= 1}`, so that now with `dfhack.screen.getWindowSize()` I can move the hotkey around and it always fits.

It would've been nice to be able to move the widget itself. IIRC that wasn't possible last time I checked (half a year ago? because of label placement in the workorder-recheck.lua).

TymurGubayev:
> what does the `20` mean? image

something changed somewhere and now we have for some reason a set bit without a name:

image

(even though `df.job_item_flags3[20] == nil`)

and this is how the list is filled from a bitfield:

```lua
local function list_flags(list, bitfield)
    for name,val in pairs(bitfield) do
        if val then
            table.insert(list, name)
        end
    end
end
```

Can this be a bug? Or maybe it should've been `unknown`?

To avoid confusion, I'll change the method above to exclude those.

gui/job-details.lua
myk002: That's because the label frame is larger than the label text. Try setting `auto_width=true` or setting an explicit width

myk002:
> what does the `20` mean?

That might require a structure update. I'll look into it

gui/job-details.lua
TymurGubayev: I improved upon this. The logic is as follows now.

0. Fix the distance from the right edge; make the widget slightly wider than necessary to allow moving the text around
1. determine the calculated x1 coordinate of the frame, which is the `screen width` minus the `distance from the right` minus the `widget width`
2. if the x1 < 0, set it to 0 --- making it the real position of the widget (I *think* this works better than built-in methods for some reason, although it's a few experiments back)
3. is the real left edge position of the widget too far to the left (`x1 < 6`), move the text to the right (`offset = 6 - x1`).

This works well enough until you try to narrow the game window to the minimum and then widen it again. Then the `self.frame.r` is changed from `102` to `77` and stays that way. That's why the 4th step is needed

4. restore original `self.frame.r`

Even with this after widening the window it doesn't immediately get the right position, but another small resizing in any direction sets it right.

myk002:
I identified the flags and pushed the updated build to the Steam `testing` branch if you'd like to test with it. Flag 20 is (unsurprisingly, given the context) "gem".

gui/job-details.lua
```diff
@@ -346,24 +367,44 @@ local function ScrJobDetails()
     return df.global.game.main_interface.job_details
 end
 
+local function ScrWorkorderConditions()
+    return df.global.game.main_interface.info.work_orders.conditions
```
myk002, line 371: this is statically allocated and can be cached. e.g.
local conditions = df.global.game.main_interface.info.work_orders.conditions

myk002: job_details above is static as well

myk002:
testing notes:

1. the alignment seems off (screen dimensions: 150x66):
image
2. maybe don't enable the Configure job inputs HotkeyLabel if there is nothing that can be configured:
image
3.  The ability to reset to defaults would be very nice
4. The ability to *remove* traits (or change the item type) could be considered an exploit. Could we only offer that capability if armok mode is enabled? (`dfhack.getHideArmokTools()` returns `false`)
5. I've learned from `showmood` that people want to see whole item quantities, not the internal counters. could we display `1` here instead of `150` (similar for cloth and thread)
image
6. The Esc hotkey hint probably isn't necessary. Most other windows don't have that.
7. "quantity: 1" would be clearer than "x1"

gui/job-details.lua
```diff
@@ -0,0 +1,635 @@
...
+        qerror("This script needs to be run from a job details or order conditions screen")
+    end
+
+    JobDetailsScreen{ job = job, context = context }:show()
+end
+
+local function is_change_possible()
+    -- we say it is if there is at least one item in the job
+    local job = get_current_job()
+    return job.items and #job.items ~= 0
```
myk002, line 487: should be `job.job_items`, not `job.items`, right?

gui/job-details.lua
```diff
@@ -0,0 +1,635 @@
...
+        prompt = 'Please select traits for input '..idx,
+        none_caption = 'no traits',
+    }:show()
+end
+
+function JobDetails:onResetChanges()
+    for _, obj in pairs(self.list.choices) do
+        local stored_obj = self.stored[obj.index]
+
+        local item_type = stored_obj.item_type
```
myk002, line 419: Attempting to reset changes for me results in this error:
```
gui/job-details.lua:419: attempt to index a nil value (local 'stored_obj')
```

myk002: looking through the code, it looks like storeInitialProperties is checking `job.items` instead of `job.job_items`

TymurGubayev: the `job.job_items` vs `job.items` is the bane of my existence 😣

because there are two structures involved, manager order and job, and one of them has both but uses the one that's distinct.

I think I fixed it now.

PR #800, opened 2023-08-13
AI-generated written text content

gistya:
# Description

This PR adds a module `gpt` with an overlay and resizable window, enabling AI-generated written content based on in-game descriptions of written content. `gpt.lua` sends written descriptions to a python helper script that runs in the background, `gptserver.py`, which in turn handles secure communication with OpenAI's servers.

## Security 

While the `gpt.lua` dfhack script communicates with the helper script using an unencrypted localhost TCP connection, no sensitive information is ever passed between dfhack and the python helper, which will ignore any connections from a hosts other than localhost (127.0.0.1).

All communications between `gptserver.py` and OpenAI's cloud are performed via OpenAI's official python module, which in turn handles all the nitty-gritty of requests. Their module employs up-to-date HTTPS security.

Your personal OpenAI key will be stored on disk in a file named `oaak.txt`. This file is not encrypted but it's only read by the python script (not by DFHack). Use common sense permissions for this file like you would for an SSH private key. 

OpenAI also allows you to set a hard limit on monthly costs, so that if this script goes in an infinite loop or somehow your API key leaks out, you won't be on the hook for thousands of dollars. Also, you can regenerate the API key at anytime on OpenAI's website and make a new key, which is recommended to be done quite frequently regardless of what security you're using.

Should these measures be seen as insufficient, we could investigate platform-specific system keychain storage for the API key, however this might make the python script less portable across platforms. Please comment below on your thoughts on this.

## Costs

Use of this script requires an OpenAI API account with a positive balance. They charge by the token, and the cost varies depending on which model is used. The default model selection is the fast, cheap one. You may elect to go with GPT-3.5 or 4.0 by launching `gptserver.py` with command-line options `-gpt3` or `-gpt4`, but be ready for slower response times and higher per-token fees.

Hopefully at some point this kind of service becomes free. 

# Remaining work to be done:

## Tests

At this point the PR contains no automated tests, hence why it's being posted as a draft PR. 

I would like to ask the preliminary reviewers for advice on testing strategy as I have not yet looked into how testing of these scripts is done. 

### Integration testing

My initial thoughts would be to use the test harness (?) to boot the gptserver.py in a "mock server" mode where it will give canned responses for canned requests. We would want to add a Pipfile.lock (if there's not already one) to ensure the sands of python dependencies don't shift out from under the test runners (might also come in handy for end users so we don't have to list the dependencies in the docs; that's not very maintainable and I felt gross doing it). 

This setup would allow us to run tests that validate the integration between the lua and python scripts hasn't been broken and error states in the python script get handled properly (right now the UI doesn't reflect a death of the python script appropriately, which is a TODO before this merges).

### End-to-end testing

If we know the knowledge descriptions of some Urist McTestbuddy in the test world, and we can macro the right screens being opened, then we could setup something like an end-to-end test with the only mocked thing being OpenAI's service.

### Unit testing

Since the script does sanitization on response strings to strip characters that DF can't render properly, and since it does heuristics to determine which type of written content a given description is of, it would be good to have some unit tests around these functions. I've tried to make those as functional as possible to facilitate testing. 

Since the script is stateful in its present form, I would also like to add tests to validate whether the state transitions are handled properly and in a robust manner. There should be no combination of states that the script cannot recover gracefully from. Ideally it could be made completely stateless except for a single state store, but I haven't gotten that far yet. (Redux FTW)

# Future directions

- addition of a configuration UI panel within DFHack (or additional command-line options for the python script) to allow configuration of various API parameters to be used for the OpenAI requests
- if/when dfhack integrates luasocket, then avoid using python altogether

I'm a bit leery of exposing implementation details specific to OpenAI within the DFHack UI, because I am guessing that a decent number of folks will want to rewrite the python script to talk to a different LLM which would likely have its own, totally different API. Having to then also update a bunch of UI code could be annoying.

### DIFF
--- a/docs/gpt.rst
+++ b/docs/gpt.rst
@@ -1,0 +1,67 @@
+gpt
+=======
+
+.. dfhack-tool::
+    :summary: AI-generated written content!
+    :tags: fort gameplay
+
+Enables a UI for submitting knowledge item descriptions to OpenAI for generating
+poetry, star charts, and excerpts from longer works such as biographies, dictionaries, 
+treatises on technological evolution, comparative biographies, cultural histories, 
+autobiographies, cultural comparisons, essays, guides, manuals, and more. 
+
+``enable gpt``
+=======
+Enables the plugin. The overlay will be shown when a knowledge item or unit view sheet is open.
+
+``disable gpt``
+=======
+Disables the plugin.
+
+Setup:
+
+1. Register for an OpenAI API account. It must be a paid or active trial account.
+2. Generate an API token for your account.
+3. Save your OpenAI API token to a file at the root of your DF directory, `oaak.txt`.
+4. Install python. We used version 3.11 installed from the Microsoft Store.
+5. Install python dependencies Flask and OpenAI: `pip install Flask` and `pip install OpenAI`.
+6. Start the local helper python app: cd into dfhack/scripts directory & run `python gptserver.py`.
+
+Once the python helper is running, you may now enable and use the gpt plugin.
+
+The python script defaults to using the fast, cheap, legacy model `text-davinci-003`.
+If you wish to use the slower, more expensive `gpt-3.5-turbo` or `gpt-4` models, you 
+can start the script with `python gptserver.py -gpt3` or `python gptserver.py -gpt4`.
+Tweaking additional OpenAI API parameters will require modifying `gptserver.py` to suit 
+your particular desires, until such time as someone may have added additional
+configuration options in a future update to DFHack :D
+
+Note: EVERY TEXT YOU GENERATE COSTS $$ if you are on a paid account. The fee is appx. $0.005 USD 
+      at the time of this writing. YMMV!
+
+Versions of python dependencies tested with:
+
+Package            Version
+------------------ ---------
+aiohttp            3.8.5
+aiosignal          1.3.1
+async-timeout      4.0.2
+attrs              23.1.0
+blinker            1.6.2
+certifi            2023.7.22
+charset-normalizer 3.2.0
+click              8.1.6
+colorama           0.4.6
+Flask              2.3.2
+frozenlist         1.4.0
+idna               3.4
+itsdangerous       2.1.2
+Jinja2             3.1.2
+MarkupSafe         2.1.3
+multidict          6.0.4
+openai             0.27.8
+requests           2.31.0
+tqdm               4.65.0
+urllib3            2.0.4
+Werkzeug           2.3.6
+yarl               1.9.2
--- a/gpt.lua
+++ b/gpt.lua
@@ -1,0 +1,555 @@
+--@ module = true
+
+local json = require('json')
+local dfhack = require('dfhack')
+local utils = require('utils')
+local luasocket = require('plugins.luasocket')
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local overlay = require('plugins.overlay')
+
+--
+-- TYPES
+--
+
+-- Enum for state of progress of the script.
+local Status = {
+  start = 0,
+  waiting = 1,
+  receiving = 2,
+  done = 3
+}
+
+local function string_from_Status(status)
+  if status == Status.start then return "start" end
+  if status == Status.waiting then return "waiting" end
+  if status == Status.receiving then return "receiving" end
+  if status == Status.done then return "done" end
+end
+
+local Content_Type = {
+  -- Non-fiction
+  manual = 'manual',
+  guide = 'guide',
+  treatise = 'treatise',
+  essay = 'essay',
+  dictionary = 'dictionary',
+  encyclopedia = 'encyclopedia',
+  star_chart = 'star chart',
+  -- Literature
+  poem = 'poem',
+  short_story = 'short story',
+  novel = 'novel',
+  alternate_history = 'alternate history',
+  -- Individual
+  letter = 'letter',
+  autobiography = 'autobiography',
+  biography = 'biography',
+  comparative_biography = 'comparative biography',
+  -- Group
+  genealogy = 'genealogy',
+  cultural_history = 'cultural history',
+  cultural_comparison = 'cultural comparison',
+  -- Unsupported
+  unsupported = 'unsupported'
+}
+
+local Progress_Symbol = { '/', '-', '\\', '|' }
+
+--
+-- CONSTS
+--
+
+-- Whether or not to print debug outpuut to the console.
+local is_debug_output_enabled = false
+
+-- Port on which to communicate with the python helper.
+local port = 5001
+
+-- Max number of empty responses from the helper after receiving data before 
+-- assuming that the response is complete. (Each line is received individually.)
+local max_retries = 5
+
+-- Whether or not the client object should be configured as blocking.
+local is_client_blocking = false
+
+-- Seconds to configure the client object's timeout.
+local client_timeout_secs = 60
+
+-- Milliseconds to configure the client object's timeout.
+local client_timeout_msecs = 0
+
+-- Total client timeout time. 
+local timeout = client_timeout_secs + client_timeout_msecs/1000
+
+-- Number of onRenderFrame events to wait before polling again.
+local polling_interval = 10
+
+-- Prompt component to use for generating excerpts of non-poetry knowledge items.
+local excerpts_prompt = 'Now, imagine two paragraphs, each one taken directly from a different section within the described book. These excerpts should seem like two of the most interesting, insightful, or groundbreaking passages in the treatise. They should read as direct quotes from the text, not as summaries/reviews or quotations of an interview with the author. They should concern minute details of the subject, as an interesting example given by the author, or a colorful anecdote within the text. The two paragraphs should be labeled, Excerpt A and Excerpt B. Two blank newlines should separate the two excerpts cleanly. The text should generally fit in the context of the game, Dwarf Fortress.'
+local star_chart_prompt = 'render an ASCII-art Dwarf Fortress star-chart inspired by that description using only Dwarvish names for stellar objects in the legend. DO NOT INCLUDE ANY references to Dwarf Fortress or the process of AI generation, the whole thing must be in-character! The star chart\'s title should match the above description!'
+
+-- Local config filename.
+local config = config or json.open('dfhack-config/gpt.json')
+
+-- User-facing list of valid content types that the script currently supports. 
+local valid_content_type_list = (function()
+  local list = 'a '
+  
+  local size = (function()
+    local count = 0
+    for _ in pairs(Content_Type) do count = count + 1 end
+    return count
+  end)()
+
+  local last_supported_index = size - 2
+  local index = 0
+
+  for key, content_type in pairs(Content_Type) do
+    if key == Content_Type.unsupported then goto continue end
+    
+    if index == last_supported_index then
+      list = list .. 'or ' .. content_type .. '.'
+    else 
+      list = list .. content_type .. ', '
+    end
+      
+    index = index + 1
+    ::continue::
+  end
+
+  return list
+end)()
+
+--
+-- STATE VARS
+--
+
+-- Tracks the state of the script to manage execution flow. 
+local current_status = Status.start
+
+-- Stores a reference to the client object while waiting/receiving a request.
+local client = nil
+
+-- Tracking to maintain polling interval.
+local poll_count = 0
+
+-- Current number of active retries.
+local retries = 0
+
+-- Cache for receiving data during polling.
+local total_data = ''
+
+-- When the request was submitted. Used for calculating timeout.
+local start_time = nil
+
+-- Text to display to the user.
+local gui_text = "Waiting for knowledge text description..."
+
+-- The most recently-submitted knowledge item. Used to avoid re-sending
+-- the same item multiple times in a row.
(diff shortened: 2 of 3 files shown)


gistya:
pre-commit.ci autofix

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+=======
+
+.. dfhack-tool::
+    :summary: AI-generated written content!
+    :tags: fort gameplay
+
+Enables a UI for submitting knowledge item descriptions to OpenAI for generating
+poetry, star charts, and excerpts from longer works such as biographies, dictionaries,
+treatises on technological evolution, comparative biographies, cultural histories,
+autobiographies, cultural comparisons, essays, guides, manuals, and more.
```
myk002, line 11:
```suggestion
When you are inspecting written works, such as poetry, star charts, or treatises on
technological evolution, this tool will generate and display actual examples of written
text based on the in-game description.

The text is generated by an online AI service, which you have to register for. See below
for required setup.
```

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+
+``enable gpt``
+=======
+Enables the plugin. The overlay will be shown when a knowledge item or unit view sheet is open.
+
+``disable gpt``
+=======
+Disables the plugin.
```
myk002: this actually isn't necessary. players can turn the overlay on and off in `gui/control-panel`.

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Setup:
```
myk002, line 21:
```suggestion
Required online account setup
-----------------------------
```

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+
+1. Register for an OpenAI API account. It must be a paid or active trial account.
```
myk002, line 23: a url would be useful here

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+2. Generate an API token for your account.
+3. Save your OpenAI API token to a file at the root of your DF directory, `oaak.txt`.
```
myk002, line 25: we try not to put any user data in the DF root directory. I suggest putting it in `dfhack-config/`.

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+4. Install python. We used version 3.11 installed from the Microsoft Store.
+5. Install python dependencies Flask and OpenAI: `pip install Flask` and `pip install OpenAI`.
+6. Start the local helper python app: cd into dfhack/scripts directory & run `python gptserver.py`.
```
myk002: I see two options here. if we deploy this with DFHack, then this python script should probably go in the DFHack/dfhack repo and get installed somewhere more appropriate (like maybe the DF root dir)

if this script gets distributed as a mod (see https://docs.dfhack.org/en/stable/docs/guides/modding-guide.html#the-structure-of-a-mod), then the python code can get included in the mod files

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+
+Once the python helper is running, you may now enable and use the gpt plugin.
```
myk002, line 30: I'd suggest having the overlay display a message like "cannot contact server -- do you have the Python helper running?" instead of requiring the player to remember to start the server first

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
...
+
+The python script defaults to using the fast, cheap, legacy model `text-davinci-003`.
+If you wish to use the slower, more expensive `gpt-3.5-turbo` or `gpt-4` models, you
+can start the script with `python gptserver.py -gpt3` or `python gptserver.py -gpt4`.
+Tweaking additional OpenAI API parameters will require modifying `gptserver.py` to suit
+your particular desires, until such time as someone may have added additional
+configuration options in a future update to DFHack :D
```
myk002: it would be fairly trivial to have a configuration screen that can open from the overlay and send options data to the python server

docs/gpt.rst
```diff
@@ -0,0 +1,67 @@
+gpt
+=======
+
+.. dfhack-tool::
+    :summary: AI-generated written content!
```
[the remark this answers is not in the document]
gistya: OK will fix thanks.

docs/gpt.rst
gistya: fixed

docs/gpt.rst
gistya: OK, makes sense

docs/gpt.rst
gistya: The overlay doesn't try to contact the server. It waits until you press the button and there is something to submit, then it tries. At that point, if it can't reach the server, the UI does display a message to the user about this.

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+    prompt_value = 'Please write a poem given the following description of the poem and its style: \n\n'..knowledge_description
+  elseif Content_Type[content_type] == Content_Type.star_chart then
+    debug_log('Creating star chart.')
+    prompt_value = 'Considering the star chart description between the >>> <<< below, ' .. star_chart_prompt .. ' >>> ' .. knowledge_description .. ' <<< '
+  elseif content_type == Content_Type.unsupported then
+    debug_log('Creating error response.')
+    prompt_value = 'Return a response stating simply, "There has been an error."'
+  else
+    debug_log('Creating prompt for non-poem/non-star-chart/non-unsupported content_type: ' .. content_type)
+    prompt_value = 'In between the four carrots is a description of a written ' .. content_type .. ': ^^^^' .. knowledge_description .. '^^^^. \n\n' .. excerpts_prompt
```
myk002, line 277: carrots -> carets

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+end
+
+function GPTBannerOverlay:onInput(keys)
+  if GPTBannerOverlay.super.onInput(self, keys) then return true end
+
+  if keys._MOUSE_R_DOWN or keys.LEAVESCREEN then
+    if view then
+      view:dismiss()
+    end
+  end
```
myk002, line 475: do we actually need to close the window if the knowledge description screen is closed? why not just leave it open so the player can appreciate it? they can always close it with a right click on the window at any time.

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+
+--
+-- GUI: Overlay
+--
+
+GPTBannerOverlay = defclass(GPTBannerOverlay, overlay.OverlayWidget)
+GPTBannerOverlay.ATTRS{
+    default_pos={x=-35,y=-2},
+    default_enabled=true,
+    viewscreens={'dwarfmode/ViewSheets/UNIT','dwarfmode/ViewSheets/ITEM'},
```
myk002, line 452: this seems too broad -- do we need more fine-grained focus strings here? it's not hard to add more context in `modules/Gui.cpp`

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+    elseif client ~= nil then
+      if poll_count == polling_interval then
+        poll(client)
+        poll_count = 0
+      else
+        poll_count = poll_count + 1
+      end
+    end
+
+    self.subviews.label:updateLayout()
```
myk002, line 522: this can get expensive. better to only update the layout when the text actually changes

gpt.lua
```diff
@@ -0,0 +1,557 @@
...
+}
+
+local function string_from_Status(status)
+  if status == Status.start then return "start" end
+  if status == Status.waiting then return "waiting" end
+  if status == Status.receiving then return "receiving" end
+  if status == Status.done then return "done" end
+end
+
+local Content_Type = {
```
myk002, line 30: this feels..odd, but I need to read through the code a few more times before I feel comfortable suggesting changes

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+}
+
+local Progress_Symbol = { '/', '-', '\\', '|' }
+
+--
+-- CONSTS
+--
+
+-- Whether or not to print debug outpuut to the console.
+local is_debug_output_enabled = false
```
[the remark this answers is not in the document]
gistya: amen 🗡️

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+
+-- Port on which to communicate with the python helper.
+local port = 5001
```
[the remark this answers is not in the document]
gistya: This is a constant (not meant to be changed).

myk002:
would it make sense to make the generated text persistent per world? then the same knowledge item would always get the same generated text per world (and it would cut down on API costs if the same item were viewed multiple times)

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+local client_timeout_secs = 60
+
+-- Milliseconds to configure the client object's timeout.
+local client_timeout_msecs = 0
+
+-- Total client timeout time.
+local timeout = client_timeout_secs + client_timeout_msecs/1000
+
+-- Number of onRenderFrame events to wait before polling again.
+local polling_interval = 10
```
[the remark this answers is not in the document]
gistya: OK will look into it thanks

gpt.lua
```diff
@@ -0,0 +1,555 @@
...
+  receiving = 2,
+  done = 3
+}
+
+local function string_from_Status(status)
+  if status == Status.start then return "start" end
+  if status == Status.waiting then return "waiting" end
+  if status == Status.receiving then return "receiving" end
+  if status == Status.done then return "done" end
+end
```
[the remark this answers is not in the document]
gistya: OK will update it tomorrow. BTW this is one thing that Swift really got perfect, how they handle enums with associated values is pure genius.

gpt.lua
gistya: Just using what was available. If you give me one for the UNIT/Skills/Knowledge I'll gladly use it

myk002:
> I would like to ask the preliminary reviewers for advice on testing strategy as I have not yet looked into how testing of these scripts is done.

we have facilities for mocking out pretty much anything. You can even mock out part of the DF game state so the script can read test data from the `df` object (see https://github.com/DFHack/scripts/blob/master/test/deteriorate.lua)

another good unit test example: https://github.com/DFHack/scripts/blob/master/test/prioritize.lua

gpt.lua
gistya: Thing is, right click usually doesn't close the knowledge description window, it just backs you to the list of known written content items, at which point we want the window closed so the user can open another knowledge item without necessarily generating it as soon as it's opened (which is what would happen if the window has remained opened). If someone is trying to economize which things they generate, that would be annoying. 

The good news is that if you click back into the same knowledge item you just left, the text is all still there, and you can always alt-tab to the python cmd console to see the output there also in case you forgot to screenshot a good one. 

I would suggest running the plugin and playing around with it, and then if you think I've misread the situation please let me know and I'll make the adjustment, or maybe we think of a third solution that's better than either of these options.

gpt.lua
gistya: Maybe this is what's making the pcall() lock up the UI updating? I'll play around with it tomorrow.

gpt.lua
```diff
@@ -0,0 +1,557 @@
+--@ module = true
+
+local json = require('json')
+local dfhack = require('dfhack')
```
gistya, line 4:
```suggestion
```

gistya (commented):
responding to the comments from myk002

gistya:
> would it make sense to make the generated text persistent per world? then the same knowledge item would always get the same generated text per world (and it would cut down on API costs if the same item were viewed multiple times)

Right now it's a bit of a compromise. It caches the last-received item for the session in memory, so that if you were to re-open the same one again many times in a row, it will show you the cached response. 

However as soon as you look at something else, now that occupies the cache, so that if you go back to the first one, it will regenerate fresh. 

This can be nice, since it saves you from regenerating an entry if you just accidentally closed it and then reopened it. But it also lets you regenerate it purposefully in case the first version was lame, by looking at something else, and then coming back to the first thing.

That could be a good thing, if you're using the default cheap/fast engine, which is more of a quantity over quality compromise. In this case you'll encounter more times where you want to regenerate the same prompt. 

However if you're running GPT4, which costs about 5x more but is also likely to produce a great result the first time every time, then you'll probably not want to regenerate the same one twice, ever. 

As to cacheing all the responses and persisting them to disk, what kind of SQL-based store do we have in project that could handle this in an efficient manner instead of, y'know, having a huge JSON object where you have to load the whole thing into memory and inefficiently search it?

gpt.lua
myk002: yeah, let me see if I can give you something better here.

docs/gpt.rst
lethosor: pcall shouldn't really be introducing much of a penalty, although errors themselves might... how often are you calling it?

gpt.lua
myk002: here we go:
image
and with a knowledge item selected:
image

myk002:
> So, if we were to consider adding support for caching every response and persisting the responses to disk, what kind of database store do we have in the project's dependencies, capable handle this in a memory-efficient manner?

how much text are we talking here? The excerpts I've seen are pretty small, and putting a few MB in JSON isn't out of the question. At runtime you'd cache in a table in Lua in memory. What are the upper bounds of the memory complexity requirements?

PR #802, opened 2023-08-26
Stocks script based on pre-v50 stocks plugin

chawk0:
This script is intended to clone the functionality of the pre-v50 stocks plugin from Falconne.  The layout of the panels and hotkeys is nearly identical, with one exception being that filters are based on alt- hotkeys rather than ctrl- hotkeys.

Remaining features to implement:
- filtering caged items and items-for-trade
- toggling trade on items
- zoom to a selected item
- item grouping

### DIFF
--- a/stocks2.lua
+++ b/stocks2.lua
@@ -1,0 +1,718 @@
+-- a v50 script remake of the older stocks plugin
+-- chawkzero, Aug 2023
+
+local gui = require("gui")
+local widgets = require("gui.widgets")
+local utils = require("utils")
+
+local item_desc_width = 38
+local item_table = {}
+
+local current_search_query = ""
+local filter_states = {
+	in_job = true,
+	rotten = true,
+	owned = true,
+	forbidden = true,
+	dump = true,
+	on_fire = true,
+	melt = true,
+	in_inv = true,
+	caged = true,
+	trade = true,
+	no_flags = true,
+}
+
+local min_qual = 0
+local max_qual = 6
+local min_wear = 0
+
+-- 0 = listed, 1 = selected
+local apply_to_mode = 0
+local apply_to_labels = {"Listed", "Selected"}
+
+local stocks_window_width = 106
+local stocks_window_height = 50
+
+StocksWindow = defclass(StocksWindow, widgets.Window)
+StocksWindow.ATTRS({
+	frame = {w = stocks_window_width, h = stocks_window_height},
+	frame_title = "Stocks",
+	autoarrange_subviews = false,
+	autoarrange_gap = 0,
+})
+
+function StocksWindow:init()
+	--dfhack.printerr("StocksWindow:init() called!")
+	--[[
+	COLOR_RESET, COLOR_BLACK, COLOR_BLUE, COLOR_GREEN, COLOR_CYAN,
+    COLOR_RED, COLOR_MAGENTA, COLOR_BROWN, COLOR_GREY, COLOR_DARKGREY,
+    COLOR_LIGHTBLUE, COLOR_LIGHTGREEN, COLOR_LIGHTCYAN, COLOR_LIGHTRED,
+    COLOR_LIGHTMAGENTA, COLOR_YELLOW, COLOR_WHITE
+	--]]
+
+	self:addviews({
+		widgets.Panel({
+			view_id = "left_panel",
+			frame = {t = 0, b = 0, l = 0, r = stocks_window_width - 76},
+			frame_style = gui.INTERIOR_FRAME,
+			subviews = {
+				widgets.EditField({
+					view_id = "search_box",
+					frame = {t = 1, l = 0},
+					label_text = "Search: ",
+					on_char = function(ch) return ch:match("[%l%s]") end,
+					on_change = function(new_text, old_text)
+						current_search_query = new_text
+						self:filter_list(current_search_query)
+					end,
+				}),
+				widgets.Label({
+					text = {{text = "Items: ", pen = COLOR_LIGHTGREEN}},
+					frame = {t = 3, l = 0},
+					view_id = "item_count_label"
+				}),
+				widgets.List({
+					view_id = "item_list",
+					--edit_below = false,
+					frame = {t = 4, b = 1, l = 1, r = 1},
+					--not_found_label = "empty?",
+					text_pen = {fg = COLOR_GREY, bg = COLOR_BLACK},
+					text_hpen = {fg = COLOR_BLACK, bg = COLOR_GREY},
+					cursor_pen = {fg = COLOR_LIGHTGREEN, bg = COLOR_DARKGREY},
+				}),
+			},
+		}),
+		widgets.Panel({
+			view_id = "right_panel",
+			frame = {t = 0, b = 0, r = 0, l = 72},
+			frame_style = gui.INTERIOR_FRAME,
+			subviews = {
+				widgets.Label({
+					text = "Filters (Alt+Key toggles)",
+					frame = {t = 0, l = 0},
+					view_id = "filters_header_label",
+				}),
+				widgets.Label({
+					text = "to be filled in...",
+					frame = {t = 1, l = 0},
+					view_id = "filters_label",
+				}),
+				widgets.Label({
+					text = {
+						{text = "Shift-C", pen = COLOR_LIGHTGREEN}, {text = ": Clear All", pen = COLOR_WHITE}, NEWLINE,
+						{text = "Shift-E", pen = COLOR_LIGHTGREEN}, {text = ": Enable All", pen = COLOR_WHITE}, NEWLINE,
+						{text = "Shift-G", pen = COLOR_LIGHTGREEN}, {text = ": Toggle Grouping", pen = COLOR_WHITE}
+					},
+					frame = {t = 8, l = 0},
+					view_id = "filters_label2",
+				}),
+				widgets.Label({
+					text = {
+						{text = "-+", pen = COLOR_LIGHTGREEN}, {text = ": Min Qual: ", pen = COLOR_WHITE}, {text = "tbd", pen = COLOR_BROWN}, NEWLINE,
+						{text = "/*", pen = COLOR_LIGHTGREEN}, {text = ": Max Qual: ", pen = COLOR_WHITE}, {text = "tbd", pen = COLOR_BROWN}, NEWLINE,
+						{text = "Shift-W", pen = COLOR_LIGHTGREEN}, {text = ": Min Wear: ", pen = COLOR_WHITE}, {text = "tbd", pen = COLOR_BROWN}
+					},
+					frame = {t = 12, l = 0},
+					view_id = "filters_qual_wear",
+				}),
+				widgets.Label({
+					text = {
+						{text = "Actions (", pen = COLOR_BROWN}, {text = "...", pen = COLOR_LIGHTGREEN}, {text = " Items)", pen = COLOR_BROWN}
+					},
+					frame = {t = 16, l = 0},
+					view_id = "actions_item_count_label",
+				}),
+				widgets.Label({
+					text = {
+						{text = "Shift-Z", pen = COLOR_LIGHTGREEN}, {text = ": Zoom    ", pen = COLOR_WHITE},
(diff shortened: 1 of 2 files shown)


stocks2.lua
```diff
@@ -0,0 +1,544 @@
+-- a v50 script remake of the older stocks plugin
```
lethosor, line 1: We should just name this "stocks", if it's a drop-in replacement and we aren't intending on resurrecting the plugin.

chawk0: I had it named stocks.lua originally but somehow DFHack was bringing up an old stocks plugin description and upon running the command, I would get errors about possible compatibility problems, etc.  For now, I just renamed it to stocks2 since I wasn't sure what the fix was.

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+                }),
+                widgets.Label({
+                    text = {
+                        {text = "-+", pen = COLOR_LIGHTGREEN}, {text = ": Min Qual: ", pen = COLOR_WHITE}, {text = "tbd", pen = COLOR_BROWN}, NEWLINE,
+                        {text = "/*", pen = COLOR_LIGHTGREEN}, {text = ": Max Qual: ", pen = COLOR_WHITE}, {text = "tbd", pen = COLOR_BROWN}, NEWLINE,
+                        {text = "Shift-W", pen = COLOR_LIGHTGREEN}, {text = ": Min Wear: ", pen = COLOR_WHITE}, {text = "tbd", pen = COLOR_BROWN}
+                    },
+                    frame = {t = 12, l = 0},
+                    view_id = "filters_qual_wear",
+                }),
```
myk002, line 108: This should be a `RangeSlider` (and accompanying `CycleHotkeyLabel`s). Example: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/zone.lua#L169

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+                    frame = {t = 16, l = 0},
+                    view_id = "actions_item_count_label",
+                }),
+                widgets.Label({
+                    text = {
+                        {text = "Shift-Z", pen = COLOR_LIGHTGREEN}, {text = ": Zoom    ", pen = COLOR_WHITE},
+                        {text = "-D", pen = COLOR_LIGHTGREEN}, {text = ": Dump", pen = COLOR_WHITE}, NEWLINE,
+                        {text = "Shift-F", pen = COLOR_LIGHTGREEN}, {text = ": Forbid  ", pen = COLOR_WHITE},
+                        {text = "-M", pen = COLOR_LIGHTGREEN}, {text = ": Melt", pen = COLOR_WHITE}, NEWLINE,
+                        {text = "Shift-T", pen = COLOR_LIGHTGREEN}, {text = ": Mark for Trade", pen = COLOR_GREY},
```
myk002, line 122: These should all be `HotkeyLabel`s. The "Shift" portion of the key is not shown, it's just printed as a capital letter, so that will save you some space.

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+end
+
+function StocksScreen:onDismiss()
+    view = nil
+end
+
+if dfhack_flags.module then
+    return
+end
+
```
myk002, line 543: need to check for a loaded fort. Example: https://github.com/DFHack/scripts/blob/master/gui/blueprint.lua#L623

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+                    view_id = "actions_label",
+                }),
+                widgets.Label({
+                    text = {
+                        {text = "Shift-A", pen = COLOR_LIGHTGREEN}, {text = ": Apply to: ", pen = COLOR_WHITE},
+                        {text = apply_to_labels[apply_to_mode + 1], pen = COLOR_BROWN},
+                    },
+                    frame = {t = 20, l = 0},
+                    view_id = "apply_to_label",
+                }),
```
myk002, line 134: This "apply to" selector seems anachronistic (which it is -- the original stocks screen was not mouse-aware). Let me finish this initial review, but it seems to me we need to reconsider the design for both keyboard and mouse accessibility. I encourage you to play with the trade and pasture assignment screens to see how the most recent screens utilize the mouse and enable multi-select of items. The critical difference between this screen and those is that in those screens, there is only one action -- mark/unmark for trade or assign/unassign to pasture/restraint/cage/pit. I suspect here we'll need to have a two-stage process: select item(s), then perform action. You can indicate that an item is selected the same way that the trade screen does it: https://github.com/DFHack/scripts/blob/master/internal/caravan/movegoods.lua#L331

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+        local item_desc = dfhack.items.getDescription(v, v:getType())
+
+        -- store item description and an item ref
+        table.insert(temp_table, { desc = item_desc, ref = v })
+    end
+
+    -- sort alphabetically
+    local spec = { key = function(v) return v.desc end }
+    local order = utils.make_sort_order(temp_table, { spec })
+    for i = 1, #order do item_table[i] = temp_table[order[i]] end
```
myk002, line 158: is this faster than `table.sort`? Example: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/zone.lua#L628

I also don't see any widgets for sorting so players can choose the sort order. I encourage you to steal the structure for sorting from `zone.lua` (where the above example links to)

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+            for i2, g2 in ipairs(g:getItem().general_refs) do
+                if g2:getType() == df.general_ref_type.UNIT_HOLDER then
+                    return false
+                end
+            end
+        end
+    end
+
+    -- lastly, don't melt masterworks
+    if item:getQuality() >= 5 then
```
myk002, line 208: use the enum values instead of magic numbers: `df.item_quality.Masterful`

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+    item.flags.melt = not item.flags.melt
+
+    if item.flags.melt then
+        utils.insert_sorted(df.global.world.items.other.ANY_MELT_DESIGNATED, item, "id")
+    else
+        utils.erase_sorted(df.global.world.items.other.ANY_MELT_DESIGNATED, item, "id")
+    end
+end
+
+function is_valid_melt_item(item)
```
myk002, line 171: this should be called from the logistics plugin instead of rewritten here. Could you open another PR to make the the logic at https://github.com/DFHack/dfhack/blob/develop/plugins/logistics.cpp#L269 callable from Lua? This just means move that logic to a static function and add it to `DFHACK_PLUGIN_LUA_FUNCTIONS` here: https://github.com/DFHack/dfhack/blob/develop/plugins/logistics.cpp#L744
then you can add `local logistics = require('plugins.logistics') to the top of this file and call it via `logistics.logistics_is_meltable_item(item)`

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+
+    -- lastly, don't melt masterworks
+    if item:getQuality() >= 5 then
+        return false
+    end
+
+    return true
+end
+
+function StocksWindow:onInput(keys)
```
myk002, line 215: once appropriate mouse-aware widgets are being used, I believe you can get rid of this function entirely

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+function StocksWindow:show_qual_and_wear()
+    local quality_labels = {"Ordinary", "WellCrafted", "FinelyCrafted", "Superior", "Exceptional", "Masterful", "Artifact"}
+    self.subviews.filters_qual_wear:setText(
+        {
+            { text = "-+", pen = COLOR_LIGHTGREEN}, { text = ": Min Qual: ", pen = COLOR_WHITE }, { text = quality_labels[min_qual + 1], pen = COLOR_BROWN }, NEWLINE,
+            { text = "/*", pen = COLOR_LIGHTGREEN}, { text = ": Max Qual: ", pen = COLOR_WHITE }, { text = quality_labels[max_qual + 1], pen = COLOR_BROWN }, NEWLINE,
+            { text = "Shift-W", pen = COLOR_LIGHTGREEN}, { text = ": Min Wear: ", pen = COLOR_WHITE }, { text = tostring(min_wear), pen = COLOR_BROWN }
+        }
+    )
+end
```
myk002, line 360: This should be handed by the RangeSlider

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+            { text = "D", pen = COLOR_MAGENTA},    { text = ": Dump    ", pen = get_pen(filter_states.dump) },
+            { text = "E", pen = COLOR_LIGHTRED},   { text = ": On Fire", pen = get_pen(filter_states.on_fire) }, NEWLINE,
+            { text = "M", pen = COLOR_BLUE},       { text = ": Melt    ", pen = get_pen(filter_states.melt) },
+            { text = "I", pen = COLOR_WHITE},      { text = ": In Inventory", pen = get_pen(filter_states.in_inv) }, NEWLINE,
+            { text = "C", pen = COLOR_LIGHTRED},   { text = ": Caged   ", pen = get_pen(filter_states.caged) },
+            { text = "T", pen = COLOR_LIGHTGREEN}, { text = ": Trade", pen = get_pen(filter_states.trade) }, NEWLINE,
+            { text = "N", pen = COLOR_GREY},       { text = ": No Flags", pen = get_pen(filter_states.no_flags) },
+        }
+    )
+end
```
myk002, line 387: These should be handled by `ToggleHotkeyLabel` options

stocks2.lua
```diff
@@ -0,0 +1,544 @@
+-- a v50 script remake of the older stocks plugin
+-- chawkzero, Aug 2023
+
+local gui = require("gui")
+local widgets = require("gui.widgets")
+local utils = require("utils")
+
+local item_desc_width = 38
+local item_table = {}
```
myk002, line 9: You shouldn't have to track this state yourself. It will be accessible as the choices in the FilteredList

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+                        current_search_query = new_text
+                        self:filter_list(current_search_query)
+                    end,
+                }),
+                widgets.Label({
+                    text = {{text = "Items: ", pen = COLOR_LIGHTGREEN}},
+                    frame = {t = 3, l = 0},
+                    view_id = "item_count_label"
+                }),
+                widgets.List({
```
myk002, line 67: This should be a FilteredList instead of just a List so you don't have to manage searching yourself.

stocks2.lua
```diff
@@ -0,0 +1,544 @@
...
+                    {text = owned_char, width = 1, pen = COLOR_GREEN},
+                    {text = forbidden_char, width = 1, pen = COLOR_RED},
+                    {text = dump_char, width = 1, pen = COLOR_LIGHTMAGENTA},
+                    {text = on_fire_char, width = 1, pen = COLOR_LIGHTRED},
+                    {text = melt_char, width = 1, pen = COLOR_BLUE},
+                    {text = in_inv_char, width = 1, pen = COLOR_WHITE}, "   ",
+                    {text = quality_label, width = 13, rjustify = false, pad_char = ' ', pen = quality_pen},
+                },
+                ref = v.ref,
+            })
```
myk002, line 513: all this can be dynamically rendered by the text entry attached to the list choice. that would also make it responsive to changes to the item from outside of this tool's control.
Example: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/zone.lua#L421

myk002 (commented):
This is a very good start, but the UX is not quite up to "modern" standards. Check out the `AssignAnimal` class in https://github.com/DFHack/dfhack/blob/develop/plugins/lua/zone.lua#L28 for example usage of HotkeyLabel, CycleHotkeyLabel, ToggleHotkeyLabel, FilteredList, searching, sorting, state management, and multi-select with shift-click.

stocks2.lua
myk002: You can copy the condition and wear sliders from here: https://github.com/DFHack/scripts/blob/master/internal/caravan/common.lua#L124

myk002:
Oh, and also please add a line to the "New tools" section of the changelog: https://github.com/DFHack/scripts/blob/master/changelog.txt#L29

PR #814, opened 2023-09-01
use reserved range for tileset

shevernitskiy:
depends on [#3716](https://github.com/DFHack/dfhack/pull/3716)


myk002 (approved):
approved for merge after https://github.com/DFHack/dfhack/pull/3716 is merged

PR #823, opened 2023-09-03
[caravan] separate search key into separate terms for move goods

myk002:


internal/caravan/common.lua
```diff
@@ -13,12 +13,16 @@ local to_pen = dfhack.pen.parse
...
-    for c in str:gmatch("[%w%s]") do
-        out = out .. c:lower()
+function add_words(words, str)
+    for word in str:gmatch("[%w]+") do
+        table.insert(words, word:lower())
     end
-    return out
+end
+
+function make_search_key(str)
```
ab9rf, line 22: is this "common" function being used in more than once place with this change, or is that a future enhancement?

myk002: the make_search_key? it's used by other places that don't need to augment the key

ab9rf (approved):

PR #828, opened 2023-09-06
Autofish: Docs update, incorrect values

AndrielChaoti:
Fixed autofish docs showing the incorrect minimum fish default value. (the default is 75, not 50)

### DIFF
--- a/docs/autofish.rst
+++ b/docs/autofish.rst
@@ -26,7 +26,7 @@ Usage
     on hand in your fortress. Fishing will be disabled when the amount of fish
     goes above this value.
 
-    ``min`` (default: 50) controls the minimum fish you want before restarting
+    ``min`` (default: 75) controls the minimum fish you want before restarting
     fishing.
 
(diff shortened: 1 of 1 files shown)


docs/autofish.rst
```diff
@@ -22,16 +23,23 @@ Usage
...
 
-    ``min`` (default: 50) controls the minimum fish you want before restarting
-    fishing.
+Options
+-------
 
-    Use ``--toggle-raw``(``-r``) (default: on) to toggle letting the script
-    also count your raw fish as part of your quota. Use it a second time to
-    disable this.
+``-r``, ``--toggle-raw``
```
myk002, line 40: This doesn't have to be changed in this PR (and we're really too close to release to add behavior changes now), but I'm realizing that this is probably not a good choice for an option. If you have a startup script with `autofish -r` in it, is that turning it on or off? What if you have two scripts that both assume they're turning the option in a particular direction? This should probably be changed into `--raw=<boolean>`

myk002 (approved):

PR #829, opened 2023-09-06
Allow raw fish option to be set from command line

AndrielChaoti:
Makes changes necessary to replace the "count raw fish" switch on the command line into a settable option, instead of a toggle.

Other minor changes to output as well.

### DIFF
--- a/autofish.lua
+++ b/autofish.lua
@@ -218,7 +218,6 @@ local positionals = argparse.processArgsGetopt(args,
             set_useRaw(toBool[optArg])
         else
             qerror("Invalid argument to --raw \"".. optArg .."\". expected boolean")
-            return
         end
     end}
 })
--- a/autofish.lua
+++ b/autofish.lua
@@ -1,5 +1,5 @@
 -- handles automatic fishing jobs to limit the number of fish the fortress keeps on hand
--- autofish [enable | disable] <max> [min] <options>
+-- autofish [enable | disable] <max> [min] [<options>]
 
 --@ enable=true
 --@ module=true
--- a/autofish.lua
+++ b/autofish.lua
@@ -1,5 +1,5 @@
 -- handles automatic fishing jobs to limit the number of fish the fortress keeps on hand
--- autofish [enable | disable] <max> [min] [--include-raw | -r]
+-- autofish [enable | disable] <max> [min] <options>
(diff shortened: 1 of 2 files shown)


autofish.lua
```diff
@@ -1,5 +1,5 @@
 -- handles automatic fishing jobs to limit the number of fish the fortress keeps on hand
--- autofish [enable | disable] <max> [min] [--include-raw | -r]
+-- autofish [enable | disable] <max> [min] <options>
```
myk002, line 2:
```suggestion
-- autofish [enable | disable] <max> [min] [<options>]
```

docs/autofish.rst
```diff
@@ -37,9 +37,9 @@ Positional Parameters
 Options
 -------
 
-``-r``, ``--toggle-raw``
-    (default: on) to toggle letting the script also count your raw fish as part
-    of your quota. Use it a second time to disable this.
+``r``, ``--raw <true | false>``
```
myk002, line 40:
```suggestion
``r``, ``--raw (true | false)``
```

docs/autofish.rst
```diff
@@ -37,9 +37,9 @@ Positional Parameters
...
+    (default: on) Set whether or not raw fish should be counted in the running
```
myk002, line 41:
```suggestion
    (default: ``true``) Set whether or not raw fish should be counted in the running
```

docs/autofish.rst
myk002, line 40: the example `-r` commandline below should be updated too

autofish.lua
```diff
@@ -206,10 +206,21 @@ if dfhack_flags and dfhack_flags.enable then
...
-    {{"r", "toggle-raw",
-    handler=function() s_useRaw = not s_useRaw end}
+    {{"r", "raw", hasArg=true,
+    handler=function(optArg)
+        optArg=string.lower(optArg)
+        if toBool[optArg] ~= nil then
+            set_useRaw(toBool[optArg])
+        else
+            qerror("Invalid argument to --raw \"".. optArg .."\". expected boolean")
+            return
```
myk002, line 221: no need for `return` -- `qerror` throws

myk002 (approved):

PR #833, opened 2023-09-11
Use pathability groups (thanks @myk002) to detect stranded citizens

azrazalea:
- [x] Create a command that can detect stranded dwarves
- [x] Add periodic checking and warning box similar to warn-starving
- [x] Add ability to react to warning box with ignore 
- [x] Save these ignores persistently
- [x] Update GUI to change view based on the ignored status of selected units.
- [x]  Add command-line options to configure ignores 
- [x]  Add command-line ability to get status
- [x]  Add command-line ability to get status + walkability group numbers 
- [x] Allow periodicity to be configured?
- [x]  Add ignore group action (toggleable? Yeah let's do it)
- [x]  Figure out how to get mouse buttons to work
- [x]  Improve UI, scrollbox, bigger window, ??
- [x] On reflection, sort the groups in the list ascending by size so that the worst off stranded are on the top.
- [x] Look into adding to autostart list?


azrazalea:
So I understand in general how we're doing the calculations to get the pathability group (using modulus 16 and getting the remainder), and I can see that it works and what the results are, but I don't know enough about the data structures involved to know why that works.

If anyone an point me to some resources or explain how map blocks/block.walkable works that'd be awesome!

azrazalea:
Fixes https://github.com/DFHack/dfhack/issues/3658

azrazalea:
I have no idea what a good time interval is for this. Without being able to ignore 4 times a day is definitely too often, but after ignores are added maybe it'll be easier?

changelog.txt
```diff
@@ -30,6 +30,9 @@ Template for new versions:
 
 ## New Features
 
+## New Scripts
+- `warn-stranded`: new repeatable maintenance script to check for stranded units, based off warn-starving
```
myk002, line 34:
```suggestion
- `warn-stranded`: new repeatable maintenance script to check for stranded units, based off `warn-starving`
```

docs/warn-stranded.rst
```diff
@@ -0,0 +1,34 @@
+warn-stranded
+=============
+
+.. dfhack-tool::
+    :summary: Reports citizens that are stranded and can't reach any other unit
```
myk002, line 5: Summary sentence has to end in a period to satisfy the parser.

docs/warn-stranded.rst
```diff
@@ -0,0 +1,34 @@
...
+    :tags: fort units
+
+If any (live) units are stranded the game will pause and you'll get a warning dialog telling you
```
myk002, line 8:
```suggestion
If any (live) units are stranded, the game will pause and you'll get a warning dialog telling you
```

docs/warn-stranded.rst
```diff
@@ -0,0 +1,34 @@
...
+--------
+
+``warn-stranded clear``
+   Clear all ignored units and then check for ones that are stranded.
+
+Options
+-------
+
+``clear``
+  Will clear all ignored units so that warnings will be displayed again.
```
myk002, line 34:
```suggestion
```

myk002: "clear" is a subcommand, not an option, and is sufficiently explained in the Examples section.

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+   dfhack.persistent.delete('warnStrandedIgnore')
+end
+
+local args = utils.invert({...})
+if args.clear then
+   clear()
+end
+
+
+warning = defclass(warning, gui.ZScreen)
```
myk002, line 22: I suggest inheriting from `gui.ZScreenModal`, since that's the intended UX here (it just sets the properties for you so you don't have to set them yourself)

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+local widgets = require 'gui.widgets'
+
+local function clear()
+   dfhack.persistent.delete('warnStrandedIgnore')
+end
+
+local args = utils.invert({...})
+if args.clear then
+   clear()
+end
```
myk002, line 19: should move this below the `isMapLoaded()` check so you don't try to clear persistent data in a world that doesn't exist yet

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+
+function warning:onClear()
+   clear()
+end
+
+function warning:onDismiss()
+    view = nil
+end
+
+function doCheck()
```
myk002, line 164: try to stick to four space indents

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+  local grouped = {}
+  local citizens = dfhack.units.getCitizens()
+
+  -- Pathability group calculation is from gui/pathable
+  for _, unit in pairs(citizens) do
```
myk002, line 169:
```suggestion
  for _, unit in ipairs(citizens) do
```

myk002: functionally equivalent, but you're dealing with a list here, not a map, so `ipairs` is more idiomatic

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+    local target = unit.pos
```
myk002, line 170: `unit.pos` might not be valid if the unit is in a cage or in some other esoteric circumstances. Use `dfhack.units.getPosition(unit)` instead

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+    local block = dfhack.maps.getTileBlock(target)
+    local walkGroup = block and block.walkable[target.x % 16][target.y % 16] or 0
+    local groupTable = grouped[walkGroup]
+
+    if groupTable == nil then
+      grouped[walkGroup] = { unit }
+    else
+      table.insert(groupTable, unit)
+    end
```
myk002, line 179:
```suggestion
    table.insert(ensure_key(grouped, walkGroup), unit)
```

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+  end
+
+  local strandedUnits = {}
+
+
+  for _, units in pairs(grouped) do
+    if #units == 1 and not unitIgnored(units[1]) then
```
myk002, line 186: do we only care if a single dwarf is stranded? what about closing a drawbridge and locking a group of foragers in the caverns? How about if we warn if we detect any bifurcation, and identify the smaller group as the "stranded"?

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+
+local function unitIgnored(unit)
+   local currentIgnore = dfhack.persistent.get('warnStrandedIgnore')
+   if currentIgnore == nil  then return false end
+
+   for index, id in pairs(currentIgnore['ints']) do
+      if id == unit.id then
+         print(id)
+         print(unit.id)
+         print('Unit Ignored')
```
myk002, line 83: leftover debug prints?

warn-stranded.lua
```diff
@@ -0,0 +1,204 @@
...
+      table.insert(choices, { text = getUnitDescription(unit), unit = unit })
+   end
+   local list = self.subviews.list
+   list:setChoices(choices, 1)
+end
+
+function warning:updateBottom()
+   self.subviews.bottom_ui:setText(
+      {
+         { key = 'SELECT', text = ': Toggle ignore unit', on_activate = self:callback('onIgnore') }, ' ',
```
myk002, line 125: please use `HotkeyLabel` widgets instead of hand-coding these "buttons". I realize you probably got this from `gui/autobutcher`, but that code is ancient and needs to be updated to modern standards. `gui/control-panel` would probably be a better model.

myk002 (commented):
looks pretty solid! great job!

myk002:
> So I understand in general how we're doing the calculations to get the pathability group (using modulus 16 and getting the remainder), and I can see that it works and what the results are, but I don't know enough about the data structures involved to know why that works.
> 
> If anyone an point me to some resources or explain how map blocks/block.walkable works that'd be awesome!

This just happens to be one of the few things we have documented about the data structures. https://docs.dfhack.org/en/latest/docs/api/Maps.html

tl;dr - blocks represent a 16x16 group of tiles, and the modulus transforms the map-local coordinates to block-local coordinates.

myk002:
> * Allow periodicity to be configured?

nah. as long as the control panel commandline runs it "often enough", that will be fine.

myk002:
maybe a `warn-stranded status` command would be useful - prints out the currently stranded units and marks those which are ignored.

warn-stranded.lua
azrazalea: I can work on that! The original issue specifically said a single dwarf but updating it should be easy

azrazalea: For that matter if we really wanted we could have a configurable threshold.

azrazalea:
> maybe a `warn-stranded status` command would be useful - prints out the currently stranded units and marks those which are ignored.

Agreed! That was the kind of thing I was thinking with the "add command line options to manage ignores" point!

I believe I have fixed all the review comments besides the indentation (my editor is set to 2, but I at least normalize it). I'll fix that shortly, I need to head to sleep!

On the topic of not just finding single dwarves: Do you think everybody not part of the biggest group (regardless of how many people are with them) should show up? That would mean if you have N groups you still get everyone who's not with the big group.

Thank you for all the great feedback, working on a new project is always FUN and you've helped a lot. I'll be sure to add you to the authors at the top next time I work on this.

myk002:
> On the topic of not just finding single dwarves: Do you think everybody not part of the biggest group (regardless of how many people are with them) should show up? That would mean if you have N groups you still get everyone who's not with the big group.

I *think* this sounds correct. From a player's perspective, wouldn't you want to know about all stranded units? Totally open for debate, though, if you have other opinions on what is best here

> Thank you for all the great feedback, working on a new project is always FUN and you've helped a lot.

Absolutely! I'm glad you found our codebase comprehensible : ) there's quite a bit to learn here.

> I'll be sure to add you to the authors at the top next time I work on this.

That is totally not necessary -- I actually try to discourage people from adding their names to these files. Over the life cycle of this script, many people will make contributions. An author's name at the top of a particular file can diminish the work of other people who have contributed. Attribution is better handled by `git blame` and the commit history.

That said, I very much encourage you to add your name to our official [authors list](https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst) so you can be immortalized there : )

gui/control-panel.lua
```diff
@@ -130,6 +130,9 @@ local REPEATS = {
     ['warn-starving']={
         desc='Show a warning dialog when units are starving or dehydrated.',
         command={'--time', '10', '--timeUnits', 'days', '--command', '[', 'warn-starving', ']'}},
+    ['warn-stranded']={
+        desc='Show a warning dialog when units are stranded from all others.'
```
myk002, line 134:
```suggestion
        desc='Show a warning dialog when units are stranded from all others.',
```

azrazalea:
> I _think_ this sounds correct. From a player's perspective, wouldn't you want to know about all stranded units? Totally open for debate, though, if you have other opinions on what is best here

I like the idea. With an eye toward being able to make a fancy UI if this becomes super popular in the future, I'm thinking of doing the following:

- List all units part of any group besides the largest count one
- Sort the list to where groups of units are together, and display an index-based group number (1, 2, 3 etc) next to the name (I know the UI framework can support various ways of making this better but I don't know which one I want to go for and it is the hardest part for me)
- Allow the user to specify a group number to ignore instead of having to ignore one by one (Input box?).
- Allow all functionality on the command line as well, including status and ignoring by unit ID (status will have to display the unit ID for them) or group.

azrazalea:
Oh, I really wanted to implement a zoom to unit. It looks like df 50 broke that for dfhack though?

myk002:
> I know the UI framework can support various ways of making this better but I don't know which one I want to go for and it is the hardest part for me)

I believe a standard `widgets.List` would suffice here, with the text generated from the group number and the unit name

> Hotkey that looks at currently selected unit and ignores their group?).

This sounds simple to implement. A HotkeyLabel that gets the currently selected list item and then iterates over that unit's group

> Oh, I really wanted to implement a zoom to unit. It looks like df 50 broke that for dfhack though?

Opening up the unit info panel is not straightforward, but you can zoom to the unit's location and highlight the tile that the unit is on.

azrazalea:
> Opening up the unit info panel is not straightforward, but you can zoom to the unit's location (`dfhack.gui.revealInDwarfmodeMap`) and highlight the tile that the unit is on.

Oooh that function is what I was looking for. The ones I found prior to this were using older code and a lot of it was commented out or just broken. I was searching for zoom though and that's why I missed it.

> I believe a standard `widgets.List` would suffice here, with the text generated from the group number and the unit name. you can attach auxiliary data to the list item by sticking a `data` table in the list choice. e.g. https://github.com/DFHack/scripts/blob/master/internal/caravan/movegoods.lua#L499

Excellent, that looks like it does everything UI-wise I want and more so I'll use it as an example for things.

One logic thing I'm working through:

- We definitely need to not use ignored units when calculating if anyone is stranded since if they are the only stranded we should say no one is.
- We can't include ignored units when calculating the biggest group  (after determining that at least one non-ignored unit is stranded) since theoretically the ignored units could cause a different group to be biggest. This will mean if your biggest group of dwarves has a lot of ignores it might not be chosen as the main group. This is a little weird but I feel like the alternative might be weirder?
- We need to add back in the ignored units at the end in the appropriate groups so we can deselect them. Ignored units from the main group should listed as group 1 I think if any exist so they are obvious?

azrazalea:
With that last commit I did the grouping refactor. Need to figure out the UI (I need a scrollbox and it currently just spills into the actions list, the window being bigger to a max when the list is bigger would be nice to) but everything is functional. I'll update my list of TODOs.

azrazalea:
I was actually playing last night (instead of just writing) and found an unexpected benefit of this script:

My current fortress is in a reanimator biome so I keep the bridged entrance closed unless I have a reason to open it.

So every time there is a migrant wave warn-stranded triggers and shows me everyone in the migrant wave! It's actually pretty helpful for my play style.

myk002:
> So every time there is a migrant wave warn-stranded triggers and shows me everyone in the migrant wave! It's actually pretty helpful for my play style.

Ha, nice : ) That information is annoyingly difficult to get in vanilla, but it's coming up more often in recent DFHack tools. The latest development (not yet released) version of the squad assignment screen can sort by arrival time, which helps, and eventually we'll have `manipulator` back, which can sort all units by arrival time and let you perform actions on them.

azrazalea:
@myk002 Alright, considering this complete if you could do a real review please!

docs/warn-stranded.rst
```diff
@@ -0,0 +1,51 @@
...
+There is a command line interface that can print status of units without pausing or bringing up a window.
+
+The GUI and command-line both also have the ability to ignore units so they don't trigger a pause and window.
+
+You can enable ``warn-stranded`` notifications in `gui/control-panel` on the "Maintenance" tab.
+
+Usage
+-----
+
+``warn-stranded -[wicg] [status|ignore|unignore] <id>``
```
myk002, line 22: this "man" style of options documentation isn't used in DFHack. Additionally, I suggest breaking the various invocation forms out into separate lines.

edit:  In writing this, it also occurs to me that many of the options might be better off as subcommands. Moreover, I suggest a few changes for ID display:

- display the group id when there is more than one unit in the group, remove `--walkgroups` option
- always display the unit id on console; never display it for the GUI; remove `--ids` option
- convert `clear` to a subcommand; remove `--clear` option
- convert `group` to a subcommand; remove `--group` option

```
Usage
-----

::

    warn-stranded
    warn-stranded status
    warn-stranded clear
    warn-stranded (ignore|ignoregroup|unignore|unignoregroup) <id>
```

warn-stranded.lua
```diff
@@ -0,0 +1,584 @@
...
+                        },
+                        widgets.HotkeyLabel{
+                            key = 'CUSTOM_SHIFT_C',
+                            label = 'Clear All Ignored',
+                            on_activate = self:callback('onClear'),
+                        },
+                        widgets.HotkeyLabel{
+                            key = 'CUSTOM_Z',
+                            label = 'Zoom to unit',
+                            on_activate = self:callback('onZoom'),
```
myk002, line 213: it seems to me that the player would *always* want to zoom to the selected unit. Could this be an `on_select` handler for the List instead?

warn-stranded.lua
```diff
@@ -0,0 +1,584 @@
...
+                    cursor_pen = { fg = COLOR_BLACK, bg = COLOR_GREEN },
+                    on_submit=self:callback('onIgnore'),
+                },
+                widgets.Panel{
+                    frame={h=5},
+                    autoarrange_subviews=true,
+                    subviews = {
+                        widgets.HotkeyLabel{
+                            key='SELECT',
+                            label='Toggle Ignore',
```
myk002, line 193: in labels, only the first letter of the first word should be capitalized

warn-stranded.lua
```diff
@@ -0,0 +1,584 @@
...
+                        },
+                        widgets.HotkeyLabel{
+                            key='CUSTOM_G',
+                            label='Toggle Group',
+                            on_activate = self:callback('onToggleGroup'),
+                        },
```
myk002, line 199: This could also be triggered by shift double clicking on a list item (`on_double_click2`)

warn-stranded.lua
```diff
@@ -0,0 +1,584 @@
...
+            frame_title='Stranded Citizen Warning',
+            resizable=true,
+            autoarrange_subviews=true,
+            subviews = {
+                widgets.List{
+                    frame={h=15},
+                    view_id = 'list',
+                    text_pen = { fg = COLOR_GREY, bg = COLOR_BLACK },
+                    cursor_pen = { fg = COLOR_BLACK, bg = COLOR_GREEN },
+                    on_submit=self:callback('onIgnore'),
```
myk002, line 185: should also call `onIgnore` when double clicking (the `on_double_click` handler)

warn-stranded.lua
```diff
@@ -0,0 +1,584 @@
...
+                            label = 'Clear All Ignored',
+                            on_activate = self:callback('onClear'),
+                        },
+                        widgets.HotkeyLabel{
+                            key = 'CUSTOM_Z',
+                            label = 'Zoom to unit',
+                            on_activate = self:callback('onZoom'),
+                        }
+                    }
+                },
```
myk002, line 216: I suggest adding some text to the bottom of the panel explaining the mouse actions, similar to the trade dialog windows.

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+    local is_ignored, index = unitIgnored(unit, ignores)
+
+    if is_ignored then
+        table.remove(ignores, index)
+    else
+        table.insert(ignores, unit.id)
+    end
+
+    dfhack.persistent.delete('warnStrandedIgnore')
+    dfhack.persistent.save({key = 'warnStrandedIgnore', value = table.concat(ignores, ' ')})
```
myk002, line 117: I think this might be an inefficient way of going about storing state. you can store each unit's data in its own key and use prefix matching to retrieve the full set of keys. the plugins tend to use this strategy, like `seedwatch`, `logistics`, `autobutcher`, and others. see `dfhack.persistent.get_all()` for the Lua version of the API.

You can also cache the state in a global that is reset on world load. example: https://github.com/DFHack/scripts/blob/master/prioritize.lua#L586

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+
+                print(text)
+            end
+        end
+    end
+
+    return false
+end
+
+if positionals[1] == 'ignore' then
```
myk002, line 514: a more standard idiom here would be to use `elseif` instead of `end if`, especially since these cases are mutually exclusive

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+        end
+    end
+
+    return true
+end
+
+-- ===============================================================
+--                   Graphical Interface
+-- ===============================================================
+warning = defclass(warning, gui.ZScreenModal)
```
myk002, line 163: Classes should start with an upper case letter (I'm aware that `warn-starving` fails to follow that pattern; I'll have to update the code there)

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+
+function warning:init(info)
+    self:addviews{
+        widgets.Window{
```
myk002, line 167: I suggest making the Window a separate class from the Screen and attaching most of your business logic to the Window subclass since it's so important *not* to override methods for the Screen subclass. Your current code doesn't do anything *wrong* here, but future editors might not be aware of the pitfalls of adding too much logic to a Screen subclass.

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+            view_id = 'main',
```
myk002, line 168: this view_id isn't actually used anywhere and can be removed

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+            min_size={w=60, h=25},
+            frame_title='Stranded Citizen Warning',
+            resizable=true,
+            autoarrange_subviews=true,
+            subviews = {
+                widgets.List{
+                    frame={h=15},
+                    view_id = 'list',
+                    text_pen = { fg = COLOR_GREY, bg = COLOR_BLACK },
+                    cursor_pen = { fg = COLOR_BLACK, bg = COLOR_GREEN },
```
myk002, line 179: we generally try to keep list coloring standard for a consistent presentation; is there a particular reason you're recoloring it here?

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+                    on_submit=self:callback('onIgnore'),
+                    on_select=self:callback('onZoom'),
+                    on_double_click=self:callback('onIgnore'),
+                    on_double_click2=self:callback('onToggleGroup'),
+                },
+                widgets.Panel{
+                    frame={h=5},
+                    autoarrange_subviews=true,
```
myk002, line 187: this layout seems to use space inefficiently. some of these could potentially be distributed horizontally with non-autoarranged custom frames.

warn-stranded.lua
```diff
@@ -0,0 +1,581 @@
...
+        return "("..sym..")"
+    else
+        return ""
+    end
+end
+
+-- Partially taken from warn-starving
+local function getUnitDescription(unit)
+    return '['..dfhack.units.getProfessionName(unit)..'] '..dfhack.TranslateName(dfhack.units.getVisibleName(unit))..
+        ' '..getSexString(unit.sex)..' Stress category: '..dfhack.units.getStressCategory(unit)
```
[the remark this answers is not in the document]
azrazalea: Oooh I didn't know this syntax, very nice. I have only used Lua occasionally.

Yeah, the stress category was an experiment but I'm not really loving it either. The gender was just an aid for quickly getting a view of the dwarves. I'm a speed reader so I do word shapes more than actually reading them so that + gender is usually enough for me to know which dwarf I'm looking at without fully parsing the line.

warn-stranded.lua
azrazalea: I think I just copied this from somewhere, removing.

warn-stranded.lua
azrazalea: Oh, awesome. Yeah this was very hacked together and I wasn't sure the best way to use the persistent API here. I'll look at those examples and get something better going.

gui/control-panel.lua
```diff
@@ -134,6 +134,9 @@ local REPEATS = {
     ['warn-starving']={
         desc='Show a warning dialog when units are starving or dehydrated.',
         command={'--time', '10', '--timeUnits', 'days', '--command', '[', 'warn-starving', ']'}},
+    ['warn-stranded']={
+        desc='Show a warning dialog when units are stranded from all others.',
+        command={'--time', '300', '--timeUnits', 'ticks', '--command', '[', 'warn-stranded', ']'}},
```
myk002, line 139:
```suggestion
        command={'--time', '0.25', '--timeUnits', 'days', '--command', '[', 'warn-stranded', ']'}},
```

docs/warn-stranded.rst
```diff
@@ -0,0 +1,53 @@
...
+
+.. dfhack-tool::
+    :summary: Reports citizens that are stranded and can't reach any other citizens.
+    :tags: fort units
+
+If any (live) groups of fort citizens are stranded from the main (largest) group,
+the game will pause and you'll get a warning dialog telling you which citizens are isolated.
+This gives you a chance to rescue them before they get overly stressed or start starving.
+
+Each ciitizen will be put into a group with the other citizens stranded together.
```
myk002, line 12:
```suggestion
Each citizen will be put into a group with the other citizens stranded together.
```

docs/warn-stranded.rst
```diff
@@ -0,0 +1,53 @@
...
+    warn-stranded
+    warn-stranded status
+    warn-stranded clear
+    warn-stranded (ignore|ignoregroup|unignore|unignoregroup) <id>
+
+Examples
+--------
+
+``warn-stranded``
+    Standard command that checks for standed citizens and causes a window to pop up with a warning if any are stranded.
```
myk002, line 34:
```suggestion
    Standard command that checks citizens and pops up a warning if any are stranded.
```

docs/warn-stranded.rst
```diff
@@ -0,0 +1,53 @@
...
+    Does nothing when there are no unignored stranded citizens.
+
+``warn-stranded status``
+    List all stranded citizens and all ignored citizens. Includes citizen unit ids.
+
+``warn-stranded clear``
+    Clear(unignore) all ignored citizens.
```
myk002, line 41:
```suggestion
    Clear (unignore) all ignored citizens.
```

changelog.txt
```diff
@@ -30,6 +30,9 @@ Template for new versions:
 
 ## New Features
 
+## New Scripts
+- `warn-stranded`: new repeatable maintenance script to check for stranded units, based off `warn-starving`
```
myk002, line 34:
```suggestion
- `warn-stranded`: new repeatable maintenance script to check for stranded units, similar to `warn-starving`
```

warn-stranded.lua
```diff
@@ -1,21 +1,26 @@
...
 -- GUI heavily based off of autobutcher
---@ module = true
+--@module = true
 
 local gui = require 'gui'
 local utils = require 'utils'
 local widgets = require 'gui.widgets'
 local argparse = require 'argparse'
 local args = {...}
+scriptPrefix = 'warn-stranded'
```
myk002, line 11:
```suggestion
local scriptPrefix = 'warn-stranded'
```

warn-stranded.lua
```diff
@@ -58,73 +64,67 @@ local function addId(text, unit)
...
+--              Persistence API
+-- ===============================================
+-- Optional refresh parameter forces us to load from API instead of using cache
+
+-- Uses persistent API. Low-level, gets all entries currently in our persistent table
+--   will return an empty array if needed. Clears and adds entries to our cache.
+-- Returns the new global ignoresCache value
+local function getIgnoredUnits()
+    local ignores = dfhack.persistent.get_all(scriptPrefix)
+    if ignores == nil then return {} end
```
myk002, line 77: this will return an empty set when there is no saved data to load, but will not set `ignoresCache`. Looking at the callers of this function, it looks like it is always expected that `ignoresCache` is set/updated after calling `getIgnoredUnit()`. This is especially important in the `onStateChange` hook, where a failure to clear the old data out of `ignoresCache` may cause strange behavior in the newly loaded fort.

on another note, it looks like the return value of `getIgnoredUnits()` is never used. should this instead be called `loadIgnoredUnits()`?

azrazalea: Interestingly I did test and loading a fort with ignores followed by loading one without any does clear it! Good notes though, I'll get this updated tonight (heading to the ren fair!)

warn-stranded.lua
```diff
@@ -0,0 +1,583 @@
...
+-- ===============================================================
+--                   Graphical Interface
+-- ===============================================================
+WarningWindow = defclass(WarningWindow, widgets.Window)
+WarningWindow.ATTRS{
+    frame={w=80, h=25},
+    min_size={w=60, h=25},
+    frame_title='Stranded Citizen Warning',
+    resizable=true,
+    autoarrange_subviews=true,
```
myk002, line 170:
```suggestion
    frame={w=60, h=25, r=2, t=18},
    resize_min={w=50, h=15},
    frame_title='Stranded citizen warning',
    resizable=true,
```

warn-stranded.lua
```diff
@@ -0,0 +1,583 @@
...
+}
+
+function WarningWindow:init(info)
+    self:addviews{
+        widgets.List{
+            frame={h=15},
```
myk002, line 176:
```suggestion
            frame={l=0, r=0, t=0, b=6},
```

warn-stranded.lua
```diff
@@ -0,0 +1,583 @@
...
+            view_id = 'list',
+            on_submit=self:callback('onIgnore'),
```
myk002, line 178:
```suggestion
```

warn-stranded.lua
```diff
@@ -0,0 +1,583 @@
...
+                    on_activate = self:callback('onClear'),
+                    auto_width=true,
+
+                },
+                widgets.WrappedLabel{
+                    frame={b=1, l=0},
+                    text_to_wrap='Click to toggle unit ignore. Shift doubleclick to toggle a group.',
+                },
+            }
+        },
```
myk002, line 221:
```suggestion
        widgets.WrappedLabel{
            frame={b=3, l=0},
            text_to_wrap='Double click to toggle unit ignore. Shift double click to toggle a group.',
        },
        widgets.HotkeyLabel{
            frame={b=1, l=0},
            key='SELECT',
            label='Toggle ignore',
            auto_width=true,
        },
        widgets.HotkeyLabel{
            frame={b=1, l=23},
            key='CUSTOM_G',
            label='Toggle group',
            on_activate = self:callback('onToggleGroup'),
            auto_width=true,
        },
        widgets.HotkeyLabel{
            frame={b=0, l=0},
            key = 'CUSTOM_SHIFT_I',
            label = 'Ignore all',
            on_activate = self:callback('onIgnoreAll'),
            auto_width=true,

        },
        widgets.HotkeyLabel{
            frame={b=0, l=23},
            key = 'CUSTOM_SHIFT_C',
            label = 'Clear all ignored',
            on_activate = self:callback('onClear'),
            auto_width=true,
```

warn-stranded.lua
```diff
@@ -0,0 +1,583 @@
...
+-- ======================================================================
+
+local function compareGroups(group_one, group_two)
+    return #group_one['units'] < #group_two['units']
+end
+
+local function getStrandedUnits()
+    local groupCount = 0
+    local grouped = {}
+    local citizens = dfhack.units.getCitizens()
```
myk002, line 310:
```suggestion
    local citizens = dfhack.units.getCitizens(true)
```

myk002 (commented):
After playtesting, I have the following notes:
- the central position of the window makes it hard to see the unit that it's zooming to. how about making the window narrower and making it appear on the right side of the screen?
- the min resize isn't set quite right, should be `resize_min`, not `min_size` (though that would have made sense to call it that if I had thought of it at the time)
- the list should expand with the window as it is resized instead of being set to a constant height
- widgets.List has some bad behavior around clicking -- it triggers on_submit() in addition to on_select(). this has the side effect that clicking on a list item to zoom has the side effect of toggling the ignored status. How about changing the ignore gesture to double click? that would match well with shift-double click toggling the group.
- insane citizens should be included (right?)

I'll see if I can express this in code change suggestions

warn-stranded.lua
```diff
@@ -0,0 +1,583 @@
...
+
+            table.insert(choices, { text = text, data = {unit = unit, group = groupIndex} })
+        end
+    end
+
+    local list = self.subviews.list
+    list:setChoices(choices)
+end
+
+function WarningWindow:onIgnore(_, choice)
```
myk002, line 248:
```suggestion
function WarningWindow:onIgnore(_, choice)
    if not choice then
        _, choice = self.subviews.list:getSelected()
    end
```

warn-stranded.lua
```diff
@@ -0,0 +1,577 @@
...
+            auto_width=true,
+
+        },
+        widgets.HotkeyLabel{
+            frame={b=0, l=23},
+            key = 'CUSTOM_SHIFT_C',
+            label = 'Clear all ignored',
+            on_activate = self:callback('onClear'),
+            auto_width=true,
+    }
```
myk002, line 213:
```suggestion
            auto_width=true,
        },
    }
```

PR #834, opened 2023-09-13
Update quantum.lua to v50

MarioWizard119:
WHAT THIS PR DO
Updates gui/quantum to be able to be used on v50 saves. A bit jank, but mostly functional

WHAT THIS PR DON'T DO, BUT PROLLY NEED TO BE DONE (BE IT BY ME OR SOMEONE FAR SMARTER)
Unfortunately I couldn't get the script to automatically order a minecart if there wasn't one. From what I could gather from the old gui/quantum file, that too was handled by quickfort. That could probably be fixed later once I figure out how to do it, but a stopgap measure it to just warn that one will need to be made beforehand or after.

Also could probably use a cleanup or optimization, I'm extremely new to this, and this is the first actual project I've contributed to (I don't count my Space Station 13 days. Writing an entire goddamn essay to explain why changing two variables won't destroy the game's balance isn't coding, It's the psychological equivalent of a root beer float, but with mint chocolate chip ice cream and orange Fanta instead), so feel free to make/have me make changes to it to adhere to community standards. And also, for some reason I couldn't get the script to intercept left clicks on the map whatsoever so I had to resort to a janky hotkey workaround, *but* it's functional.


gui/quantum.lua
```diff
@@ -11,262 +11,212 @@ local quickfort = reqscript('quickfort')
...
 
-QuantumUI = defclass(QuantumUI, guidm.MenuOverlay)
+
+-- UI Layout
+QuantumUI = defclass(QuantumUI, gui.ZScreen)
 QuantumUI.ATTRS {
-    frame_inset=1,
     focus_path='quantum',
-    sidebar_mode=df.ui_sidebar_mode.LookAround,
+	defocusable=true
```
myk002, line 19: `defocusable` is true by default, so you can remove that one. I also suggest you follow the example of `gui/blueprint` and set
```lua
pass_movement_keys=true,
pass_mouse_clicks=false,
```

gui/quantum.lua
```diff
@@ -275,80 +225,57 @@ local function order_minecart(pos)
     quickfort_orders.enqueue_additional_order(quickfort_ctx, 'wooden minecart')
     quickfort_orders.create_orders(quickfort_ctx)
 end
+--Key Input
+function QuantumUI:onInput(keys)
```
myk002, line 229: Unfortunately, this will break the ZScreen stack (that is, input will not be propagated properly). Try splitting out the Window widget into its own class and attaching input processing code there. Check out the example architecture in the docs here: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#zscreen-class

gui/quantum.lua
```diff
@@ -11,262 +11,212 @@ local quickfort = reqscript('quickfort')
...
-            widgets.TooltipLabel{
-                text_to_wrap='Note that enabling refuse will cause clothes' ..
+            text_pen=COLOR_GREY
+		},
+		widgets.HotkeyLabel{
+            key='CUSTOM_Q',
+            label='Place QSP under cursor',
+            on_activate=self:callback('commit')
+		},
+        widgets.Panel{autoarrange_subviews=true,
```
myk002, line 45: indents should be four spaces

gui/quantum.lua
```diff
@@ -11,262 +11,212 @@ local quickfort = reqscript('quickfort')
...
     }
 
     self:addviews{main_panel}
 end
 
+-- UI Functions
 function QuantumUI:get_help_text()
-    if not self.feeder then
+    if not dfhack.gui.getSelectedStockpile() then
         return 'Please select the feeder stockpile with the cursor or mouse.'
```
myk002, line 120: there isn't really a cursor option for this anymore. directions should be to use the mouse

gui/quantum.lua
```diff
@@ -11,262 +11,212 @@ local quickfort = reqscript('quickfort')
...
+	output_vert = qspOutputVert[dump_dir]
+	output_type = qspOutputType[allow_refuse]
+	trackstop_dir = qspTrackstopDir[dump_dir]
+	if
+		name == ''
+	then
+		name = 'Unnamed QSP'
+	end
+	placed_output = output_type..'{quantum=true}'
+	placed_trackstop = trackstop_dir..'{take_from='..input_id..' route="'..name..'"}'
```
myk002, line 195: the trackstop itself should get a name, something like "$name dumper"

setting the route property to a name that already exists will attach to an existing route, which is never what you want for a qsp. I suggest either enforcing a unique name or just not naming the route

gui/quantum.lua
```diff
@@ -11,262 +11,212 @@ local quickfort = reqscript('quickfort')
...
+	}
+	
+	quickfort.apply_blueprint{
+		mode='build',
+		pos=dfhack.gui.getMousePos(), 
+		data={
+			[0]={
+				[0]={[0]=placed_trackstop}
+			}
+		}
```
myk002, line 214:
```suggestion
		data=placed_trackstop
```

gui/quantum.lua
```diff
@@ -11,262 +11,212 @@ local quickfort = reqscript('quickfort')
...
+	}
+	dfhack.run_command('assign-minecarts all')
+	self:updateLayout()
+	message= 'QSP Placed!'
+	dialogs.MessageBox{text=message:wrap(70)}:show()
```
myk002, line 219: does this need to be a popup? could we perhaps instead write it to the gui/quantum window panel?

gui/quantum.lua
```diff
@@ -11,262 +11,212 @@ local quickfort = reqscript('quickfort')
...
+	output_hor = qspOutputHor[dump_dir]
+	output_vert = qspOutputVert[dump_dir]
+	output_type = qspOutputType[allow_refuse]
+	trackstop_dir = qspTrackstopDir[dump_dir]
+	if
+		name == ''
+	then
+		name = 'Unnamed QSP'
+	end
+	placed_output = output_type..'{quantum=true}'
```
myk002, line 194: should be of the form:
`ry{name="$name quantum" quantum=true}:+all`
or
`c{name="$name quantum" quantum=true}:+all`

myk002:
> Unfortunately I couldn't get the script to automatically order a minecart if there wasn't one. From what I could gather from the old gui/quantum file, that too was handled by quickfort.

`quickfort` can still be responsible for ordering the minecart. that's what the `quickfort orders` command is for.

myk002 (commented):
could you possibly fix the formatting? See summary of errors here: https://results.pre-commit.ci/run/github/54066125/1699332446.rdyxy0-8R_eBoKIIwgdL8w

MarioWizard119:
pre-commit.ci autofix

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
 local quickfort_command = reqscript('internal/quickfort/command')
 local quickfort_orders = reqscript('internal/quickfort/orders')
 
-QuantumUI = defclass(QuantumUI, guidm.MenuOverlay)
+
+-- UI Layout
+QuantumUI = defclass(QuantumUI, gui.ZScreen)
```
myk002, line 16: Try not to mix the functions between two classes. It makes the logic of the file much harder to follow. Put the `ZScreen` subclass at the bottom and the `Window` subclass (and all other business logic) at the top

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
+    resizable = true,
+    autoarrange_subviews=true
+}
 
-    local main_panel = widgets.Panel{autoarrange_subviews=true,
-                                     autoarrange_gap=1}
-    main_panel:addviews{
-        widgets.Label{text='Quantum'},
+function QuantumWin:init()
+    cart_count = #assign_minecarts.get_free_vehicles()
```
myk002, line 32: `cart_count` should be an instance variable, not a global. call it `self.cart_count` instead

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
+
+    self:addviews{
+        widgets.Label{
+            text='Quantum'
+        },
```
myk002, line 37: this is in the frame title now and doesn't need to be a label

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
-                initial_option=false},
-            widgets.TooltipLabel{
-                text_to_wrap='Note that enabling refuse will cause clothes' ..
+            text_pen=COLOR_GREY
+        },
+        widgets.HotkeyLabel{
+            key='CUSTOM_Q',
+            label='Place QSP under cursor',
+            on_activate=self:callback('commit')
+        },
```
myk002, line 46: I'd say don't bother with the keyboard cursor for now. Few people know how to use it and it complicates the UI. Go with mouse clicks, like `gui/quantum` and `gui/blueprint` and other map tools do

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
+        widgets.Panel{autoarrange_subviews=true,
+            frame = { w=75, h=5 },
+            resizable = true,
```
myk002, line 49: `resizable` only makes sense for `Window` subclasses

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
+            subviews={
+                widgets.ToggleHotkeyLabel{
+                    view_id='refuse',
+                    key='CUSTOM_R',
+                    label='Allow refuse/corpses',
+                    initial_option=false
+                },
+                widgets.TooltipLabel{
+                    text_to_wrap='Note that enabling refuse will cause clothes' ..
```
myk002, line 96: this behavior was removed from DF in 50.07 (or thereabouts). You can take this option out and just always have refuse/corpses enabled

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
 }
 
-function QuantumUI:init()
-    local cart_count = #assign_minecarts.get_free_vehicles()
+QuantumWin = defclass(QuantumWin, widgets.Window)
+QuantumWin.ATTRS {
+    frame_title = 'Quantum Stockpile',
+    frame = { w=60, h=26 },
+    resizable = true,
+    autoarrange_subviews=true
```
myk002, line 28: Could you post a screenshot of the UI? I suspect the dimensions could use some tweaking. also, you might benefit from setting `autoarrange_gap=1`

gui/quantum.lua
myk002, line 1: it's very likely the tool docs will need updating -- `docs/gui/quantum.rst`

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
+function QuantumUI:init()
+    self:addviews{QuantumWin{}}
 end
 
-function QuantumUI:get_help_text()
-    if not self.feeder then
-        return 'Please select the feeder stockpile with the cursor or mouse.'
+-- UI Functions
+function QuantumWin:get_help_text()
+    if not dfhack.gui.getSelectedStockpile() then
```
myk002, line 123: pass `true` to prevent console error spam when a stockpile is not selected

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
-    return move..move_to_less, move_back..move_to_greater
-end
+    quickfort.apply_blueprint{
+        mode='build',
+        pos=dfhack.gui.getMousePos(),
+        data={
+            [0]={
+                [0]={[0]=placed_trackstop}
+            }
+        }
```
myk002, line 215:
```suggestion
        data=placed_trackstop,
```

gui/quantum.lua
```diff
@@ -275,80 +231,42 @@ local function order_minecart(pos)
...
-local function create_quantum(pos, qsp_pos, feeder_tiles, name, trackstop_dir,
-                              allow_refuse)
-    local dsg = allow_refuse and 'yr' or 'c'
-    local stats = quickfort.apply_blueprint{mode='place', data=dsg, pos=qsp_pos}
-    if stats.place_designated.value == 0 then
-        error(('failed to place quantum stockpile at (%d, %d, %d)')
-              :format(qsp_pos.x, qsp_pos.y, qsp_pos.z))
+--Key Input
+function QuantumWin:onInput(keys)
+    pos = dfhack.gui.getSelectedStockpile()
```
myk002, line 236: leftover line? it doesn't seem to do anything other than set an unused global variable

gui/quantum.lua
```diff
@@ -275,80 +231,42 @@ local function order_minecart(pos)
...
-    if stats.build_designated.value == 0 then
-        error(('failed to build trackstop at (%d, %d, %d)')
-              :format(pos.x, pos.y, pos.z))
+    local pos = nil
+    if
+        keys.CUSTOM_Q
+    then
+        self:commit()
+    elseif
+        keys._MOUSE_L_DOWN and not self:getMouseFramePos()
```
myk002, line 250: should be _MOUSE_L, not _MOUSE_L_DOWN

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
-                                'Right', 'Left')
-
-    local quantumstop_name_part, quantum_name_part = '', ''
-    if name ~= '' then
-        quantumstop_name_part = (' name="%s quantum"'):format(name)
-        quantum_name_part = ('{givename name="%s dumper"}'):format(name)
+    if
+        cart_count == 0
+    then
+        dfhack.run_command('workorder "{\"amount_left\":1,\"amount_total\":1,\"frequency\":\"OneTime\",\"id\":6,\"is_active\":false,\"is_validated\":true,\"item_subtype\":\"ITEM_TOOL_MINECART\",\"job\":\"MakeTool\",\"material_category\":[\"wood\"]}"')
```
myk002, line 221: why are you ordering a minecart via `workorder` instead of the `order_minecart` function below?

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
 
-function QuantumUI:select_stockpile(pos)
-    local flags, occupancy = dfhack.maps.getTileFlags(pos)
-    if not flags or occupancy.building == 0 then return end
-    local bld = dfhack.buildings.findAtTile(pos)
-    if not bld or bld:getType() ~= df.building_type.Stockpile then return end
-
-    local tiles = {}
+--UI Data Tables
+qspOutputHor = {
```
myk002, line 152: these should all be declared `local`

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
 
-    local function get_feeder_overlay_char(pos)
-        return safe_index(zlevel, pos.y, pos.x) and 'X'
+--Actually making the QSP
+function QuantumWin:commit()
+    name = self.subviews.name.text
+    dump_dir = self.subviews.dir:getOptionLabel()
+    allow_refuse = self.subviews.refuse:getOptionValue()
+    input_id = dfhack.gui.getSelectedStockpile().id
+    output_hor = qspOutputHor[dump_dir]
```
myk002, line 186: these should all be declared `local`

gui/quantum.lua
```diff
@@ -11,262 +11,218 @@ local quickfort = reqscript('quickfort')
...
-function QuantumUI:get_qsp_pos(cursor)
-    local offsets = self.subviews.dir:getOptionValue()
-    return {
-        x = cursor.x + (offsets.x or 0),
-        y = cursor.y + (offsets.y or 0),
-        z = cursor.z
+    placed_output = output_type..'{name="'..name..'" quantum=true}:+all'
+    placed_trackstop = trackstop_dir..'{name="'..name..'" take_from='..input_id..'}'
+
+    quickfort.apply_blueprint{
```
myk002, line 198: this needs the error checking that was in the original code -- if the player has selected an invalid tile for the QSP, we should not blindly move forward

gui/quantum.lua
MarioWizard119: I had left the keyboard option in cause there's a part o the community that gets cranky when there's not keyboard shortcuts for stuff. The Q key places it where the mouse cursor is, so I can see why it would be redundant, and I understand it not being worth it to cater to that small subsection of the community.

gui/quantum.lua
MarioWizard119: Basic answer was it was what I could get to work. The order Minecart function was one of the things I couldn't seem to get to work properly before, but now that the class stuff's been sorted out I'll see what I can do about it. I chose work order specifically because it wasn't reliant on any external blueprint or work order file and I could use a json string in the script itself. I can change it if you want me to.

gui/quantum.lua
MarioWizard119: This was somewhat of a deliberate design choice. There are a few uses for a pseudo QSP where the output tile isn't valid to put a stockpile on, for example, dumping goblin corpses down a mass grave out of sight and out of mind, dumping ores down a huge shaft down to the magma sea and magma forges, and dumping iron minecarts onto a roller system to fill them with magma. I've even used quantum to make storm drains and water reactors via a dummy 1x1 input stockpile with no settings, since they both use the same mechanism of a dumping Minecart.

As for the error checking, wouldn't it be redundant? From what I could tell from the old version, the error checking went place trackstop (dry run), if dry placed entities = 0 then display error message, else place trackstop. What's stopping us from place trackstop, if placed entities = 0 then display error message? I could see it being necessary for the trackstop itself, rather than the output stockpile because regardless of the above niche uses or an actual QSP, neither will work without the trackstop, unless it causes problems in the backend I'm not aware of.

PR #836, opened 2023-09-16
update fix/general-strike

ab9rf:
make less aggressive

should close dfhack/dfhack#3779

### DIFF
--- a/fix/general-strike.lua
+++ b/fix/general-strike.lua
@@ -6,10 +6,16 @@ local argparse = require('argparse')
 local function fix_seeds(quiet)
     local count = 0
     for _,v in ipairs(df.global.world.items.other.SEEDS) do
-        if not v.flags.in_building then
+        if (not v.flags.in_job) and (not v.flags.in_building) then
             local bld = dfhack.items.getHolderBuilding(v)
             if bld and bld:isFarmPlot() then
                 v.flags.in_building = true
+                for _,i in ipairs(bld.contained_items) do
+                    print (('%d %d'):format(i.item.id, v.id))
+                    if i.item.id == v.id then
+                        i.use_mode = 2
+                    end
+                end
                 count = count + 1
             end
         end
--- a/fix/general-strike.lua
+++ b/fix/general-strike.lua
@@ -11,7 +11,6 @@ local function fix_seeds(quiet)
             if bld and bld:isFarmPlot() then
                 v.flags.in_building = true
                 for _,i in ipairs(bld.contained_items) do
-                    print (('%d %d'):format(i.item.id, v.id))
                     if i.item.id == v.id then
                         i.use_mode = 2
                     end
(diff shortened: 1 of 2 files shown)


fix/general-strike.lua
```diff
@@ -6,10 +6,15 @@ local argparse = require('argparse')
...
     local count = 0
     for _,v in ipairs(df.global.world.items.other.SEEDS) do
-        if not v.flags.in_building then
+        if (not v.flags.in_job) and (not v.flags.in_building) then
             local bld = dfhack.items.getHolderBuilding(v)
             if bld and bld:isFarmPlot() then
                 v.flags.in_building = true
+                for _,i in ipairs(bld.contained_items) do
+                    if i.item.id == v.id then
+                        i.use_mode = 2
```
myk002, line 15: I take it you discovered the seeds had the wrong use_mode?

ab9rf: in rare cases, yes

when seed planting starts, the seed is added to the building's inventory vector in use mode 0 (the same as for any workshop, and farms do act like a workshop in this regard), with `in_job` and `in_building` set. when the plant seed job completes, the `in_job` flag is cleared and the seed's use mode changes to 2

the problem that this tool is seeking to correct is when the plant seed job is cancelled incorrectly. normally, when the job is cancelled, the `in_job` flag is cleared but the use mode is unchanged. in _most_ cases the seed is immediately removed from the building and everything is fine, but in _uncommon_ cases the seed gets stuck in the building. i'm guessing that there's a cancellation code path in toady's code that doesn't cancel plant seed jobs properly leaving them in a limbo state where they can neither be removed nor completed, which results (i suspect) in planters spamming trying to replant the seed but not being able to do so because it's locked into that building somehow. my assumption is that toady has (as he is wont to do) cut and paste job cancellation code into diverse locations, and updated some but not all such cancellation pathways in a way that has an differential impact here

the script attempts to detect and fix this latter case by changing the seed to "planted", but the version putnam provided is both too aggressive and insufficient: it tries to change them to "planted" _while they're still being planted_, and it fails to change their use mode to 2, which results in the game being inconsistent about whether the seed is actually planted or not. this results in overplanting the field (i've seen them shove as many as nine seeds into a field with four spaces), and inconsistent on-screen displays it appears that the seeds in use_mode 0 actually _do_ grow and eventually get harvested, but they don't _look_ like they're in the farm and are not, at least for some purposes, treated as if they are in the farm, and i can understand why players feel like their seeds have disappeared because that's what it looks like on the map view

the determination that `use_mode` 2 was required was based both on experimenting with poking other `use_mode`s into the inventory vector using `gm-editor` and by watching seeds that appeared to be planted correctly. every mode other than mode 2 yielded the "purple infinity" on the farm plot detail view, while _only_ mode 2 resulted in the map view showing the plot as occupied, and only mode 2 prevents the dwarfs from trying to plant another seed in the same spot

i ran a test fort for two seasons with the code above, during which i saw only a very small number (but not zero) of corrections and saw none of the weirdness i was seeing with the previous version

PR #837, opened 2023-09-16
Control panel orders-reevaluate option

Halifay:
Added orders-reevaluate option to the control panel that request manager orders conditions recheck once per month.


gui/control-panel.lua
```diff
@@ -127,6 +127,9 @@ local REPEATS = {
     ['orders-sort']={
         desc='Sort manager orders by repeat frequency so one-time orders can be completed.',
         command={'--time', '1', '--timeUnits', 'days', '--command', '[', 'orders', 'sort', ']'}},
+    ['orders-reevaluate']={
+        desc='Invalidates manager orders making it necessary to recheck conditions.',
```
myk002, line 131: I would specifically say "once a month" here. otherwise people might be afraid that their orders will be in a constant reevaluation loop

Halifay: Done

myk002 (approved):
approved; but holding off on merging until https://github.com/DFHack/dfhack/pull/3781 is merged

gui/control-panel.lua
```diff
@@ -127,6 +127,9 @@ local REPEATS = {
     ['orders-sort']={
         desc='Sort manager orders by repeat frequency so one-time orders can be completed.',
         command={'--time', '1', '--timeUnits', 'days', '--command', '[', 'orders', 'sort', ']'}},
+    ['orders-reevaluate']={
+        desc='Invalidates work orders once a month forcing manager to recheck conditions.',
```
myk002, line 131:
```suggestion
        desc='Invalidates work orders once a month, allowing conditions to be rechecked.',
```

PR #839, opened 2023-09-25
Misc improvement: `suspendmanager.lua` has new suspension reason for unsupported constructions.

master-spike:
This change to suspendmanager deals with https://github.com/DFHack/dfhack/issues/3601. I've added a new suspension reason `UNSUPPORTED` and added some checks relating to it. The way it works is based on an observation: each tiletype shape that can provide support provides it in one of two ways - either it supports like a wall (all 6 ortho axes as well as orthogonally on the layer above, with an invisible floor), or it supports like a floor (orthogonally on the same level). The only other thing that can provide support is a support building (pillar), which provides support to floors above it or walls both above and below.

I have tested this to a standard that I hope is rigorous enough although @plule says he's willing to take a shot at testing some other cases I may have missed.

### DIFF
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -172,7 +172,7 @@ local CONSTRUCTION_WALL_SUPPORT = utils.invert{
 }
 
 local CONSTRUCTION_FLOOR_SUPPORT = utils.invert{
-    df.construction_type.FLOOR,
+    df.construction_type.Floor,
     df.construction_type.DownStair,
     df.construction_type.Ramp,
     df.construction_type.TrackN,
@@ -447,7 +447,7 @@ local function constructionIsUnsupported(job)
     local wall_would_support = {}
     local floor_would_support = {}
     local supportbld_would_support = {}
-    
+
     if CONSTRUCTION_FLOOR_SUPPORT[constr_type] then
         wall_would_support = neighboursWallSupportsFloor(pos)
         floor_would_support = neighboursFloorSupportsFloor(pos)
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -409,6 +409,13 @@ local function neighboursFloorSupportsFloor(pos)
     }
 end
 
+local function hasWalkableNeighbour(pos)
+    for _,n in pairs(neighbours(pos)) do
+        if (walkable(n)) then return true end
+    end
+    return false
+end
+
 local function tileHasSupportWall(pos)
     local tt = dfhack.maps.getTileType(pos)
     if tt then
@@ -442,6 +449,10 @@ local function constructionIsUnsupported(job)
     if not building or building:getType() ~= df.building_type.Construction then return false end
 
     local pos = {x=building.centerx, y=building.centery,z=building.z}
+
+    -- if no neighbour is walkable it can't be constructed now anyways,
+    -- this early return helps reduce "spam"
+    if not hasWalkableNeighbour(pos) then return false end
 
     -- find out what type of construction
     local constr_type = building:getSubtype()
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -56,7 +56,7 @@ REASON_DESCRIPTION = {
     [REASON.RISK_BLOCKING] = 'May block another build job',
     [REASON.ERASE_DESIGNATION] = 'Waiting for carve/smooth/engrave',
     [REASON.DEADEND] = 'Blocks another build job',
-    [REASON.UNSUPPORTED] = 'Construction is unsupported'
+    [REASON.UNSUPPORTED] = 'Would collapse immediately'
 }
 
 --- Suspension reasons from an external source
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -452,7 +452,7 @@ local function constructionIsUnsupported(job)
 
     -- if no neighbour is walkable it can't be constructed now anyways,
     -- this early return helps reduce "spam"
-    if not hasWalkableNeighbour(pos) then return false end
+    -- if not hasWalkableNeighbour(pos) then return false end -- commented out pending `walkable()` fix
 
     -- find out what type of construction
     local constr_type = building:getSubtype()
--- a/suspendmanager.lua
+++ b/suspendmanager.lua
@@ -36,6 +36,7 @@ REASON = {
     ERASE_DESIGNATION = 4,
     --- Blocks a dead end (either a corridor or on top of a wall)
     DEADEND = 5,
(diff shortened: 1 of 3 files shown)


suspendmanager.lua
```diff
@@ -266,6 +340,134 @@ local function neighbours(pos)
...
+
+    -- find out what type of construction
+    local constr_type = building:getSubtype()
+    if CONSTRUCTION_FLOOR_SUPPORT[constr_type] then
+        for _,n in pairs(neighboursWallSupportsFloor(pos)) do
+            local tt = dfhack.maps.getTileType(n)
+            if tt then
+                local attrs = df.tiletype.attrs[tt]
+                if TILETYPE_SHAPE_WALL_SUPPORT[attrs.shape] then  return false end
+            end
```
donhth, line 436: You might consider breaking these bits out into a function.

plule:
Maybe I'm mixing up something but it does not seem to correctly detect this situation, buidling a gold bar floor:

image

below:

image

above:

image

suspendmanager.lua
```diff
@@ -161,6 +164,77 @@ local CONSTRUCTION_IMPASSABLE = utils.invert{
...
 
+local CONSTRUCTION_WALL_SUPPORT = utils.invert{
+    df.construction_type.Wall,
+    df.construction_type.Fortification,
+    df.construction_type.UpStair,
+    df.construction_type.UpDownStair,
+}
+
+local CONSTRUCTION_FLOOR_SUPPORT = utils.invert{
+    df.construction_type.FLOOR,
```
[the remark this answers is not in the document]
master-spike: Good catch, that must be why it doesn't work for floors!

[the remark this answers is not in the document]
master-spike: I've made the change now - should work for that case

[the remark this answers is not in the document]
plule: It does :)

suspendmanager.lua
```diff
@@ -433,6 +639,11 @@ function SuspendManager:refresh()
             end
         end
 
+        -- Check for construction jobs which may be unsupported
```
plule, line 642: This block should be moved under the ```Internal reasons to suspend a job```.

The terminology is not very good, but everything before this check is what is already suspended by dwarf fortress, and suspendmanager (/unsuspend) will never unsuspend it, and everything after is what suspendmanager is actively suspending.

suspendmanager.lua
```diff
@@ -36,6 +36,7 @@ REASON = {
     ERASE_DESIGNATION = 4,
     --- Blocks a dead end (either a corridor or on top of a wall)
     DEADEND = 5,
+    UNSUPPORTED = 6,
```
plule, line 39:
```suggestion
    --- Would cave-in immediately on completion
    UNSUPPORTED = 6,
```

suspendmanager.lua
master-spike: Ok, I misunderstood what `preventBlocking` meant in this context. I think maybe that might warrant a rename due to the expanded scope of this script? perhaps something like `preventRiskyJobs` or something?

plule: It's actually even wider than that, the suspended jobs on smoothing designation also fall in that category, they are not risky, they would just erase a smoothing job. Maybe falling back on the broader "smart" term would be best.

This issue is spread all across the suspension script, even in the UI and the `suspend`/`unsuspend` command line options, I suggest not to do that in that pr, maybe a separate issue?

plule:
With floors and tracks (at least), it won't suspend it if there is a floor above, but it does cave-in:

image

above:
image

below:
image

master-spike:
> With floors and tracks (at least), it won't suspend it if there is a floor above, but it does cave-in:
> 
> image
> 
> above: image
> 
> below: image

I think I made a mistake copying and pasting the coordinates and forgetting to change them 🤦 I think I've fixed this and I've added comments on the directions to make it easier to quickly verify correctness

suspendmanager.lua
```diff
@@ -266,6 +341,138 @@ local function neighbours(pos)
...
+
+---
+local function constructionIsUnsupported(job)
+    if job.job_type ~= df.job_type.ConstructBuilding then return false end
+
+    local building = dfhack.job.getHolder(job)
+    if not building or building:getType() ~= df.building_type.Construction then return false end
+
+    local pos = {x=building.centerx, y=building.centery,z=building.z}
+
```
plule, line 445: One small quality of life thing: It feels a bit "spammy" to suspend unreachable constructions:

image

Maybe you could add an early return here if none of the direct neighbour is walkable to make suspendmanager less intrusive in that situation.

plule:
For the terminology, I think that "would collapse" is the best for any user facing text (so reason description «Would collapse immediately», doc and changelog too?), it's the term used in the native UI «Something has collapsed on the surface!». «Unsupported» feels a bit ambiguous, and cave in is not that obvious to me. Disclaimer I'm not a native speaker though

master-spike:
> For the terminology, I think that "would collapse" is the best for any user facing text (so reason description «Would collapse immediately», doc and changelog too?), it's the term used in the native UI «Something has collapsed on the surface!». «Unsupported» feels a bit ambiguous, and cave in is not that obvious to me. Disclaimer I'm not a native speaker though

You're right, something about this was bothering me too. "Unsupported" sounds like it could be mistaken for a technical term (as in, it could be read like "this script doesn't support this type of construction"), rather than a game thing.

suspendmanager.lua
plule: mmh this broke it actually, walkable() returns false for floor bars, the tile type shape is «empty» for walkable constructions.

I'm trying to understand the consequences for suspendmanager in general, this could cause other bugs

plule: @master-spike My brain is too fried to understand/find the consequences of the `walkable()` bug, maybe you can revert/comment out this part for now.

As far as I'm concerned except this, it's all good :)

donhth:
You can also say it lacks a foundation, has no foundation, needs foundation, or requires a foundation, etc.

suspendmanager.lua
master-spike: Ah, it's because `walkable()` is incomplete - we don't have definitions of walkable buildings (floor bars/grates, retracted bridges, etc) like we do for impassable ones. I'll roll back this change and maybe see if there's a better way to check for walkable tiles.

master-spike: Probably better to fix `walkable()` in a separate issue

plule (approved):
Nice!

myk002 (approved):

PR #841, opened 2023-09-26
suspendmanager: walkable() now correctly identifies all walkable squares and broke out a new function to explicitly exclude cases where that tile is a tree branch

master-spike:
Fixes https://github.com/DFHack/dfhack/issues/3814

This change fixes this issue. Decision of where to apply `walkable()` and where to apply `isSuitableAccess()` was discussed in the discord

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -31,6 +31,7 @@ Template for new versions:
 ## New Features
 
 ## Fixes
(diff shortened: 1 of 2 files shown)


plule (approved):
Thank you!

myk002 (commented):
this looks like it could use a changelog entry

myk002 (approved):

PR #843, opened 2023-09-26
reinstate startdwarf and add scrollbar overlay

myk002:
Fixes https://github.com/DFHack/dfhack/issues/3789
depends on https://github.com/DFHack/dfhack/pull/3816 and https://github.com/DFHack/df-structures/pull/664 (on linux, at least)

### DIFF
--- a/startdwarf.lua
+++ b/startdwarf.lua
@@ -1,17 +1,79 @@
--- change number of dwarves on initial embark
+--@ module=true
 
-local addr = dfhack.internal.getAddress('start_dwarf_count')
-if not addr then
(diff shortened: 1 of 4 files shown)


startdwarf.lua
```diff
@@ -1,17 +1,79 @@
--- change number of dwarves on initial embark
+--@ module=true
```
lethosor, line 1: The first line should be kept for `ls` output, I think.

myk002: This is only true if there is no accompanying .rst doc file. In-tree scripts don't need any in-script docs.

PR #851, opened 2023-10-06
add --oldest and --clear features to `prioritize`

ab9rf:


prioritize.lua
```diff
@@ -200,6 +200,33 @@ local function boost(job_matchers, opts)
     end
 end
 
+local function boost_oldest(job_matchers, opts)
+    local quiet = opts.quiet
+    local count = opts.oldest_count
+    local job = df.global.world.jobs.list
+    while ( job.next ) do
+        job = job.next
+        if (job.item and not job.item.flags.do_now and job.item.order_id == -1) then
```
myk002, line 209: I see this excludes manager orders. should it also exclude repeat jobs? what about jobs that already have a worker attached and are in progress?

myk002 (commented):
can i haz rst docs? also changelog?

it would be nice to be able to see *what* the oldest N jobs are before they are prioritized. I'm also a little unsure how to promote/explain the oldest jobs functionality so that it gets used appropriately. should we add a repeat command to the control panel that calls it once a <time quantum>?

prioritize.lua
ab9rf: it probably should preclude repeat jobs, i don't use those in my forts except very rarely so it did not come up for me

the others are also good ideas

ab9rf:
> it would be nice to be able to see _what_ the oldest N jobs are before they are prioritized. I'm also a little unsure how to promote/explain the oldest jobs functionality so that it gets used appropriately. should we add a repeat command to the control panel that calls it once a ?

all of these are "maybe? dunno yet, need more experiences" to tell, i'm still experimenting with numbers to use here

a reporting tool might be a good idea, but i'm not sure how to go about that

i have a separate script that does something along these lines but it is very much not ready to be shared with the community

PR #860, opened 2023-10-13
move unavailable scripts into cold storage

myk002:
so they don't clutter up the tool indices

https://github.com/DFHack/dfhack/issues/3503


lethosor (commented):
This will also prevent running the scripts at all, which is a change. Is that what you want?

myk002:
Yeah, that's the debate in the issue. Also discord: https://discord.com/channels/793331351645323264/849036993019117599/1162264840014925824

I was going to at least look into moving the tools that were currently completely non-functional

You know, now that I think about it more, what about instead of moving the scripts, we just remove the tags? I think that might satisfy the request in the issue. We can back up the tags in that online spreadsheet, remove all tags other than "unavailable", and just leave everything else as it is. How's that?

PR #861, opened 2023-10-14
Updated transform-unit.lua to try to select the currently selected un…

RosaryMala:
…it if none is provided.

It still can crash the game, however.


myk002 (approved):
PR looks fine for what it is. If you do get this script working, there are some cleanup tasks that need to be done as well (e.g. doc and tag updates, script formatting, etc.)

PR #865, opened 2023-10-15
[prioritize] refuse to automatically prioritize dig jobs

myk002:
since so many jobs can bork the DF job scheduler. print a warning instead.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -43,6 +43,7 @@ Template for new versions:
 - `hide-tutorials`: fix the embark tutorial prompt sometimes not being skipped
 
 ## Misc Improvements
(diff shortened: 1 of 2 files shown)


ab9rf:
i would suggest that a better idea might be to refuse to prioritize any combination of jobs where the number of jobs affected exceeds the number of units in the fort available to do labors

myk002:
Marking a lot of jobs by itself is not usually an issue. If we mark 500 BringItemToDepot jobs as `do_now`, that's perfectly fine. The problem with dig jobs is that they continuously trickle in, which is far more disruptive to the fortress than getting all hands on deck for a single-shot burst.

PR #870, opened 2023-10-18
Update unit-syndromes.lua

Pebob:
added missing return in the getSyndromeName function


myk002 (commented):
nice : ) could you add a line to changelog.txt indicating the (player-visible) issue that this fixes?

myk002 (approved):

PR #871, opened 2023-10-20
Use lua time class for date math.

dhthwy:
Initial. I did a quick grep thru the files. gui/petitions and list-agreements seems to be the main ones to benefit from this class, but I didn't look very hard either. There might be other candidates.


dhthwy:
PR depends on https://github.com/DFHack/dfhack/pull/3907

PR #873, opened 2023-10-23
WIP New lua script: a gui pregnancy tool

Pebob:
TODO:
1.Add a help string.
2.The exit function also isnt working as intended right now.
3.changelog.txt would also need to be update

anything else?

Would appreciate any feedback (new to lua / dfhack scripts)


gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+        end
+    else return ('No father selected')
+    end
+end
+
+function PregnancyGui:findName(unit)
+    if unit.name.has_name then
+        return dfhack.TranslateName(unit.name)
+    elseif unit.name.nickname ~= "" then
+        return unit.name.nickname
```
myk002, line 184: doesn't TranslateName already handle this case?

gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+    end
+end
+
+function PregnancyGui:CreatePregnancy()
+	local genes,father_id,father_caste,father_name
+    local bypass = true
+    local force = self.subviews.Force:getOptionValue()
+
+    local count = #self.msg
+    for i=0, count do self.msg[i]=nil end --empty self.msg
```
myk002, line 247:
```suggestion
    self.msg = {}
```

gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+    end
+    if culled then 
+        return {'Unknown','culled'}
+    else 
+        return {dfhack.TranslateName(historical_spouse.name),spouse_loc,historical_spouse}
+    end
+end
+
+function PregnancyGui:CreatePregnancy()
+	local genes,father_id,father_caste,father_name
```
myk002, line 242:
```suggestion
    local genes,father_id,father_caste,father_name
```

gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+        else 
+            table.insert(self.msg, ('FAILED:%s%s is already pregnant, no father is recorded'):format(
+                NEWLINE,
+                self:findName(self.mother)
+        ))
+        end
+    end
+    -- self.success = false
+    if bypass or force then
+        --TODO add GUI to select the number of months for pregnancy timer
```
myk002, line 297: You can use a RangeSlider for this. See example here: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/zone.lua#L169-L217

gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+            frame={t=12},
+            frame_style=gui.FRAME_INTERIOR,
+            autoarrange_subviews=1,
+            subviews={
+                widgets.HotkeyLabel{
+                    frame={l=0},
+                    key='CUSTOM_SHIFT_P',
+                    label="Create pregnancy",
+                    on_activate=self:callback('CreatePregnancy'),
+                    enabled=function() return self.mother or self.father and self.father_historical end
```
myk002, line 68: this condition doesn't make sense to me. it's enabled if the mother is set OR if the (father and father_historical) is set? How can you create a pregnancy without a mother being set?

myk002 (commented):
Looks pretty good! Thank you for writing this! What about the case of creating a pregnancy without a specified father? You already have the logic for duplicating the female genes in the case of a historical father. You could extend this to support creating a pregnancy without a selected father at all.

gui/pregnancy.lua
Pebob: Hmm... i tried this but only seem to get empty strings in the label widgets...

gui/pregnancy.lua
Pebob: Thanks! I got it working (kinda) i had to reverse index my values and use the index for the get_left_idx_fn and get_right_idx_fn.

gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+
+function PregnancyScreen:init()
+    self:addviews{PregnancyGui{}}
+end
+
+function PregnancyScreen:onDismiss()
+    view = nil
+end
+
+view = view and view:raise() or PregnancyScreen{}:show()
```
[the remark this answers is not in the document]
Pebob: not quite sure what that means, sorry!!

really appreciate the feedback.
Thank you!!!!!!!

gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+                    text_to_wrap=self.msg,
+                    show_tooltip=true
+                },
+
+                widgets.ToggleHotkeyLabel{
+                    view_id='Force',
+                    label='Force',
+                    options={{label='On', value=true, pen=COLOR_GREEN},
+                    {label='Off', value=false, pen=COLOR_RED}},
+                    initial_option=false
```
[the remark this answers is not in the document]
Pebob: The defaults are:
```
options={{label='On', value=true, pen=COLOR_GREEN},
             {label='Off', value=false}}
```
I only really wanted to add the red colour to the off option, is there a better way to do that?

Pebob:
Yeah, I was planning on adding a option to not have a father.

I did do some searching but I couldnt find the genes or historical figures (i found their attributes etc). Do the genes exist? or do you (or anyone else) know if there are base genes for a entity?

Thank you very much for the detailed feedback.

gui/pregnancy.lua
Pebob: Re the RangeSlider - I have some issues dragging the slider to the minimum value (for me I have to slide all the way to the right to select the minimum).
Its not the biggest deal.

I did have to make the window pretty wide (initially i was planning to give the options in weeks). I guess RangeSlider is most suited for when there are <10 or so options?

I am curious if the reverse indexing was correct too.

gui/pregnancy.lua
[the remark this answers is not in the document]
myk002: To put it more directly, you need a blank line at the end of the file

gui/pregnancy.lua
```diff
@@ -0,0 +1,328 @@
...
+            subviews={
+                widgets.HotkeyLabel{
+                    frame={l=0},
+                    key='CUSTOM_SHIFT_P',
+                    label="Create pregnancy",
+                    on_activate=self:callback('CreatePregnancy'),
+                    enabled=function() return self.mother or self.father and self.father_historical end
+                },
+                widgets.TooltipLabel{
+                    text_to_wrap=self.msg,
```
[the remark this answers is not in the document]
Pebob: ohhh, i missed this last week, ill try this this evening / tomorrow (in combination with `self.msg = {}`)

gui/pregnancy.lua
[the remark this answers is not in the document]
myk002: docs are still required

gui/pregnancy.lua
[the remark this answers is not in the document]
myk002: no, there isn't this is good

gui/pregnancy.lua
```diff
@@ -3,18 +3,32 @@ local widgets = require('gui.widgets')
 
 PregnancyGui = defclass(PregnancyGui, widgets.Window)
 PregnancyGui.ATTRS {
-    frame_title='My Window',
-    frame={w=50, h=45},
+    frame_title='Pregnancy manager',
+    frame={w=64, h=35},
     resizable=true, -- if resizing makes sense for your dialog
     resize_min={w=50, h=20}, -- try to allow users to shrink your windows
```
myk002, line 9: are these numbers still appropriate?

gui/pregnancy.lua
```diff
@@ -3,18 +3,32 @@ local widgets = require('gui.widgets')
...
 }
 
 function PregnancyGui:init()
-    self.mother = false
+    if dfhack.gui.getSelectedUnit(true).sex == df.pronoun_type.she then
```
myk002, line 13: this will throw if there is no unit selected when the UI is opened. You should check for a nil return value from `getSelectedUnit()` before using it

gui/pregnancy.lua
```diff
@@ -3,18 +3,32 @@ local widgets = require('gui.widgets')
...
+        self.mother = dfhack.gui.getSelectedUnit(true)
+    else self.mother = false
```
myk002, line 15: unless a field is always intended to be a boolean, it's better to leave unset values unset so they are `nil` instead of `false`.

gui/pregnancy.lua
```diff
@@ -26,14 +40,14 @@ function PregnancyGui:init()
                 },
                 widgets.HotkeyLabel{
                     frame={l=0},
-                    label="Select Mother",
+                    label="Set mother to selected unit",
```
myk002, line 43: you could also set `enabled=function() local unit = dfhack.gui.getSelectedUnit(true) return unit and unit.sex == df.pronoun_type.she end,`

myk002: and similarly for the father HotkeyLabel

gui/pregnancy.lua
```diff
@@ -42,47 +56,82 @@ function PregnancyGui:init()
...
                 widgets.ToggleHotkeyLabel{
+                    frame={l=1, t=1},
                     view_id='Force',
-                    label='Force',
+                    label='Replace existing pregnancy',
                     options={{label='On', value=true, pen=COLOR_GREEN},
                     {label='Off', value=false, pen=COLOR_RED}},
                     initial_option=false
                 },
+                widgets.TooltipLabel{
```
myk002, line 91: Can you use a plain `Label` here? The attributes you're setting seem to undo everything that makes it a `TooltipLabel`

gui/pregnancy.lua
```diff
@@ -3,18 +3,32 @@ local widgets = require('gui.widgets')
...
     self.father = false
     self.father_historical = false
     self.msg = {}
-    -- self.success = false
+
+    local term_options = {}
+    local term_index = {}
+    local months
+    for months=0,10 do
+        -- table.insert(term_options,{label=('%s months'):format(months),value=months}) --I tried this to add labels, probably doing something wrong, it broke the range widget
```
myk002, line 25: I notice below you were getting the label to index an array. this should work fine if you get the option value instead

myk002:
Is this PR still in progress?

myk002:
I'll see if I can get this fixed up and merged

PR #884, opened 2023-11-06
Add trackstop and rollers overlays

robob27:
Add overlay for changing dump direction and friction of track stops after construction. If this is accepted, I can do a follow-up to remove the old trackstop.cpp stuff.

https://github.com/DFHack/scripts/assets/2104555/8390e43f-b9cf-472f-8529-fcc43422137e

Some screenshots showing how it looks with various suspend manager/building plan states:

image

image

Also adds rollers overlay (same position, size, most of the same options):

https://github.com/DFHack/scripts/assets/2104555/7778ef61-b9b3-443f-ba91-6102ab34d81a

The directions _appear_ to be flipped here but I double checked them against freshly built ones. When you press the arrow up and build the result is "FromSouth" and so on.

### DIFF
--- a/docs/trackstop.rst
+++ b/docs/trackstop.rst
@@ -2,9 +2,9 @@ trackstop
 =========
 
 .. dfhack-tool::
-    :summary: Overlay to allow changing track stop friction and dump direction and roller direction and speed after construction.
-    :tags: fort gameplay buildings interface
+    :summary: Add dynamic configuration options for track stops.
+    :tags: fort buildings interface
 
-This script provides 2 overlays that are managed by the `overlay` framework.
+This script provides 2 overlays that are managed by the `overlay` framework. The script does nothing when executed.
 The trackstop overlay allows the player to change the friction and dump direction of a selected track stop after it has been constructed.
-The rollers overlay allows the player to change the roller direction and speed of a selected track stop after it has been constructed.
+The rollers overlay allows the player to change the roller direction and speed of a selected roller after it has been constructed.
--- a/docs/trackstop.rst
+++ b/docs/trackstop.rst
@@ -1,0 +1,9 @@
+trackstop
+=========
+
+.. dfhack-tool::
+    :summary: Overlay to allow changing track stop friction and dump direction after construction
+    :tags: fort gameplay buildings interface
+
+This script provides an overlay that is managed by the `overlay` framework.
+The overlay allows the player to change the friction and dump direction of a selected track stop after it has been constructed.
--- a/docs/trackstop.rst
+++ b/docs/trackstop.rst
@@ -2,8 +2,9 @@ trackstop
 =========
 
 .. dfhack-tool::
-    :summary: Overlay to allow changing track stop friction and dump direction after construction
+    :summary: Overlay to allow changing track stop friction and dump direction and roller direction and speed after construction.
     :tags: fort gameplay buildings interface
 
-This script provides an overlay that is managed by the `overlay` framework.
-The overlay allows the player to change the friction and dump direction of a selected track stop after it has been constructed.
+This script provides 2 overlays that are managed by the `overlay` framework.
+The trackstop overlay allows the player to change the friction and dump direction of a selected track stop after it has been constructed.
+The rollers overlay allows the player to change the roller direction and speed of a selected track stop after it has been constructed.
--- a/trackstop.lua
+++ b/trackstop.lua
@@ -1,13 +1,19 @@
 -- Overlay to allow changing track stop friction and dump direction after construction
 --@ module = true
+
+if not dfhack_flags.module then
+  qerror('trackstop cannot be called directly')
+end
+
 local gui = require('gui')
 local widgets = require('gui.widgets')
 local overlay = require('plugins.overlay')
+local utils = require('utils')
 
-local NORTH = 'North'
-local EAST = 'East'
-local SOUTH = 'South'
-local WEST = 'West'
+local NORTH = 'North ^'
+local EAST = 'East >'
+local SOUTH = 'South v'
+local WEST = 'West <'
 
 local LOW = 'Low'
 local MEDIUM = 'Medium'
@@ -25,10 +31,7 @@ local FRICTION_MAP = {
   [MAX] = 50000,
 }
 
-local FRICTION_MAP_REVERSE = {}
-for k, v in pairs(FRICTION_MAP) do
-  FRICTION_MAP_REVERSE[v] = k
-end
+local FRICTION_MAP_REVERSE = utils.invert(FRICTION_MAP)
 
 local SPEED_MAP = {
   [LOW] = 10000,
@@ -38,10 +41,7 @@ local SPEED_MAP = {
   [MAX] = 50000,
 }
 
-local SPEED_MAP_REVERSE = {}
-for k, v in pairs(SPEED_MAP) do
-  SPEED_MAP_REVERSE[v] = k
-end
+local SPEED_MAP_REVERSE = utils.invert(SPEED_MAP)
 
 local DIRECTION_MAP = {
   [NORTH] = df.screw_pump_direction.FromSouth,
@@ -50,17 +50,14 @@ local DIRECTION_MAP = {
   [WEST] = df.screw_pump_direction.FromEast,
 }
 
-local DIRECTION_MAP_REVERSE = {}
-for k, v in pairs(DIRECTION_MAP) do
-  DIRECTION_MAP_REVERSE[v] = k
-end
+local DIRECTION_MAP_REVERSE = utils.invert(DIRECTION_MAP)
 
 TrackStopOverlay = defclass(TrackStopOverlay, overlay.OverlayWidget)
 TrackStopOverlay.ATTRS{
-  default_pos={x=-71, y=29},
+  default_pos={x=-73, y=29},
   default_enabled=true,
   viewscreens='dwarfmode/ViewSheets/BUILDING/Trap',
-  frame={w=27, h=4},
+  frame={w=25, h=4},
   frame_style=gui.MEDIUM_FRAME,
   frame_background=gui.CLEAR_PEN,
 }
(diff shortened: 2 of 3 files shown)


myk002 (commented):
I suggest adding a CLI so this script can be used to programmatically alter the properties of track stops, for example as part of a dwarfputing setup

Also needs docs (docs/trackstop.rst) and a changelog entry (changelog.txt)

robob27:
> I suggest adding a CLI so this script can be used to programmatically alter the properties of track stops, for example as part of a dwarfputing setup
> 
> Also needs docs (docs/trackstop.rst) and a changelog entry (changelog.txt)

Added docs and changelog entry! CLI would be cool. You think it would be alright if we do that in a separate PR? I won't be able to finish that up today and not sure exactly when I'll get around to it. Could be very soon... or not 😅. Also realized that trackstop.cpp dealt with rollers too, so I should probably investigate those at some point too.

robob27:
Updated with rollers. trackstop.cpp handled both trackstops and rollers without being named in such a way to indicate that. Should we change the name of this script to something like trackstoprollers.lua or leave as is?

myk002:
Leave it as it is, I think.

myk002:
> The directions appear to be flipped here but I double checked them against freshly built ones. When you press the arrow up and build the result is "FromSouth" and so on.

Is that backwards? If the arrow is up, then you're dumping to the north, and "FromSouth" == "ToNorth"

robob27:
> > The directions appear to be flipped here but I double checked them against freshly built ones. When you press the arrow up and build the result is "FromSouth" and so on.
> 
> Is that backwards? If the arrow is up, then you're dumping to the north, and `FromSouth` == "ToNorth"



> > The directions appear to be flipped here but I double checked them against freshly built ones. When you press the arrow up and build the result is "FromSouth" and so on.
> 
> Is that backwards? If the arrow is up, then you're dumping to the north, and `FromSouth` == "ToNorth"

The direction thing I mentioned only applies to the roller direction:


https://github.com/DFHack/scripts/assets/2104555/fe2ee2fe-4ffe-416a-81dd-7ef33cb73561

Dump direction for stops is handled by `dump_x_shift` and `dump_y_shift` instead of a screw pump direction.

docs/trackstop.rst
```diff
@@ -0,0 +1,10 @@
+trackstop
+=========
+
+.. dfhack-tool::
+    :summary: Overlay to allow changing track stop friction and dump direction and roller direction and speed after construction.
```
myk002, line 5: Summaries should be 53 characters max

docs/trackstop.rst
```diff
@@ -0,0 +1,10 @@
...
+    :tags: fort gameplay buildings interface
+
+This script provides 2 overlays that are managed by the `overlay` framework.
```
myk002, line 8: maybe mention that the script does nothing when run

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+  [LOW] = 50,
+  [MEDIUM] = 500,
+  [HIGH] = 10000,
+  [MAX] = 50000,
+}
+
+local FRICTION_MAP_REVERSE = {}
+for k, v in pairs(FRICTION_MAP) do
+  FRICTION_MAP_REVERSE[v] = k
+end
```
myk002, line 31:
```suggestion
local FRICTION_MAP_REVERSE = utils.invert(FRICTION_MAP)
```

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+  [MEDIUM] = 20000,
+  [HIGH] = 30000,
+  [HIGHER] = 40000,
+  [MAX] = 50000,
+}
+
+local SPEED_MAP_REVERSE = {}
+for k, v in pairs(SPEED_MAP) do
+  SPEED_MAP_REVERSE[v] = k
+end
```
myk002, line 44:
```suggestion
local SPEED_MAP_REVERSE = utils.invert(SPEED_MAP)
```

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+  [NORTH] = df.screw_pump_direction.FromSouth,
+  [EAST] = df.screw_pump_direction.FromWest,
+  [SOUTH] = df.screw_pump_direction.FromNorth,
+  [WEST] = df.screw_pump_direction.FromEast,
+}
+
+local DIRECTION_MAP_REVERSE = {}
+for k, v in pairs(DIRECTION_MAP) do
+  DIRECTION_MAP_REVERSE[v] = k
+end
```
myk002, line 56: utils.invert

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+
+TrackStopOverlay = defclass(TrackStopOverlay, overlay.OverlayWidget)
+TrackStopOverlay.ATTRS{
+  default_pos={x=-71, y=29},
+  default_enabled=true,
```
myk002, line 61: since this is a gameplay mod that lets you do things that are not normally possible, I'm hesitant to enable it by default. Instead, I think you should document in the help that it needs to be turned on in the Overlays tab in `gui/control-panel`.

@ab9rf @lethosor what do you think? is this the right call?

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+  friction_cycle:setOption(FRICTION_MAP_REVERSE[friction])
+
+  self.subviews.dump_direction:setOption(self:getDumpDirection())
+
+  TrackStopOverlay.super.render(self, dc)
+end
+
+function TrackStopOverlay:shouldRender()
+  local building = dfhack.gui.getSelectedBuilding()
+  return building and building.trap_type == df.trap_type.TrackStop
```
myk002, line 143: if you expect building can be nil, then be sure to call `getSelectedBuilding(true)` instead of just `getSelectedBuilding()` so you don't spam the console with errors

myk002: on the other hand, though, why would this ever return false? You're already attached to `viewscreens='dwarfmode/ViewSheets/BUILDING/Trap'`

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+  end
+  TrackStopOverlay.super.onInput(self, keys)
+end
+
+function TrackStopOverlay:init()
+  self:addviews{
+    widgets.Label{
+      frame={t=0, l=0},
+      text='Dump',
+    },
```
myk002, line 158: The usual method is to set the `label` property of the `CycleHotkeyLabel` instead of having a separate `Label` label

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+    widgets.CycleHotkeyLabel{
+      frame={t=0, l=9},
+      key='CUSTOM_CTRL_X',
+      options={NONE, NORTH, EAST, SOUTH, WEST},
```
myk002, line 162: I'd give NONE a different color, like maybe blue?

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+      view_id='dump_direction',
+      on_change=function(val) self:setDumpDirection(val) end,
+    },
+    widgets.Label{
+      frame={t=1, l=0},
+      text='Friction',
+    },
```
myk002, line 169: here too, you could set `label='Friction',` in the `CycleHotkeyLabel`

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+    widgets.CycleHotkeyLabel{
+      frame={t=1, l=9},
+      key='CUSTOM_CTRL_F',
+      options={NONE, LOW, MEDIUM, HIGH, MAX},
```
myk002, line 173: These could get colors according to their strength. Blue for NONE, then green, yellow, red, lightred

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+      view_id='friction',
+      on_change=function(val) self:setFriction(val) end,
+    },
+  }
+end
+
+RollerOverlay = defclass(RollerOverlay, overlay.OverlayWidget)
+RollerOverlay.ATTRS{
+  default_pos={x=-71, y=29},
+  default_enabled=true,
```
myk002, line 183: same here, since it's a gameplay mod, shouldn't be enabled by default and instead documented about how to turn on

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+function RollerOverlay:render(dc)
+  local building = dfhack.gui.getSelectedBuilding()
+
+  self.subviews.direction:setOption(DIRECTION_MAP_REVERSE[building.direction])
+  self.subviews.speed:setOption(SPEED_MAP_REVERSE[building.speed])
+
+  TrackStopOverlay.super.render(self, dc)
+end
+
+function RollerOverlay:init()
```
myk002, line 225: similar notes here. labels as CycleHotkeyLabel properties, colorize options

trackstop.lua
```diff
@@ -0,0 +1,259 @@
...
+  }
+end
+
+OVERLAY_WIDGETS = {
+  trackstop=TrackStopOverlay,
+  rollers=RollerOverlay,
+}
+
+if not dfhack_flags.module then
+  main{...}
```
myk002, line 258: there is no `main`, no?

trackstop.lua
ab9rf: if all this does is enable reconfiguring a track stop without reconstructing it, i don't have a problem with it being on by default

trackstop.lua
robgoodberry: Yeah I assume the building will always be there

trackstop.lua
robob27: Ah yeah, that shouldn't be there yet

trackstop.lua
myk002: Ok, how about removing the `gameplay` tag, then. I'd like to keep a clear policy that "gameplay mods are not on by default", and if we're classifying this as "on by defaultable", then I'd like it to not be in the "gameplay" group.

trackstop.lua
myk002: so maybe this function can be removed entirely (and onInput() as well)

robob27:
I also added `^` `>` `<` `v` in the directions while I was adding the colors to the other ones and I think I like it. Let me know what y'all think.

https://github.com/DFHack/scripts/assets/2104555/54d5274c-46b8-476e-b14a-b93ddc0ac13f

trackstop.lua
robob27: Removed the gameplay tag as discussed in another comment

myk002 (approved):

myk002:
will merge once the `trackstop` plugin is removed (otherwise the docs will conflict)

trackstop.lua
```diff
@@ -0,0 +1,258 @@
...
+  local building = dfhack.gui.getSelectedBuilding()
+  return building and building.trap_type == df.trap_type.TrackStop
+end
+
+function TrackStopOverlay:onInput(keys)
+  if not self:shouldRender() then
+    return
+  end
+  TrackStopOverlay.super.onInput(self, keys)
+end
```
myk002, line 148: Oh, I forgot -- can this function be removed?

trackstop.lua
```diff
@@ -0,0 +1,258 @@
...
+    building.use_dump = 1
+    building.dump_x_shift = -1
+    building.dump_y_shift = 0
+  end
+end
+
+function TrackStopOverlay:render(dc)
+  if not self:shouldRender() then
+    return
+  end
```
myk002, line 125: can this be removed?

trackstop.lua
robob27: Hmm, so there is currently a spot where the console can get spammed with errors with the way I have this written.

If you have gm-editor open and try to edit a field that opens a dialog to input the new value, then `getSelectedBuilding()` is returning `nil` and the console is spammed with errors. Maybe I should use `getSelectedBuilding(true)` and guard against it being `nil` after all?

myk002: maybe that's a good idea, at least for the time being. once the dialogs are updated to use ZScreens, this will no longer be an issue, but I don't know when I'm going to get around to doing that.

myk002 (approved):

PR #889, opened 2023-11-08
WIP Add reorderstops overlay

robob27:
Need to do some testing with this and figure out if changing the ids breaks stuff, or if anything else needs to be cleared out like the guide paths. Putting up what I have for now.


trackstop.lua
```diff
@@ -52,6 +52,23 @@ local DIRECTION_MAP = {
 
 local DIRECTION_MAP_REVERSE = utils.invert(DIRECTION_MAP)
 
+local function swapElements(tbl, index1, index2)
+  tbl[index1], tbl[index2] = tbl[index2], tbl[index1]
+  return tbl
+end
+
+local function reset_guide_paths(conditions)
+  for _i, condition in ipairs(conditions) do
```
myk002, line 61: for unused variables, the Lua convention is just to use an underscore with no other characters.

trackstop.lua
```diff
@@ -236,7 +260,170 @@ function RollerOverlay:init()
...
+      view_id='routes',
+      frame={t=1,l=1},
+      choices={},
+      on_select=function(index, item)
+        if not item then return end
+        if item.type == 'stop' then
+          local item_pos = df.global.plotinfo.hauling.routes[item.route_index].stops[item.stop_index].pos
+          df.global.game.main_interface.recenter_indicator_m.x = item_pos.x
+          df.global.game.main_interface.recenter_indicator_m.y = item_pos.y
+          df.global.game.main_interface.recenter_indicator_m.z = item_pos.z
```
myk002, line 292: pass another `, true` to `revealInDwarfmodeMap` to get this for free

trackstop.lua
```diff
@@ -236,7 +260,170 @@ function RollerOverlay:init()
   }
 end
 
+selected_stop = selected_stop or nil
```
myk002, line 263: why does this need to be a global instead of an instance variable in ReorderStopsWindow?

trackstop.lua
```diff
@@ -236,7 +260,170 @@ function RollerOverlay:init()
...
+
+ReorderStopsWindow = defclass(ReorderStopsWindow, widgets.Window)
+ReorderStopsWindow.ATTRS {
+  frame={t=4,l=60,w=45, h=28},
+  frame_title='Reorder Stops',
+  view_id='main',
```
myk002, line 269: since this is never referenced, you don't need to give it an explicit view_id

trackstop.lua
```diff
@@ -236,7 +260,170 @@ function RollerOverlay:init()
   }
 end
 
+selected_stop = selected_stop or nil
+
+ReorderStopsWindow = defclass(ReorderStopsWindow, widgets.Window)
+ReorderStopsWindow.ATTRS {
+  frame={t=4,l=60,w=45, h=28},
```
myk002, line 267: since this window contains a list with a variable number of elements, it would be useful to set `resizable=true`

trackstop.lua
```diff
@@ -236,7 +260,170 @@ function RollerOverlay:init()
...
+          local hauling = df.global.plotinfo.hauling
+          local routes = hauling.routes
+          local view_stops = hauling.view_stops
+          local route = routes[item.route_index]
+
+          -- rearrange stops
+          if item.type == 'stop' then
+            local stop_index = item.stop_index
+
+            -- don't allow moving stops to a different route
```
myk002, line 307: why not?

trackstop.lua
robob27: Ah yeah that's it. Was wondering why it still showed greyed out in the editor but hadn't looked into it yet. ty!

trackstop.lua
robob27: Siick

trackstop.lua
robob27: Can fix!

trackstop.lua
robob27: Yeah this might be fine, I'll revisit it for sure.

trackstop.lua
robob27: I guess with the way modals work I don't need the other global variable either. Nice. Removed that too.

trackstop.lua
robob27: Now possible

changelog.txt
```diff
@@ -28,7 +28,10 @@ Template for new versions:
 
 ## New Tools
 - `sync-windmills`: synchronize or randomize movement of active windmills
-- `trackstop`: new overlay to allow changing track stop dump direction and friction and roller direction and speed after construction
+- `trackstop`: provides 3 new overlays:
+    - trackstop: allow changing track stop dump direction and friction
+    - rollers: allow changing roller direction and speed
+    - reorderstops: reorder stops in hauling routes
```
myk002, line 34: this kind of formatting doesn't translate well to the final rendered changelog. each line should be independent. If you start each of them with ``` `trackstop`: ``` then they will get combined into a list similar to your intention here

trackstop.lua
```diff
@@ -236,7 +268,197 @@ function RollerOverlay:init()
...
+  df.global.game.main_interface.recenter_indicator_m.x = -30000
+  df.global.game.main_interface.recenter_indicator_m.y = -30000
+  df.global.game.main_interface.recenter_indicator_m.z = -30000
+end
+
+ReorderStopsOverlay = defclass(ReorderStopsOverlay, overlay.OverlayWidget)
+ReorderStopsOverlay.ATTRS{
+  default_pos={x=6, y=6},
+  default_enabled=true,
+  viewscreens='dwarfmode/Hauling',
```
myk002, line 444: is this specific enough, or should it disappear on hauling subscreens? We might need to extend Gui.cpp if this is the case

trackstop.lua
```diff
@@ -236,7 +268,197 @@ function RollerOverlay:init()
...
 
+ReorderStopsWindow = defclass(ReorderStopsWindow, widgets.Window)
+ReorderStopsWindow.ATTRS {
+  frame={t=4,l=60,w=49, h=26},
+  frame_title='Reorder Stops',
+  resizable=true,
+}
+
+local SELECT_STOP_HINT = 'Select a stop to move'
+local SELECT_ANOTHER_STOP_HINT = 'Select another stop to swap or same to cancel'
```
myk002, line 279: right click should also cancel out of a partially completed reorder operation. you can do this by handling `keys._MOUSE_R` in `ReorderStopsWindow:onInput(keys)` and returning `true` if you have handled the right click and don't want the window to close.

myk002 (commented):
I have to admit this scares me a little bit, but if @ab9rf gives the ok that we've handled all the potential places where the id is stored, then I'll allow it.

ab9rf:
> I have to admit this scares me a little bit, but if @ab9rf gives the ok that we've handled all the potential places where the id is stored, then I'll allow it.

the function where this all takes place in the game is one of the monsters that ghidra cannot decompile (i left it running for six hours and it was unable to do so). clearing the guide path after reordering the stop resolved the odd behavior we were seeing previously. there aren't any other fields involved and since we're just reordering the elements in the vector, pointers to the individual stops are unaffected

the one thing i'm not sure this covers is the `vehicle_current_stop` (bay12 name) vector in `hauling_routest` (also bay12 name). basically if you change the order of the stops, you also have to update `hauling_route::vehicle_stops` to reflect that the index where that vehicle is currently has changed. i haven't reviewed this code to confirm that it actually does this

myk002:
also, make sure you test to see what happens if a minecart is en route to a reordered stop, both in the pushing case and the guided case

robgoodberry:
Thanks to both of you! Will update and ensure to test these scenarios.

robob27:
I was wondering why I got a reply for my own notification, replied on my other account 😅 haha

trackstop.lua
robob27: It looks like it's okay on all of them except them item selection one. Will fix.

trackstop.lua
robob27: Fixed

robob27:
I think the code now handles updating `hauling_route::vehicle_stops` properly when reordering stops within the same route. I don't currently do anything with `hauling_route::vehicle_stops` when swapping stops between different routes. I think this works out okay?

If I have this setup:

image

And I swap Stop 1 and Stop 4 (different physical locations):

image

From the same initial setup, if I swap Stop 1 and Stop 5 (same physical locations):

image

Again from that setup, if I swap Stop 2 and Stop 4 (different physical locations):

image

Again from that setup, if I swap Stop 1 and Stop 2 (same route, same result regardless of which stop is selected first):

image

So, if after swapping stops, the cart for that route is not physically at the stop index it was at before the swap, it will be moved there. If the swapped stop is at the same physical location, the cart will be considered in the right spot. If the stops are swapped within the same route, the vehicle index is updated appropriately.

Still working on more playtesting.

myk002:
I have notes on the UX, but they can wait until you have the basic functionality tested

PR #890, opened 2023-11-10
unforbid: ignore tattered items by default

AndrielChaoti:
also adds option to disable ignoring tattered items.
this is very much a first pass with minimal testing :)

Fixes https://github.com/DFHack/dfhack/issues/3967

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -33,6 +33,7 @@ Template for new versions:
 ## New Features
 - `gui/design`: show selected dimensions next to the mouse cursor when designating with vanilla tools, for example when painting a burrow or designating digging
 - `quickfort`: new blueprint mode for designating burrows
+- `unforbid`: now ignores worn and tattered items by default (X/XX), use -x to bypass
 
 ## Fixes
 - `gui/unit-syndromes`: show the syndrome names properly in the UI
--- a/changelog.txt
+++ b/changelog.txt
@@ -33,7 +33,7 @@ Template for new versions:
 ## New Features
 - `gui/design`: show selected dimensions next to the mouse cursor when designating with vanilla tools, for example when painting a burrow or designating digging
 - `quickfort`: new blueprint mode for designating burrows
-- `unforbid`: now ignores worn and tattered items by default (X/XX), use -x to bypass
+- `unforbid`: now ignores worn and tattered items by default (X/XX), use -X to bypass
 
 ## Fixes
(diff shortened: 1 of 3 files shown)


unforbid.lua
```diff
@@ -2,10 +2,10 @@
 
 local argparse = require('argparse')
 
-local function unforbid_all(include_unreachable, quiet)
+local function unforbid_all(include_unreachable, include_tattered, quiet)
     if not quiet then print('Unforbidding all items...') end
 
-    local citizens = dfhack.units.getCitizens(true)
+    local citizens = dfhack.units.getCitizens()
```
myk002, line 8: You can change this to `false` if you think you have a case for it, but don't remove the parameter. It causes a crash when in debug mode.

unforbid.lua
```diff
@@ -22,6 +22,11 @@ local function unforbid_all(include_unreachable, quiet)
                     if not quiet then print(('  unreachable: %s (skipping)'):format(item)) end
                     goto skipitem
                 end
+
+                if ((not options.include_tattered) and item.wear >= 3) then
```
myk002, line 26: The threshold is debatable. Should it be configurable? I can see the player wanting it to match whatever `cleanowned` is set to. What should be the default? We should probably match the default `cleanowned` command in the control panel.

myk002 (commented):
Overall, looks great! Don't forget docs and the changelog

docs/unforbid.rst
```diff
@@ -24,3 +24,6 @@ Options
 
 ``-q``, ``--quiet``
     Suppress non-error console output.
+
+``-x``, ``-include-worn``
```
myk002, line 28: the actual option character is capital `X` -- which one would you prefer?

changelog.txt
```diff
@@ -33,6 +33,7 @@ Template for new versions:
 ## New Features
 - `gui/design`: show selected dimensions next to the mouse cursor when designating with vanilla tools, for example when painting a burrow or designating digging
 - `quickfort`: new blueprint mode for designating burrows
+- `unforbid`: now ignores worn and tattered items by default (X/XX), use -x to bypass
```
myk002, line 36:
```suggestion
- `unforbid`: now ignores worn and tattered items by default (X/XX), use -X to bypass
```

myk002 (approved):

PR #896, opened 2023-11-17
add `item` bulk forbid/dump/melt tool

chdoc:
This adds a generic item manipulation tool, allowing to bulk forbid/dump/melt items based on various criteria or remove the aforementioned designations. Also provides an API to serve as backend for other scripts.

### DIFF
--- a/docs/item.rst
+++ b/docs/item.rst
@@ -43,6 +43,10 @@ Options
     Get an accurate count of the items that would be affected, including the
     implicit filters of the selected main action.
 
+``-a, --include-artifacts``
+    Include artifacts in the item list. Regardless of this setting, artifacts
+    are never dumped or melted.
+
 ``-i, --inside <burrow>``
     Only include items inside the given burrow.
 
@@ -74,10 +78,6 @@ Options
     (cf. https://www.lua.org/manual/5.4/manual.html#6.4.1), so "cave spider
     silk" will match both "cave spider silk web" as well as "cave spider silk
     cloth". Use ``^pattern$`` to match the entire description.
-
-``-a, --include-artifacts``
-    Include artifacts in the item list. Regardless of this setting, artifacts
-    are never dumped or melted.
 
 ``-w, --min-wear <integer>``
     Only include items whose wear/damage level is at least ``integer``. Useful
@@ -116,3 +116,85 @@ Options
 
 ``--hidden``
     Only include hidden items.
+
+
+API
+---
+
+The item script can be called programmatically by other scripts, either via the
+commandline interface with ``dfhack.run_script()`` or via the API functions
+defined in :source-scripts:`item.lua`, available from the return value of
+``reqscript('item')``:
+
+* ``item.execute(action, conditions, options)``
+* ``item.executeWithPrinting(action, conditions, options)``
+
+Performs ``action`` (``forbid``, ``melt``, etc.) on all items satisfying
+``conditions`` (a table containing functions from item to boolean). ``options``
+is a table containing the boolean flags ``artifact``, ``dryrun``, and
+``bytype``, which correspond to the (filter) options described above.
+
+The function ``item.execute`` performs no output, while the ``WithPrinting``
+variant performs the same output as the ``item`` tool.
+
+The API provides a number of helper functions to aid in the construction of the
+filter table. The first argument ``tab`` is always the table to which the filter
+should be added.
+
+* ``item.condition_burrow(tab,burrow, outside)``
+    Corresponds to ``--inside`` or ``--outside`` (when ``outside=true``). The
+    ``burrow`` argument must be a burrow object, not a string.
+
+* ``item.condition_type(tab, match)``
+    If ``match`` is a string, this corresponds to ``--type <match>``. Also
+    accepts numbers, matching against ``item:getType()``
+
+* ``item.condition_reachable(tab)``
+    Corresponds to ``--reachable``
+
+* ``item.condition_unreachable(tab)``
+    Corresponds to ``--unreachable``
+
+* ``item.condition_description(tab, pattern)``
+    Corresponds to ``--description <pattern>``
+
+* ``item.condition_material(tab, match)``
+    Corresponds to ``--material <match>``
+
+* ``item.condition_matcat(tab, match)``
+    Corresponds to ``--mat-category <match>``
+
+* ``item.condition_wear(tab, lower, upper)``
+    Selects items with wear level between ``lower`` and ``upper`` (Range 0-3, see above).
+
+* ``item.condition_quality(tab, lower, upper)``
+    Selects items with quality between ``lower`` and ``upper`` (Range 0-5, see above).
+
+* ``item.condition_stockpiled(tab, invert)``
+    Selects stockpiled items, or scattered items when ``invert=true``.
+
+* ``item.condition_forbidden(tab)``
+    Checks for ``item.flags.forbid``
+
+* ``item.condition_melt(tab)``
+    Checks for ``item.flags.melt``
+
+* ``item.condition_dump(tab)``
+    Checks for ``item.flags.dump``
+
+* ``item.condition_hidden(tab)``
+    Checks for ``item.flags.hidden``
+
+* ``item.condition_visible(tab)``
+    Checks for ``not item.flags.hidden``
+
+ API usage example::
+
+   local item = reqscript('item')
+   local cond = {}
+
+   item.condition_type(cond, "BOULDER")
+   item.execute('unhide', cond, {}) -- reveal all boulders
+
+   item.condition_stockpiled(cond, true)
+   item.execute('hide', cond, {})   -- hide all boulders not in stockpiles
--- a/docs/item.rst
+++ b/docs/item.rst
@@ -136,7 +136,7 @@ commandline interface with ``dfhack.run_script()`` or via the API functions
 defined in :source-scripts:`item.lua`, available from the return value of
 ``reqscript('item')``:
 
-* ``execute(action, conditions, options)``
+* ``execute(action, conditions, options [, return_items])``
 
 Performs ``action`` (``forbid``, ``melt``, etc.) on all items satisfying
 ``conditions`` (a table containing functions from item to boolean). ``options``
@@ -146,7 +146,7 @@ and ``owned`` which correspond to the (filter) options described above.
 The function ``execute`` performs no output, but returns three values:
 
 1. the number of matching items
-2. a table containing all matched items, if the action is ``count``
+2. a table containing all matched items, if ``return_items`` is provided and true.
 3. a table containing a mapping from numeric item types to their occurrence
    count, if ``options.bytype=true``
 
--- a/docs/item.rst
+++ b/docs/item.rst
@@ -13,7 +13,7 @@ the number of items that matched the filters and were modified.
 Usage
 -----
 
-``item [ count | [un]forbid | [un]dump | [un]melt ] <filter options>``
+``item [ count | [un]forbid | [un]dump | [un]hide | [un]melt ] <filter options>``
 
 Action names should be self explanatory. All actions other than ``help`` and
 ``count`` have implicit filters associated with them. For instance, ``item
@@ -31,8 +31,7 @@ Examples
     flood-fill to create a burrow covering an entire cavern layer).
 
 ``item melt -t weapon -m steel --max-quality 3``
-    Designate all steel weapons whose quality is less than "exceptional" for
-    melting.
+    Designate all steel weapons whose quality is at most superior for melting.
 
 Options
 -------
@@ -66,11 +65,12 @@ Options
     "metal"). Use ``:lua df.dfhack_material_category`` to get a list of all
     material categories.
 
-``-d, --description <string>``
-    Filter by item description (singular form without stack sizes). Example:
-    "cave spider silk web". Note: Due to a bug, this is of limited use for some
-    animal products such as wool, because their description always includes the
-    stack size.
+``-d, --description <pattern>``
+    Filter by item description (singular form without stack sizes). The
+    ``pattern`` is a Lua pattern
+    (cf. https://www.lua.org/manual/5.4/manual.html#6.4.1), so "cave spider
+    silk" will match both "cave spider silk web" as well as "cave spider silk
+    cloth". Use ``^pattern$`` to match the entire description.
 
 ``-a, --include-artifacts``
     Include artifacts in the item list. Regardless of this setting, artifacts
@@ -91,3 +91,18 @@ Options
 ``-Q, --max-quality <integer>``
     Only include items whose quality level is at most ``integer``. Useful
     values are 0 (ordinary) to 5 (masterwork).
+
+``--forbidden``
+    Only include forbidden items.
+
+``--melting``
+    Only include items designated for melting.
+
+``--dumping``
+    only include items designated for dumping.
+
+``--visible``
+    Only include visible items (i.e., ignore hidden items).
+
+``--hidden``
+    Only include hidden items.
--- a/docs/item.rst
+++ b/docs/item.rst
@@ -1,0 +1,93 @@
+item
+====
+
+.. dfhack-tool::
+    :summary: Perform bulk operations on items based on various properties.
+    :tags: fort productivity items
+
+Filter items in you fort by various properties (e.g., item type, material,
+wear-level, quality, ...), and perform the bulk operations forbid, dump, melt,
+and their inverses. By default, the tool does not consider artifacts. Outputs
+the number of items that matched the filters and were modified.
+
+Usage
+-----
+
+``item [ help | count | [un]forbid | [un]dump | [un]melt ] <filter option>``
+
+Action names should be self explanatory. All actions other than ``help`` and
+``count`` have implicit filters associated with them. For instance, ``item
+forbid --unreachable`` will neither report nor test reachability of items that
+are already forbidden.
+
+Examples
+--------
+
+``item forbid --unreachable``
+    Forbid all items that cannot be reached by any of your citizens.
+
+``item unforbid --inside Cavern1 --type wood``
+    Unforbid/reclaim all logs inside the burrow named "Cavern1" (Hint: use 3D
+    flood-fill to create a burrow covering an entire cavern layer)
+
+``item melt -t weapon -m steel --max-quality 3``
+    Designate all steel weapons whose quality is less than "exceptional" for
+    melting.
+
+Options
+-------
+
+``-n, --dry-run``
+    Use get an accurate count of the items that would be affected, including the
+    implicit filters of the selected main action.
+
+``-i, --inside <burrow>``
+    Only include items inside the given burrow.
+
+``-o, --outside <burrow>``
+    Only include items outside the given burrow.
+
+``-r, --reachable``
+    Only include items reachable by one of your citizens.
+
+``-u, --unreachable``
+    Only include items not reachable by one of your citizens.
+
+``-t, --type <string>``
+    Filter by item type (e.g., BOULDER, CORPSE, ...). Also accepts lower case
+    spelling (e.g. "corpse"). Use ``:lua @df.item_type`` to get the list of all
+    item types.
+
+``-m, --material <string>``
+    Filter by material the item is made out of (e.g., "iron").
+
+``-c, --mat-category <string>``
+    Filter by material category of the material item is made out of (e.g.,
+    "metal"). Use ``:lua df.dfhack_material_category`` to get have a list of all
+    material categories.
+
+``-d, --description <string>``
+    Filter by item description (singular form without stack sizes). Example:
+    "cave spider silk web". Note: Due to a bug, this is of limited use for some
+    animal products such as wool, because their description always includes the
+    stack size.
+
+``-a, --include-artifacts``
+    Include artifacts in the item list. Regardless of this setting, artifacts
+    are never dumped or melted
+
+``-w, --min-wear <integer>``
+    Only include items whose wear/damage level is at least ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``-W, --max-wear <integer>``
+    Only include items whose wear/damage level is at most ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``-q, --min-quality <integer>``
+    Only include items whose quality level is at least ``integer``. Useful
+    values are 0 (standard) to 5 (masterwork).
+
+``-Q, --max-quality <integer>``
+    Only include items whose quality level is at most ``integer``. Useful
+    values are 0 (standard) to 5 (masterwork).
--- a/docs/item.rst
+++ b/docs/item.rst
@@ -58,7 +58,7 @@ Options
     are never dumped or melted.
 
 ``--include-owned``
-    Include items owned by units (e.g., your dwarfs or visitors)
+    Include items owned by units (e.g., your dwarves or visitors)
 
 ``-i, --inside <burrow>``
     Only include items inside the given burrow.
@@ -192,7 +192,7 @@ the filter is described.
 * ``condition_stockpiled(tab, negate)``
     Corresponds to ``--stockpiled``.
 
-* ``condition_[forbid|melt|dump|hidden|owned](tab,negate)``
+* ``condition_[forbid|melt|dump|hidden|owned](tab, negate)``
     Selects items with the respective flag set to ``true`` (e.g.,
     ``condition_forbid`` checks for ``item.flags.forbid``).
 
--- a/docs/item.rst
+++ b/docs/item.rst
@@ -2,11 +2,11 @@ item
 ====
 
 .. dfhack-tool::
-    :summary: Perform bulk operations on items based on various properties.
+    :summary: Perform bulk operations on groups of items.
     :tags: fort productivity items
 
 Filter items in you fort by various properties (e.g., item type, material,
-wear-level, quality, ...), and perform the bulk operations forbid, dump, melt,
+wear-level, quality, ...), and perform bulk operations like forbid, dump, melt,
 and their inverses. By default, the tool does not consider artifacts. Outputs
 the number of items that matched the filters and were modified.
 
@@ -15,10 +15,11 @@ Usage
 
 ``item [ count | [un]forbid | [un]dump | [un]hide | [un]melt ] <filter options>``
 
-Action names should be self explanatory. All actions other than ``help`` and
-``count`` have implicit filters associated with them. For instance, ``item
-forbid --unreachable`` will neither report nor test reachability of items that
-are already forbidden.
+The ``count`` action just counts up the items that would be matched by the
+given filter options. Otherwise, the named property is set (or unset) on all
+the items that the filter options match. The counts reported when you actually
+apply a property might differ from what is output by the ``count`` action if
+some items already have the property set to the desired state.
 
 Examples
 --------
@@ -40,8 +41,9 @@ Options
 -------
 
 ``-n, --dry-run``
-    Get an accurate count of the items that would be affected, including the
-    implicit filters of the selected main action.
+    Get a count of the items that would be modified by an operation, which will be the
+    number returned by the ``count`` action minus the number of items with the desired
+    property already set.
 
 ``-a, --include-artifacts``
     Include artifacts in the item list. Regardless of this setting, artifacts
@@ -69,7 +71,7 @@ Options
 
 ``-c, --mat-category <string>``
     Filter by material category of the material item is made out of (e.g.,
-    "metal"). Use ``:lua df.dfhack_material_category`` to get a list of all
+    "metal"). Use ``:lua @df.dfhack_material_category`` to get a list of all
     material categories.
 
 ``-d, --description <pattern>``
@@ -134,14 +136,14 @@ Performs ``action`` (``forbid``, ``melt``, etc.) on all items satisfying
 is a table containing the boolean flags ``artifact``, ``dryrun``, and
 ``bytype``, which correspond to the (filter) options described above.
 
-The function ``item.execute`` performs no output, while the ``WithPrinting``
+The function ``execute`` performs no output, while the ``WithPrinting``
 variant performs the same output as the ``item`` tool.
 
 The API provides a number of helper functions to aid in the construction of the
 filter table. The first argument ``tab`` is always the table to which the filter
 should be added.
 
-* ``item.condition_burrow(tab,burrow, outside)``
+* ``item.condition_burrow(tab, burrow, outside)``
     Corresponds to ``--inside`` or ``--outside`` (when ``outside=true``). The
     ``burrow`` argument must be a burrow object, not a string.
 
@@ -190,11 +192,11 @@ should be added.
 
  API usage example::
 
-   local item = reqscript('item')
+   local itemtools = reqscript('item')
    local cond = {}
 
-   item.condition_type(cond, "BOULDER")
-   item.execute('unhide', cond, {}) -- reveal all boulders
+   itemtools.condition_type(cond, "BOULDER")
+   itemtools.execute('unhide', cond, {}) -- reveal all boulders
 
-   item.condition_stockpiled(cond, true)
-   item.execute('hide', cond, {})   -- hide all boulders not in stockpiles
+   itemtools.condition_stockpiled(cond, true)
+   itemtools.execute('hide', cond, {})   -- hide all boulders not in stockpiles
--- a/docs/item.rst
+++ b/docs/item.rst
@@ -13,7 +13,7 @@ the number of items that matched the filters and were modified.
 Usage
 -----
 
-``item [ help | count | [un]forbid | [un]dump | [un]melt ] <filter option>``
+``item [ count | [un]forbid | [un]dump | [un]melt ] <filter options>``
 
 Action names should be self explanatory. All actions other than ``help`` and
 ``count`` have implicit filters associated with them. For instance, ``item
@@ -28,7 +28,7 @@ Examples
 
 ``item unforbid --inside Cavern1 --type wood``
     Unforbid/reclaim all logs inside the burrow named "Cavern1" (Hint: use 3D
-    flood-fill to create a burrow covering an entire cavern layer)
+    flood-fill to create a burrow covering an entire cavern layer).
 
 ``item melt -t weapon -m steel --max-quality 3``
     Designate all steel weapons whose quality is less than "exceptional" for
@@ -38,7 +38,7 @@ Options
 -------
 
 ``-n, --dry-run``
-    Use get an accurate count of the items that would be affected, including the
+    Get an accurate count of the items that would be affected, including the
     implicit filters of the selected main action.
 
 ``-i, --inside <burrow>``
@@ -51,7 +51,7 @@ Options
     Only include items reachable by one of your citizens.
 
 ``-u, --unreachable``
-    Only include items not reachable by one of your citizens.
+    Only include items not reachable by any of your citizens.
 
 ``-t, --type <string>``
     Filter by item type (e.g., BOULDER, CORPSE, ...). Also accepts lower case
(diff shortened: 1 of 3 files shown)


chdoc:
Obviously, this is still in a very rough shape. General feedback is welcome; let's not get lost in the nitty-gritty.

item.lua
```diff
@@ -0,0 +1,294 @@
...
+            item.flags.melt = true
+            changed = changed + 1
+        elseif action == 'unmelt' and item.flags.melt then
+            item.flags.melt = false
+            changed = changed + 1
+        end
+        :: skipitem ::
+    end
+    print(count, 'items matched the filter options')
+    print(invisible, 'invisible items were skipped')
```
myk002, line 214: I don't think we should report on invisible items. This is a productivity tool, not an armok tool

myk002 (commented):
Overall, looks great!

There is overlap with `buildingplan` for the filtering by material/quality, but I don't know how much benefit there is to sharing that logic. Maybe if there are performance issues with filtering large amounts of items. Or maybe if you also want to offer category filters like "stone".

I also noted overlap with `logistics` that I feel more strongly should be shared

item.lua
chdoc: Sure, this was for me because I wanted to understand what was happening outside visible range.

item.lua
```diff
@@ -0,0 +1,294 @@
...
+            changed = changed + 1
+        elseif action == 'undump' and item.flags.dump then
+            item.flags.dump = false
+            changed = changed + 1
+        elseif action == 'melt' and not item.flags.melt then
+            item.flags.melt = true
+            changed = changed + 1
+        elseif action == 'unmelt' and item.flags.melt then
+            item.flags.melt = false
+            changed = changed + 1
```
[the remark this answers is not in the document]
chdoc: I wasn't aware that there is double bookkeeping required for melting. In this case, maybe the melting logic should be encapsulated somewhere, potentially somewhere more basic than `logistics`.

docs/item.rst
```diff
@@ -0,0 +1,58 @@
...
+    :tags: fort productivity items
+
+Filter items in you fort by various properties (e.g., item type, material,
+wear-level, quality, ...), and perform the bulk operations forbid, dump, melt,
+and their inverses. By default, the tool does not act on artifacts.
+
+Usage
+-----
+
+``item [ help | count | [un]forbid | [un]dump | [un]melt ] <filter option>``
```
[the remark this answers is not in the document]
chdoc: Yes, this is basically why I stayed away from this for the moment. Really want this though.

item.lua
```diff
@@ -0,0 +1,294 @@
...
+        local wgroup = dfhack.maps.getWalkableGroup(unit.pos)
+        cgroups[wgroup] = true
+    end
+    cgroups[0] = false -- exclude unwarkable tiles
+    return cgroups
+end
+
+-- see: https://discord.com/channels/793331351645323264/807443277132595211/1173979892036141137
+--- @param item item
+local function isMeltable(item)
```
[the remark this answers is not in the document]
chdoc: I wasn't aware that this was in `logistics`. Just like the act of staging something for melting, this should probably be someplace more basic. Also, I was expecting the reachability tests to belong elsewhere as well.

item.lua
```diff
@@ -0,0 +1,294 @@
...
+            goto skipitem
+        end
+
+        -- check conditions provided via options
+        for _, condition in pairs(conditons) do
+            if not condition(item) then goto skipitem end
+        end
+
+        -- only try to melt things that are actually meltable
+        -- TOTHINK: treated as implicit filter, should we count the item?
```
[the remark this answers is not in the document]
chdoc: I'm thinking that maybe we should add more action-implied filters (e.g. don't forbid already forbidden items) and perform this check before the potentially more expensive user-provided filters. This would make those actions provide lower counts than running `item count` with the same filters. Maybe a `--dry-run` option is in order.

item.lua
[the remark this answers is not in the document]
chdoc: @myk002 I took a second look at this, and I don't even fully understand what is required here. I'm inclined to just hold back on melting until there is an API for melting that can be called safely. Maybe this logic can be moved to (what is exported as) `dfhack.items`? There is already some trade logic there.

item.lua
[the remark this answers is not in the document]
chdoc: @myk002 Upon inspection of the code in `logistics`, the conditions used here (which are derived from the discussion on discord) do not match the conditions in (`logistics`). I am not checking for artifacts or container conditions while the code in `logistics` does not check for metal corpse pieces.

item.lua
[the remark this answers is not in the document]
chdoc: I implemented `--dry-run.`

docs/item.rst
```diff
@@ -0,0 +1,58 @@
...
+    Filter by item type (e.g., BOULDER, CORPSE, ...). Also accepts lower case
+    spelling (e.g. "corpse")
+
+``-m, --material <string>``
+    Filter by material the item is made out of (e.g., "iron").
+
+``-d, --description <string>``
+    Filter by item description (singular form without stack sizes). Example:
+    "cave spider silk web". Note: This does not work for animal products such as
+    wool, because their description always includes the stack size.
```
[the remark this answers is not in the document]
chdoc: I reformulated this to make it more clear.

item.lua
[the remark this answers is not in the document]
myk002: it makes sense to expose that as a library function call; both `canMelt()` and `setMelt(bool)`

item.lua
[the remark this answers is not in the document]
myk002: That's the value of reconciling two independent implementations ; )
metal corpse pieces can be melted? I didn't know that.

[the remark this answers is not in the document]
chdoc: Supposedly, they cannot. So `canMelt()` should rule them out. Maybe I misread your check, and it makes a positive check for item categories that can be melted?

item.lua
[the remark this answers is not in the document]
chdoc: @myk002 Would you be willing to add those two to the Steam `testing` release. Then I will rework the PR to make use of them.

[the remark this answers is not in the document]
myk002: I will, but after -r4 gets tagged later this week. We need to make a quick release to fix some current impactful bugs.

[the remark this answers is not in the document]
chdoc: @myk002 Can we make progress on this?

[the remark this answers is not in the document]
chdoc: @myk002 Just to avoid any potential duplication of work: I have started to work on the C++ side of this. I should be able to create a PR before Wednesday.

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+    Only include items inside the given burrow.
+
+``-o, --outside <burrow>``
+    Only include items outside the given burrow.
+
+``-r, --reachable``
+    Only include items reachable by one of your citizens.
+
+``-u, --unreachable``
+    Only include items not reachable by one of your citizens.
```
myk002, line 43:
```suggestion
    Only include items not reachable by any of your citizens.
```

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+Usage
+-----
+
+``item [ help | count | [un]forbid | [un]dump | [un]melt ] <filter option>``
+
+Action names should be self explanatory. All actions other than ``help`` and
+``count`` have implicit filters associated with them. For instance, ``item
+forbid --unreachable`` will neither report nor test reachability of items that
+are already forbidden.
+
```
myk002, line 22: needs "Examples" section here, before "Options". it would be a good idea to include a `--description` example for dealing with wool so the warning below has something to point to for how to do it "right"

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+Options
+-------
+
+``-h, --help``
+    Print help.
+
```
myk002, line 28: while it's usually a good idea to implement --help, help options are not usually documented since the canonical way to get tool help is `help toolname`

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+``-a, --include-artifacts``
+    Include artifacts in the item list. Regardless of this setting, artifacts
+    are never dumped or melted
+
+``--min-wear <integer>``
+    Only include items whose wear/damage level is at least ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``--min-wear <integer>``
```
myk002, line 67:
```suggestion
``--max-wear <integer>``
```

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+    Filter by item description (singular form without stack sizes). Example:
+    "cave spider silk web". Note: Due to a bug, this is of limited use for some
+    animal products such as wool, because their description always includes the
+    stack size.
+
+``-a, --include-artifacts``
+    Include artifacts in the item list. Regardless of this setting, artifacts
+    are never dumped or melted
+
+``--min-wear <integer>``
```
myk002, line 63: wear limits are actually frequently used and should probably be given short options. maybe `-w` and `-W`?

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+    Only include items whose wear/damage level is at most ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``--min-quality <integer>``
+    Only include items whose quality level is at least ``integer``. Useful
+    values are 0 (standard) to 5 (masterwork).
+
+``--min-quality <integer>``
```
myk002, line 75:
```suggestion
``--max-quality <integer>``
```

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+
+``--min-wear <integer>``
+    Only include items whose wear/damage level is at least ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``--min-wear <integer>``
+    Only include items whose wear/damage level is at most ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``--min-quality <integer>``
```
myk002, line 71: quality options are frequently used as well and could use short forms, say `-q` and `-Q`

item.lua
```diff
@@ -0,0 +1,300 @@
...
+end
+
+-- check whether the item is reachable by one of the citizens
+local function isReachable(item)
+    local reachable = false
+    local x, y, z = dfhack.items.getPosition(item)
+    if x then -- item has a valid position
+        local citizens = dfhack.units.getCitizens(true)
+        for _, unit in pairs(citizens) do
+            if dfhack.maps.canWalkBetween(unit.pos, xyz2pos(x, y, z)) then
```
myk002, line 22: use `dfhack.units.getPosition(unit)` instead of `unit.pos` so units in odd circumstances (e.g. caged) are handled properly

item.lua
```diff
@@ -0,0 +1,300 @@
...
+--- @return boolean
+local function fastReachable(item,wgroups)
+    local x, y, z = dfhack.items.getPosition(item)
+    if x then -- item has a valid position
+        local igroup = dfhack.maps.getWalkableGroup(xyz2pos(x, y, z))
+        if wgroups[igroup] then
+            return true
+        else
+            return false
+        end
```
myk002, line 44: if returning something truthy is enough, consider `return wgroups[igroup]`. If it needs to be a boolean, consider `return not not wgroups[igroup]`

item.lua
```diff
@@ -0,0 +1,300 @@
...
+end
+
+--- @return table<integer,boolean>
+local function citizenWalkabilityGroups()
+    local cgroups = {}
+    for _, unit in pairs(dfhack.units.getCitizens(true)) do
+        local wgroup = dfhack.maps.getWalkableGroup(unit.pos)
+        cgroups[wgroup] = true
+    end
+    cgroups[0] = false -- exclude unwarkable tiles
```
myk002, line 57: note that this is already falsey since `cgroups[0]` is `nil` even before you set it to `false`

item.lua
```diff
@@ -0,0 +1,300 @@
...
+        df.item_type[type] ~= 'PET' and
+        df.item_type[type] ~= 'FOOD' and
+        df.item_type[type] ~= 'EGG'
+end
+
+local function addCondition(tab, cond)
+    if cond then
+        table.insert(tab, cond)
+    end
+end
```
myk002, line 84: this function is equivalent to `table.insert(tab, cond)`. inserting `nil` is a noop. I suppose this function could still be useful if `cond` can be the boolean value `false`, though.

docs/item.rst
```diff
@@ -0,0 +1,77 @@
...
+    Only include items whose quality level is at least ``integer``. Useful
+    values are 0 (standard) to 5 (masterwork).
+
+``--min-quality <integer>``
+    Only include items whose quality level is at most ``integer``. Useful
+    values are 0 (standard) to 5 (masterwork).
```
myk002, line 77: since you have a public API, it should be documented in an `API` section here

item.lua
```diff
@@ -0,0 +1,300 @@
...
+-----------------------------------------------------------------------
+-- external API: helpers to assemble filters and `item.act` to execute.
+-----------------------------------------------------------------------
+item = {}
```
myk002, line 88: public APIs are normally expressed via global functions. They will already be namespaced from the local variable that holds the `reqscript` return value. With your API held inside the `item` table, clients will call your code like this:

```lua
local itemtools = reqscript('item`)

local conditions = {}
itemtools.item.condition_type(conditions, 'amulet')
itemtools.item.act('melt', conditions, {})
```

This may be cleaner if there were two top level public functions:

```lua
function add_condition(conditions, type, match)
    ...
end

function apply(action, conditions, options)
    options = options or {}
    ...
end

**Comment by myk002** [comment]
more ideas for useful filters here: http://www.bay12forums.com/smf/index.php?topic=164123.msg8521571#msg8521571

File: item.lua (hunk 16 above)
Comment (chdoc) [reply item.lua:57]: Are you sure? I thought that a citizen in mid air or stuck in a fortification would be in walkability group 0.

File: item.lua (hunk 19 above)
Comment (chdoc) [reply item.lua:88]: Indeed, looking back at this I'm not happy with the API. Part of this is due to me being upset that, while functional, Lua doesn't handle partial applications very well. However, that was before I saw `dfhack.curry`. With that, factoring things as
```lua
local function condition_material (material, item) return ...    -- definition
table.insert(conditions, curry(condition_material,material))     -- argument parsing / `add_condition`
```
looks more reasonable.

What do you imagine for the `type` argument of the global `add_condition` function. Strings or the fields of something like a "pseudo enum":
```
Condition = { material = {}, description = {}, ... }
```

chdoc: I ultimately decided not to bunch all the filter-generating functions into one. The different functions take, in addition to the obligatory table, 0-2 arguments of different types. So I don't think this would make a lot of sense.

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Filter items in you fort by various properties (e.g., item type, material,
+wear-level, quality, ...), and perform the bulk operations forbid, dump, melt,
+and their inverses. By default, the tool does not consider artifacts. Outputs
+the number of items that matched the filters and were modified.
+
+Usage
+-----
+
+``item [ help | count | [un]forbid | [un]dump | [un]melt ] <filter option>``
```
myk002, line 16:
```suggestion
``item [ count | [un]forbid | [un]dump | [un]melt ] <filter options>``
```

myk002: supporting `help` as a command is fine, but the canonical way to get help is `help item`, so we don't want to promote alternatives.

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Examples
+--------
+
+``item forbid --unreachable``
+    Forbid all items that cannot be reached by any of your citizens.
+
+``item unforbid --inside Cavern1 --type wood``
+    Unforbid/reclaim all logs inside the burrow named "Cavern1" (Hint: use 3D
+    flood-fill to create a burrow covering an entire cavern layer)
```
myk002, line 31:
```suggestion
    flood-fill to create a burrow covering an entire cavern layer).
```

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+
+``item melt -t weapon -m steel --max-quality 3``
```
myk002, line 33: I suggest supporting words for quality levels (in addition to numbers) and using the words in the example here. People aren't going to remember the numeric mappings.

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+    Designate all steel weapons whose quality is less than "exceptional" for
```
myk002, line 34: if the parameter is named "max', the wording here should be "at most" not "is less than"

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+    melting.
+
+Options
+-------
+
+``-n, --dry-run``
+    Use get an accurate count of the items that would be affected, including the
```
myk002, line 41:
```suggestion
    Get an accurate count of the items that would be affected, including the
```

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+    implicit filters of the selected main action.
```
myk002, line 42: the implicit filters need documenting : )

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+    Filter by item type (e.g., BOULDER, CORPSE, ...). Also accepts lower case
+    spelling (e.g. "corpse"). Use ``:lua @df.item_type`` to get the list of all
+    item types.
+
+``-m, --material <string>``
+    Filter by material the item is made out of (e.g., "iron").
+
+``-c, --mat-category <string>``
+    Filter by material category of the material item is made out of (e.g.,
+    "metal"). Use ``:lua df.dfhack_material_category`` to get have a list of all
```
myk002, line 66:
```suggestion
    "metal"). Use ``:lua df.dfhack_material_category`` to get a list of all
```

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+    material categories.
+
+``-d, --description <string>``
+    Filter by item description (singular form without stack sizes). Example:
+    "cave spider silk web". Note: Due to a bug, this is of limited use for some
+    animal products such as wool, because their description always includes the
+    stack size.
```
myk002, line 73: maybe give an example of what the wool description is and how to deal with it

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+
+``-a, --include-artifacts``
+    Include artifacts in the item list. Regardless of this setting, artifacts
+    are never dumped or melted
```
myk002, line 77:
```suggestion
    are never dumped or melted.
```

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+    Only include items whose wear/damage level is at least ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``-W, --max-wear <integer>``
+    Only include items whose wear/damage level is at most ``integer``. Useful
+    values are 0 (pristine) to 3 (XX).
+
+``-q, --min-quality <integer>``
+    Only include items whose quality level is at least ``integer``. Useful
+    values are 0 (standard) to 5 (masterwork).
```
myk002, line 89:
```suggestion
    values are 0 (ordinary) to 5 (masterwork).
```

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+
+``-Q, --max-quality <integer>``
+    Only include items whose quality level is at most ``integer``. Useful
+    values are 0 (standard) to 5 (masterwork).
```
myk002, line 93:
```suggestion
    values are 0 (ordinary) to 5 (masterwork).
```

docs/item.rst
myk002: The script API still needs documenting; see https://docs.dfhack.org/en/stable/docs/tools/quickfort.html#api for an example

myk002 (commented):
doc review done. I'll look at the code once we resolve some things there

docs/item.rst
chdoc: There is already a paragraph right under "Usage". Are you saying that's not enough?

docs/item.rst
chdoc: I'm not even sure myself which words I should be accepting here. There isn't even perfect agreement between the wiki and DFHack: "Superior" vs "Superior quality". Some of the levels would have dashes (finely-crafted). Also, is it "masterful" or "masterwork". I think numbers are easier and easier to get right.

docs/item.rst
chdoc: I don't think there's a good way to deal with "wool", at least not without changing the meaning of the description filter. The description provided by DF always involves the stack size. I guess one could work around this issue by removing any potential stack size at the end of the string provided by DF or matching the argument against parts of the description provided by DF.

chdoc:
I don't know why, but your comment about the API still needing to be documented does not offer me the option to reply/resolve. I have taken note of this, and I will document the API once we are in agreement what it should look like.

myk002:
> I don't know why, but your comment about the API still needing to be documented does not offer me the option to reply/resolve. I have taken note of this, and I will document the API once we are in agreement what it should look like.

I was actually hoping the documentation would facilitate the conversation. It's hard for me to understand your intent without seeing the docs about how you see the API being used.

item.lua
myk002: sorry, yes, you're absolutely right

docs/item.rst
myk002: I'm not sure I'd call not reporting items that are already forbidden an "implicit filter". If `item forbid --unreachable` included the count of items that were already forbidden, I'd just call that a bug : P

What about items that are owned, in a job, rotten, on fire, part of a building, etc.? Many other tools implicitly filter these out. I thought this was more what you meant.

item.lua
```diff
@@ -0,0 +1,340 @@
...
+        table.insert(
+            tab,
+            function (item) return df.item_type[item:getType()] == string.upper(match) end
+        )
+    elseif type(match) == "number" then
+        table.insert(
+            tab,
+            function (item) return item:getType() == type end
+        )
+    else qerror("match argument must be string or number")
```
myk002, line 81: `qerror` makes sense for when this is called from the commandline processing code, but if another tool is using the API, this should be `error` so problems will generate a stack trace

item.lua
```diff
@@ -0,0 +1,340 @@
...
+        tab,
+        function(item) return dfhack.items.getDescription(item, 1) == desc end)
+end
+
+--- @param tab conditions
+--- @param material string
+function condition_material(tab, material)
+    table.insert(
+        tab,
+        function(item) return dfhack.matinfo.decode(item):toString() == material end)
```
myk002, line 112: can this be a prefix match (at colon boundaries)? Otherwise people will have to specify ":NONE" a lot

item.lua
```diff
@@ -0,0 +1,340 @@
...
+end
+
+--- @param tab conditions
+--- @param match string
+function condition_matcat(tab, match)
+    if table.contains(df.dfhack_material_category,match)
```
myk002, line 118: why table.contains() instead of `if df.dfhack_material_category[match]`?

docs/item.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Filter items in you fort by various properties (e.g., item type, material,
+wear-level, quality, ...), and perform the bulk operations forbid, dump, melt,
+and their inverses. By default, the tool does not consider artifacts. Outputs
+the number of items that matched the filters and were modified.
+
+Usage
+-----
+
+``item [ count | [un]forbid | [un]dump | [un]melt ] <filter options>``
```
myk002, line 16: there have been requests on Reddit for a tool to hide all boulders not in stockpiles. that could be worked into this tool pretty easily, I think. new subcommands [un]hide, new condition: inside/outside stockpile

item.lua
myk002: the new refactored API looks good to me

item.lua
chdoc: Because `df.dfhack_material_category` maps numbers *to* material categories, and I didn't realize that the table is bidirectional. Given that 0, the value for "plant", is falsey, I guess the correct way is `if df.dfhack_material_category[match] ~= nil`

docs/item.rst
chdoc: Indeed, adding `[un]hide` should be easy. That's just a flag, isn't it? 

Regarding stockpiles: The use-case you outlined requires a (fast) test whether an item is in *any* stockpile. I'm not sure whats the best way to go about this.

item.lua
chdoc: I don't understand, the string being matched here is just the material (i.e., "iron" or "steel").

docs/item.rst
chdoc: I ended up resolving this by (a) removing any trailing stack size that gets reported by DF and (b) using pattern/regex matching rather than string equality for the ``--description`` filter.

docs/item.rst
myk002: I think a scan of all stockpiles and a hashmap might be required. we have `dfhack.buildings.getStockpileContents(stockpile)` to get the list of stockpile contents, so iterating over all stockpiles and calling that would do it. It shouldn't be much worse than the existing iteration over all items.

myk002: also have to iterate over the contents of containers returned by `getStockpileContents`

item.lua
myk002: sorry, I got that confused with `:getToken()` -- you're right, `toString()` returns a much friendlier name

item.lua
myk002: the number `0` is truthy in Lua. the only falsey values are `nil` and `false`

docs/item.rst
chdoc: That, and iterate over the contents of containers (i.e. bags) contained in containers. Seems to be working reasonably fast now, even without jumping through extra hoops.

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+
+``item hide -t boulder --scattered``
+    Hide all scattered boulders, i.e. those that are not in stockpiles.
+
+Options
+-------
+
+``-n, --dry-run``
+    Get an accurate count of the items that would be affected, including the
+    implicit filters of the selected main action.
```
myk002, line 44:
```suggestion
    Get a count of the items that would be modified by an operation, which will be the
    number returned by the ``count`` action minus the number of items with the desired
    property already set.
```

docs/item.rst
myk002: after reading the actual code, I see what you meant by "implicit filter" I suggested some rewording to make it (I think) clearer

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+    Filter by item type (e.g., BOULDER, CORPSE, ...). Also accepts lower case
+    spelling (e.g. "corpse"). Use ``:lua @df.item_type`` to get the list of all
+    item types.
+
+``-m, --material <string>``
+    Filter by material the item is made out of (e.g., "iron").
+
+``-c, --mat-category <string>``
+    Filter by material category of the material item is made out of (e.g.,
+    "metal"). Use ``:lua df.dfhack_material_category`` to get a list of all
```
myk002, line 72:
```suggestion
    "metal"). Use ``:lua @df.dfhack_material_category`` to get a list of all
```

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+    material categories.
+
+``-d, --description <pattern>``
+    Filter by item description (singular form without stack sizes). The
+    ``pattern`` is a Lua pattern
+    (cf. https://www.lua.org/manual/5.4/manual.html#6.4.1), so "cave spider
```
myk002, line 78: DFHack uses Lua 5.3. I don't think it matters for patterns, but it would still be good to point people to the docs that match the version we're using

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+API
+---
+
+The item script can be called programmatically by other scripts, either via the
+commandline interface with ``dfhack.run_script()`` or via the API functions
+defined in :source-scripts:`item.lua`, available from the return value of
+``reqscript('item')``:
+
+* ``item.execute(action, conditions, options)``
+* ``item.executeWithPrinting(action, conditions, options)``
```
myk002, line 130:
```suggestion
``reqscript('item')``::

    local itemtools = reqscript('item')
    itemtools.execute(action, conditions, options)
    itemtools.executeWithPrinting(action, conditions, options)
```

myk002: I'm reluctant to call the variable that holds the `reqscript` return value "item" in the example since that is used in so many scripts as a local reference to an actual item.

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+
+Performs ``action`` (``forbid``, ``melt``, etc.) on all items satisfying
+``conditions`` (a table containing functions from item to boolean). ``options``
+is a table containing the boolean flags ``artifact``, ``dryrun``, and
+``bytype``, which correspond to the (filter) options described above.
+
+The function ``item.execute`` performs no output, while the ``WithPrinting``
```
myk002, line 137:
```suggestion
The function ``execute`` performs no output, while the ``WithPrinting``
```

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+variant performs the same output as the ``item`` tool.
+
+The API provides a number of helper functions to aid in the construction of the
+filter table. The first argument ``tab`` is always the table to which the filter
+should be added.
+
+* ``item.condition_burrow(tab,burrow, outside)``
```
myk002, line 144:
```suggestion
* ``item.condition_burrow(tab, burrow, outside)``
```

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+    Selects stockpiled items, or scattered items when ``invert=true``.
+
+* ``item.condition_forbidden(tab)``
+    Checks for ``item.flags.forbid``
+
+* ``item.condition_melt(tab)``
+    Checks for ``item.flags.melt``
+
+* ``item.condition_dump(tab)``
+    Checks for ``item.flags.dump``
```
myk002, line 183: shouldn't these take a boolean so the inverse can be matched?

docs/item.rst
```diff
@@ -0,0 +1,200 @@
...
+ API usage example::
+
+   local item = reqscript('item')
+   local cond = {}
+
+   item.condition_type(cond, "BOULDER")
+   item.execute('unhide', cond, {}) -- reveal all boulders
+
+   item.condition_stockpiled(cond, true)
+   item.execute('hide', cond, {})   -- hide all boulders not in stockpiles
```
myk002, line 200:
```suggestion
   local itemtools = reqscript('item')
   local cond = {}

   itemtools.condition_type(cond, "BOULDER")
   itemtools.execute('unhide', cond, {}) -- reveal all boulders

   itemtools.condition_stockpiled(cond, true)
   itemtools.execute('hide', cond, {})   -- hide all boulders not in stockpiles
```

docs/item.rst
myk002: also, there seem to be a mix of `condition.foo()` + `condition.notfoo()` and `condition.foo(bool)`. I'd prefer to standardize on the latter, taking a boolean to indicate whether the relevant property should be true or false. that is foo(true) should match foo being true. I'd prefer to avoid invert=true semantics, like what is currently on the `stockpiled` condition.

docs/item.rst
myk002: or prefix with `itemtools`, that would be ok too

docs/item.rst
chdoc: I agree, the API for condition creation still looks somewhat haphazard and that `condition.foo(...,bool)` is the way to go. However, I would still advocate for `negate=true` semantics. This way, the argument can be naturally elided when the condition is to be matched positively, which is probably the majority of cases. Otherwise, this would require `true` and `nil` to behave the same and `false` to stand for negation.

item.lua
```diff
@@ -0,0 +1,384 @@
...
+--- @param options { help : boolean, artifact : boolean, dryrun : boolean, by_type : boolean }
+--- @return number, table<number,number>
+function execute(action, conditions, options)
+    local count = 0
+    local types = {}
+
+    for _, item in pairs(df.global.world.items.other.IN_PLAY) do
+        -- never act on items used for constructions/building materials and carried by hostiles
+        -- also skip artifacts, unless explicitly told to include them
+        if item.flags.construction or
```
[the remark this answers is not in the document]
chdoc: After looking at the comments in df-structures I would hope/assume that `IN_PLAY` implies `garbage_collect=false`, no? Likewise, `removed` items should already be filtered by not having a valid position on a visible tile. 

I agree with excluding `trader` and `on_fire`. 

I certainly want to include `in_job` `spider_web` items. One of my main use cases is to forbid things in caverns, *in particular* when someone is already on the way to collect it. 

Not sure about `owned (that should probably be optional), and `in_chest` (don't know enough about its weird semantics do decide whether it matters or not)

myk002 (commented):
This is looking fantastic! I'm really excited to be merging this!

item.lua
```diff
@@ -0,0 +1,416 @@
...
+            (item.flags.artifact and not options.artifact) or
+	    item.flags.on_fire or
+	    item.flags.trader or
+	    (item.flags.owned and not options.owned)
+        then
+            goto skipitem
+        end
+
+        -- implicit filters:
+        if action == 'melt' and (item.flags.melt or not dfhack.items.canMelt(item)) or
```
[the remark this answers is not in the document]
chdoc: I'm fairly certain it does, and disagree with redundant parentheses increasing readability/clarity.

item.lua
```diff
@@ -0,0 +1,416 @@
...
+            action == 'dump' and (item.flags.dump or item.flags.artifact) or
+            action == 'undump' and not item.flags.dump or
+            action == 'hide' and item.flags.hidden or
+            action == 'unhide' and not item.flags.hidden
+        then
+            goto skipitem
+        end
+
+        -- check conditions provided via options
+        for _, condition in pairs(conditions) do
```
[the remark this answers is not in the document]
chdoc: This is actually a conscious decision: when using the API, `conditions` is provided by client code, and it is perfectly legal to craft this table yourself. Sure, the `@param` requests a sequence, but `pairs` is more robust in this context. I don't foresee anyone calling this with some DFHack bidirectional table for which `pairs` doesn't work as one might expect.

item.lua
```diff
@@ -0,0 +1,416 @@
...
+--- @param options { help : boolean, artifact : boolean, dryrun : boolean, bytype : boolean, owned : boolean }
+function executeWithPrinting (action, conditions, options)
+    local count, _ , types = execute(action, conditions, options)
+    if action == "count" then
+	print(count, 'items matched the filter options')
+    elseif options.dryrun then
+	print(count, 'items would be modified')
+    else
+	print(count, 'items were modified')
+    end
```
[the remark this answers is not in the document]
chdoc: I'm not sure where you see a tab character up there. Also, only one of those print statements will execute. So there is nothing to align here.

item.lua
```diff
@@ -0,0 +1,416 @@
...
+
+function condition_owned(tab, negate)
+    local pred = function(item) return item.flags.owned end
+    addPositiveOrNegative(tab, pred, negate)
+end
+
+--- @param tab conditions
+--- @param negate { negate : boolean }|nil
+function condition_stockpiled(tab, negate)
+    local stocked = {}
```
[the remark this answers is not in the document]
chdoc: This is slow enough as it is. On my 25K items fort (14K of which are stockpiled), this is the only condition with a somewhat noticeable delay (<1s). Nice to be able to quickly check these numbers. :smile:

item.lua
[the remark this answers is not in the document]
myk002: print('a', 'b') will result in the output "a\tb", which will be converted into a "a b" by `gui/launcher` before display -- so actually I guess that's not bad and would work fine here, so no change necessary.

item.lua
```diff
@@ -0,0 +1,421 @@
...
+    if action == "count" then
+        print(count, 'items matched the filter options')
+    elseif options.dryrun then
+        print(count, 'items would be modified')
+    else
+        print(count, 'items were modified')
+    end
+    if options.bytype and count > 0 then
+        local sorted = {}
+        for tp, ct in pairs(types) do
```
myk002, line 299:
```suggestion
        for tp, ct in ipairs(types) do
```

PR #852, opened 2023-10-06
[combine] reduce aggressiveness, increase feature set

myk002:
Proposed changes:
- no more dynamic stack sizes; experience has shown that they tend to grow to absurd proportions
- limit containers to their advertised volume
- add handling of items inside built containers
- add handling of thread/cloth
- add handling of partial bars in smelters
- only combine non-plantable seeds
- the cheapest hospital thread will be chosen if possible to absorb the deductions, otherwise use the cheapest thread in the fort

discord discussion: https://discord.com/channels/793331351645323264/807444515194798090/1159933907538559016

proposed new hard limits:
```lua
local valid_types_map = {
    all    = { },
    ammo   = {[df.item_type.AMMO]        ={type_id=df.item_type.AMMO,         max_size=25}},
    parts  = {[df.item_type.CORPSEPIECE] ={type_id=df.item_type.CORPSEPIECE,  max_size=1}},
    drink  = {[df.item_type.DRINK]       ={type_id=df.item_type.DRINK,        max_size=math.huge}}, -- max size governed by container capacity
    fat    = {[df.item_type.GLOB]        ={type_id=df.item_type.GLOB,         max_size=5},
                  [df.item_type.CHEESE]      ={type_id=df.item_type.CHEESE,       max_size=5}},
    fish   = {[df.item_type.FISH]        ={type_id=df.item_type.FISH,         max_size=5},
                  [df.item_type.FISH_RAW]    ={type_id=df.item_type.FISH_RAW,     max_size=5},
                  [df.item_type.EGG]         ={type_id=df.item_type.EGG,          max_size=5}},
    food   = {[df.item_type.FOOD]        ={type_id=df.item_type.FOOD,         max_size=20}},
    meat   = {[df.item_type.MEAT]        ={type_id=df.item_type.MEAT,         max_size=5}},
    plant  = {[df.item_type.PLANT]       ={type_id=df.item_type.PLANT,        max_size=5},
                  [df.item_type.PLANT_GROWTH]={type_id=df.item_type.PLANT_GROWTH, max_size=5}},
    powder = {[df.item_type.POWDER_MISC] ={type_id=df.item_type.POWDER_MISC,  max_size=10}},
    seed   = {[df.item_type.SEEDS]       ={type_id=df.item_type.SEEDS,        max_size=5}},
}
```

note that the `seed` category is now explicitly non-plantable seeds. corpse part stack limit of 1 seems non-useful, so that should either be raised or the category should be removed.

PR #897, opened 2023-11-19
realistic limits for combine

silverflyone:
Refer to #852.
Implemented reduced stack sizes.

### DIFF
--- a/combine.lua
+++ b/combine.lua
@@ -10,12 +10,11 @@ local opts, args = {
     dry_run = false,
     types = nil,
     quiet = false,
+    unlimited_stack = false,
     verbose = 0,
   }, {...}
 
 -- default max stack size of 30
-local MAX_ITEM_STACK=30
-local MAX_AMMO_STACK=25
 local MAX_CONT_ITEMS=30
 local MAX_MAT_AMT=30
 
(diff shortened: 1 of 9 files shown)


combine.lua
```diff
@@ -445,20 +444,36 @@ local function stacks_add_items(stockpile, stacks, items, container, ind)
     end
 end
 
+local function unlimited_stacks(types)
+    log(4, 'Unlimited stacks\n')
+
+    if opts.unlimited_stack then
```
myk002, line 450: It might be cleaner/clearer if this `if` statement guarded the call to this function in populate_stacks instead of the body here

combine.lua
```diff
@@ -738,6 +758,7 @@ local function parse_commandline(opts, args)
             {'t', 'types', hasArg=true, handler=function(optarg) opts.types=parse_types_opts(optarg) end},
             {'d', 'dry-run', handler=function() opts.dry_run = true end},
             {'q', 'quiet', handler=function() opts.quiet = true end},
+            {'u','unlimited-stack',handler=function() opts.unlimited_stack = true end},
```
myk002, line 761: Needs documenting in the rst file

myk002 (commented):
Looks good! Could you delete the commented-out lines? I don't think we're going to need them again.

Also, looks like some leftover print statements in the unlimited stacks fn

silverflyone:
pre-commit.ci autofix

silverflyone:
@myk002 I've yet to complete the container volume allocations, but have put in the capture of the container volumes, as well as the item volumes.

Yet to update the update log file.

I've done the other review comments. Let me know if there are other changes needed.

myk002:
I saw that about the container volumes. I figured that would be completed in a later PR. The stack size limits in this PR is already good progress.

docs/combine.rst
```diff
@@ -45,29 +45,32 @@ Options
...
+        ``fish``:  FISH, FISH_RAW and EGG. Max 5.
 
-        ``food``:  FOOD
+        ``food``:  FOOD. Max 20.
 
-        ``meat``:  MEAT
+        ``meat``:  MEAT. Max 5.
 
-        ``parts``: CORPSEPIECE
+        ``parts``: CORPSEPIECE. Max 1.
```
myk002, line 60: Can corpsepiece be removed from this list if the max stack size is 1?

Edit: oh I see, it's still useful for -u.

Would this be more useful in general if it just handled teeth and the default stack size was large?

docs/combine.rst
```diff
@@ -45,29 +45,32 @@ Options
...
 
-        ``plant``: PLANT and PLANT_GROWTH
+        ``plant``: PLANT and PLANT_GROWTH. Max 5.
 
-        ``powders``: POWDERS_MISC
+        ``powders``: POWDERS_MISC. Max 10.
 
-        ``seed``: SEEDS
+        ``seed``: SEEDS. Max 1.
```
myk002, line 66: This can be 5 if we filter for only non-plantable seeds. Also, we really should filter for non-plantable seeds or else `-u` could destroy farming

myk002 (commented):
I'm a bit concerned about the -u flag. As currently written, `combine all -u` will have negative consequences.

I suggest one of the following:
1. Change unlimited to a custom maximum
2. Change seeds to only act on non-plantable seeds and change default size to 5
3. Change body parts to teeth and change default size to 20

OR

Remove -u, corpse parts, and seeds from combine for now and implement the first option later

I'm thinking the second option would be best, given the short timeframe for -r4 (and I really want to get the max stack size changes at least into -r4)

silverflyone:
Okay, I'll pull those out into another later branch.

What is the timing for the next release?

I'll take out corpse and seeds until I can figure out how to do non plantable seeds and teeth.

myk002:
Next release maybe this coming Thursday. Cuz https://github.com/DFHack/dfhack/issues/4040

silverflyone:
Yeah that's close. I'll have a look at it tonight.

silverflyone:
"pre-commit.ci autofix"

silverflyone:
@myk002 I've reverted back to the realistic combine. Not sure why the pre-commit.ci isn't working. I'm heading to bed now. can have another look tomorrow evening.

silverflyone:
pre-commit.ci autofix

combine.lua
```diff
@@ -14,10 +14,19 @@ local opts, args = {
   }, {...}
 
 -- default max stack size of 30
-local MAX_ITEM_STACK=30
-local MAX_AMMO_STACK=25
 local MAX_CONT_ITEMS=30
-local MAX_MAT_AMT=30
+
+-- TODO:
+-- 1. Combine plantable seeds only, currently removed seeds as option.
```
myk002, line 20:
```suggestion
-- 1. Combine non-plantable seeds only, currently removed seeds as option.
```

myk002 (approved):

PR #898, opened 2023-11-19
Updated markdown script to work with Steam version

glotov4:
- Added support of screens: - Unit/Health/Description - Unit/Personality/Traits - Item/Description
- Added name, age & profession of units to description
- Name of the default file now is “markdown_/worldName/_export.md”
- Improved markdown formatting
- Removed code for unsupported screens
- Refactored the code


markdown.lua
```diff
@@ -1,227 +1,191 @@
...
+    local item_raw_description = df.global.game.main_interface.view_sheets.raw_description
+    log:write('### ' .. dfhack.df2utf(item_raw_name) .. '\n\n#### Description: \n' .. reformat(dfhack.df2utf(item_raw_description)) .. '\n')
+    print('Exporting description of the ' .. item_raw_name)
+
+elseif unit then   
+    -- Unit processing
+    -- Simulate UI interactions to load data into memory (click through tabs)
+    local screen = dfhack.gui.getDFViewscreen()
+    -- Click "Personality" 
+    df.global.gps.mouse_x = 145
```
myk002, line 162: The x coordinates will depend on the screen resolution since the view sheets panel moves relative to the right screen edge. Subtract a constant from the width returned from dfhack.screen.getWindowSize()

markdown.lua
```diff
@@ -1,227 +1,191 @@
...
+
+    -- Click "Health/Description"
+    df.global.gps.mouse_x = 142
+    df.global.gps.mouse_y = 15
+    gui.simulateInput(screen, '_MOUSE_L')
+
+    local unit_description_raw = df.global.game.main_interface.view_sheets.unit_health_raw_str[0].value
+    local unit_personality_raw = df.global.game.main_interface.view_sheets.personality_raw_str
+
+    if unit_description_raw or unit_personality_raw then
```
myk002, line 179: Note that zero length strings are truthy in lua, so this will always be true

myk002 (commented):
I'll need to do a more thorough review when I'm not on my phone, but this looks like an excellent start! Great job!

markdown.lua
glotov4: Done, and tested on Linux machine with other resolution

glotov4:
I've acted on all comments and rewritten almost everything. Thank you!

However, this PR was opened from my markdown-fix branch. I'm creating a new pull request from my master to DFHack's master - #899 
It didn't have an initial pull request merged until now, sorry.

PR #899, opened 2023-11-20
Markdown.lua - Steam version update 

glotov4:
**Updated script to work with the Steam version**

Features:
- Added support of screens:
    - Unit/Health/Description
    - Unit/Personality/Traits
    - Item/Description
- Added name, age & profession of units to description
- Improved markdown formatting (headers, line dividers)
- By default, the file name is now “markdown_/Your_World_Name/_export.md”

Docs:
- Moved the helpstr to markdown.rst
- Added tags
- Improved docs
- Added error messages with instructions

Refactoring:
- Refactored the code
- Removed code for processing of unsupported screens (it had time & month processing, as well as logic for other screens - might be useful in the future)
- Argument '-n' is now '-o' for "overwrite" and it can be placed before or after the filename
- Replacing whitespaces from user input with underscores
- Changed success message based on write mode    
- Bug fixes (of the new script):
    - Clicks on tabs, used for data loading, now use relative coordinates to support different screen resolutions
    - Removed redundant check "if unit or item"

### DIFF
--- a/docs/markdown.rst
+++ b/docs/markdown.rst
@@ -2,56 +2,106 @@ markdown
 ========
 
 .. dfhack-tool::
-    :summary: Exports the text you see on the screen for posting online.
-    :tags: unavailable
+    :summary: Save descriptions of selected units/items in markdown format (e.g., for Reddit).
+    :tags: fort | inspection | units
 
-This tool saves a copy of a text screen, formatted in markdown, for posting to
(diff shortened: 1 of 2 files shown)


myk002:
you actually had it right the first time. you should not be starting PRs from your master branch. See https://docs.dfhack.org/en/stable/docs/dev/Contributing.html#pull-request-guidelines

image

PR #902, opened 2023-11-22
script for debugging items being assigned to multiple squads

chdoc:
For what its worth, here is the script I wrote to debug the equipment situation in one of my forts. 

Is would be nice if someone could check that the logic is sound. 

Of course, it would be nice to automatically fix the conflicts, but doing so requires further analysis.


myk002 (commented):
ugh, I forgot to actually submit my review. sorry for the delay.

PR #903, opened 2023-11-26
Ignore all fruit gatherers when checking stranded

azrazalea:
Did some testing on this. It definitely works to ignore citizens gathering fruit. I haven't ran into any issues of false positives so far, but we should still keep an eye on it

### DIFF
--- a/warn-stranded.lua
+++ b/warn-stranded.lua
@@ -333,8 +333,9 @@ local function getStrandedUnits()
                 or getWalkGroup(xyz2pos(unitPos.x+1, unitPos.y+1, unitPos.z))
                 or 0
         end
-
-        if unitIgnored(unit) then
+		
+        -- Ignore units who are gathering plants to avoid errors with stepladders
+        if unitIgnored(unit) or unit.job.current_job.job_type == df.job_type.GatherPlants then
             table.insert(ensure_key(ignoredGroup, walkGroup), unit)
         else
             table.insert(ensure_key(grouped, walkGroup), unit)
--- a/warn-stranded.lua
+++ b/warn-stranded.lua
@@ -269,7 +269,7 @@ function WarningWindow:onZoom()
     local unit = choice.data['unit']
 
     local target = xyz2pos(dfhack.units.getPosition(unit))
-    dfhack.gui.revealInDwarfmodeMap(target, false, true)
+    dfhack.gui.revealInDwarfmodeMap(target, true, true)
 end
 
 function WarningWindow:onToggleGroup()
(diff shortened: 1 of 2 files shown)


warn-stranded.lua
```diff
@@ -333,8 +333,9 @@ local function getStrandedUnits()
                 or getWalkGroup(xyz2pos(unitPos.x+1, unitPos.y+1, unitPos.z))
                 or 0
         end
-
-        if unitIgnored(unit) then
+		
+        -- Ignore units who are gathering plants to avoid errors with stepladders
+        if unitIgnored(unit) or unit.job.current_job.job_type == df.job_type.GatherPlants then
```
ab9rf, line 338: are you guaranteed that `current_job` is non-nil at this point?

azrazalea: Good call. I think in my test fortress everyone was super busy so that didn't get caught. I'll check into that and update it as needed.

azrazalea:
I think ideally we need to pick up on whatever the DF is using to have these dwarves pathfind out of difficult situations by climbing and check _that_ instead but I have no clue what that is

ab9rf:
> I think ideally we need to pick up on whatever the DF is using to have these dwarves pathfind out of difficult situations by climbing and check _that_ instead but I have no clue what that is

the trigger for that is a pathfinding failure, which would require us to invoke the pathfinder, which (a) we cannot do and (b) would cause the very lag we're trying to avoid

ab9rf (approved):

warn-stranded.lua
```diff
@@ -334,7 +334,8 @@ local function getStrandedUnits()
                 or 0
         end
 
-        if unitIgnored(unit) then
+        -- Ignore units who are gathering plants to avoid errors with stepladders
+        if unitIgnored(unit) or (unit.job.current_job and unit.job.current_job.job_type == df.job_type.GatherPlants) then
```
myk002, line 338: I suggest also ignoring mining job types since they can un-strand themselves. There's a handy attr in df.job_types you can check instead of having to itemize all the dig job types

myk002 (commented):
Don't forget the changelog entry!

warn-stranded.lua
myk002: `df.job_type.attrs[unit.job.current_job.job_type].type == 'Digging'`

azrazalea: Gotcha, I'll get that added tonight

azrazalea:
image

myk002 (approved):

PR #905, opened 2023-11-28
Fix creature alcohol in ban-cooking.lua

Droseran:
The creature ID is currently nil due to being an uninitialized variable, causing ban-cooking to error if a creature alcohol has the [EDIBLE_COOKED] token. This replaces the uninitialized variable with the one containing the creature ID.


myk002 (commented):
Thanks! Please add this fix to changelog.txt too

Also, in a separate PR, please add yourself to the DFHack authors list in https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst

Droseran:
Alright, those are done, thanks!

myk002 (approved):

PR #914, opened 2024-01-02
update confirm to use overlays and Lua-based confirmation prompts

myk002:
Fixes https://github.com/DFHack/dfhack/issues/4107
Fixes https://github.com/DFHack/dfhack/issues/3282
Fixes https://github.com/DFHack/dfhack/issues/3106
Fixes https://github.com/DFHack/dfhack/issues/2814

### DIFF
--- a/confirm.lua
+++ b/confirm.lua
@@ -1,0 +1,495 @@
+--@ module = true
+
+local gui = require('gui')
+local json = require('json')
+local overlay = require('plugins.overlay')
+local widgets = require("gui.widgets")
+
+local CONFIG_FILE = 'dfhack-config/confirm.json'
+
+registry = registry or {}
+
+------------------------
+-- Confirmation configs
+
+ConfirmConf = defclass(ConfirmConf)
+ConfirmConf.ATTRS{
+    id=DEFAULT_NIL,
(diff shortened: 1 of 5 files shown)


confirm.lua
```diff
@@ -0,0 +1,535 @@
...
+
+local function has_caravans()
+    for _, caravan in pairs(df.global.plotinfo.caravans) do
+        if caravan.time_remaining > 0 then
+            return true
+        end
+    end
+end
+
+if not next(registry) then
```
lethosor, line 70: I think we should try to avoid checks like this because it makes development harder. Making any changes to the `ConfirmConf` invocations below would require manually clearing `registry` or restarting DF. Previously I think reloading the plugin (or the plugin lua module) was enough to apply changes.

myk002: would it be better to take out the line that checks for duplicate IDs in `ConfirmConf:init()`? That was intended to catch copy paste errors, but if it were taken out, then the conf definitions would get automatically refreshed on every run

myk002: suggested change in https://github.com/DFHack/scripts/pull/915/commits/1bf799631920940b017079e055d7ee6431158e2c

PR #920, opened 2024-01-04
persist liquidSources table

Heneman:
Fixes https://github.com/DFHack/dfhack/issues/4131


source.lua
```diff
@@ -1,10 +1,19 @@
 --@ module = true
 local repeatUtil = require('repeat-util')
 
-liquidSources = liquidSources or {}
+liquidSources = json.decode(persist.GlobalTable[GLOBAL_KEY] or {})
```
myk002, line 4: you'll also need to define hooks to load the state when a map is loaded and re-register with repeat-util

Heneman: Apologies for creating this prematurely, it's been awhile since I've used GitHub and I was just refamiliarizing myself with it! I'll avoid creating `DRAFT/WIP` PRs in the future :sweat_smile: 

I'll be pushing more updates soon

myk002: no worries! it's fine to create a draft PR. It's a good way to get continuous feedback when trying something new. There's a button for that, though -- you don't have to put DRAFT in the title. I'll click it for you.
image

Heneman: I appreciate the help!

So as things currently are, I have the Sources persisting between saves as long as the game isn't closed, but closing and relaunching the game causes the Sources to all be removed again.

Do I need to hook into something else? :thinking: 

Also I'm happy to hear any nitpicks and cleanup recommendations for my changes, this is my first endeavor into DFHack and I'm happy to learn how y'all do things.

source.lua
```diff
@@ -1,10 +1,28 @@
...
+local GLOBAL_KEY = 'source' -- used for state change hooks and persistence
+
+g_state = g_state or {}
 
 local sourceId = 'liquidSources'
 
+enabled = enabled or false
+
+function isEnabled()
+    return enabled
```
myk002, line 15: now that I look at it, you might be better off following a model closer to `prioritize`. this tool will be "enabled" if there are sources in the state map.

Heneman: that looks like it was the golden ticket

Heneman:
persisting but sources will stack instead of overwrite if one is added to a tile that already has one. i imagine we don't want more than once source on a tile so i can add that in, but this _is_ currently functional :tada:

Heneman:
It's working! IT'S WORKING!

source.lua
```diff
@@ -1,38 +1,68 @@
 --@ module = true
 local repeatUtil = require('repeat-util')
+local json = require('json')
+local persist = require('persist-table')
 
-liquidSources = liquidSources or {}
+local GLOBAL_KEY = 'source' -- used for state change hooks and persistence
 
 local sourceId = 'liquidSources'
```
myk002, line 8: you can get rid of `sourceId` and use `GLOBAL_KEY` for this as well. no need to have two separate and different IDs

source.lua
```diff
@@ -1,38 +1,68 @@
...
 
+g_sources_list = g_sources_list or {}
+local function retrieve_state()
+    return g_sources_list
+end
```
myk002, line 13: unless you're planning on writing unit tests, you don't need this extra level of indirection. you can use `g_sourcess_list` directly.

source.lua
```diff
@@ -1,38 +1,68 @@
...
+
+local function has_elements(collection)
+    for _,_ in pairs(collection) do return true end
+    return false
+end
```
myk002, line 18: Since I wrote `prioritize` I found a better way to write this. instead of `has_elements(t)` you can write `next(t)` and take advantage of the built-in `next` keyword

source.lua
```diff
@@ -1,38 +1,68 @@
...
+
+function isEnabled()
+    return has_elements(retrieve_state())
```
myk002, line 21: e.g.
```lua
function isEnabled()
    return next(g_sources_list)
end
```

source.lua
```diff
@@ -1,38 +1,68 @@
...
 end
 
-function AddLiquidSource(pos, liquid, amount)
-    table.insert(liquidSources, {
-        liquid = liquid,
-        amount = amount,
-        pos = copyall(pos),
-    })
+function add_liquid_source(pos, liquid, amount)
+    new_source = {liquid = liquid, amount = amount, pos = copyall(pos)}
```
myk002, line 42: need a `local` here, both for the function and the `new_source` variable

source.lua
```diff
@@ -1,38 +1,68 @@
...
+local function persist_state(liquidSources)
+    persist.GlobalTable[GLOBAL_KEY] = json.encode(liquidSources)
+end
+
 local function formatPos(pos)
     return ('[%d, %d, %d]'):format(pos.x, pos.y, pos.z)
 end
 
-function IsFlowPassable(pos)
+function is_flow_passable(pos)
```
myk002, line 33: should be a `local` function

source.lua
```diff
@@ -1,38 +1,68 @@
...
 
-function AddLiquidSource(pos, liquid, amount)
-    table.insert(liquidSources, {
-        liquid = liquid,
-        amount = amount,
-        pos = copyall(pos),
-    })
+function add_liquid_source(pos, liquid, amount)
+    new_source = {liquid = liquid, amount = amount, pos = copyall(pos)}
+    print(("Adding %d %s to [%d, %d, %d]"):format(amount, liquid, pos.x, pos.y, pos.z))
```
myk002, line 43: since you already have `formatPos`, this could be
```lua
print(('Adding %d %s to %s'):format(amount, liquid, formatPos(pos)))
```

source.lua
```diff
@@ -127,18 +181,52 @@ function main(args)
...
+        return
+    end
+    local persisted_data = json.decode(persist.GlobalTable[GLOBAL_KEY] or '') or {}
+    -- sometimes the keys come back as strings; fix that up
+    for k,v in pairs(persisted_data) do
+        if type(k) == 'string' then
+            persisted_data[tonumber(k)] = v
+            persisted_data[k] = nil
+        end
+    end
```
myk002, line 205: you shouldn't have this problem because your keys are list indices -- prioritize stores a map

source.lua
```diff
@@ -127,18 +181,52 @@ function main(args)
...
+        on_new_job=on_new_job,
+        status=status,
+        boost=boost,
+        boost_and_watch=boost_and_watch,
+        remove_watch=remove_watch,
+        print_current_jobs=print_current_jobs,
+        print_registry=print_registry,
+        parse_commandline=parse_commandline,
+    }
+end
```
myk002, line 222: this is used for unit tests, which you don't have. you can remove this block

source.lua
```diff
@@ -127,18 +181,52 @@ function main(args)
...
+
+if dfhack_flags.module then
+    return
 end
```
myk002, line 226: this if block is redundant with the one below it

source.lua
```diff
@@ -127,18 +181,52 @@ function main(args)
...
 
 if not dfhack_flags.module then
     main({...})
 end
+
+persist_state(g_sources_list)
```
myk002, line 232: this should be inside the if block with `main` since its purpose is to persist any state changes that `main` made

source.lua
```diff
@@ -54,31 +84,55 @@ function AddLiquidSource(pos, liquid, amount)
...
+        end
+        sources[key] = nil
+    end
+    persist_state(sources)
+    load_liquid_source()
+end
+
+function delete_liquid_source(pos)
+    print(("Deleting Source at [%d, %d, %d]"):format(pos.x, pos.y, pos.z))
+    for k, v in pairs(retrieve_state()) do
```
myk002, line 109: since you're using `table.insert` to populate the table, it's a list, not a map. using `ipairs` instead of `pairs` here (and elsewhere in this file) would be semantically clearer.

myk002 (commented):
looks like all the major pieces are in place! great job! just needs some cleanup. Thank you for getting this done! I can do another review pass once this is cleaned up.

Heneman:
pre-commit.ci autofix

Heneman:
o.O 

I thought I'd be triggering a cleanup of the trailing whitespace that the precommit check was showing failed, I didn't intend for a build to be triggered. Apologies for any mess that caused :grimacing:

myk002:
> I didn't intend for a build to be triggered. Apologies for any mess that caused

A build is *supposed* to be triggered after every change -- that's how we detect potential failures before we merge them. Not a mess. Working as intended.

source.lua
```diff
@@ -54,31 +77,54 @@ function AddLiquidSource(pos, liquid, amount)
...
+        local block = dfhack.maps.getTileBlock(v.pos)
+        if block and is_flow_passable(v.pos) then
+            local flags = dfhack.maps.getTileFlags(v.pos)
+            flags.flow_size = 0
+            dfhack.maps.enableBlockUpdates(block, true)
+        end
+        g_sources_list[key] = nil
     end
+    persist_state(g_sources_list)
+    load_liquid_source()
```
myk002, line 96: neither of these lines should be necessary. you'll be persisting state later, and you don't need to register the repeat when deleting a source from the list

source.lua
```diff
@@ -54,31 +77,54 @@ function AddLiquidSource(pos, liquid, amount)
...
+    print("Clearing all Sources")
+    for k, v in ipairs(g_sources_list) do
+        delete_source_at(k)
+    end
+end
+
+local function list_liquid_sources()
     print('Current Liquid Sources:')
-    for _,v in pairs(liquidSources) do
+    for _,v in pairs(g_sources_list) do
```
myk002, line 118:
```suggestion
    for _,v in ipairs(g_sources_list) do
```

source.lua
```diff
@@ -127,18 +173,30 @@ function main(args)
...
         print(('Added %s %d at %s'):format(liquidArg, amountArg, formatPos(targetPos)))
         return
     end
 
-    print(dfhack.script_help())
+    persist_state(g_sources_list)
+end
+
+dfhack.onStateChange[GLOBAL_KEY] = function(sc)
+    if sc ~= SC_MAP_LOADED or df.global.gamemode ~= df.game_mode.DWARF then
```
myk002, line 189: you should also clear your state when a world is unloaded

Heneman:
Should be ready for :eyes: again! The rearrangement of functions and some cleanup means some of the comments aren't resolving, but I _think_ I got everything that was active

source.lua
```diff
@@ -56,29 +55,64 @@ function AddLiquidSource(pos, liquid, amount)
...
+    if v then
+        local block = dfhack.maps.getTileBlock(v.pos)
+        if block and is_flow_passable(v.pos) then
+            local flags = dfhack.maps.getTileFlags(v.pos)
+            flags.flow_size = 0
+            dfhack.maps.enableBlockUpdates(block, true)
+        end
+        g_sources_list[key] = nil
+    end
+    load_liquid_source()
```
myk002, line 70:
```suggestion
```

myk002: no need to register with repeat-util when removing a source

source.lua
```diff
@@ -127,18 +161,35 @@ function main(args)
...
 
-    print(dfhack.script_help())
+dfhack.onStateChange[GLOBAL_KEY] = function(sc)
+    if sc ~= SC_MAP_LOADED or df.global.gamemode ~= df.game_mode.DWARF then
+        return
+    end
+
+    if sc == SC_WORLD_UNLOADED then
+        clear_liquid_sources()
+    end
```
myk002, line 181: this needs to go above the other if statement, otherwise it will never be run

myk002 (commented):
looks good! remember to add a line to the Misc Improvements section (under the "Future" version header) of [changelog.txt](https://github.com/DFHack/scripts/blob/master/changelog.txt#L42C2-L42C2). Something like:

```
- `source`: now saves its state with the fort
```

myk002 (approved):

myk002:
Great job with this! Thank you for working with me through all the comments!

Heneman:
Happy to help! 

Headed back to my failed megaproject with my newfound magic powers:

!(so long)[https://media4.giphy.com/media/gIN6MT17NtapJqFr25/giphy.gif]

PR #929, opened 2024-01-06
workorder-detail-fix: Enforce order item details for buggy job types

flashy-man:
The fix: An 'eventful' handler that checks each job after it's dispatched to a workshop, correcting the details of any job items that have been altered from its parent order's. 

The bug: For certain work orders that have been changed via gui/job-details (or other means), jobs created from the order will revert the changes. This is because a job_item's type doesn't exactly match what is expected for the job. Particularly, the "any" type will pedantically only match other "any" types, so for jobs like PrepareMeal, a "plant" or "liquid" type job_item will trigger a revert when it's dispatched to the workshop. ("plant" =/= "any", technically)

With that fixed, it's easier to order lavish meals of meticulous selection, or silk images sewn specifically on hemp ropes.

One possible issue: Eventful frequency is set to max (every tick) for job initiated event, which will affect every other script using it. 

Before, I had it set up as a 100-tick periodic script, but there was some hassle with active jobs. Eventful's JOB_INITIATED check needs to run frequently enough to ensure that the job is caught before anyone starts gathering items. The items vector can be cleared if caught in time, just not sure if it's wise. I had to go in and manually un-task the cleared job items, and who knows what else could go wrong. Doing it just every 10 ticks might be fine as long as fastdwarf isnt on.

### DIFF
--- a/workorder-detail-fix.lua
+++ b/workorder-detail-fix.lua
@@ -1,0 +1,114 @@
+local script_name = "workorder-detail-fix"
+local eventful = require 'plugins.eventful'
+if not handler_ref then local handler_ref = nil end 
+
+local function get_job_id(match)
+    for val, name in ipairs(df.job_type) do 
+        if name == match then return val end
+    end
+end
+
+-- all jobs with the "any" (-1) type in its default job_items may be a problem
+local offending_jobs = {
+    [get_job_id("EncrustWithGems")] = true,
+    [get_job_id("EncrustWithGlass")] = true,
+    [get_job_id("StudWith")] = true,
+    [get_job_id("PrepareMeal")] = true,
+    [get_job_id("DecorateWith")] = true,
+    [get_job_id("SewImage")] = true,
+    -- list may be incomplete
+}
+
+-- copy order.item fields/flags over to job's job_item
+-- only the essentials: stuff that is editable via gui/job-details
+local function correct_item_details(job_item, order_item)
+    local fields = {'item_type', 'item_subtype', 'mat_type', 'mat_index'}
+    for _, field in pairs(fields) do 
+        job_item[field] = order_item[field] 
+    end
+
+    local flags_names = {'flags1', 'flags2', 'flags3', 'flags4', 'flags5'}
+    for _, flags in pairs(flags_names) do 
+        local order_flags = order_item[flags]
+        if type(order_flags) == "number" then 
+            job_item[flags] = order_flags
+        else -- copy over the flags one by one
+            for o_flag, val in pairs(order_flags) do 
+                job_item[flags][o_flag] = val
+            end
+        end
+    end
+end
+
+-- correct each job as it's initialized 
+-- this is the handler, running after the job is dispatched
+local function enforce_order_details(job)
+    if not job.job_items then return end -- never happens (error here?)
+    local order_id = job.order_id -- only jobs with an ORDER ID 
+    if (order_id == -1) or (order_id == nil) then return end
+
+    -- only jobs with the item type issue. encrusting, sewing, cooking, etc.
+    if not offending_jobs[job.job_type] then return end
+
+    local order = nil -- get the order ref from order id
+    for _, ord in ipairs(df.global.world.manager_orders) do 
+        if ord.id == order_id then order = ord; break end
+    end
+
+    if not order then return end -- oops, no order
+    if not order.items then return end -- no order item details to enforce
+
+    -- copy the item details over when the types don't match
+    for idx, job_item in ipairs(job.job_items) do
+        local order_item = order.items[idx]
+        if not order_item then break end -- never happens (error here?)
+        if job_item.item_type ~= order_item.item_type then
+            -- dfhack's isSuitableItem function will allow the orders we want,
+            -- but disallow insane combinations like meals made of shoes
+            local suitable = dfhack.job.isSuitableItem(
+                job_item, order_item.item_type, order_item.item_subtype )
+            if suitable then 
+                correct_item_details(job_item, order_item)
+            else --[[ error on unsuitable item?]] end
+        end
+    end
+end
+
+local function enable() 
+    print(script_name.." ENABLED")
+    -- set eventful onJobInitiated handler to run every tick (frequency 0)
+    eventful.enableEvent(eventful.eventType.JOB_INITIATED, 0) 
+    eventful.onJobInitiated.workorder_detail_fix = enforce_order_details
+    handler_ref = eventful.onJobInitiated.workorder_detail_fix
+end
+
+local function disable()
+    print(script_name.." DISABLED")
+    eventful.onJobInitiated.workorder_detail_fix = nil
+    handler_ref = nil
+end
+
+local function status()
+    local status = "DISABLED"
+    local handler = eventful.onJobInitiated.workorder_detail_fix
+    if handler ~= nil then
+        -- ensure the handler still matches the one copied back from eventful
+        if handler == handler_ref then 
+            status = "ENABLED"
+        else
+            status = "ERROR: Handler overwritten!"
+            print("why is this here:", handler)
+            print("should be", handler_ref)
+        end
+    end
+    print(script_name.." status: "..status)
+end
+
+args={...}
+
+cmd_table = { ["enable"]=enable, ["disable"]=disable, ["status"]=status }
+
+cmd = cmd_table[args[1]:lower()]
+if cmd then cmd() else
+    print(script_name.." valid cmds: enable, disable, status")
+end
--- a/changelog.txt
+++ b/changelog.txt
@@ -40,6 +40,7 @@ Template for new versions:
 - `warn-stranded`: Update onZoom to use df's centering functionality
 - `ban-cooking`: fix banning creature alcohols resulting in error
 - `confirm`: properly detect clicks on the remove zone button even when the unit selection screen is also open (e.g. the vanilla assign animal to pasture panel)
+- `workorder-detail-fix`: fix item types not being passed properly on some modified work order jobs
 
 ## Misc Improvements
 - `gui/control-panel`: reduce frequency for `warn-stranded` check to once every 2 days
--- a/docs/workorder-detail-fix.rst
+++ b/docs/workorder-detail-fix.rst
@@ -1,0 +1,24 @@
+workorder-detail-fix
+=========
+
+.. dfhack-tool::
+    :summary: Fixes a bug with modified work orders creating incorrect jobs.
+    :tags: fort bugfix workorders
+
+Some work order jobs have a bug when their input item details have been modified.
+
+Example 1: a Stud With Iron order, modified to stud a cabinet, instead creates a job to stud any furniture.
+
+Example 2: a Prepare Meal order, modified to use all plant type ingredients, instead creates a job to use any ingredients.
+
+This fix forces these jobs to properly inherit the item details from their work order.
+
+Usage
+-----
+
+``workorder-detail-fix enable``
+    enables the fix
+``workorder-detail-fix disable``
+    disables the fix
+``workorder-detail-fix status``
+    print fix status


workorder-detail-fix.lua
```diff
@@ -0,0 +1,114 @@
...
+
+local function get_job_id(match)
+    for val, name in ipairs(df.job_type) do 
+        if name == match then return val end
+    end
+end
+
+-- all jobs with the "any" (-1) type in its default job_items may be a problem
+local offending_jobs = {
+    [get_job_id("EncrustWithGems")] = true,
```
TymurGubayev, line 13: you can just do `df.job_type.EncrustWithGems` etc instead.

btw,
```lua
type( df.job_type.EncrustWithGems ) -- 86, number
type( df.job_type[86] ) -- EncrustWithGems, string
```

flashy-man: Much better, it's done. But i accidentally changed offending_jobs to a non-local variable in the commit. Does that matter? I've seen other scripts do it, but I haven't noticed a pattern as to why

TymurGubayev: generally speaking, it's *usually* better to use local variables where possible, as it usually leads to cleaner code (and also helps avoid bugs if various parts of a program use the same global variable).

Also, in Lua accessing a local variable is faster. You'd see stuff like `local print = print` before tight loops for this reason.

As for reasons to use globals in dfhack scripts specifically: the value is then stored between script executions. This can be useful f.e. for the script's configuration.

In your specific case, it doesn't matter too much (but ultimately it'd probably be @myk002 who'll approve this PR, so it's actually up to them). Personally, I would use a local (because the script as a whole won't run often, only the one function from it). Also personally, in case I do decide to make it global, I'd make it like this:

```lua
offending_jobs = offending_jobs or {
   [df.job_type.EncrustWithGems] = true,
   -- ... etc
}
```

This reuses a previously created table `offending_jobs` if there already is one, or creates one if this is the first time the script is run (and the value of `offending_jobs` is `nil` when we read on the right side of that assignment).

TymurGubayev:
Btw, the pre-commit checks are failing because of trailing whitespace, mixed line ending, and something about the end of file.

myk002 (commented):
I haven't looked at the code yet, but here are some higher-order notes:

- script should be in the `fix/` subdir: e.g. `fix/workorder-details`
- If this is useful to run for all players all the time, it should be added as an entry in the [control panel registry](https://github.com/DFHack/scripts/blob/master/internal/control-panel/registry.lua#L51) and enabled by default
- you'll need to implement the [enabled API](https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#enabling-and-disabling-scripts)

myk002:
Ok, I finally got a chance to look at this properly.

So, for my context, is this a problem that only comes up when using `gui/job-details`, or can this happen with purely vanilla UI actions? Did you get `gui/job-details` from https://github.com/DFHack/scripts/pull/799 ?

Is the code in this PR *required* for `gui/job-details` to function correctly? If so, maybe this should be part of that tool and be automatically engaged instead of being a separate "fix' command.

@TymurGubayev what do you think about this?

flashy-man:
Yes the `gui/job-details` from #799 is the only tool I know to *directly* edit the order details like this, and this fix is required for the full expected functionality there. Shouldn't affect vanilla users at all.

However `gui/job-details` might not be the only tool making use of the fix. In theory it should also affect work orders imported via the `orders` plugin. (though my testing says that `orders` currently destroys item details on export/import anyway, so it won't matter unless that's changed)

TymurGubayev:
The `workorder` script behaving similarly to the `orders` plugin is kind of as expected because it's basically just a part of the plugin translated into Lua, plus minor adjustments about exactly how the imported manager orders are handled.

It'd be good to adjust both of them before the `job-details` PR is merged, otherwise, players are in for nasty surprises.

As for the strategy about when should this fix run: since there are 3 different sources of possible problems (`orders` plugin, `workorder`, and `job-details` scripts), we need to either always have it on; or periodically check the manager orders for problematic things and then turn the main fix on (is there maybe a hook we can use, f.e. when new manager order is created or the manager is about to perform their job?); or we adjust the 3 sources to make them enable this fix in some way when necessary (maybe set and store a flag in save data).

flashy-man:
>periodically check the manager orders for problematic things and then turn the main fix on

I like this one. Then it's not running a bunch of handlers when it doesn't need to, and no other scripts need to know about it. In lieu of  some appropriate hook, it could be a periodic script like:

- Load save, no handlers on yet
- Check periodically for a funky order in orders list
- Fix handler is enabled for the rest of the session if so (we don't need to check for funky orders anymore)

And then make that make sense with enabled API

flashy-man:
As outlined, now it checks the order list first before arming the handler. Still to do:

- Do some structured testing, auto-generate lots of orders, run checker every tick, etc, to weed out any missed edge cases or errors.

- Think about checking orders list once on init. It should be trivial (just add it as a recommended/default init somewhere in dfhack config.) It's just a matter of whether you guys think it's warranted. I've been annoyed by this bug for a while, so of course I think it is. 

  - ^ the init function from this could be used by other tools wanting to make use of the fix.

- This could be a full daemon that auto-disarms the handlers when not needed. But with no way (yet) to undo the side-effects to the `eventful` handler frequency, it wouldn't make me feel much better.

<details><summary>Some semi-relevant ramblings i took out of my comments: (click me)</summary>
<p>Only SewImage orders have the item-to-improve in the first slot of their items list.
SewImage also isn't affected by <a href=http://www.bay12games.com/dwarves/mantisbt/view.php?id=10092>Bug#10092</a>
Whereas all other improvement jobs are. Coincidence? Perhaps not.<p>
<p>I looked deep into that particular bug and got stuck looking for the source of "some stack variable" that appeared to come in way before the manager order checker process touched it. Would make a lot of sense if it's related to the index of the improved item.</p>
<p>I do have a half-fix for 10092, however it isn't able to complete periodic orders yet. If there's not an "issue" for it yet I may raise one (along with others that came up in this PR)</p>
</details>

ab9rf:
i'm not thrilled with the idea of a script traversing the job list with each tick, might want to do some load testing to see what the performance impact is

myk002:
> i'm not thrilled with the idea of a script traversing the job list with each tick, might want to do some load testing to see what the performance impact is

at least the iteration is done in cpp and not here. On the other hand, this will be a 5x increase over `prioritize`, which sets the period to 5 ticks for JOB_INITIATED.

I want to think about this a little more to see if there is a way to solve the problems in the source tools instead of as a periodic fixup. Alternately, is there behavior that could be merged into vanilla when creating the jobs? If we can identify legitimately "wrong" behavior, Putnam might make a vanilla fix.

TymurGubayev:
> If we can identify legitimately "wrong" behavior, Putnam might make a vanilla fix.

well, some job types are spawned with adjusted details, while others aren't, so IMHO it is a bug, even though it's not possible in vanilla.

But can't we use `onJobInitiated(job)` instead of checking every job every tick, or is there some technical limitation I'm not aware of?

TymurGubayev:
I think there are 3 possibilities
* this gets recognized as a bug and is fixed in vanilla;
* this fix is run every tick - since it only affects new jobs, maybe it's not that bad?
* this fix is run less frequently, in which case it needs to deal with started item gathering, which gets us back to
> The items vector can be cleared if caught in time, just not sure if it's wise. I had to go in and manually un-task the cleared job items, and who knows what else could go wrong.

I think the manual part here can be automated, and then maybe it'll just work?

Another possibility to solve the cleaning up is to cancel all related jobs, including the main one, and create a new one instead with the fix already applied.

flashy-man:
I saw the word 'hook' near onJobInitiated event in eventful.cpp, so i *hoped* that it traverses a list of events collected via hooks, rather than traversing the whole job list. The only performance check i did was handler CPU time, which i can scarcely measure. (Better performance testing is on my list, also trying to understand eventful.cpp)

Also a period of 5-10 ticks could be just fine for this fix to be ~99% effective. I'm probably thinking too much of the odd dwarf who happens to be near the workshop at just the wrong time. He'd have to pick up the new job, start working on it, and get to the stage where the actual items are designated from `job.job_items`, all within X ticks. Worst case scenario here is that the wrong items are gathered, because someone probably had fastdwarf on, and the script has to have some spaghetti code to fix it.

Re: Editing the affected source tools instead. I've looked hard for ways to beat this bug at the source on order creation. I couldn't find any way to make the item types pass properly from the order, other than the binpatching I was originally doing. There's an instruction in there that really does not want to let any item type pass other than -1 for the 'bugged' jobs.  If it would consider `job_item.vector_id` like `dfhack.job.isSuitableItem` does, and also consider that the "any" type should match *any type*, it would probably be fine.

As far as asking for a vanilla fix: @TymurGubayev  said it, it's basically *almost* a vanilla bug. Maybe if we ask really nicely. We just want to make our nobles happy right?

ab9rf:
> I saw the word 'hook' near onJobInitiated event in eventful.cpp, so i _hoped_ that it traverses a list of events collected via hooks, rather than traversing the whole job list.

unfortunately the only way we have to detect new jobs is... to traverse the job list and look for any jobs that weren't on it the last time we traversed it.  (see https://github.com/DFHack/dfhack/blob/180fd34a96e7ef2bbdb77c2a172ba43ade182464/library/modules/EventManager.cpp#L435C1-L463) so subscribing the "job initialized" hook forces the core event system to do a job list traversal at the frequency requested.

flashy-man:
Got the manager schedule (thanks to whoever found `manager_timer`). 
It was really A) `manager_timer == 0 and tick % 150 == 30`. Timer decrements on the latter condition, starting at 10 until it reaches 0 (then it dispatches the jobs, and resets.) Ends up so it goes 11 times between dispatches. (11 * 150 = 1650)

Now the fix code is different. *Still needs more testing*, but I went through some different saves to make sure i could sync with all the manager schedules. (it even works on 47.05). Here is the flow:

1. Enable: determine next tick of manager job dispatch & schedule the handler for that tick
	```Lua
	next_dispatch = tick + (-(tick - 30) % 150) + manager_timer * 150
	```

2. Handler: runs every 150 ticks. Only on conditions A)***, B) manager exists and C) bugged orders exist, will it check the job list for new jobs, and correct them on the same tick they were dispatched. So order-traversing once every 1650 ticks, and only job-traversing then if we find a bugged order.

*** A) is at the top. Also, the script runs after the manager process, so the condition is slightly different. Not sure about sub-tick timing yet but it seems Lua scripts always run after game code

3. Very small fix for the active job issue. It still happens on that first run.
(i didnt test this at all aside from having done it several times in gm-editor)
	```Lua
	-- experimental: remove the items and un-task them
	while job.items[0] do
	    job.items[0].item.flags.in_job = false
	    job.items:erase(0)
	end```

Still a periodic script technically, but it's better. Since the main concern is doing less to fix it, i'll make an overdue write-up of the bug with some snapshots of the code. Could be something I missed, but I couldn't find any way to bypass that filtering process in the job dispatch.

binpatching was best for my personal use, but with all these OS and steam versions and blah blah, probably not the way to go.

(commits incoming)

myk002:
>    ```lua
>    -- experimental: remove the items and un-task them
>    --while job.items[0] do -- bad idea
>    while #job.items ~= 0 do
>        job.items[0].item.flags.in_job = false
>        job.items:erase(0)
>    end
>    ```

In cpp we have `Job::disconnectJobItem()`, which is more thorough than this code (e.g. it removes specific refs on the item). it's not exposed to Lua yet, but it could be exposed and used by this script, if needed.

fix/workorder-detail-fix.lua
```diff
@@ -0,0 +1,199 @@
...
+        repeatutil.cancel(schedule_key)
+        disable(true)
+    end
+    if #managers == 0 then return end
+    if df.global.plotinfo.manager_timer ~= 10 then return end
+    local orders = get_bugged_orders()
+    if not orders then return end -- no bugs to fix
+    local highest = last_job_id
+    for _, job in utils.listpairs(df.global.world.jobs.list) do
+        if job.id <= last_job_id then goto nextjob end
```
myk002, line 118: repeat jobs reuse their ids, but that should be ok here, right?

fix/workorder-detail-fix.lua
```diff
@@ -0,0 +1,199 @@
...
+        highest = math.max(job.id, highest)
+        if job.order_id == -1 then goto nextjob end
+        local order = orders[job.order_id]
+        if not order then goto nextjob end -- order wasn't bugged
+        -- job in progress: only happens on the first run.
+        -- experimental fix: remove the items and un-task them
+        while #job.items ~= 0 do
+            job.items[0].item.flags.in_job = false
+            job.items:erase(0)
+        end
```
myk002, line 128: needs to remove specific refs on the item; see my previous comment on exposing `Job::disconnectJobItem`

fix/workorder-detail-fix.lua
```diff
@@ -0,0 +1,199 @@
...
+enabled = enabled or false -- enabled API
+function isEnabled() return enabled end
+
+local managers = df.global.plotinfo.main.fortress_entity.assignments_by_type.MANAGE_PRODUCTION
+if not managers then error("NO MANAGERS VECTOR!") end
+local last_job_id = -1
+local jobs_corrected = 0
+
+-- all jobs with the NONE (-1) type in its default job_items may be a problem
+local offending_jobs = {
```
myk002, line 17: use `utils.invert` for simpler syntax here

fix/workorder-detail-fix.lua
```diff
@@ -0,0 +1,199 @@
...
+            for _, _item in ipairs(order.items) do
+                if _item.item_type ~= NONE then goto fix end
+            end
+            goto nextorder
+        end
+
+        -- All other types are improve jobs; only the improved item is checked
+        -- Only SewImage has the item-to-improve at items[0]
+        improve_item_index = (order_job_type ~= SewImage) and 1 or 0
+        item = order.items[improve_item_index]
```
myk002, line 92: missing `local` keywords

fix/workorder-detail-fix.lua
```diff
@@ -0,0 +1,199 @@
...
+    [df.job_type.DecorateWith] = true,
+    [df.job_type.SewImage] = true,
+    -- list may be incomplete
+}
+
+-- copy order.item fields/flags over to job's job_item
+-- only the essentials: stuff that is editable via gui/job-details
+local function correct_item_details(job_item, order_item)
+    local fields = {'item_type', 'item_subtype', 'mat_type', 'mat_index'}
+    for _, field in pairs(fields) do
```
myk002, line 31:
```suggestion
    for _, field in ipairs(fields) do
```

fix/workorder-detail-fix.lua
flashy-man: had a weird time with `goto` jumping into locals' scope. Now they're inside a do/end block so `goto` can skip over it

fix/workorder-detail-fix.lua
flashy-man: right, GUI doesn't allow setting repeat on order jobs anyway

fix/workorder-detail-fix.lua
```diff
@@ -0,0 +1,199 @@
...
+    return orders
+end
+
+-- correct newly dispatched work order jobs
+local disable
+local function on_dispatch_tick()
+    if df.global.cur_year_tick % 150 ~= 30 then
+        print(script_name.." desynced from dispatch tick")
+        repeatutil.cancel(schedule_key)
+        disable(true)
```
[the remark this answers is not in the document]
flashy-man: got it re-syncing now. Surprisingly the fix works with it on. I wouldn't think `timestream` handles work orders well in the first place

Also, the already-gathering-items "fix" is gone for now because I got a couple crashes (!), either relating to the general refs or the `while` loop checking the vector. It's not crucial anyway

flashy-man:
Woohoo review-- Main things: 

0. New name: `fix/workorder-details`

1. Now re-syncing when dispatch tick is lost (only happens using `timestream` that I have seen)

2. No more trying to fix already-gathered items. It was crashing either due to the vector checking `while` loop, or not clearing item refs.

PR #933, opened 2024-01-07
[control-panel] track autostart across multiple sites in the same world

myk002:
until we figure out what a per-site persistence API might look like


myk002:
superseded by https://github.com/DFHack/dfhack/pull/4153 and https://github.com/DFHack/scripts/pull/934

PR #955, opened 2024-01-23
move down `local points` assigment on gui/design

MagistrumT-T:
fixes DFHack/dfhack#4202 by moving the assignment of `points` to after the center dragging logic which was manipulating self.marks.


myk002 (commented):
looks good! There looks to be an errant tab character that ci.pre-commit won't let pass. Please also add an entry in changelog.txt

PR #957, opened 2024-01-23
[gui/launcher] add note about accessing the in-game launcher

myk002:
since the DFHack logo is not shown on that page.


myk002:
oh, I already added this text...

PR #958, opened 2024-01-25
automatic milking/shearing at closest workshop (WIP)

chdoc:
This is a draft for a tool that will queue milking/shearing orders at the closest farmer's workshop on the same z-level. In order for the tool to shear/milk an animal, there must be a farmer's workshop on the same z-level as the animal. Only the workshop closest to the animal will be used. 

Currently, I run this with `repeat --time 1500 --command [ husbandry ]`, but I plan to make it use the `enable` infrastructure with proper callbacks. 

Part of the logic is copied from `workorder`, so I guess some refactoring may be required.

### DIFF
--- a/husbandry.lua
+++ b/husbandry.lua
@@ -1,0 +1,128 @@
+
+-- From workorder.lua
+---------------------------8<-----------------------------
+
+-- local utils = require 'utils'
+
+local world = df.global.world
+local uu = dfhack.units
+
+local function isValidAnimal(u)
+    -- this should also check for the absence of misc trait 55 (as of 50.09), but we don't
+    -- currently have an enum definition for that value yet
+    return uu.isOwnCiv(u)
+        and uu.isAlive(u)
+        and uu.isAdult(u)
+        and uu.isActive(u)
+        and uu.isFortControlled(u)
+        and uu.isTame(u)
+        and not uu.isMarkedForSlaughter(u)
+        and not uu.getMiscTrait(u, df.misc_trait_type.Migrant, false)
+end
+
+-- true/false or nil if no shearable_tissue_layer with length > 0.
+local function canShearCreature(u)
+    local stls = world.raws.creatures
+        .all[u.race]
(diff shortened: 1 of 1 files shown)


myk002:
It's fine to work the logic out here, but I think this should eventually be moved into `workorder` and the choice of general workorder vs. at closest shop should be controllable with a commandline option. We can add both versions to `gui/control-panel` so players have a choice.

husbandry.lua
```diff
@@ -0,0 +1,128 @@
...
+    local closest = nil
+    local distance = nil
+    for _, workshop in pairs(workshops_by_z[unit.pos.z] or {}) do
+        local d = distance2(unit.pos.x, unit.pos.y, workshop.centerx, workshop.centery)
+        if not closest or d < distance then
+            closest = workshop
+            distance = d
+        end
+    end
+
```
myk002, line 116: and if there is no workshop on the animal's level?

chdoc:
> It's fine to work the logic out here, but I think this should eventually be moved into `workorder` and the choice of general workorder vs. at closest shop should be controllable with a commandline option. We can add both versions to `gui/control-panel` so players have a choice.

I'm not so sure about this. The `workorder` tool is primarily about "Creat[ing] manager workorders". Is just happens to have some convenience functionality for dealing with milking/shearing. Besides using the same predicates for checking whether an animal can be milked/sheared -- something that arguably should be moved to a library -- I don't think there's any overlap between the tools. 

Now that milking and shearing is working like I would like it to, the next step for me would be to automate cheese making, because, once again, there is no way to reasonably do this using manager work orders.

husbandry.lua
chdoc: Then no milking job gets queued, because farmer's workshops that are not next to a pasture are probably intended for other things (processing plants or spinning thread).

chdoc:
Has been superseded by the long merged `husbandry` tool.

PR #959, opened 2024-01-26
[makeown] fix field names changed in structures update

myk002:
ref: https://github.com/DFHack/df-structures/pull/684


myk002:
this highlights a code best practice: prefer `{new=true}` to `df.longtypename.T_whatever:new()` so code doesn't break when types are renamed

PR #960, opened 2024-01-26
[item] fix enumeration of (present) item type ids

chdoc:
@myk002 Turns out your one-line change breaks things: `types` is not a sequence, the keys are item type ids matching the filters, hardly continuous...

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,6 +18,7 @@ Template for new versions:
 
 ## Fixes
 - `makeown`: fix error when adopting units that need a historical figure to be created
+- `item`: fix missing item categories when using ``--by-type``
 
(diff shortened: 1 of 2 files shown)


myk002:
I'm terribly sorry - I thought I tested that after changing : /

It's a new version, though, so this counts as a "Fix". Could you add a line to the changelog?

chdoc:
> I'm terribly sorry - I thought I tested that after changing : /

Happens, but imagine my stumped look when something I tried at least 100 times stopped working after the merge. ^^
 
> It's a new version, though, so this counts as a "Fix". Could you add a line to the changelog?

Done.

myk002 (approved):

PR #961, opened 2024-01-27
Add Animal Options To Valid Animal Viewsheet (WIP)

LightHardt:
This is the start of resolving https://github.com/DFHack/dfhack/issues/4043

Wanted to start this request to potentially get some feedback on things.

## Current problems
- Only seems to have correct placement when in full screen mode. When testing in windowed it is moved off screen.
- Still need to add geld and adoption widgets however, I don't see the way I'm currently doing it with HotkeyLabel being able to aesthetically fit all the options.


animal-options.lua
```diff
@@ -0,0 +1,80 @@
...
+-- set available for adoption flag
+local function set_adoption_flag()
+    local unit = dfhack.gui.getSelectedUnit(true)
+    unit.flags3.available_for_adoption = 1
+end
+
+creature_screen=defclass(creature_screen, overlay.OverlayWidget)
+creature_screen.ATTRS {
+    desc = "Add options to tamed animals view sheet",
+    default_pos={x=127,y=9},
```
myk002, line 50: For a view sheet, you will need to anchor to the right side of the screen. This means using a negative `x` value.

animal-options.lua
```diff
@@ -0,0 +1,80 @@
...
+    frame={w=18, h=3},
+}
+function creature_screen:init()
+    self:addviews{
+        widgets.Panel{
+            frame_background=gui.CLEAR_PEN,
+            frame_style=gui.FRAME_MEDIUM,
+            visible=check_valid_unit,
+            subviews={
+                widgets.HotkeyLabel{
```
myk002, line 62: Since this can be toggled, perhaps use a ToggleHotkeyLabel

If you want to get fancy, you can implement a ToggleHotkeyButton and have it display the vanilla icon.

animal-options.lua
```diff
@@ -0,0 +1,80 @@
+--@ module = true
```
myk002, line 1: There is related functionality in `zone.lua` (https://github.com/DFHack/dfhack/blob/develop/plugins/lua/zone.lua). I suggest you add this overlay there instead of creating a new top-level script.

LightHardt: Should i just go ahead and close this pull request and make my changes there?

animal-options.lua
LightHardt: Thank you completed that!

animal-options.lua
LightHardt: Actually can I finish making changes here for review then after I will close?

animal-options.lua
LightHardt: Implemented ToggleHotkeyLabel I do think at some point I will try ToggleHotkeyButton as using the vanilla icons was something i had in my head when i first started.

LightHardt:
Fixed noted previous issues

Before I move to the zone.lua I want to try and find out how to make the initial_option be dependent on the unit's flag. I tried just passing the functions to it but that did not work out. It seems when I toggle one unit's butcher/geld/adopt option all other units are set to On as well. This doesn't actually set the other unit's flag its purely a visual thing.

For a place to put the options i went with this

image

I have never seen a tame animal have unmet needs in this box before however if any recommendations or thoughts please let me know.

myk002:
> how to make the initial_option be dependent on the unit's flag

The overlay is not recreated on every showing. Therefore, the state of the widgets only gets initialized once, and it will be when the overlay is instantiated at program start, not when it is first made visible.

You'll need to dynamically sync the widget state to the selected unit. A good place to do this is a `render` override. Example: https://github.com/DFHack/scripts/blob/4e9d2b1dde964c315a9b318e7cf806cc2ae02c28/trackstop.lua#L130

animal-options.lua
myk002: Sure

animal-options.lua
```diff
@@ -0,0 +1,140 @@
...
+    if unit.flags3.available_for_adoption == 1 then
+        return true
+    else
+        return false
+    end
+end
+
+creature_screen=defclass(creature_screen, overlay.OverlayWidget)
+creature_screen.ATTRS {
+    desc = "Add options to tamed animals view sheet",
```
myk002, line 94: How about also adding an option for assigning a trainer? The logic for that is written (for the `logistics` plugin), but I might have to make it available to lua

animal-options.lua
```diff
@@ -0,0 +1,140 @@
...
+-- The above function already handles checking if valid unit
+-- so just set slaughter flag
+local function set_slaughter_flag()
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    if unit.flags2.slaughter == 1 then
+        unit.flags2.slaughter = 0
+    else
+        unit.flags2.slaughter = 1
+    end
```
myk002, line 31:
```suggestion
local function set_slaughter_flag(val)
    local unit = dfhack.gui.getSelectedUnit(true)
    if not unit then return end
    unit.flags2.slaughter = val
```

myk002: Similar for other callbacks

animal-options.lua
```diff
@@ -0,0 +1,140 @@
...
+                widgets.ToggleHotkeyLabel{
+                    frame={t=0,l=0},
+                    label='Butcher',
+                    key='CUSTOM_CTRL_B',
+                    initial_option=false,
+                    on_change = set_slaughter_flag
+                },
+                widgets.ToggleHotkeyLabel{
+                    frame={t=1,l=0},
+                    label='Geld',
```
myk002, line 116: Needs an `enabled` attribute set to a function that determines whether the current unit is geldable

animal-options.lua
```diff
@@ -0,0 +1,140 @@
...
+                    key='CUSTOM_CTRL_G',
+                    initial_option=false,
+                    on_change = set_gel_flag
+                },
+                widgets.ToggleHotkeyLabel{
+                    frame={t=2,l=0},
+                    label='Adopt',
+                    key='CUSTOM_CTRL_A',
+                    initial_option=false,
```
myk002, line 125: You can get rid of the initial options - they won't matter once the realtime state sync is in place

LightHardt: Removed initial_option

animal-options.lua
LightHardt: Added check to callbacks

animal-options.lua
LightHardt: Added check

animal-options.lua
LightHardt: Interesting after the base capability is flushed out and gets merged in, I'll make it the next thing I work on. Would be nice to just have all the options there since already have 3 of them.

LightHardt:
If anything else i need to update/change before moving to zone.lua just lmk

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+
+local function is_geldable()
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    -- If the unit is not male or already gelded do nothing
+    if unit.sex ~= 1 or unit.flags3.gelded == true then
+        return false
+    else
+        return true
+    end
```
myk002, line 28:
```suggestion
    return dfhack.units.isGeldable(unit)
```

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+end
+
+creature_screen=defclass(creature_screen, overlay.OverlayWidget)
```
myk002, line 31: "screen" is something specific in DF (an element on the viewscreen stack), and this is not it. `AnimalActionsWidget` might be more appropriate (also notice PascalCase class naming conventions)

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+creature_screen.ATTRS {
+    desc = "Add options to tamed animals view sheet",
+    default_pos={x=-42,y=37},
+    default_enabled=false,
```
myk002, line 35: This can be on by default

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+
+    self.subviews.butcher_animal:setOption(self:get_butcher(unit))
+    self.subviews.geld_animal:setOption(self:get_geld(unit))
+    self.subviews.adopt_animal:setOption(self:get_adopt(unit))
+
+    creature_screen.super.render(self, dc)
+end
+
+function creature_screen:init()
+self:addviews{
```
myk002, line 123: needs indent

animal-options.lua
```diff
@@ -0,0 +1,175 @@
+--@ module = true
+
+local overlay = require('plugins.overlay')
+local widgets = require('gui.widgets')
+local gui = require('gui')
+
+local YES = 'Yes'
+local NO = 'No'
+
+-- Make sure an animal unit of your civ is selected as well as not a pet
```
myk002, line 10: why not a pet? pets are still geldable

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+    end
+end
+
+-- Use render to set On/Off dynamically for each unit
+function creature_screen:render(dc)
+    local unit = dfhack.gui.getSelectedUnit(true)
+
+    self.subviews.butcher_animal:setOption(self:get_butcher(unit))
+    self.subviews.geld_animal:setOption(self:get_geld(unit))
+    self.subviews.adopt_animal:setOption(self:get_adopt(unit))
```
myk002, line 117:
```suggestion
    if unit then
        self.subviews.butcher_animal:setOption(dfhack.units.isMarkedForSlaughter(unit))
        self.subviews.geld_animal:setOption(dfhack.units.isMarkedForGelding(unit))
        self.subviews.adopt_animal:setOption(dfhack.units.isAvailableForAdoption(unit))
    end
```

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+        frame_style=gui.FRAME_MEDIUM,
+        visible=check_valid_unit,
+        subviews={
+            widgets.CycleHotkeyLabel{
+                frame={t=0,l=0},
+                label='Butcher',
+                key='CUSTOM_CTRL_B',
+                options={
+                    {label=NO, value=NO, pen=COLOR_WHITE},
+                    {label=YES, value=YES, pen=COLOR_GREEN},
```
myk002, line 135:
```suggestion
                    {label=NO, value=false, pen=COLOR_WHITE},
                    {label=YES, value=true, pen=COLOR_GREEN},
```

myk002: you can use `true` and `false` directly here as the option values and avoid the need to translate in each of the "set" functions.

also, yellow might be more appropriate than green for the "On" setting. Green means more "this is the good/safe setting", which isn't exactly applicable here

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+                view_id='adopt_animal',
+                on_change = function(val) self:set_adoption_flag(val) end
+            },
+        },
+    },
+}
+end
+
+OVERLAY_WIDGETS = {
+    creaturescreen=creature_screen,
```
myk002, line 169: how about `animal_actions` as the widget name?

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+            widgets.ToggleHotkeyLabel{
+                frame={t=2,l=0},
+                label='Adopt',
+                key='CUSTOM_CTRL_A',
+                options={
+                    {label=NO, value=NO, pen=COLOR_WHITE},
+                    {label=YES, value=YES, pen=COLOR_GREEN},
+                },
+                view_id='adopt_animal',
+                on_change = function(val) self:set_adoption_flag(val) end
```
myk002, line 161:
```suggestion
                on_change=self:callback('set_adoption_flag'),
```

myk002: and similar for the other wigets

animal-options.lua
```diff
@@ -0,0 +1,175 @@
...
+                    {label=NO, value=NO, pen=COLOR_WHITE},
+                    {label=YES, value=YES, pen=COLOR_GREEN},
+                },
+                view_id='geld_animal',
+                enabled=is_geldable,
+                on_change = function(val) self:set_geld_flag(val) end
+            },
+            widgets.ToggleHotkeyLabel{
+                frame={t=2,l=0},
+                label='Adopt',
```
myk002, line 154: also needs an `enabled` attribute that responses to `dfhack.units.isAvailableForAdoption(unit)`

myk002 (commented):
I'm adding trainer assignment functions in https://github.com/DFHack/dfhack/pull/4223 so you can support training if you like. Valid animals are:
```lua
dfhack.units.isTamable(unit) and not dfhack.units.isDomesticated(unit)
```
(in addition to your current validity check)
and the property query is `dfhack.units.isMarkedForTraining(unit)`

animal-options.lua
LightHardt: Switched `creature_screen` to suggested `AnimalActionsWidget`

animal-options.lua
LightHardt: turned on by default

animal-options.lua
LightHardt: Made indent

myk002:
When you do make the PR for integrating this into zone.lua, please remember to make a branch first. When your local fork diverges from upstream on the main branches, bad things happen.

animal-options.lua
LightHardt: I started with just butchering and forgot to account for gelding. Added `enabled` fields to butcher and adopt to make sure not a pet.

animal-options.lua
LightHardt: Made colors yellow and switched values with booleans.

animal-options.lua
LightHardt: Made change

animal-options.lua
LightHardt: Made changes

animal-options.lua
LightHardt: When I add that check it disables the option and seems to only reenable if i go into the creatures screen and turn on adoption there

myk002: what is your `enable` logic?

LightHardt:
```
local function is_avail_adoption()
    local unit = dfhack.gui.getSelectedUnit(true)

    return dfhack.units.isAvailableForAdoption(unit)
end
```

myk002: ohohoh, it returns whether the flag is already set, not whether it *can* be set. Hrm.. what's the logic for whether it *can* be set? just your existing validity check and race != cat?

LightHardt: Oh i did not know cats couldn't be adopted (noob) ill add that in and that will end up being the check

myk002: indeed, cats adopt dwarves, not the other way around

PR #962, opened 2024-01-28
Find Units Belonging To Certain Group

LightHardt:
Made this script as I went to try and find what units were in a troupe that wanted me to grant them residency and realized I would have click through them all (quite a bit of visitors) to find who was in it (at least I believe so, newer player maybe there is an easier way).
 
image
image

I think end goal instead of this being a standalone script maybe it could be incorporated into sort? I did take a look sort and it was a little intimidating. Thoughts?


myk002:
This could be a feature of a `gui/units` (or maybe we'll stick with the historical name `manipulator`), where you have a list of units that you can filter and sort and do things with

LightHardt:
> This could be a feature of a `gui/units` (or maybe we'll stick with the historical name `manipulator`), where you have a list of units that you can filter and sort and do things with

Gotcha in this case I may just close this for now as just spitting out names then having to go and search for them anyway to actually view their skills, personalities, etc would still be tedious.

Off the top of your head do you know if there would be a way to open up their character sheets from the script as it finds them? Might be a good interim solution.

myk002:
There is no current way to reliably open up a unit's info sheet programmatically. DF has bespoke code for many different UI transitions that we'd have to replicate. You can *zoom* to the unit's position, but actually clicking on the unit would be done by the player.

it might be technically possible to simulate clicks to bring up the character sheet, but that can get complicated if there are many units/items/buildings on the target tile.

LightHardt:
Thank you for the info!

PR #964, opened 2024-01-31
Fix `gui/design` conflicting center point behavior

MagistrumT-T:
This solves #4234 by moving the center dragging logic up and adding a separate check for no movement movement case for dragging marks.

Now clicking on the center mark a second time without moving will let the function flow and enter mark editing mode.

Fixes: https://github.com/DFHack/dfhack/issues/4234

### DIFF
--- a/gui/design.lua
+++ b/gui/design.lua
@@ -1475,6 +1475,27 @@ function Design:onInput(keys)
             self.placing_mirror = false
             self.needs_update = true
         else
+            -- Clicking center point
+            if #self.marks > 0 then
+                local center = self.shape:get_center()
+                if pos == center and not self.prev_center then
+                    self.start_center = pos
+                    self.prev_center = pos
+                    return true
+                elseif self.prev_center then
+                    --If there was no movement presume user wanted to click the mark underneath instead and let the flow through.
+                    if pos == self.start_center then
+                        self.start_center = nil
+                        self.prev_center = nil
+                    else
+                    -- Since it moved let's just drop the shape here.
+                        self.start_center = nil
+                        self.prev_center = nil
+                        return true
+                    end
+                end
+            end
+            
             if self.shape.basic_shape and #self.marks == self.shape.max_points then
                 -- Clicking a corner of a basic shape
                 local shape_top_left, shape_bot_right = self.shape:get_point_dims()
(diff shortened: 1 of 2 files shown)


gui/design.lua
```diff
@@ -1475,6 +1475,27 @@ function Design:onInput(keys)
             self.placing_mirror = false
             self.needs_update = true
         else
+            -- Clicking center point
+            if #self.marks > 0 then
+                local center = self.shape:get_center()
+                if pos == center and not self.prev_center then
```
myk002, line 1481: `pos == center` is only true if they point to the same table. Is this what you mean, or do you want `same_xyz(pos, center)`?

gui/design.lua
```diff
@@ -1475,6 +1475,27 @@ function Design:onInput(keys)
...
+                    self.start_center = pos
+                    self.prev_center = pos
+                    return true
+                elseif self.prev_center then
+                    --If there was no movement presume user wanted to click the mark underneath instead and let the flow through.
+                    if pos == self.start_center then
```
myk002, line 1487: same here. you probably want `same_xyz`

myk002 (commented):
don't forget the changelog entry (unless you think this is adequately covered by what's already there)

gui/design.lua
MagistrumT-T: As I understand in this line they are both `Point` objects, and the `==` operator was overloaded with the __eq() function in here:

https://github.com/DFHack/scripts/blob/ead5eaa603a4d7c97908206414fd06155d07a05e/internal/design/util.lua#L77-L81

It is showing the behavior I expect, but I don't understand Lua metatable shenanigans very well, so there could be implications I don't understand.

gui/design.lua
MagistrumT-T: I am reasonably sure that `self.start_center` will also be a `Point` here and make use of Coscker's handy compare function,  but once again, I am not that well-versed in this and may have misunderstood how this usage works.

MagistrumT-T:
Also I had totally forgot the changelog entry, sorry.

myk002:
pre-commit.ci autofix

myk002 (approved):

PR #997, opened 2024-02-21
[fix/stuck-worship] new tool: prevent units getting stuck worshipping

myk002:
when they have multiple deities

addresses DF bug https://dwarffortressbugtracker.com/view.php?id=10918

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -27,6 +27,7 @@ Template for new versions:
 # Future
 
 ## New Tools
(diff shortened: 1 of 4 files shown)


myk002:
@quietust could you look this over and tell me if this algorithm for need redistribution raises any red flags? I've been testing it, and it seems to work reliably, but this is the first time I've dealt with religious needs and I'd like some assurance that I'm not doing something silly.

PR #1016, opened 2024-03-03
New and improved gui/mechanisms

Bumber64:
Gui script for unlinking and freeing mechanisms from buildings

Also fixed typos of "seize" in confirm, and ``focus_string`` for ``gui/teleport`` being "autodump".

Requires DFHack/dfhack#4319 to move freed mechanisms to ground properly. (Buildings may deconstruct if dwarves are allowed pull out the mechanisms themselves. Don't know what causes this, as it only happens sometimes.)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -33,6 +33,7 @@ Template for new versions:
 - `instruments`: provides information on how to craft the instruments used by the player civilization
 - `modtools/item-trigger`: (reinstated) modder's resource for triggering scripted content when specific items are used
 - `modtools/if-entity`: (reinstated) modder's resource for triggering scripted content depending on the race of the loaded fort
+- `devel/block-borders`: (reinstated) highlights boundaries of map blocks or embark tile blocks
 
 ## New Features
 - `gui/settings-manager`: add import, export, and autoload for work details
--- a/devel/block-borders.lua
+++ b/devel/block-borders.lua
@@ -1,19 +1,8 @@
 -- overlay that displays map block borders
 
---[====[
-
-devel/block-borders
-===================
-
-An overlay that draws borders of map blocks. See :doc:`/docs/api/Maps` for
-details on map blocks.
-
-]====]
-
-local gui = require "gui"
+local gui = require('gui')
 local guidm = require "gui.dwarfmode"
-
-local ui = df.global.plotinfo
+local widgets = require('gui.widgets')
 
 local DRAW_CHARS = {
     ns = string.char(179),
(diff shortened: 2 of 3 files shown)


internal/confirm/specs.lua
```diff
@@ -162,9 +162,9 @@ ConfirmSpec{
 }
 
 ConfirmSpec{
-    id='trade-sieze',
-    title='Sieze merchant goods',
-    message='Are you sure you want size marked merchant goods? This will make the merchant unwilling to trade further and will damage relations with the merchant\'s civilization.',
+    id='trade-seize',
+    title='Seize merchant goods',
+    message='Are you sure you want seize marked merchant goods? This will make the merchant unwilling to trade further and will damage relations with the merchant\'s civilization.',
```
myk002, line 167: wow, three typos in three lines. I was on a roll.

gui/mechanisms.lua
```diff
@@ -0,0 +1,601 @@
...
+                break
+            end
+        end
+    elseif b._type == df.building_doorst or b._type == df.building_hatchst then
+        b.door_flags.forbidden = false
+        b.door_flags.operated_by_mechanisms = false
+
+        local mb = dfhack.maps.getTileBlock(b.pos)
+        if mb then
+            mb.flags.repath_on_melt = true --fix pathing
```
Bumber64, line 83: Is there a better way to clean up a tile's pathability than this?

gui/mechanisms.lua
```diff
@@ -0,0 +1,598 @@
...
+        if gref._type == df.general_ref_building_triggerst or
+            gref._type == df.general_ref_building_triggertargetst then
+                return k
+        end
+    end
+end
+
+local function get_mech_link(m) --return target building or nil
+    local x = get_trigger_index(m)
+    x = x and df.building.find(m.general_refs[x].building_id) or NULL
```
myk002, line 66: `NULL` -> `nil`

gui/mechanisms.lua
myk002: You'll likely need to set `df.global.world.reindex_pathfinding = true` to have DF recalculate the pathability flood fill

myk002 (commented):
I like the idea of being able to unlink mechanisms, but I'd rather see this as an overlay than a GUI tool. Specifically, it could draw action buttons to the left or right of the list of linked buildings on this panel:
image

The overlay could be added to https://github.com/DFHack/dfhack/blob/develop/plugins/lua/buildingplan/mechanisms.lua where we already have mechanism-related code

example of overlay that adds widgets to a vanilla list: `PreferenceOverlay` in https://github.com/DFHack/dfhack/blob/develop/plugins/lua/sort/diplomacy.lua

gui/mechanisms.lua
Bumber64: Doesn't seem sufficient.

Bumber64:
> I like the idea of being able to unlink mechanisms, but I'd rather see this as an overlay than a GUI tool. Specifically, it could draw action buttons to the left or right of the list of linked buildings on this panel [...]

How do we indicate that the unlink buttons are a DFHack feature?

The auto-freeing settings would go on that panel. Manual freeing buttons would probably go on the "Show items" panel, along with a "Free all unlinked" option there somewhere.

Converting the script is probably more effort than I want to do at the moment, however.

myk002:
> How do we indicate that the unlink buttons are a DFHack feature?

the usual method is to either put them in a DFHack-labeled frame or to put red square brackets around it (`widgets.TextButton` or `widgets.BannerPanel`):
image

Bumber64:
> > How do we indicate that the unlink buttons are a DFHack feature?
> 
> the usual method is to either put them in a DFHack-labeled frame or to put red square brackets around it

What if we were putting a button here:
button

myk002:
that looks like a reasonable spot. A TextButton like `[Unlink]` instead of an icon might be better to make it clearer that it's not a vanilla button.

The player shouldn't have to press anything else to make the button appear -- the overlay will just render it when in the correct context. I notice that right now this context is represented as `dwarfmode/ViewSheets/BUILDING/Trap/Lever` and it doesn't differentiate between the "Show linked buildings" and "Shot items" tabs. We can modify Gui.cpp to differentiate the context. Then you can attach the overlay to `dwarfmode/ViewSheets/BUILDING/Trap/Lever/LinkedBuildings`

Bumber64:
Ran into a DF bug: https://dwarffortressbugtracker.com/view.php?id=12721

The linked buildings page doesn't scroll.

Bumber64:
> Then you can attach the overlay to `dwarfmode/ViewSheets/BUILDING/Trap/Lever/LinkedBuildings`

Is there a way to specify any `dwarfmode/*/LinkedBuildings`, or do we need to specify every linked building screen manually?

myk002:
> Is there a way to specify any matching `*/LinkedBuildings`, or do we need to specify every linked building screen manually?

if there are too many to list, you can specify the longest common prefix as your `viewscreens` attribute and then check the suffix in your `render` and `onInput` overrides. e.g. (from `gui/notify`):
```lua
function NotifyOverlay:render(dc)
    if not CONFLICTING_TOOLTIPS[mi.current_hover] then
        NotifyOverlay.super.render(self, dc)
    end
end
```

you can get the current focus strings from `dfhack.gui.getFocusStrings(dfhack.gui.getDFViewscreen(true))`

Bumber64:
How does one embed buttons in a scrolling list? The scroll position is defined by `df.global.game.main_interface.view_sheets.scroll_position_item`, but I'm not sure what the bounds for the list should be considering UI scale, etc. Presumably we have to disable rendering the buttons if they're out of bounds. By default, I think they just sit at the edge of the overlay's rect.

myk002:
implement `preUpdateLayout(parent_rect)` for your overlay class and dynamically adjust your height based on `parent_rect.height`. You can then calculate how many visible buttons there are and where to render them. Example: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/sort.lua#L989

Bumber64:
Is there any way to detect the overlay activating? I've got state data when swapping between `*/LinkedBuildings` and `*/Items`, and I'd prefer not to rebuild my lists every render frame. (I've been checking if the `viewing_bldid` is different, but that doesn't help when changing tabs on the same building.)

myk002:
> (I've been checking if the `viewing_bldid` is different, but that doesn't help when changing tabs on the same building.)

would that work if a building was selected, deselected, and selected again after the state had changed?

unless the performance is a significant blocker, drawing the UI state directly from DF state and keeping the overlay as stateless as you can is a good idea. Example: https://github.com/DFHack/dfhack/blob/develop/plugins/lua/orders.lua#L391-L415

Bumber64:
> would that work if a building was selected, deselected, and selected again after the state had changed?
No, that's the same problem were I'd need to detect the overlay activating. It only catches swapping buildings without closing the view sheet.

> unless the performance is a significant blocker, drawing the UI state directly from DF state and keeping the overlay as stateless as you can is a good idea.
It's O(n^2), iterating `contained_items` and their `general_refs` (if the item is a mechanism.) I'm not sure how many mechanisms you'd need to cram into a lever to make it a problem.

What's the difference between `render` and `onRenderFrame` besides the later receiving a `rect`?

myk002:
> What's the difference between `render` and `onRenderFrame` besides the later receiving a `rect`?

for the purpose of "do something on every rendered frame", there's no difference. `render` is the `View` method that is called first. The default implementation of `render` calls `onRenderFrame` for widgets.

PR #1020, opened 2024-03-04
new tool: instruments

chdoc:
This tool provides information on how to craft the instruments used by the player civilization.


chdoc:
Sorry for the noise, I accidentally pushed with `-d` rather than `-f` :man_shrugging:

PR #1021, opened 2024-03-04
new tool: instruments

chdoc:
This tool provides information on how to craft the instruments used by the player civilization.

### DIFF
--- a/docs/instruments.rst
+++ b/docs/instruments.rst
@@ -1,0 +1,20 @@
+instruments
+===========
+
+.. dfhack-tool::
+    :summary: Show how to craft instruments.
+    :tags: fort productivity
+
+This tool provides information on how to craft the instruments used by the
+player civilization. For single-piece instruments, it shows the skill and
+material needed to craft it. For multi-piece instruments, it displays the skill
+used in its assembly as well as information on how to craft the necessary
+pieces. It also shows whether the instrument is handheld or placed as a
+building.
+
+Usage
+-----
+
+::
+
+    instruments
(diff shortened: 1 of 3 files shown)


docs/instruments.rst
```diff
@@ -0,0 +1,20 @@
+instruments
+===========
+
+.. dfhack-tool::
+    :summary: Show how to craft instruments.
+    :tags: fort productivity
```
myk002, line 6: since this tool doesn't actually generate workorders, I'd classify it as `inspection` instead of `productivity`

myk002 (commented):
straightforward logic and super clean code. Love it!

I'll add it to the official list at https://docs.google.com/spreadsheets/d/1hiDlo8M_bB_1jE-5HRs2RrrA_VZ4cRu9VXaTctX_nwk/edit#gid=1774645373

docs/instruments.rst
myk002: edit: maybe copy the tags from `necronomicon`: `fort inspection productivity items`.

chdoc: Hmm, upon some thought, I'd argue that `necronomicon` is the one that's misclassified. At a glance, it seems to be the only tool tagged `productivity` that is not actually modifying gamestate. Also, this tool, unlike `necronomicon`, does not actually "interact with in-game items". So I think your first reaction was the correct one: `fort inspection`.

myk002: that's a good point. I'll remove `productivity` from `necronomicon`

chdoc: @myk002 I think you moved too fast and merged it tagged `fort productivity` rather than `fort inspection`.

chdoc: I was trying to avoid creating multiple commits in case you disagreed. So I thought you'd change that before merging. :man_shrugging:

myk002: oop, will fix in situ

PR #1026, opened 2024-03-07
Made two functions global for third party use.

ch3mbot:
Made get_newregion_viewscreen and get_modlist_fields global, so that mod related data is accessible through the dfhack api.


myk002 (approved):

myk002:
see https://steamcommunity.com/app/2346660/discussions/0/4293690852332196768/ for the background for this PR

PR #1049, opened 2024-03-17
fix/ownership feature

LightHardt:
For details, please refer to https://github.com/DFHack/dfhack/issues/4373

I was able to add to the control panel by following `fix/protect-nicks` closely however I can't actually enable it from the control panel. Any help on this would be appreciated.

Also, I left some additional information/question on the issue discussing this and if any feedback or thoughts on it please let me know.


myk002:
Since this is a stateless tool, a better, simpler model to follow is `fix/stuck-instruments`. You can control the cycle with a `repeat` configuration in the control panel registry

LightHardt:
Thank you that did the trick!

fix/ownership.lua
```diff
@@ -0,0 +1,49 @@
...
+the constant looping of the "Store owned item" job. This may show as a unit
+repeatedly trying to put an item in their cabinet and they cant causing them
+to keep picking it up and trying to put it in.
+
+Usage:
+fix/ownership
+fix/ownership help
+
+]====]
+
```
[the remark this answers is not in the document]
myk002: there are some leftover doc strings in the DFHack scripts that we haven't removed yet, but we shouldn't be adding more. Docs should now only be in separate .rst files.

fix/ownership.lua
```diff
@@ -0,0 +1,49 @@
+--@ module = true
```
myk002, line 1: You should only mark the file as a module if it can actually be loaded as a module (that is, it does not have side effects when loaded). this line can be removed.

internal/control-panel/registry.lua
```diff
@@ -68,6 +68,9 @@ COMMANDS_BY_IDX = {
     {command='fix/general-strike', group='bugfix', mode='repeat', default=true,
         desc='Prevent dwarves from getting stuck and refusing to work.',
         params={'--time', '1', '--timeUnits', 'days', '--command', '[', 'fix/general-strike', '-q', ']'}},
+    {command='fix/ownership', group='bugfix', mode='repeat', default=true,
+        desc='Fixes instances of units claiming the same item or an item they don\'t own',
```
myk002, line 72:
```suggestion
        desc='Fixes instances of units claiming the same item or an item they don\'t own.',
```

fix/ownership.lua
```diff
@@ -0,0 +1,49 @@
...
+repeatedly trying to put an item in their cabinet and they cant causing them
+to keep picking it up and trying to put it in.
+
+Usage:
+fix/ownership
+fix/ownership help
+
+]====]
+
+-- unit thinks they own the item but the item doesnt hold the proper
```
myk002, line 21:
```suggestion
-- unit thinks they own the item but the item doesn't hold the proper
```

fix/ownership.lua
```diff
@@ -0,0 +1,49 @@
...
+            local item = df.item.find(unit.owned_items[index])
+            if not item then goto continue end
+
+            for _, ref in ipairs(item.general_refs) do
+                if df.general_ref_unit_itemownerst:is_instance(ref) then
+                    -- make sure the ref belongs to unit
+                    if ref.unit_id == unit.id then goto continue end
+                end
+            end
+            print('Erasing ' .. dfhack.TranslateName(unit.name) .. ' claim on item #' .. item.id)
```
myk002, line 35:
```suggestion
            print('Erasing ' .. dfhack.TranslateName(unit.name) .. ' invalid claim on item #' .. item.id)
```

myk002 (approved):

PR #1072, opened 2024-04-14
Add peafowl to gui/control-panel poultry list

nsimmonds:
Adds a control-panel line that sets the targets for peafowl similarly to other poulry.


myk002:
pre-commit.ci autofix

myk002 (approved):

PR #1079, opened 2024-04-20
Added ability to request gems by their value

Hiranus:


myk002:
could you update your fork? it appears to be out of date. For the future, it is better practice to create a branch on your fork before submitting a PR. That way you have the ability to work on *other* things while this PR is still under review.

internal/caravan/tradeagreement.lua
```diff
@@ -26,6 +37,59 @@ local function diplomacy_toggle_cat()
...
+
+    table.sort(gemValuesUnique)
+    return gemPrices, gemValuesUnique
+end
+
+local function MinimumGemValue()
+    local gemPrices, gemValuesUnique = GetCivGemPrices()
+    local list = {}
+    for index, value in ipairs(gemValuesUnique) do
+        list[index] = tostring(value)
```
myk002, line 81: just displaying the number is a little spartan. could we add some more information here? for example, it would be useful to know how many items are at each value tier.

internal/caravan/tradeagreement.lua
```diff
@@ -35,4 +99,14 @@ function TradeAgreementOverlay:init()
             on_activate=diplomacy_toggle_cat,
         },
     }
+    self:addviews {
+        widgets.HotkeyLabel {
+            frame = { t = 1, l = 0 },
+            label = 'Select gems at least as valuable as',
```
myk002, line 105: Instead of "select at least as valuable as", how about "Select by value"? Only items with the shown value would be toggled on click. This would also be relevant beyond gems. Being able to select the cheapest leather, for example, is a common request.

myk002 (commented):
needs documentation in the Overlays section of `docs/caravan.rst`

internal/caravan/tradeagreement.lua
```diff
@@ -1,23 +1,83 @@
...
-    frame_background=gui.CLEAR_PEN,
+TradeAgreementOverlay.ATTRS {
+    desc = 'Adds select all/none functionality when requesting trade agreement items.',
+    default_pos = { x = 45, y = -6 },
+    default_enabled = true,
+    viewscreens = 'dwarfmode/Diplomacy/Requests',
+    frame = { w = 40, h = 4 },
+    frame_style = gui.MEDIUM_FRAME,
+    frame_background = gui.CLEAR_PEN,
 }
```
myk002, line 16: Inside of tables, the convention (at least inside of DFHack code) is to have no spaces around the `=`

internal/caravan/tradeagreement.lua
```diff
@@ -26,13 +86,71 @@ local function diplomacy_toggle_cat()
...
+            table.insert(matValuesUnique, val)
+        else
+            filter[matPrice].count = filter[matPrice].count + 1
+        end
+    end
+    table.sort(matValuesUnique, function(a, b) return a.value < b.value end)
+    return matPrices, matValuesUnique
+end
+
+local function ValueSelector()
```
myk002, line 120: convention for local functions is snake_case, not PascalCase (which is generally used for class names)

internal/caravan/tradeagreement.lua
```diff
@@ -26,13 +86,71 @@ local function diplomacy_toggle_cat()
...
+    local currTab = GetTab()
+    local matPrices, matValuesUnique = GetCivMatPrices(currTab)
+    local list = {}
+    for index, value in ipairs(matValuesUnique) do
+        list[index] = tostring(value.value) ..
+            " - " .. tostring(value.count) .. " " .. (value.count == 1 and currTab.labelSingular or currTab.labelPlural)
```
myk002, line 126: for multi-part strings like this, it is usually cleaner to use a format string:
```lua
('%d - %d %s'):format(value.value, value.count,
    value.count == 1 and currTab.labelSingular or currTab.labelPlural)
```

myk002: I'd also suggest adding the "gear" symbol after the value to indicate price. the number-heavy string label also feels a bit user-unfriendly. we might want to post some more screenshots to the #feature-discuss Discord channel to get feedback on that.

internal/caravan/tradeagreement.lua
```diff
@@ -1,23 +1,83 @@
...
 local diplomacy = df.global.game.main_interface.diplomacy
+local lasttab, lastresult = -1, nil
+
+local function tabsdef(pages, labelSingular, labelPlural, getCivResourceList, funcGetValue)
+    return {
+        pages = pages,
+        labelSingular = labelSingular,
+        labelPlural = labelPlural,
+        getCivResourceList = getCivResourceList,
+        funcGetValue = funcGetValue,
```
myk002, line 26: the convention for this type of variable is to name it like `get_value_fn`

myk002:
ping

PR #1086, opened 2024-04-25
gui/autobutcher changes

Ajhaa:
- Add a warning to enable autobutcher to make changes
- Add features to butcher and unbutcher all animals

Fixes: https://github.com/DFHack/dfhack/issues/4220

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -33,16 +33,10 @@ Template for new versions:
 ## Fixes
 
 ## Misc Improvements
+- `gui/autobutcher`: add functionality to butcher and unbutcher all animals
+- `gui/autobutcher`: add warning if autobutcher is disabled
 
 ## Removed
-
-# 50.13-r2
-
-## New Features
-- `gui/autobutcher`: add functionality to butcher and unbutcher all animals
-
-## Misc Improvements
-- `gui/autobutcher`: add warning if autobutcher is disabled
 
 # 50.13-r1.1
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -36,16 +36,19 @@ Template for new versions:
 
 ## Removed
 
-# 50.13-r1.1
+# 50.13-r2
+
 ## New Features
 - `gui/autobutcher`: add functionality to butcher and unbutcher all animals
+
+## Misc Improvements
+- `gui/autobutcher`: add warning if autobutcher is disabled
+
+# 50.13-r1.1
 
 ## Fixes
 - `gui/quantum`: accept all item types in the output stockpile as intended
 - `deathcause`: fix error on run
-
-## Misc Improvements
-- `gui/autobutcher`: add warning if autobutcher is disabled
 
 # 50.13-r1
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -37,10 +37,15 @@ Template for new versions:
 ## Removed
 
 # 50.13-r1.1
+## New Features
+- `gui/autobutcher`: add functionality to butcher and unbutcher all animals
 
 ## Fixes
 - `gui/quantum`: accept all item types in the output stockpile as intended
 - `deathcause`: fix error on run
+
+## Misc Improvements
+- `gui/autobutcher`: add warning if autobutcher is disabled
 
 # 50.13-r1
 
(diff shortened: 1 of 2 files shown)


gui/autobutcher.lua
```diff
@@ -306,6 +306,13 @@ function WatchList:init()
                         frame={t=1, l=82},
                         text='ordered'
                     },
+                    widgets.Label{
+                        view_id='disabled_warning',
+                        visible=not plugin.isEnabled,
```
myk002, line 311:
```suggestion
                        visible=function() return not plugin.isEnabled() end,
```

myk002: you can add a similar (but inverted) `visible` attribute to the list; then you can remove the code added to `refresh()`

the list one can just look like this:
```lua
visible=plugin.isEnabled
```
since it's already a function with the correct signature

gui/autobutcher.lua
```diff
@@ -316,7 +323,7 @@ function WatchList:init()
             },
             widgets.Panel{
                 view_id='footer',
-                frame={l=0, r=0, b=0, h=6},
+                frame={l=0, r=0, b=0, h=7},
```
myk002, line 326: if you change `h` here, you'll also need to change the `b` value for `list_panel`

gui/autobutcher.lua
```diff
@@ -402,9 +409,25 @@ function WatchList:init()
                         auto_width=true,
                         on_activate=self:callback('onSetRow'),
                     },
+                    widgets.HotkeyLabel{
+                        view_id='butcher_all',
+                        frame={t=5, l=0},
```
myk002, line 414: When we first released this tool, the labels were unaligned in the footer and we got a lot of complaints about how messy it looked. I suggest reorganizing the labels so it looks like this:

```
f: f kids        F: F adults        w: Toggle watch  r: Set row targets to 0
m: m kids        M: M adults        x: Delete row    R: Set row targets to N
b: Butcher race  B: Unbutcher race  e: Butcher all   E: Unbutcher all
A: Autobutcher is Enabled           W: Autowatch is Enabled
```

myk002 (commented):
good job!

changelog.txt
```diff
@@ -36,6 +36,14 @@ Template for new versions:
...
 ## Removed
 
+# 50.13-r2
+
+## New Features
+- `gui/autobutcher`: add functionality to butcher and unbutcher all animals
+
+## Misc Improvements
+- `gui/autobutcher`: add warning if autobutcher is disabled
+
```
myk002, line 46: we don't add a version header to the changelog until we tag the release. it's possible that the version this goes into will not be named `50.13-r2`, even that's what the current plan is. These should go under the `Future` section.

changelog.txt
```diff
@@ -36,6 +36,14 @@ Template for new versions:
 
 ## Removed
 
+# 50.13-r2
+
+## New Features
+- `gui/autobutcher`: add functionality to butcher and unbutcher all animals
```
myk002, line 42: this doesn't quite qualify as a "new feature"; it should be in the Misc Improvements section instead

gui/autobutcher.lua
```diff
@@ -709,6 +734,29 @@ function WatchList:onButcherRace()
...
+function WatchList:onUnbutcherAll()
+    for _, data in ipairs(plugin.autobutcher_getWatchList()) do
+        plugin.autobutcher_unbutcherRace(data.id)
+    end
+
+    self:refresh()
+end
+
+function WatchList:onButcherAll()
+    dlg.showYesNoPrompt(
```
myk002, line 746: After playing around with the UI changes, I'm going back and forth on whether this prompt is necessary. Marking all animals for butchering is non-destructive -- you can easily unmark all for butchering.

However, if a player *accidentally* hits `e`, then it could be an unpleasant surprise when all the player's animals are subsequently slaughtered.

Butchering all animals should be an infrequent enough action such that the prompt probably won't become annoying, so I think it can stay.

myk002 (commented):
looks great! thanks!

just a few requested changes to the changelog, and looks ready to merge

Some followup thoughts:
- Please feel free to add your name to https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst?plain=1
- For future PRs, please consider making a branch instead of committing to `master`. It prevents merge conflicts and allows you to work on multiple PRs at once.

myk002 (approved):

PR #1099, opened 2024-05-03
fix engravings that have gone out of sync

myk002:
Fixes: https://github.com/DFHack/dfhack/issues/4421

PR #1101, opened 2024-05-04
fixups for fix/engravings

myk002:
continued from https://github.com/DFHack/scripts/pull/1099


myk002:
pre-commit.ci autofix

PR #1117, opened 2024-05-12
extinguish.lua: plant_flags.is_burning no longer used

Bumber64:
`plant_flags.is_burning` isn't used by DF anymore, renamed `plant_flags.unused_01`. Commented out the code.

Not sure if existing code is sufficient, or something else needs to be done.

DFHack/df-structures/pull/753


myk002 (approved):

myk002:
confirmed that this flag no longer is needed to put out plant fires

PR #1118, opened 2024-05-14
Fix gm-unit for skills above level 20

tatoyoda600:
Fixes DFHack/dfhack#4579
image

### DIFF
--- a/internal/gm-unit/editor_skills.lua
+++ b/internal/gm-unit/editor_skills.lua
@@ -24,9 +24,9 @@ function list_skills(unit, learned_only)
                     u_skill={rating=-1,experience=0}
                 end
                 local rating
-                if u_skill.rating >15 then
-                    local legendary_bonus= u_skill.rating - 15
-                    rating=df.skill_rating.attrs[15]
+                if u_skill.rating > df.skill_rating.Legendary then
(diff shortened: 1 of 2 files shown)


internal/gm-unit/editor_skills.lua
```diff
@@ -24,7 +24,13 @@ function list_skills(unit, learned_only)
                     u_skill={rating=-1,experience=0}
                 end
                 local rating
-                if u_skill.rating >=0 then
+                if u_skill.rating >15 then
```
myk002, line 27: Please use `df.skill_rating.Legendary` instead of the magic number `15`

myk002 (commented):
Thanks! please add a line to the `Misc Improvements` section under the "Future" version in `changelog.txt`

Also, please feel free to add yourself to https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst (in a separate PR)

internal/gm-unit/editor_skills.lua
tatoyoda600: I've replaced the magic number and added an entry in `changelog.txt`.
I'm unsure if I made the changelog entry correctly though.

myk002 (approved):
looks great! Thanks!

PR #1120, opened 2024-05-16
[item] don't match spider webs

myk002:
@chdoc FYI


myk002:
that's fair; will change

PR #1122, opened 2024-05-17
devel/tree-info

Bumber64:
Script that prints a visualization of a tree's `tree_info` struct. Will be in color if the DFHack console is used.

Useful to see if the tree thinks there are roots where there aren't any, or which tiles are considered blocked.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -28,6 +28,7 @@ Template for new versions:
 
 ## New Tools
 - `gui/sitemap`: list and zoom to nearby people, locations, and artifacts
+- ``devel/tree-info``: print a visualization of tree_info data
 
 ## New Features
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -28,7 +28,7 @@ Template for new versions:
 
 ## New Tools
 - `gui/sitemap`: list and zoom to nearby people, locations, and artifacts
-- ``devel/tree-info``: print a visualization of tree_info data
+- `devel/tree-info`: print a visualization of tree_info data
 
 ## New Features
 
--- a/devel/tree-info.lua
+++ b/devel/tree-info.lua
@@ -1,0 +1,163 @@
+--Print a tree_info visualization of the tree at the cursor.
+--@module = true
+
+local branch_dir =
+{
+    [0] = ' ',
+    [1] = string.char(26), --W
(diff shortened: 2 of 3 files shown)


changelog.txt
```diff
@@ -28,6 +28,7 @@ Template for new versions:
 
 ## New Tools
 - `gui/sitemap`: list and zoom to nearby people, locations, and artifacts
+- ``devel/tree-info``: print a visualization of tree_info data
```
myk002, line 31:
```suggestion
- `devel/tree-info`: print a visualization of tree_info data
```

docs/devel/tree-info.rst
```diff
@@ -0,0 +1,17 @@
+devel/tree-info
+===============
+
+.. dfhack-tool::
+    :summary: Print a tree_info visualization of the tree at the cursor.
+    :tags: adventure fort inspection map plants
```
myk002, line 6:
```suggestion
    :tags: dev
```

devel/tree-info.lua
```diff
@@ -0,0 +1,163 @@
...
+    [8] = string.char(24), --S
+    [9] = string.char(191), --WS
+    [10] = string.char(179), --NS
+    [11] = string.char(180), --WNS
+    [12] = string.char(218), --ES
+    [13] = string.char(194), --WES
+    [14] = string.char(195), --NES
+    [15] = string.char(197), --WNES
+}
+
```
myk002, line 23: in case the script has an error while accessing the data, it would keep the console in a better state if the print statements were self-contained in their color changes. That is, have a function like:

```lua
local function print_color(color, s)
    dfhack.color(color)
    dfhack.print(s)
    dfhack.color(COLOR_RESET)
end
```

and use that instead of `dfhack.print` directly

myk002 (commented):
nice

docs/devel/tree-info.rst
myk002: I'll register this script at https://docs.google.com/spreadsheets/d/1hiDlo8M_bB_1jE-5HRs2RrrA_VZ4cRu9VXaTctX_nwk/edit#gid=1774645373

myk002 (approved):

PR #1127, opened 2024-05-24
update to canonicalized structures

myk002:
ref: https://github.com/DFHack/df-structures/pull/763

### DIFF
--- a/internal/quickfort/zone.lua
+++ b/internal/quickfort/zone.lua
@@ -23,15 +23,15 @@ end
 
 local function parse_gather_props(zone_data, props)
     if props.pick_trees == 'false' then
-        ensure_keys(zone_data, 'zone_settings', 'gather').pick_trees = false
+        ensure_keys(zone_data, 'zone_settings', 'gather', 'flags').pick_trees = false
         props.pick_trees = nil
     end
     if props.pick_shrubs == 'false' then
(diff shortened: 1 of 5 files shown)


myk002:
merging so we can run a full test suite in https://github.com/DFHack/dfhack/pull/4631

internal/quickfort/zone.lua
```diff
@@ -105,9 +105,9 @@ local zone_db_raw = {
     b={label='Bedroom', default_data={type=df.civzone_type.Bedroom}},
     h={label='Dining Hall', default_data={type=df.civzone_type.DiningHall}},
     n={label='Pen/Pasture', default_data={type=df.civzone_type.Pen,
-       assign={zone_settings={pen={unk=1}}}}},
+       assign={zone_settings={pen={check_occupants=true}}}}},
     p={label='Pit/Pond', props_fn=parse_pit_pond_props, default_data={type=df.civzone_type.Pond,
-       assign={zone_settings={pit_pond=df.building_civzonest.T_zone_settings.T_pit_pond.top_of_pit}}}},
+       assign={zone_settings={pond={flag={keep_filled=true}}}}}},
```
quietust, line 110: Strictly speaking, the old `top_of_pit` value should be equivalent to `check_occupants=True` and `keep_filled=False`.

PR #1139, opened 2024-06-02
Adventure Mode fix/gifted-pets script

Crystalwarrior:
Fixes gifted pets data not being adjusted to match the unit the pet was gifted to, creating a situation where fast-travelling unloads the unit with no way to access that unit again in any context


fix/gifted-pets.lua
```diff
@@ -0,0 +1,43 @@
...
+            for u, link in ipairs(site_links) do
+                local new_site_link = df.new(link)
+                new_site_link.site = link.site
+                new_site_link.sub_id = link.sub_id
+                new_site_link.entity = link.entity
+
+                companion_record.figure.site_links:insert('#', new_site_link)
+            end
+            -- pet_owner_record.companions:insert('#', companion_record.id)
+            alreadyProcessed[tostring(companion_record.id)] = true
```
myk002, line 33: no need to convert the key to a string. integer/numeric IDs work just as well

fix/gifted-pets.lua
```diff
@@ -0,0 +1,43 @@
...
+
+            companion_record.unit.relationship_ids.GroupLeader = pet_owner_record.unit_id
+            companion_record.group_leader_id = pet_owner
+            
+            for u, link in ipairs(site_links) do
+                local new_site_link = df.new(link)
+                new_site_link.site = link.site
+                new_site_link.sub_id = link.sub_id
+                new_site_link.entity = link.entity
+
```
myk002, line 29:
```suggestion
                local new_site_link = link:new()
```

myk002: calling `new` on an existing object copies that object

fix/gifted-pets.lua
```diff
@@ -0,0 +1,43 @@
...
+        end
+    end
+end
+
+-- clean up our records
+for index = #adventurer_record.companions-1, 0, -1 do
+    if alreadyProcessed[tostring(adventurer_record.companions[index])] then
+        adventurer_record.companions:erase(index)
+    end
+end
```
myk002, line 43:
```suggestion
end

```

myk002:
the trick here is to integrate it into the UI so the player knows to click a button to run it at the right time or else make it fully automatic so that the player doesn't need to do anything to benefit from the fix

fix/gifted-pets.lua
```diff
@@ -0,0 +1,112 @@
...
+
+local function findInParty(party, histfig_id)
+    for k, v in ipairs(party) do
+        if v == histfig_id then
+            return true
+        end
+    end
+    return false
+end
```
myk002, line 40: this could be replaced with a call to `utils.linear_index` (or `utils.binsearch` if the vector is sorted)

fix/gifted-pets.lua
```diff
@@ -0,0 +1,112 @@
...
+        link._type == df.histfig_site_link_home_site_realization_buildingst or
+        link._type == df.histfig_site_link_lairst or
+        link._type == df.histfig_site_link_home_site_realization_sulst or
+        link._type == df.histfig_site_link_home_site_saved_civzones then
+            -- Make a copy of the link
+            local new_site_link = df.new(link)
+            -- Assign the same data
+            new_site_link.site = link.site
+            new_site_link.sub_id = link.sub_id
+            new_site_link.entity = link.entity
```
myk002, line 75: should already be done by the `df.new(link)` (maybe make that `link:new()` to make it clearer)

fix/gifted-pets.lua
```diff
@@ -0,0 +1,112 @@
...
+            print("nemesis "..nem_id.." is present in nemesis record but not part of the adventurer's party, checking if it's a pet")
+            local pet_owner_record = findPetOwnerOf(companion_record)
+            if pet_owner_record then
+                print("pet owner confirmed! Identified nemesis "..pet_owner_record.id.." as owner.")
+                local party = party
+                local owner_in_party = findInParty(party.party_core_members, companion_record.figure.id) or findInParty(party.party_extra_members, companion_record.figure.id)
+                if not owner_in_party then
+                    print("the pet's owner is NOT a companion. We need to associate the pet with the owner correctly so the pet's unit is not inaccessible.")
+                    fixGiftedPet(companion_record, pet_owner_record)
+                    queueCompanionErase[pet_record.id] = true
```
myk002, line 99: `pet_record` is not defined

myk002 (commented):
what are your thoughts for making this tool either discoverable by the player or automatic? What screen would need to be monitored or action hooked?

Crystalwarrior:
> what are your thoughts for making this tool either discoverable by the player or automatic? What screen would need to be monitored or action hooked?

I need to keep track of unretirement, adding a member to party, and removing pet from party.
Another thing I could keep track of is speech options like "gift pet" being processed far enough that the game sets all its data

myk002 (commented):
I think this should be made part of `advtools` and not be a separate script.

My reasoning is that Toady is aware of this issue, and may have a fix soon. Adding and then removing a script is a bit of a waste of a tombstone slot. Moreover, we're advertising `advtools` as "the place to look for adventure mode commandline stuff", and this fix might be more discoverable there.

docs/advtools.rst
```diff
@@ -25,6 +26,25 @@ When you run this command, you will get a list of your extra party members and c
...
+
+``advtools pets``
+    Fixes companion pets and pets you have gifted to prevent buggy behavior and crashes.
+
+``pets`` Command
+----------------
+
+Run this command after you gift a pet or after you hire a companion who has their own pets following them.
+This should fix pets not joining you during fast travel, potential duplication issues, and gifted pets
+still being part of your party despite being given away.
```
myk002, line 40:
```suggestion
```

myk002: this seems to duplicate what is said in the next two paragraphs

myk002:
pre-commit.ci autofix

myk002 (approved):

PR #1146, opened 2024-06-04
New script - add to core party

Crystalwarrior:
* Pick one of your extra members to convert to your core party, allowing control
* Also converts those core party members into an "Adventurer"
* some code shamelessly copy pasted from unretire-anyone.lua

there's also an issue where the "brain" icon deciding on manual/automatic control is somewhat broken in the `c`ompanions screen.

research leads me to believe the data is stored per-unit or unit ID, need to figure out where that data is stored exactly. Might be one of the unknown variables?

### DIFF
--- a/add-to-core-party.lua
+++ b/add-to-core-party.lua
@@ -1,0 +1,65 @@
+local dialogs = require 'gui.dialogs'
+local utils = require 'utils'
+local viewscreen = dfhack.gui.getDFViewscreen(true)
+if viewscreen._type ~= df.viewscreen_dungeonmodest then
+    qerror("This script can only be used during adventure mode!")
+end
+
+local function addToCoreParty(nemesis)
+    -- Adds them to the party core members list
+    local party = df.global.adventure.interactions
+    -- problem: the "brain" icon deciding on manual/automatic control is somewhat broken.
+    -- research leads me to believe the data is stored per-unit or unit ID, need to figure out
+    -- where that data is stored exactly. Might be one of the unknown variables?
+    party.party_core_members:insert('#', nemesis.figure.id)
+    local extra_member_idx, _ = utils.linear_index(party.party_extra_members, nemesis.figure.id)
+    if extra_member_idx then
+        party.party_extra_members:erase(extra_member_idx)
+    end
+    -- Adds them to unretire list
+    nemesis.flags.ADVENTURER = true
+    nemesis.flags.ACTIVE_ADVENTURER = true
+end
+
+local function showExtraPartyPrompt(advSetUpScreen)
+    local choices = {}
+    for _, figure_id in ipairs(df.global.adventure.interactions.party_extra_members) do
+        -- shamelessly copy-pasted from unretire-anyone.lua
+        local histFig = df.historical_figure.find(figure_id)
+        local nemesis = df.nemesis_record.find(histFig.nemesis_id)
+        local histFlags = histFig.flags
+        local creature = df.creature_raw.find(histFig.race).caste[histFig.caste]
+        local name = creature.caste_name[0]
+        if histFig.died_year >= -1 then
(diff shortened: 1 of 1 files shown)


myk002 (commented):
before I look at the code, this needs docs.

the file would go in `docs/add-to-core-party.rst`, but I'm concerned that we're creating too many similar adventurer party tools that would be better combined into one. would such a combined tool make sense? what would be the API?

Crystalwarrior:
> before I look at the code, this needs docs.
> 
> the file would go in `docs/add-to-core-party.rst`, but I'm concerned that we're creating too many similar adventurer party tools that would be better combined into one. would such a combined tool make sense? what would be the API?

I think it would make sense especially for scripts that only work in adventure and not fort mode. But I dunno how to organize that yet

add-to-core-party.lua
```diff
@@ -0,0 +1,65 @@
...
+        end
+        local sym = df.pronoun_type.attrs[creature.sex].symbol
+        if sym then
+            name = name .. ' (' .. sym .. ')'
+        end
+        if histFig.name.has_name then
+            name = dfhack.TranslateName(histFig.name) ..
+                " - (" .. dfhack.TranslateName(histFig.name, true) .. ") - " .. name
+        end
+        table.insert(choices, { text = name, nemesis = nemesis, search_key = name:lower() })
```
myk002, line 57: if these histfigs always have a unit (do they?) then you can do
```lua
local hf = df.historical_figure.find(histfig_id)
if not hf then goto continue end
local nemesis, unit = df.nemesis_record.find(hf.nemesis_id), df.unit.find(hf.unit_id)
if not nemesis or not unit then goto continue end
local name = dfhack.units.getReadableName(unit)
table.insert(choices, {text=name, nemesis=nemesis, search_key=dfhack.toSearchNormalized(name)})
::continue::
```

add-to-core-party.lua
```diff
@@ -0,0 +1,65 @@
...
+    -- research leads me to believe the data is stored per-unit or unit ID, need to figure out
+    -- where that data is stored exactly. Might be one of the unknown variables?
+    party.party_core_members:insert('#', nemesis.figure.id)
+    local extra_member_idx, _ = utils.linear_index(party.party_extra_members, nemesis.figure.id)
+    if extra_member_idx then
+        party.party_extra_members:erase(extra_member_idx)
+    end
+    -- Adds them to unretire list
+    nemesis.flags.ADVENTURER = true
+    nemesis.flags.ACTIVE_ADVENTURER = true
```
myk002, line 21: is just adding them to the core party enough to turn `ACTIVE_ADVENTURER` to `true`? or is that just for the one that you're actively controlling?

add-to-core-party.lua
Crystalwarrior: if they don't have a unit, then adding that dead/unitless person to your core party would actually achieve nothing but trouble

add-to-core-party.lua
Crystalwarrior: this is an oversight yah

myk002 (commented):
this can now be moved to the `advtools` script as a sub command

docs/advtools.rst
```diff
@@ -3,25 +3,27 @@ advtools
...
-    Add the selected "extra" party member to your core party.
+``advtools party``
+    Shows a dialog prompt to promote your extra party members to your core (controllable) party.
 
-``party`` Options
+``party`` Command
 -----------------
 
-TBD
+When you run this command, it will show you a list of your extra party and let you choose
```
myk002, line 24:
```suggestion
When you run this command, you will get a list of your extra party members and can choose
```

myk002 (approved):

myk002:
pre-commit.ci autofix

PR #1167, opened 2024-06-10
adapt trackstop overlay to df-structures/pull/783

chdoc:


myk002 (approved):
thank you for getting this. I'm not sure how I missed it when I updated the rest of the file.

PR #1172, opened 2024-06-12
add the Tracks gui, hotkey friendly

Crystalwarrior:
Support for showing directions, hiding your tracks and hiding other tracks

### DIFF
--- a/gui/tracks.lua
+++ b/gui/tracks.lua
@@ -1,0 +1,180 @@
+local repeatUtil = require('repeat-util')
+
+local utils = require('utils')
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local guidm = require('gui.dwarfmode')
+local world_map = df.global.world.map
+local adventure = df.global.adventure
+
+local to_pen = dfhack.pen.parse
+local others_track = to_pen{
+    ch='+',
+    fg=COLOR_YELLOW,
+    tile=dfhack.screen.findGraphicsTile('CURSORS', 5, 23),
+}
+
+local yours_track = to_pen{
+    ch='+',
+    fg=COLOR_CYAN,
+    tile=dfhack.screen.findGraphicsTile('CURSORS', 5, 22),
+}
+
+--tile=dfhack.screen.findGraphicsTile('INTERFACE_BITS_BUILDING_PLACEMENT', 1, 0)
+local pen_direction = {
+    to_pen{ch='v',fg=COLOR_YELLOW,}, -- tile=dfhack.screen.findGraphicsTile('ROLLERS', 8, 2),},     -- 0 South
+    to_pen{ch='>',fg=COLOR_YELLOW,}, -- tile=dfhack.screen.findGraphicsTile('ROLLERS', 7, 2),},     -- 1 East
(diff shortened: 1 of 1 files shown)


Crystalwarrior:
for `dfhack.keybindings.init` I guess it could be
`keybinding add Ctrl-U@dungeonmode/Default gui/tracks`

gui/tracks.lua
```diff
@@ -0,0 +1,180 @@
...
+                            table.insert(visible_tracks, {x=x, y=y, z=z, flags=flags})
+                        end
+                    end
+                end
+            end
+        end
+    end
+end
+
+TestOverlay = defclass(TestOverlay, widgets.Window)
```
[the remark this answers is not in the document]
myk002: in addition, it's not an overlay. The common pattern is to name the screen FooScreen and the Window Foo. E.g. `TracksScreen` and `Tracks`

gui/tracks.lua
```diff
@@ -0,0 +1,180 @@
...
+    pass_mouse_clicks=true,
+}
+
+function TestScreen:init()
+    local window = TestOverlay{}
+    self:addviews{
+        window,
+    }
+    onTimer()
+    repeatUtil.scheduleEvery(timerId, 1, 'ticks', onTimer)
```
myk002, line 160: I suggest not using `repeatUtil` here. That makes cleanup messy. Instead, update in your own `render` method when the tick counter has incremented.

gui/tracks.lua
```diff
@@ -0,0 +1,180 @@
...
+                end
+            end
+        end
+    end
+end
+
+TestOverlay = defclass(TestOverlay, widgets.Window)
+TestOverlay.ATTRS {
+    default_pos={x=20, y=0},
+    default_enabled=true,
```
myk002, line 67: these two aren't Window attributes and can be removed

gui/tracks.lua
```diff
@@ -0,0 +1,180 @@
...
+    frame_title='Test Overlay',
+    frame={b = 4, r = 4, w = 50, h = 12},
+    viewscreens= {
+        'dungeonmode/Default',
+    },
```
myk002, line 72: also can be removed

gui/tracks.lua
```diff
@@ -0,0 +1,180 @@
...
+    local adv_x, adv_y, adv_z = pos.x, pos.y, pos.z
+    return ("Your position: %s, %s, %s"):format(
+        adv_x, adv_y, adv_z
+    )
+end
+
+function TestOverlay:init()
+    self.directional = false
+    self.show_yours = true
+    self.show_others = true
```
myk002, line 86: you don't need to copy widget state into your object. you can access the widget state directly. Then you also don't need to have `on_change` handlers

myk002 (commented):
I think this will need a few more rounds of review, but first it needs docs so I know what I'm looking at and its intended use case

Bumber64:
Looks like adventure beta 23 finally added tracking to Steam version? Do any changes need to be made to account for this?

Crystalwarrior:
Nope! This is safe to close now that steam version added tracking properly.

PR #1177, opened 2024-06-14
Fix sleeping army controllers in Adventure Mode not responding to intruders

Crystalwarrior:
Very old script cleaned up for version 51


fix/sleepers.lua
```diff
@@ -0,0 +1,21 @@
+-- Number of fixed army controller(s) that may be shared by multiple units
+local num_fixed = 0
+
+-- Loop through all the active units currently loaded
+for k, unit in pairs(df.global.world.units.active) do
```
myk002, line 5: is a camp a site? does that mean that the fix is only relevant in sites? can we trigger this script to run when a site is loaded? that is, can we make this script enableable and watch for state change events, automatically running when the SC_MAP_LOADED event fires and there is an active site (and we're in adventure mode)?

myk002 (commented):
could you also remove docs/adv-fix-sleepers.rst and add a tombstone to docs/about/Removed.rst (see https://github.com/DFHack/dfhack/pull/4714 for an example)

fix/sleepers.lua
Crystalwarrior: There are camps that are created for travelling armies as well (for example, a unit stopped in the middle of nowhere to create a cloth block tent and is sleeping)
Yes, good idea! We can make this trigger either A) instantly react to adventurer, B) react to adventurer who isn't sneaking and present in close vicinity to the target, or C) React to receiving Pain (not bloodloss/painless attacks tho, so you can still rob someone by wrestling their items off)

Crystalwarrior: I think we should keep this manual for now until we decide on when the "fix" should apply, let's study the general sleeping behavior in adv mode and how people react to being attacked when asleep.

changelog.txt
```diff
@@ -35,6 +35,7 @@ Template for new versions:
 - `gui/tiletypes`: interface for modifying map tiles and tile properties
 - `fix/population-cap`: fixes the situation where you continue to get migrant waves even when you are above your configured population cap
 - `fix/occupancy`: fixes issues where you can't build somewhere because the game tells you an item/unit/building is in the way but there's nothing there
+- `fix/sleepers`: (reinstated) fixes sleeping units belonging to a camp that never wake up.
```
myk002, line 38: Please also add `adv-fix-sleepers` to the Removed section below

fix/sleepers.lua
myk002, line 5:
```suggestion
for _, unit in ipairs(df.global.world.units.active) do
```

fix/sleepers.lua
```diff
@@ -0,0 +1,21 @@
...
+            army_controller.data.goal_camp.camp_flag.ALARM_INTRUDER = true
+            num_fixed = num_fixed + 1
+        end
+    end
+end
+
+if num_fixed == 0 then
+    print ("No sleepers with the fixable bug were found, sorry.")
+else
+    print ("Fixed " .. num_fixed .. " bugged army_controller(s).")
```
myk002, line 20:
```suggestion
    print ("Fixed " .. num_fixed .. " group(s) of sleeping units.")
```

changelog.txt
```diff
@@ -74,6 +75,7 @@ Template for new versions:
 - `max-wave`: merged into `pop-control`
 - `devel/find-offsets`, `devel/find-twbt`, `devel/prepare-save`: remove development scripts that are no longer useful
 - `fix/item-occupancy`, `fix/tile-occupancy`: merged into `fix/occupancy`
+- `adv-fix-sleepers`: renamed into fix/sleepers
```
myk002, line 78:
```suggestion
- `adv-fix-sleepers`: renamed to `fix/sleepers`
```

myk002 (approved):

PR #1182, opened 2024-06-14
Adv-Rumors additional conversation options

Crystalwarrior:
Adds "ask whereabouts of" conversation options for ALL your histfig relationships


myk002:
is this PR still a draft or is it ready for review?

Crystalwarrior:
@myk002 ready for review

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+    choice.print_string.text:insert("#", text)
+
+    if keywords ~= nil then
+        addKeywords(choice, keywords)
+    end
+    return choice
+end
+
+local function addHistFigWhereaboutsChoice(profile)
+    for i, c in pairs(adventure.conversation.conv_choice_info) do
```
myk002, line 80:
```suggestion
    for i, c in ipairs(adventure.conversation.conv_choice_info) do
```
```suggestion
    for _, c in pairs(adventure.conversation.conv_choice_info) do
```

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+        if c.cc.type == df.talk_choice_type.AskWhereabouts and
+          c.cc.invocation_target_hfid ~= -1 and
```
myk002, line 82: if `profile.histfig_id` is never equal to -1, then  you can remove the check for -1

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+    end
+
+    local histfig = df.historical_figure.find(profile.histfig_id)
+    local name = ""
+    local creature = df.creature_raw.find(histfig.race)
+    if creature then
+        local caste = creature.caste[histfig.caste]
+        name = caste.caste_name[0]
+    end
+    local title = "Ask for the whereabouts of the " .. name .. " " .. dfhack.TranslateName(histfig.name, true)
```
myk002, line 96: do we only need the english name here? are they never referenced by their native name?

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+    if profile._type == df.relationship_profile_hf_historicalst then
+        title = title .. " (Heard of)"
+    end
+    local choice = new_choice(df.talk_choice_type.AskWhereabouts, title, dfhack.TranslateName(histfig.name):split())
```
myk002, line 100: why are we adding keywords for the untranslated name if the translated name appears in the text itself?

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+     -- insert before the last choice, which is usually "back"
+    adventure.conversation.conv_choice_info:insert(#adventure.conversation.conv_choice_info-1, choice)
+    choice.cc.invocation_target_hfid = histfig.id
+end
+
+local function addIdentityWhereaboutsChoice(identity)
+    for i, c in pairs(adventure.conversation.conv_choice_info) do
```
myk002, line 107:
```suggestion
    for _, c in ipairs(adventure.conversation.conv_choice_info) do
```
```suggestion
    for i, c in pairs(adventure.conversation.conv_choice_info) do
```

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+        if c.cc.type == df.talk_choice_type.AskWhereabouts and
+           c.cc.invocation_target_hfid ~= -1 and
```
myk002, line 109: same as above -- if `identity.impersonated_hf` is never -1, then we don't need this check

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+            local caste = creature.caste[histfig.caste]
+            name = caste.caste_name[0]
+        end
+    end
+    local title = "Ask for the whereabouts of the " .. name .. " " .. dfhack.TranslateName(identity_name, true)
+    local choice = new_choice(df.talk_choice_type.AskWhereabouts, title)
+    -- insert before the last choice, which is usually "back"
+    adventure.conversation.conv_choice_info:insert(#adventure.conversation.conv_choice_info-1, choice, dfhack.TranslateName(identity_name):split())
+    choice.cc.invocation_target_hfid = identity.impersonated_hf
+end
```
myk002, line 135: this duplicates too much of the logic in `addHistFigWhereaboutsChoice`; please refactor

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+        -- This option will likely always fail unless the false identity is impersonating someone
+        -- but giving away the false identity's true historical figure feels cheap.
+        for _, profile in pairs(identity) do
+            addIdentityWhereaboutsChoice(df.identity.find(profile.identity_id))
+        end
+
+        -- Historical entities go last so as to not give away fake identities
+        for _, profile in pairs(historical) do
+            addHistFigWhereaboutsChoice(profile)
+        end
```
myk002, line 165: with the loop through `adventure.conversation.conv_choice_info`, each of these has quadradic runtime complexity. perhaps it would be a good idea to make a single loop through `adventure.conversation.conv_choice_info` and cache the results? (and update the cache as you add more entries)

myk002: also, `ipairs` is semantically appropriate for vectors, not `pairs`

internal/advtools/convo.lua
```diff
@@ -61,8 +61,111 @@ local function generateKeywordsForChoice(choice)
...
+    end
+
+    -- generate extra keywords for all options
```
myk002, line 168: if this is being run unconditionally down here, why are you adding keywords in `new_choice`? can the call to `add_keywords` be removed from `new_choice`?

myk002 (commented):
previous comments appear to be unaddressed

internal/advtools/convo.lua
Crystalwarrior: It's possible for identity to be -1

internal/advtools/convo.lua
Crystalwarrior: this is because new_choice might have its own "vanilla"-style added keywords which is just a few, while this section ensures more searchability

internal/advtools/convo.lua
Crystalwarrior: In vanilla convo system yeah they're only referenced by their English name

internal/advtools/convo.lua
Crystalwarrior: Vanilla keyword behavior

internal/advtools/convo.lua
myk002: is that what you *want*, though? @Ozzatron what do you think?

internal/advtools/convo.lua
myk002: and the translated name is already in the keyword list?

changelog.txt
```diff
@@ -44,6 +44,7 @@ Template for new versions:
 - `gui/notify`: new notification type: injured citizens; click to zoom to injured units; also displays a warning if your hospital is not functional (or if you have no hospital)
 - `prioritize`: new info panel on under-construction buildings showing if the construction job has been taken and by whom. click to zoom to builder; toggle high priority status for job if it's not yet taken and you need it to be built ASAP
 - `gui/pathable`: new "Depot" mode that shows whether wagons can path to your trade depot
+- `advtools`: convo - add a new conversation option to "ask whereabouts of" for all your relationships (before, you could only ask whereabouts of people involved in rumors only)
```
myk002, line 47:
```suggestion
- `advtools`: automatically add a conversation option to "ask whereabouts of" for all your relationships (before, you could only ask whereabouts of people involved in rumors)
```

myk002 (approved):

PR #1188, opened 2024-06-15
gui/armies - Script to display armies on the world map

Crystalwarrior:
This is a missing feature from 47 brought back for 51 beta.
Depends on https://github.com/DFHack/dfhack/pull/4767

Might be a very short-lived script depending on how long adv-beta takes lol
more useful as an example of the PR working


Crystalwarrior:
No longer relevant. A separate script could be created to obtain *army data*, instead.

PR #1192, opened 2024-06-18
Feature/all-unmet-needs (only list)

rslgp:


myk002 (commented):
The `allneeds` script already sorts by focus impact. Moreover, I'm not keen on adding coded nicknames like this. I'd rather either implement bulk nicknaming as part of the in-progress `gui/manipulator` or extend autolabor to set labors automatically so units attend to their specific needs

myk002:
How about if you instead turn this into a notification that appears on the notification panel? It can list at most the top two or three needs of the fort (as long as they are above some threshold). Maybe also report on how focused the fort is overall. 

E.g.

Fort is 92% focused
Fort is 79% focused; top need: Pray or meditate
Fort is 45% focused; top needs: Relax, Martial training

Clicking on the message would zoom to the dwarves who have especially high needs

rslgp:
i removed the renaming part from code, that script is a different approach from allneeds, focused only on unmet filtering out the others needs

myk002:
If it's just a filter, then can this be implemented as an option for `allneeds`? It doesn't seem worth adding an entirely new tool for a minor variation in behavior.

myk002:
closing this PR for now; please reopen when you'd like to take it further. I think the next steps are to integrate the functionality into `allneeds` as an option.

PR #1199, opened 2024-06-20
Fix ordering, hide values, and block invalid combos

tatoyoda600:
Fixes the "More Options" menu to have consistent ordering.
Hides values that have a greater level of complexity
Block some invalid combinations that were previously allowed


myk002 (commented):
could you also align the order of the More options options in `docs/gui/tiletypes.rst`? I ordered them by what I saw in the UI at the time

myk002 (approved):

PR #1203, opened 2024-06-22
Fix clear-smoke memory leaks

Crystalwarrior:
Delete the instances of smoke that are already erased from the `flows` vectors

thanks @Bumber64 for the help!


Crystalwarrior:
Could potentially expand this script to be  general-purpose `clear-flows` at a later date

clear-smoke.lua
```diff
@@ -1,25 +1,28 @@
...
+local flows_to_delete = {}
 
 function clearSmoke(flows)
+    local success
     for i = #flows - 1, 0, -1 do
-        if flows[i].type == df.flow_type.Smoke then
+        local flow = flows[i]
+        if flow.type == df.flow_type.Smoke then
             flows:erase(i)
+            flows_to_delete[flow] = true
```
myk002, line 9: why can't this be deleted inline? other tools successfully do it that way, e.g. https://github.com/DFHack/dfhack/blob/develop/plugins/lua/dwarfvet.lua#L129-L137

Bumber64:
Given Discord discussion, the script should now be something like this:
```lua
function removeFlow(flow)
    flow.flags.DEAD = true

    local block = dfhack.maps.getTileBlock(flow.pos)
    if block then
        block.flow_pool.flags.active = true
    else
        df.global.world.orphaned_flow_pool.flags.active = true
    end
end

function removeFlows(flow_type)
    local count = 0
    for _,flow in ipairs(df.global.flows) do
        if not flow.flags.DEAD and flow.type == flow_type then
            removeFlow(flow)
            count = count + 1
        end
    end

    return count
end

function clearSmoke()
    if dfhack.isWorldLoaded() then
        print(('%d smoke flows removed.'):format(removeFlows(df.flow_type.Smoke)))
    else
        qerror('World not loaded!')
    end
end

clearSmoke()
```
I think we could probably turn this into a module to allow the `removeFlow` functions to be reused elsewhere. Otherwise, we could just implement those in `Maps` and add them to the Lua API.

clear-smoke.lua
```diff
@@ -1,25 +1,37 @@
...
-    clearSmoke(block.flows)
+function clearSmoke()
+    if dfhack.isWorldLoaded() then
+        print(('%d smoke flows removed.'):format(removeFlows(df.flow_type.Smoke)))
+    else
+        qerror('World not loaded!')
+    end
 end
+
+clearSmoke()
```
Bumber64, line 37: Should be
```lua
    clearSmoke()
end
```
so it doesn't run when used as a module.

myk002:
pre-commit.ci autofix

clear-smoke.lua
myk002, line 37:
```suggestion
if not dfhack_flags.module then
    clearSmoke()
end
```

myk002 (approved):

PR #1207, opened 2024-06-26
Fix gui/notify up for adventure mode

Crystalwarrior:
Fix gui/notify eating the Enter input for adventure mode
Fix wildlife showing up as a notification even though they're part of your party


internal/notify/notifications.lua
```diff
@@ -159,6 +159,8 @@ local function for_wildlife_adv(fn, reverse)
             not unit.flags1.caged and
             not unit.flags1.chained and
             not dfhack.units.isHidden(unit) and
+            unit.relationship_ids.GroupLeader ~= dfhack.world.getAdventurer().id and
```
myk002, line 162: you can factor out the `dfhack.world.getAdventurer().id` by putting a
`local adv_id = dfhack.world.getAdventurer().id`
above the `for_iter` call.
notifications take a significant amount of processing time, so we need to be careful about inefficiencies

gui/notify.lua
```diff
@@ -21,6 +21,22 @@ NotifyOverlay.ATTRS{
...
+            if not choice.data.on_click then return end
+            local prev_state = self.state[choice.data.name]
+            self.state[choice.data.name] = choice.data.on_click(prev_state, false)
+        end
+        on_submit2 = function(_, choice)
+            if not choice.data.on_click then return end
+            local prev_state = self.state[choice.data.name]
+            self.state[choice.data.name] = choice.data.on_click(prev_state, true)
+        end
+    end
```
myk002, line 38: there are two problems here:
1. it will not have the intended effect for adventure mode since `init` runs at program start (when `dfhack.world.isAdventureMode()` is false
2. it removes click to zoom functionality in addition to enter handling

instead, you can leave `init` unchanged and add a `NotifyOverlay:onInput` function that intercepts the enter (`keys.SELECT`) key and returns false before calling `return NotifyOverlay.super.onInput(self, keys)`

I'd be ok with ignoring SELECT regardless of mode so the fort mode panel behaves the same as the adventure mode panel.

gui/notify.lua
```diff
@@ -53,6 +54,14 @@ function NotifyOverlay:init()
     }
 end
 
+function NotifyOverlay:onInput(keys)
+    if keys.SELECT then return false end
+
+    if NotifyOverlay.super.onInput(self, keys) then
+        return true
+    end
```
myk002, line 62:
```suggestion
    return NotifyOverlay.super.onInput(self, keys)
```

gui/notify.lua
```diff
@@ -32,13 +32,14 @@ function NotifyOverlay:init()
                     frame={t=0, b=0, l=0, r=0},
                     -- disable scrolling with the keyboard since some people
                     -- have wasd mapped to the arrow keys
+                    -- don't do this for adventure mode to not eat inputs
```
myk002, line 35:
```suggestion
```

gui/notify.lua
```diff
@@ -32,13 +32,14 @@ function NotifyOverlay:init()
...
                     scroll_keys={},
-                    on_submit=function(_, choice)
+                    on_submit = function(_, choice)
```
myk002, line 37:
```suggestion
                    on_submit=function(_, choice)
```

gui/notify.lua
```diff
@@ -32,13 +32,14 @@ function NotifyOverlay:init()
...
                         if not choice.data.on_click then return end
                         local prev_state = self.state[choice.data.name]
                         self.state[choice.data.name] = choice.data.on_click(prev_state, false)
                     end,
-                    on_submit2=function(_, choice)
+                    on_submit2 = function(_, choice)
```
myk002, line 42:
```suggestion
                    on_submit2=function(_, choice)
```

myk002 (commented):
could you also make a note in the changelog under.. I guess under Fixes .. that the notification panel no longer responds to Enter so the Enter key is passed through to the vanilla UI?

myk002:
> could you also make a note in the changelog under.. I guess under Fixes .. that the notification panel no longer responds to Enter so the Enter key is passed through to the vanilla UI?

I'm not thrilled that there is no keyboard control for the notification panel, but I think removing the Enter key handling is absolutely the right thing to do.  maybe we can use Alt-up/down for scrolling and alt-left/right for zooming once those keys are available.

changelog.txt
```diff
@@ -59,6 +59,7 @@ Template for new versions:
 - `quickfort`: allow farm plots to be built on muddy stone (as per vanilla behavior)
 - `suspend`: remove broken ``--onlyblocking`` option; restore functionality to ``suspend all``
 - `gui/create-item`: allow creation of adamantine thread, wool, and yarn
+- `gui/notify`: the notification panel no longer responds to SELECT so the SELECT key is passed through to the vanilla UI
```
myk002, line 62:
```suggestion
- `gui/notify`: the notification panel no longer responds to the Enter key so Enter key is passed through to the vanilla UI
```

myk002 (approved):

PR #1212, opened 2024-06-28
Makeown-plus

Crystalwarrior:
Able to make snatchers your own now
Able to make giants your own to an extent (though their Large Predator  caste flag gets in the way)

### DIFF
--- a/makeown.lua
+++ b/makeown.lua
@@ -2,6 +2,11 @@
 
 local utils = require('utils')
 
+-- List of professions to convert From-To
+local convert_professions = {}
+convert_professions[df.profession.MERCHANT] = df.profession.TRADER
+convert_professions[df.profession.THIEF] = df.profession.STANDARD
+
 local function get_translation(race_id)
     local race_name = df.global.world.raws.creatures.all[race_id].creature_id
     for _,translation in ipairs(df.global.world.raws.language.translations) do
@@ -76,8 +81,41 @@ local function fix_unit(unit)
 
     unit.civ_id = df.global.plotinfo.civ_id;
 
-    if  unit.profession == df.profession.MERCHANT then  unit.profession = df.profession.TRADER end
-    if unit.profession2 == df.profession.MERCHANT then unit.profession2 = df.profession.TRADER end
+    if convert_professions[unit.profession] then unit.profession2 = df.profession.STANDARD end
+    if convert_professions[unit.profession2] then unit.profession2 = df.profession.STANDARD end
+end
+
+local function fix_army(unit)
+    -- Disassociate them from their army controller
+    if unit.enemy.army_controller then
+        if unit.enemy.army_controller.commander_hf == unit.hist_figure_id then
+            unit.enemy.army_controller.commander_hf = -1
+        end
+        unit.enemy.army_controller_id = -1
+        unit.enemy.army_controller = nil
+    end
+
+    if unit.enemy.enemy_status_slot ~= -1 then
+        local status_cache = df.global.world.enemy_status_cache
+        local status_slot = unit.enemy.enemy_status_slot
+
+        unit.enemy.enemy_status_slot = -1
+        status_cache.slot_used[status_slot] = false
+
+        for index, _ in pairs(status_cache.rel_map[status_slot]) do
+            status_cache.rel_map[status_slot][index] = -1
+        end
+
+        for index, _ in pairs(status_cache.rel_map) do
+            status_cache.rel_map[index][status_slot] = -1
+        end
+
+        -- TODO: what if there were status slots taken above status_slot?
+        -- does everything need to be moved down by one?
+        if status_cache.next_slot > status_slot then
+            status_cache.next_slot = status_slot
+        end
+    end
 end
 
 local function add_to_entity(hf, eid)
@@ -205,6 +243,7 @@ function make_own(unit)
     dfhack.units.makeown(unit)
 
     fix_unit(unit)
+    fix_army(unit)
     fix_histfig(unit)
     fix_clothing_ownership(unit)
 
--- a/makeown.lua
+++ b/makeown.lua
@@ -231,6 +231,14 @@ local function fix_histfig(unit)
             entity_link(hf, eid, true, false, k)
             ::continue::
         end
+        -- If you're makeown-ing an enemy of your civilization or group, people will feel vengeful without this
+        if df.histfig_entity_link_enemyst:is_instance(el) then
+            local eid = el.entity_id
+            if eid == civ_id or eid == group_id then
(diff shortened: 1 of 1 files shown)


makeown.lua
```diff
@@ -2,6 +2,11 @@
 
 local utils = require('utils')
 
+-- List of professions to convert From-To
+local convert_professions = {}
+convert_professions[df.profession.MERCHANT] = df.profession.TRADER
+convert_professions[df.profession.THIEF] = df.profession.STANDARD
```
myk002, line 8:
```suggestion
local convert_professions = {
    [df.profession.MERCHANT]=df.profession.TRADER,
    [df.profession.THIEF]=df.profession.STANDARD,
}
```

makeown.lua
```diff
@@ -76,8 +81,41 @@ local function fix_unit(unit)
 
     unit.civ_id = df.global.plotinfo.civ_id;
 
-    if  unit.profession == df.profession.MERCHANT then  unit.profession = df.profession.TRADER end
-    if unit.profession2 == df.profession.MERCHANT then unit.profession2 = df.profession.TRADER end
+    if convert_professions[unit.profession] then unit.profession2 = df.profession.STANDARD end
+    if convert_professions[unit.profession2] then unit.profession2 = df.profession.STANDARD end
```
myk002, line 85:
```suggestion
    if convert_professions[unit.profession] then unit.profession = convert_professions[unit.profession] end
    if convert_professions[unit.profession2] then unit.profession2 = convert_professions[unit.profession2] end
```

makeown.lua
```diff
@@ -215,6 +262,8 @@ function make_own(unit)
     else
         unit.flags1.tame = true
         unit.training_level = df.animal_training_level.Domesticated
+        -- No more "vengeful" spam when civilizing Giants!
+        unit.enemy.caste_flags.LARGE_PREDATOR = false;
```
myk002, line 266: does this "stick"? I thought `unit.enemy.caste_flags` gets overwritten with actual race caste flags (plus syndrome flags) periodically

makeown.lua
```diff
@@ -76,8 +81,41 @@ local function fix_unit(unit)
...
 
-    if  unit.profession == df.profession.MERCHANT then  unit.profession = df.profession.TRADER end
-    if unit.profession2 == df.profession.MERCHANT then unit.profession2 = df.profession.TRADER end
+    if convert_professions[unit.profession] then unit.profession2 = df.profession.STANDARD end
+    if convert_professions[unit.profession2] then unit.profession2 = df.profession.STANDARD end
+end
+
+local function fix_army(unit)
+    -- Disassociate them from their army controller
+    if unit.enemy.army_controller then
```
myk002, line 90: is there a membership list in the army itself that we need to scrub?

makeown.lua
```diff
@@ -76,8 +81,41 @@ local function fix_unit(unit)
...
+        for index, _ in pairs(status_cache.rel_map) do
+            status_cache.rel_map[index][status_slot] = -1
+        end
+
+        -- TODO: what if there were status slots taken above status_slot?
+        -- does everything need to be moved down by one?
+        if status_cache.next_slot > status_slot then
+            status_cache.next_slot = status_slot
+        end
+    end
```
myk002, line 118: evil copy and paste! Please make this a function and modify `fix/loyaltycascade` to call it. (and add a comment that it is also used by `fix/loyaltycascade`)

myk002:
also needs changelog entry

makeown.lua
Crystalwarrior: yah it doesn't stick, might as well remove it

makeown.lua
Crystalwarrior: Not sure? I don't think so but needs more research

myk002:
you can set vscode to remove trailing space on save. That might help you pass the pre-commit.ci check with more consistency.

makeown.lua
```diff
@@ -76,8 +82,45 @@ local function fix_unit(unit)
...
+            unit.enemy.army_controller.commander_hf = -1
+        end
+        unit.enemy.army_controller_id = -1
+        unit.enemy.army_controller = nil
+    end
+    fixUnitEnemyStatus(unit)
+end
+
+-- Also used by the script fix/loyaltycascade
+function fixUnitEnemyStatus(unit)
```
myk002, line 102: "fix" is subjective. I think this should be named after what it actually does: `clearUnitEnemyStatus`

makeown.lua
myk002: could you add a TODO so when it goes wrong, we remember where to start looking?

changelog.txt
```diff
@@ -48,6 +48,7 @@ Template for new versions:
 - `gui/pathable`: new "Depot" mode that shows whether wagons can path to your trade depot
 - `advtools`: automatically add a conversation option to "ask whereabouts of" for all your relationships (before, you could only ask whereabouts of people involved in rumors)
 - `gui/design`: all-new visually-driven UI for much improved usability
+- `makeown`: goblin snatchers and giants can now be makeown-ed more reliably
```
myk002, line 51: needs to get moved to changelog for current version. also, should likely go in `Fixes`, and say something like "Fix enemy creatures continuing to be hostile after `makeown`"

myk002 (commented):
LGTM other than the changelog

makeown.lua
```diff
@@ -279,6 +279,14 @@ local function fix_histfig(unit)
             entity_link(hf, eid, true, false, k)
             ::continue::
         end
+        -- If you're makeown-ing an enemy of your civilization or group, people will feel vengeful without this
+        if df.histfig_entity_link_enemyst:is_instance(el) then
```
myk002, line 283: this is unsafe. `el` can be deleted by this point

makeown.lua
```diff
@@ -280,7 +280,7 @@ local function fix_histfig(unit)
             ::continue::
         end
         -- If you're makeown-ing an enemy of your civilization or group, people will feel vengeful without this
-        if df.histfig_entity_link_enemyst:is_instance(el) then
+        if el and df.histfig_entity_link_enemyst:is_instance(el) then
```
myk002, line 283: I mean `el` can be a stale reference. the call to `entity_link` above can delete the memory that `el` points to.

myk002: You need to reorganize this loop such that  you guarantee nothing touches `el` after a call to `entity_link`.

The API of `entity_link` doesn't make it clear at all that there is a possible side effect of deleting the memory that `el` points to, so a "proper" fix would be to rearchitect this section of the code, but for the purposes of this PR, just not crashing in the case where `entity_link` deletes `el` would be enough.

PR #1217, opened 2024-06-30
Fixes shooting in adventure mode

Crystalwarrior:
Will probably be patched out in the next adv-beta build lmao


fix/shooting.lua
```diff
@@ -0,0 +1,34 @@
+-- Fixes shooting/throwing options stacking on each other, causing null pointers and crashes
+--@ enable = true
```
myk002, line 2: this script does not implement the enabled API, so this marker won't do anything.
relevant docs here: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#enabling-and-disabling-scripts

fix/shooting.lua
```diff
@@ -0,0 +1,34 @@
...
+--@ module = true
+
+local aim_projectile = df.global.game.main_interface.adventure.aim_projectile
+
+local overlay = require('plugins.overlay')
+
+FixShootingOverlay = defclass(FixShootingOverlay, overlay.OverlayWidget)
+FixShootingOverlay.ATTRS{
+    desc='Adds keywords to conversation entries.',
```
myk002, line 11: copypasta

fix/shooting.lua
```diff
@@ -0,0 +1,34 @@
...
+    default_enabled=true,
+    viewscreens={
+        'dungeonmode/Inventory',
+        'dungeonmode/AimProjectile'
+    },
+    frame={w=0, h=0},
+    overlay_onupdate_max_freq_seconds=0,
```
myk002, line 18: do you really need to run it this frequently, or just in response to input?

fix/shooting.lua
```diff
@@ -0,0 +1,34 @@
...
+function FixShootingOverlay:overlay_onupdate()
+    if aim_projectile.open == false then
+        aim_projectile.shooter_it = nil
+        aim_projectile.ammo_it = nil
+        aim_projectile.thrown_it = nil
+        aim_projectile.shooting = false
+    elseif aim_projectile.shooting then
+        aim_projectile.thrown_it = nil
+    end
+end
```
myk002, line 30: can this be replaced with
```lua
function FixShootingOverlay:onInput()
    if aim_projectile.open == false then
        aim_projectile.shooter_it = nil
        aim_projectile.ammo_it = nil
        aim_projectile.thrown_it = nil
        aim_projectile.shooting = false
    elseif aim_projectile.shooting then
        aim_projectile.thrown_it = nil
    end
end
```
that would run just before a shooting action since it would run in response to both keystrokes and mouse clicks

internal/control-panel/registry.lua
```diff
@@ -95,6 +95,8 @@ COMMANDS_BY_IDX = {
     {command='preserve-tombs', group='bugfix', mode='enable', default=true},
     {command='reaction-gloves', help_command='tweak', group='bugfix', mode='tweak', default=true,
         desc='Fixes reactions not producing gloves in sets with correct handedness.'},
+    {command='fix/shooting', group='bugfix', mode='tweak', default=true,
```
myk002, line 98: this is neither a command nor a tweak. you can remove it from the control-panel registry entirely since it will be controlled by the overlay framework

myk002 (commented):
this overlay would be better off attached to `advtools`

advtools.lua
```diff
@@ -1,9 +1,11 @@
 --@ module=true
 
 local convo = reqscript('internal/advtools/convo')
+local shooting = reqscript('internal/advtools/shooting')
 
 OVERLAY_WIDGETS = {
     conversation=convo.AdvRumorsOverlay,
+    shooting=shooting.FixShootingOverlay,
```
myk002, line 8: I'd use `fix_shooting` for the key so people know it's a fix

internal/advtools/shooting.lua
```diff
@@ -0,0 +1,28 @@
...
+    desc='Fixes Shooting your weapon breaking your Throw action.',
+    default_enabled=true,
+    viewscreens={
+        'dungeonmode/Inventory',
+        'dungeonmode/AimProjectile'
+    },
+    frame={w=0, h=0},
+}
+
+function FixShootingOverlay:onInput()
```
myk002, line 19: did you test that this works as expected? I *think* it should, but it definitely needs testing

myk002 (commented):
please add a changelog entry (New Features)

internal/advtools/shooting.lua
Crystalwarrior: Yep, confirmed

changelog.txt
```diff
@@ -28,6 +28,7 @@ Template for new versions:
 
 ## New Tools
 - `advtools`: collection of useful commands and overlays for adventure mode
+- `advtools`: added a fix for the shooting bug that corrupts throwing/shooting and can cause crashes between save loads
```
myk002, line 31:
```suggestion
- `advtools`: added an overlay that automatically fixes corrupt throwing/shooting state, preventing save/load crashes
```

myk002 (approved):

PR #1226, opened 2024-07-03
Gui/journal line wrap fix

wiktor-obrebski:
# Description
This MR introduce redesign of coordinate system and suite of tests for journal gui script.

Basing on issue found by @myk002 (like this: https://github.com/DFHack/scripts/pull/1208#discussion_r1661408870) I realized that my current coordination system do not fit well the editor needs and long term plans.

Here, I introduced dedicated class for managing `Wrapped lines` text and migrate to offset cursor and selection (where before it was {x, y} cursor and selection).

This MR do not brings nay new features to the journal script, its intention is preparation for future work and bugfixes.
Also, I improved journal docs.

# Known issue
I added tests, but `simulateInput` do not support `_MOUSE_M_DOWN` event which is crucial for mouse drag (selection).
This MR fix it but I am not sure if it not brings some unintended consequences.

https://github.com/DFHack/dfhack/pull/4797

### DIFF
--- a/docs/gui/journal.rst
+++ b/docs/gui/journal.rst
@@ -20,22 +20,31 @@ Supported Features
 ------------------
 
 - Cursor Control: Navigate through text using arrow keys (left, right, up, down) for precise cursor placement.
-- Fast Rewind: Use 'Shift+Left'/'Ctrl+B' and 'Shift+Right'/'Ctrl+F'to move the cursor one word back or forward
+- Fast Rewind: Use :kbd:`Shift+Left`/:kbd:`Ctrl+B` and :kbd:`Shift+Right`/:kbd:`Ctrl+F`to move the cursor one word back or forward
 - Longest X Position Memory: The cursor remembers the longest x position when moving up or down, making vertical navigation more intuitive.
 - Mouse Control: Use the mouse to position the cursor within the text, providing an alternative to keyboard navigation.
 - New Lines: Easily insert new lines using the :kbd:`Enter` key, supporting multiline text input.
 - Text Wrapping: Text automatically wraps within the editor, ensuring lines fit within the display without manual adjustments.
 - Backspace Support: Use the backspace key to delete characters to the left of the cursor.
-- Delete Character: 'Ctrl+D' deletes the character under the cursor.
-- Line Navigation: 'Ctrl+H' (like "Home") moves the cursor to the beginning of the current line, and 'Ctrl+E' (like "End") moves it to the end.
-- Delete Current Line: 'Ctrl+U' deletes the entire current line where the cursor is located.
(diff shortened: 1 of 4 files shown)


myk002:
As I wrote in the other review, I'm concerned that handling _down and middle mouse click in `simulateInput` will have unintended side effects. Would it be possible to use `view:onInput(keys)` instead of `gui.simulateInput` for the unit tests? example: https://github.com/DFHack/scripts/blob/master/test/gui/blueprint.lua#L300

docs/gui/journal.rst
```diff
@@ -20,22 +20,32 @@ Supported Features
 ------------------
 
 - Cursor Control: Navigate through text using arrow keys (left, right, up, down) for precise cursor placement.
-- Fast Rewind: Use 'Shift+Left'/'Ctrl+B' and 'Shift+Right'/'Ctrl+F'to move the cursor one word back or forward
+- Fast Rewind: Use :kbd:`Shift+Left`/:kbd:`Ctrl+B` and :kbd:`Shift+Right`/:kbd:`Ctrl+F` to move the cursor one word back or forward.
```
myk002, line 23: this will render as `Shift+Left` on one "key". You should probably write this as
```
:kbd:`Shift`:kbd:`Left`
```
etc.

gui/journal.lua
```diff
@@ -761,6 +819,8 @@ function JournalScreen:onDismiss()
     view = nil
 end
 
+view = nil
```
myk002, line 822: this resets `view` on every script invocation or module reload. if you want a top-level declaration for clarity's sake, use `view = view or nil` (but it's not required, and you can just delete this line entirely)

gui/journal.lua
```diff
@@ -1,10 +1,80 @@
...
+--     (wrapped {x, y} coordinates are required).
+--  2. When mouse input needs to be converted to the original text position.
+
+-- Using this class in other scenarios may lead to issues with the component's
+-- behavior when the text is wrapped.
+WrappedText = defclass(WrappedText)
+
+WrappedText.ATTRS{
+    text = '',
+    wrap_width = 0,
```
myk002, line 25: `0` isn't a valid width to pass to `string:wrap`; you might as well use `wrap_width = DEFAULT_NIL,` here

gui/journal.lua
```diff
@@ -98,13 +171,20 @@ function TextEditor:updateScrollbar()
...
     self.editor:render(dc)
     self.scrollbar:render(dc)
 end
 
+function TextEditor:onInput(keys)
+    if (self.scrollbar.is_dragging) then
+        return self.scrollbar:onInput(keys)
+    end
+
+    return self:inputToSubviews(keys)
```
myk002, line 184: is is more common to write this as `return TextEditor.super.onInput(self, keys)`

myk002 (approved):

PR #1232, opened 2024-07-05
[open-legends] don't intercept text

myk002:
Fixes: https://github.com/DFHack/dfhack/issues/4790


myk002:
Slight correction is that it's the ZScreen that gains focus, not the panel itself. Screen focus is a different mechanism from the input focus managed by the View hierarchy (the mechanism used by, for example, widgets.EditField)

PR #1241, opened 2024-07-25
Efficiency increase for ban-cooking and change to 'seeds' option

Droseran:
Increase the efficiency of ban-cooking by breaking the iteration of the material search when it finds the material ID being looked for (specifically 'STRUCTURAL' in this case).

The 'seeds' option now bans cooking for all seed producing plants and growths instead of just items that produce seeds and are brewable. The previous method could cause replantability to be lost for plants with cookable seed producing items, affecting milling in vanilla and other industries in mods. It may be a good idea to split this into two functions in the future, since the documentation only mentions that this function bans cooking seeds and not the items that produce them. This is also the only option that affects two significantly different groups of items, which is a bit inconsistent.

### DIFF
--- a/ban-cooking.lua
+++ b/ban-cooking.lua
@@ -137,17 +137,20 @@ funcs.seeds = function()
         if p.material_defs.type.seed == -1 or p.material_defs.idx.seed == -1 or p.flags.TREE then goto continue end
         ban_cooking(p.name .. ' seeds', p.material_defs.type.seed, p.material_defs.idx.seed, df.item_type.SEEDS, -1)
         for _, m in ipairs(p.material) do
-            if m.id == "STRUCTURAL" and m.flags.EDIBLE_COOKED then
-                local has_drink = false
-                local has_seed = false
-                for _, s in ipairs(m.reaction_product.id) do
-                    has_seed = has_seed or s.value == "SEED_MAT"
-                    has_drink = has_drink or s.value == "DRINK_MAT"
+            if m.id == "STRUCTURAL" then
+                if m.flags.EDIBLE_COOKED then
+                    local has_drink = false
+                    local has_seed = false
+                    for _, s in ipairs(m.reaction_product.id) do
+                        has_seed = has_seed or s.value == "SEED_MAT"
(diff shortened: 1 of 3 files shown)


ban-cooking.lua
```diff
@@ -137,30 +137,29 @@ funcs.seeds = function()
...
+                    local has_seed = false
+                    for _, s in ipairs(m.reaction_product.id) do
+                        has_seed = has_seed or s.value == "SEED_MAT"
+                    end
+                    if has_seed then
+                        local matinfo = dfhack.matinfo.find(p.id, m.id)
+                        ban_cooking(p.name .. ' ' .. m.id, matinfo.type, matinfo.index, df.item_type.PLANT, -1)
+                    end
                 end
+                break
```
myk002, line 151: this assumes that a plant cannot have more than one STRUCTURAL part. is this guaranteed, even in mods?

myk002 (commented):
could you also update the description of the `seeds` option in `docs/ban-cooking.rst`?

ban-cooking.lua
Droseran: Materials IDs (such as STRUCTURAL) have to be unique or items begin flip-flopping between being different materials on save and reload. If a modded plant does have more than one STRUCTURAL material, the save is going to have worse issues than ban-cooking not working on a plant.

changelog.txt
```diff
@@ -93,6 +93,8 @@ Template for new versions:
 - `clear-smoke`: properly tag smoke flows for garbage collection to avoid memory leak
 - `warn-stranded`: don't warn for babies carried by mothers who happen to be gathering fruit from trees
 - `prioritize`: also boost priority of already-claimed jobs when boosting priority of a job type so those jobs are not interrupted
+- `ban-cooking`: stop iterating through plant materials when the searched-for material is found
```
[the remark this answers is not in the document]
Droseran: Alright, I've removed that line from the changelog and updated the documentation file to clarify which items the 'seeds' option affects.

myk002 (approved):

PR #1242, opened 2024-07-28
Fix gui/autodump dumping into walls, etc.

Bumber64:
Prevent `gui/autodump` from attempting to dump into or select within unallocated tiles (e.g., below HFS), which causes a script error. Also prevent dumping into hidden tiles.

Prevent dumping into walls. Set proper projectile flags so falling items don't get destroyed on impact.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -38,7 +38,7 @@ Template for new versions:
 - `prioritize`: fix incorrect loading of persisted data on some OS types
 - `list-waves`: no longer gets confused by units that leave the map and then return (e.g. squads who go out on raids)
 - `fix/dead-units`: fix error when removing dead units from burrows and the unit with the greatest ID was dead
-- `gui/autodump`: fix dumping into invalid tiles, creating item projectiles that get destroyed
+- `gui/autodump`: prevent dumping into walls or invalid map area, set proper projectile flags to prevent items from being destroyed
 
 ## Misc Improvements
 - `build-now`: if `suspendmanager` is running, run an unsuspend cycle immediately before scanning for buildings to build
--- a/changelog.txt
+++ b/changelog.txt
@@ -38,6 +38,7 @@ Template for new versions:
 - `prioritize`: fix incorrect loading of persisted data on some OS types
 - `list-waves`: no longer gets confused by units that leave the map and then return (e.g. squads who go out on raids)
 - `fix/dead-units`: fix error when removing dead units from burrows and the unit with the greatest ID was dead
+- `gui/autodump`: fix dumping into invalid tiles, creating item projectiles that get destroyed
 
 ## Misc Improvements
 - `build-now`: if `suspendmanager` is running, run an unsuspend cycle immediately before scanning for buildings to build
--- a/changelog.txt
+++ b/changelog.txt
@@ -38,7 +38,8 @@ Template for new versions:
 - `prioritize`: fix incorrect loading of persisted data on some OS types
 - `list-waves`: no longer gets confused by units that leave the map and then return (e.g. squads who go out on raids)
 - `fix/dead-units`: fix error when removing dead units from burrows and the unit with the greatest ID was dead
-- `gui/autodump`: prevent dumping into walls or invalid map area, set proper projectile flags to prevent items from being destroyed
+- `gui/autodump`: prevent dumping into walls or invalid map area, as well as selecting in unallocated blocks
+- `gui/autodump`: set proper projectile flags to prevent items from being destroyed
 
 ## Misc Improvements
 - `build-now`: if `suspendmanager` is running, run an unsuspend cycle immediately before scanning for buildings to build
--- a/gui/autodump.lua
+++ b/gui/autodump.lua
@@ -229,7 +229,7 @@ function Autodump:select_box(bounds)
             for x=bounds.x1,bounds.x2 do
                 local block = dfhack.maps.getTileBlock(xyz2pos(x, y, z))
                 local block_str = tostring(block)
-                if not seen_blocks[block_str] then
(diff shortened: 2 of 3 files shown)


gui/autodump.lua
```diff
@@ -306,14 +306,50 @@ function Autodump:onRenderFrame(dc, rect)
...
 function Autodump:do_dump(pos)
     pos = pos or dfhack.gui.getMousePos()
-    if not pos then return end
-    local tileattrs = df.tiletype.attrs[dfhack.maps.getTileType(pos)]
-    local basic_shape = df.tiletype_shape.attrs[tileattrs.shape].basic_shape
-    local on_ground = basic_shape == df.tiletype_shape_basic.Floor or
-            basic_shape == df.tiletype_shape_basic.Stair or
-            basic_shape == df.tiletype_shape_basic.Ramp
+    if not pos then
+        print('No cursor')
```
myk002, line 337: this should more specifically say "Please hover the mouse cursor over the target tile", or something to that effect

also should probably use `dfhack.printerr` instead of `print`

gui/autodump.lua
```diff
@@ -306,14 +306,50 @@ function Autodump:onRenderFrame(dc, rect)
...
+        return
+    end
+
+    local tt = dfhack.maps.getTileType(pos)
+    if not tt then
+        print('No map block')
```
myk002, line 343: this needs a more user-friendly message.

maybe we should require that the tile be visible (not hidden), which would also cover this case. Then the error message could reference the visibility of the tile.

gui/autodump.lua
```diff
@@ -306,14 +306,50 @@ function Autodump:onRenderFrame(dc, rect)
...
+        return
+    end
+
+    local on_ground, in_air = tile_props(pos, tt)
+    if not (on_ground or in_air) then
+        print('Dump tile blocked')
```
myk002, line 349: this also needs a better message (and use of `dfhack.printerr`)

gui/autodump.lua
```diff
@@ -306,14 +306,50 @@ function Autodump:onRenderFrame(dc, rect)
     end
 end
 
+local function tile_props(pos, tt) --Returns is_ground, is_open_air
+    local shape_attrs = df.tiletype_shape.attrs[df.tiletype.attrs[tt].shape]
+    if shape_attrs.walkable then --TODO: don't dump on statues, etc.?
```
myk002, line 311: dumping on statues is probably fine. the game allows it, so we can too.

myk002:
also needs a doc update defining what an acceptable target tile is

Bumber64:
Renaming branches closes PRs, apparently.

gui/autodump.lua
Bumber64: Should we allow dumping on fortifications? Projectiles can end up there.

Bumber64:
Looks like we need to be setting some projectile flags. The falling items are sometimes getting destroyed regardless.

gui/autodump.lua
Bumber64: Looks like this shouldn't actually happen, since the button is disabled if the mouse isn't over a target tile.

gui/autodump.lua
myk002: I'd say allow it. If the player selects a spot that is even remotely viable, then we should respect it

changelog.txt
```diff
@@ -38,6 +38,7 @@ Template for new versions:
 - `prioritize`: fix incorrect loading of persisted data on some OS types
 - `list-waves`: no longer gets confused by units that leave the map and then return (e.g. squads who go out on raids)
 - `fix/dead-units`: fix error when removing dead units from burrows and the unit with the greatest ID was dead
+- `gui/autodump`: prevent dumping into walls or invalid map area, set proper projectile flags to prevent items from being destroyed
```
myk002, line 41:
```suggestion
- `gui/autodump`: prevent dumping into walls or invalid map area
- `gui/autodump`: set proper projectile flags to prevent items from being destroyed when teleporting items into mid-air
```

gui/autodump.lua
```diff
@@ -308,12 +308,25 @@ end
...
 function Autodump:do_dump(pos)
     pos = pos or dfhack.gui.getMousePos()
-    if not pos then return end
-    local tileattrs = df.tiletype.attrs[dfhack.maps.getTileType(pos)]
-    local basic_shape = df.tiletype_shape.attrs[tileattrs.shape].basic_shape
-    local on_ground = basic_shape == df.tiletype_shape_basic.Floor or
-            basic_shape == df.tiletype_shape_basic.Stair or
-            basic_shape == df.tiletype_shape_basic.Ramp
+    if not pos then --We check this before calling
+        dfhack.printerr('Autodump:do_dump called with bad pos!')
```
myk002, line 312: this should be an `error`, otherwise `getTileType` will throw anyway when you pass `nil` to it

gui/autodump.lua
```diff
@@ -331,7 +344,15 @@ function Autodump:do_dump(pos)
...
             if not on_ground then
-                dfhack.items.makeProjectile(item)
+                local proj = dfhack.items.makeProjectile(item)
+                if proj then
+                    proj.flags.no_impact_destroy = true
+                    proj.flags.bouncing = true
+                    proj.flags.piercing = true
+                    proj.flags.parabolic = true
+                    proj.flags.no_adv_pause = true
+                    proj.flags.no_collide = true
```
myk002, line 354: should this be done inside of `makeProjectile`?

Bumber64: This is correct for falling items, but not for all projectiles.

myk002: ah ok

myk002 (approved):

PR #1243, opened 2024-07-28
Add table of contents to Gui/journal

wiktor-obrebski:
# Description
screen

Add table of contents and undo/redo features to the `gui/journal` script.
Notable features:
 - adding line starting from `#` give 1-th level ToC header. `##` and more give more indented ToC links
 - `Ctrl+O` toggle ToC
 - visible "Shifter" button to show/hide ToC
 - ToC panel can be resized by mouse
 - click in ToC entry scroll to given header in the text
 - `CTRL+Z` undo, `CTRL+Y` redo

Bugfixes:
 - 2846 numbers are not longer interpreted as cursor moves

# TODO later
- add buttons to insert new header of 1/2/3 indentation (to ToC panel?)
- stop render text on resize, it is slow
- consider hiding ToC panel (now panel without STYLE_FRAME can not be resized)

### DIFF
--- a/docs/gui/journal.rst
+++ b/docs/gui/journal.rst
@@ -47,7 +47,7 @@ Supported Features
   - replace selected text, if available
   - If no text is selected, paste text in the cursor position
 - Scrolling behaviour for long text build-in
-- Table of contests (:kbd:`Ctrl` + :kbd:`T`), with headers line prefixed by `#`, e.g. `# Fort history`, `## Year 1`
+- Table of contests (:kbd:`Ctrl` + :kbd:`O`), with headers line prefixed by `#`, e.g. `# Fort history`, `## Year 1`
 
 Usage
 -----
--- a/docs/gui/journal.rst
+++ b/docs/gui/journal.rst
@@ -20,7 +20,7 @@ Supported Features
 ------------------
 
 - Cursor Control: Navigate through text using arrow keys (left, right, up, down) for precise cursor placement.
-- Fast Rewind: Use :kbd:`Shift` + :kbd:`Left` / :kbd:`Ctrl` + :kbd:`B` and :kbd:`Shift` + :kbd:`Right` / :kbd:`Ctrl` + :kbd:`F` to move the cursor one word back or forward.
+- Fast Rewind: Use :kbd:`Ctrl` + :kbd:`Left` / :kbd:`Ctrl` + :kbd:`B` and :kbd:`Ctrl` + :kbd:`Right` / :kbd:`Ctrl` + :kbd:`F` to move the cursor one word back or forward.
 - Longest X Position Memory: The cursor remembers the longest x position when moving up or down, making vertical navigation more intuitive.
 - Mouse Control: Use the mouse to position the cursor within the text, providing an alternative to keyboard navigation.
 - New Lines: Easily insert new lines using the :kbd:`Enter` key, supporting multiline text input.
@@ -47,7 +47,7 @@ Supported Features
   - replace selected text, if available
   - If no text is selected, paste text in the cursor position
 - Scrolling behaviour for long text build-in
-- Table of contests (:kbd:`Ctrl` + :kbd:`O`), with headers line prefixed by '#', e.g. '# Fort history', '## Year 1'
+- Table of contents (:kbd:`Ctrl` + :kbd:`O`), with headers line prefixed by '#', e.g. '# Fort history', '## Year 1'
 
 Usage
 -----
--- a/docs/gui/journal.rst
+++ b/docs/gui/journal.rst
@@ -48,6 +48,7 @@ Supported Features
   - If no text is selected, paste text in the cursor position
 - Scrolling behaviour for long text build-in
 - Table of contents (:kbd:`Ctrl` + :kbd:`O`), with headers line prefixed by '#', e.g. '# Fort history', '## Year 1'
+- Table of contents navigation: jump to previous/next section by :kbd:`Ctrl` + :kbd:`Up` / :kbd:`Ctrl` + :kbd:`Down`
 
 Usage
 -----
--- a/docs/gui/journal.rst
+++ b/docs/gui/journal.rst
@@ -46,6 +46,7 @@ Supported Features
   - replace selected text, if available
   - If no text is selected, paste text in the cursor position
 - Scrolling behaviour for long text build-in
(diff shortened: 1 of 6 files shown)


myk002:
Copying here from Discord:

You can give it a frame style that renders gui.CLEAR_PEN like this:
```lua
local INVISIBLE_FRAME = {
    frame_pen=gui.CLEAR_PEN,
    signature_pen=false,
}
```

Also, there should be hotkeys to jump to prev/next heading for the keyboard-only users

Hotkey buttons could be added to the bottom, underneath a horizontal `widgets.Divider`, until we get icons for them, at which time they can be moved to the top

The vertical widgets.Divider for the ToC can link into the bottom horizontal Divider if you use the same frame_style (`gui.FRAME_THIN`)

docs/gui/journal.rst
```diff
@@ -46,6 +47,7 @@ Supported Features
   - replace selected text, if available
   - If no text is selected, paste text in the cursor position
 - Scrolling behaviour for long text build-in
+- Table of contests (:kbd:`Ctrl` + :kbd:`O`), with headers line prefixed by '#', e.g. '# Fort history', '## Year 1'
```
myk002, line 50:
```suggestion
- Table of contents (:kbd:`Ctrl` + :kbd:`O`), with headers line prefixed by '#', e.g. '# Fort history', '## Year 1'
```

gui/journal.lua
```diff
@@ -48,69 +177,184 @@ function JournalWindow:sanitizeFrame(frame)
...
+end
+
+function JournalWindow:onPanelResizeEnd()
+    self.resizing_panels = false
+    self:ensurePanelsRelSize()
+
+    self:updateLayout()
+end
+
+function JournalWindow:onRenderBody(painter)
```
myk002, line 195: it would be useful here to set the current ToC list selection to the currently section, or rather, the section that the 5th line from the top of the screen contains if the cursor is not onscreen, and the section that contains the cursor if the cursor is onscreen.

myk002 (commented):
this is awesome ; )

gui/journal.lua
```diff
@@ -48,69 +150,174 @@ function JournalWindow:sanitizeFrame(frame)
...
+end
+
+function JournalWindow:ensurePanelsRelSize()
+    local toc_panel = self.subviews.table_of_contents_panel
+    local editor = self.subviews.journal_editor
+    local divider = self.subviews.table_of_contents_divider
+
+    toc_panel.frame.w = math.min(
+        math.max(toc_panel.frame.w, toc_panel.resize_min.w),
+        self.frame.w - editor.resize_min.w
```
myk002, line 211:
```sh
/home/myk/src/dfhack/scripts/gui/journal.lua:211: attempt to perform arithmetic on a nil value (field 'w')
stack traceback:
        /home/myk/src/dfhack/scripts/gui/journal.lua:211: in method 'ensurePanelsRelSize'
        /home/myk/src/dfhack/scripts/gui/journal.lua:218: in local 'fun'
        ./hack/lua/class.lua:110: in function <./hack/lua/class.lua:105>
        (...tail calls...)
        ./hack/lua/gui.lua:790: in method 'updateLayout'
        ./hack/lua/gui.lua:779: in method 'updateSubviewLayout'
        ./hack/lua/gui.lua:799: in method 'updateLayout'
        ./hack/lua/gui.lua:976: in method 'onResize'
        ./hack/lua/gui.lua:902: in local 'fun'
```

Note that `frame.w` is not only not guaranteed to exist, but also does not represent the current state of the window. The frame represents the *desired* dimensions, not the *actual* dimensions. I think you might want `self.frame_body.width` or `self.frame_parent_rect.width`.

both fields are set as part of `updateLayout`, so they might not exist on the first call to `preUpdateLayout`.

wiktor-obrebski: Can you try to repeat the issue and explain when you exactly get this error?

There was a refactor error in the code for 20 minutes, you could test it on the bugged version.

wiktor-obrebski: with my understand as I always set `journal_editor` and `table_of_contents_panel` in `JournalWindow:init` - they should be available from the beginning in my case.

wiktor-obrebski: I see now, its on server automatic tests. Checking.

wiktor-obrebski:
@myk002 there is a problem with `A_MOVE_N_DOWN` and other new test keys.
I added them, as we discussed, to control sections, but tests run on game version without beta yet (if I understand correctly), so the keys are not supported. So, in tests I get 
> error: ./hack/lua/gui.lua:65: Invalid keycode: A_MOVE_N_DOWN

And similar errors

gui/journal.lua
wiktor-obrebski: should be fixed now - it was a problem with missing defaults for layout restoring.

myk002:
Notes from playtesting:

The initial empty screen looks stark and scary. I think we need to add some help text:

1. In the main text area, add a short intro about what the tool is, and direct the player's attention to the `?` button for further info. This text disappears as soon as there is user text
2. In the ToC area, show information about how to add headers. This text disappears once there is a header to show

The help blurbs can be implemented as additional `widgets.WrappedLabel` widgets that overlap the regular text areas and whose `visible` properties are set to functions that detect when they should be shown.

Possible main area help text:
```
Welcome to gui/journal! You can use this tool to take notes, track progress on your plans, or anything else! The text you write here is saved with your fort. Please click on the ? button in the upper right corner to see more information, including a list of keyboard navigation hotkeys.
```

Possible ToC help text:
```
Start a line with # symbols to create headers. Those headers will appear here, and you can click on them to jump to them in the text. Click on the ? button in the upper right corner for more details.
```

internal/journal/table_of_contents.lua
```diff
@@ -12,45 +12,81 @@ TableOfContents = defclass(TableOfContents, widgets.Panel)
...
         widgets.List{
-            frame={l=1, t=0, r=1, b=0},
+            frame={l=0, t=0, r=0, b=3},
             view_id='table_of_contents',
             choices={},
             on_submit=self.on_submit
         },
+        widgets.HotkeyLabel{
+            frame={b=0},
+            key='A_MOVE_N_DOWN',
```
myk002, line 29: you'll have to gate this on the DF version as well. the keys listed here must be valid for the DF version in order for the UI to load

myk002: fixed

myk002:
merged so people can start testing a bit. We can figure out the help text in a different PR

PR #1253, opened 2024-08-03
Show a simple tutorial for gui/journal on first use

wiktor-obrebski:
Show a simple tutorial for gui/journal on first use.

There are 2 tutorials:

1. Main text are tutorial. Its render one line below the text input. Its render only on first usage of the journal in given fort. Dissapearing when user provide any text to the text area
2. Table of contents tutorial - visible always when there is no section in text

Also, from now the table of contents panel is visible by default.

### DIFF
--- a/gui/journal.lua
+++ b/gui/journal.lua
@@ -11,7 +11,22 @@ local table_of_contents = reqscript('internal/journal/table_of_contents')
 
 local RESIZE_MIN = {w=32, h=10}
 
-JOURNAL_PERSIST_KEY = 'journal'
+local JOURNAL_PERSIST_KEY = 'journal'
+
+local JOURNAL_WELCOME_COPY =  [=[
+Welcome to gui/journal, the chronicler's tool for Dwarf Fortress!
+
+Here, you can carve out notes, sketch your grand designs, or record the history of your fortress.
+The text you write here is saved together with your fort.
+
+For guidance on navigation and hotkeys, tap the ? button in the upper right corner.
+Happy digging!
+]=]
+
+local TOC_WELCOME_COPY =  [=[
+Start a line with # symbols and a space to create header.
+Those headers will appear here, and you can click on them to jump to them in the text.
+]=]
 
 journal_config = journal_config or json.open('dfhack-config/journal.json')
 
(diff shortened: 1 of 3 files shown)


gui/journal.lua
```diff
@@ -11,7 +11,22 @@ local table_of_contents = reqscript('internal/journal/table_of_contents')
...
+
+Here, you can carve out notes, sketch your grand designs, or record the history of your fortress.
+The text you write here is saved together with your fort.
+
+For guidance on navigation and hotkeys, tap the ? button in the upper right corner.
+Happy digging!
+]=]
+
+local TOC_WELCOME_COPY =  [=[
+Start a line with # symbols and a space to create header.
```
myk002, line 27:
```suggestion
Start a line with # symbols and a space to create a header.
```

gui/journal.lua
```diff
@@ -11,7 +11,22 @@ local table_of_contents = reqscript('internal/journal/table_of_contents')
...
+Those headers will appear here, and you can click on them to jump to them in the text.
```
myk002, line 28: I think this could use an example, such as:
```
Start a line with # symbols and a space to create a header. For example:

# My section heading

or 

## My section subheading
```

gui/journal.lua
```diff
@@ -49,7 +65,15 @@ function JournalWindow:init()
             on_resize_begin=self:callback('onPanelResizeBegin'),
             on_resize_end=self:callback('onPanelResizeEnd'),
 
-            on_submit=self:callback('onTableOfContentsSubmit')
+            on_submit=self:callback('onTableOfContentsSubmit'),
+            subviews={
```
myk002, line 69: adding it as a subview? brilliant.

gui/journal.lua
```diff
@@ -164,9 +201,9 @@ function JournalWindow:loadConfig()
 
     local table_of_contents = copyall(journal_config.data.toc or {})
     table_of_contents.width = table_of_contents.width or 20
-    table_of_contents.visible = table_of_contents.visible or false
+    table_of_contents.visible = table_of_contents.visible or true
```
myk002, line 204: this expression is equivalent to `table_of_contents.visible = true`. is that what you want here?

myk002 (approved):

PR #1260, opened 2024-08-06
Embark Anyone

ProtonNumber:
Allows you to embark as any civilization, including dead and non-dwarven ones through a unretire-anyone style UI.

Currently suffers from some issues with counting sites, especially with non-dwarven civs. This is ultimately a cosmetic issue, but I'd be interested if anyone has a more accurate method of doing this.

### DIFF
--- a/docs/embark-anyone.rst
+++ b/docs/embark-anyone.rst
@@ -1,0 +1,19 @@
+embark-anyone
+========
+
+.. dfhack-tool::
+    :summary: Allows you to embark as any civilisation, including dead and non-dwarven ones.
+    :tags: embark gameplay armok
+
+Usage
+-----
+
+::
+
+    embark-anyone
+
+Note
+-----
+Non-dwarven civs have their own mechanics which can render fortress mode difficult
+to unplayable. Preparing carefully is advised, and cheating in some items may be
+required.
--- a/docs/embark-anyone.rst
+++ b/docs/embark-anyone.rst
@@ -3,7 +3,12 @@ embark-anyone
 
 .. dfhack-tool::
     :summary: Allows you to embark as any civilisation, including dead and non-dwarven ones.
-    :tags: embark gameplay armok
+    :tags: embark armok
+
+This script can be ran on the embark screen and presents a list of civilisations that
+are not normally playable to add to the "Select Origin Civ" menu. This includes dead
+civilisations, the normal non-dwarven civilisations, and subterranean animal people.
+
 
 Usage
 -----
--- a/docs/embark-anyone.rst
+++ b/docs/embark-anyone.rst
@@ -5,9 +5,9 @@ embark-anyone
     :summary: Allows you to embark as any civilisation, including dead and non-dwarven ones.
     :tags: embark armok
 
-This script can be ran on the embark screen and presents a list of civilisations that
-are not normally playable to add to the "Select Origin Civ" menu. This includes dead
-civilisations, the normal non-dwarven civilisations, and subterranean animal people.
+This script must be run on the embark screen when choosing an origin civilization.
+When run, you can add any civilization, including dead or non-dwarven civilizations,
+to the list of choices.
 
 
 Usage
@@ -20,4 +20,5 @@ Usage
 Note
 -----
 Non-dwarven civs have their own mechanics which can render fortress mode difficult
-to unplayable. Preparing carefully is advised, and some crucial items may need to be spawned in.
+or unplayable. Preparing carefully is advised, and some crucial items may need to be
+spawned in with other DFHack tools (e.g. `gui/create-item`).
--- a/embark-anyone.lua
+++ b/embark-anyone.lua
@@ -1,4 +1,5 @@
 local dialogs = require 'gui.dialogs'
+local utils = require 'utils'
 local choices = {}
 
 function addCivToEmbarkList(info)
@@ -37,8 +38,8 @@ function embarkAnyone()
       -- Civs keep links to sites they no longer hold, so check owner
       -- We also take the opportunity to count population
       for j, link in ipairs(civ.site_links) do
-         local site = df.global.world.world_data.sites[link.target]
-         if site.civ_id == civ.id then
+         local site = df.world_site.find(link.target)
+         if site ~= nil and site.civ_id == civ.id then
             sites = sites + 1
 
             -- DF stores population info as an array of groups of residents (?).
@@ -49,16 +50,18 @@ function embarkAnyone()
          end
 
          -- Count living nemeses
-         for _, nem in ipairs (civ.nemesis_ids) do
-            if df.global.world.nemesis.all[nem].figure.died_year == -1 then
+         for _, nem_id in ipairs (civ.nemesis_ids) do
+            local nem = df.nemesis_record.find(nem_id)
+            if nem ~= nil and nem.figure.died_year == -1 then
                nemeses = nemeses + 1
             end
          end
 
          -- Count living histfigs
          -- Used for death detection. May be redundant.
-         for _, fig in ipairs (civ.histfig_ids) do
-            if df.global.world.history.figures[fig].died_year == -1 then
+         for _, fig_id in ipairs (civ.histfig_ids) do
+            local fig = df.historical_figure.find(fig_id)
+            if fig ~= nil and fig.died_year == -1 then
                histfigs = histfigs + 1
             end
          end
@@ -89,17 +92,20 @@ function embarkAnyone()
 
       ::continue::
    end
-   dialogs.ListBox{
-      frame_title = 'Embark Anyone',
-      text = 'Select a civilisation to add to the list of origin civs:',
-      text_pen = COLOR_WHITE,
-      choices = choices,
-      on_select = function(id, choice)
-         addCivToEmbarkList(choice.info)
-      end,
-      with_filter = true,
-      row_height = 4,
-   }:show()
+   if #choices then
+      dialogs.ListBox{
+         frame_title = 'Embark Anyone',
+         text = 'Select a civilisation to add to the list of origin civs:',
+         text_pen = COLOR_WHITE,
+         choices = choices,
+         on_select = function(id, choice)
+            addCivToEmbarkList(choice.info)
+         end,
+         with_filter = true,
+         row_height = 4,
+      }:show()
+   end
+
 end
 
 embarkAnyone()
--- a/embark-anyone.lua
+++ b/embark-anyone.lua
@@ -1,6 +1,5 @@
-local dialogs = require 'gui.dialogs'
-local utils = require 'utils'
-local choices = {}
+local dialogs = require('gui.dialogs')
+local utils = require('utils')
 
 function addCivToEmbarkList(info)
    local viewscreen = dfhack.gui.getDFViewscreen(true)
@@ -13,7 +12,7 @@ end
 
 function embarkAnyone()
    local viewscreen = dfhack.gui.getDFViewscreen(true)
-   local existing_civs = {}
+   local choices, existing_civs = {}, {}
 
    for _,existing_civ in ipairs(viewscreen.start_civ) do
       existing_civs[existing_civ.id] = true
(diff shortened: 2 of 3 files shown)


docs/embark-anyone.rst
```diff
@@ -0,0 +1,19 @@
...
+
+::
+
+    embark-anyone
+
+Note
+-----
+Non-dwarven civs have their own mechanics which can render fortress mode difficult
+to unplayable. Preparing carefully is advised, and cheating in some items may be
+required.
```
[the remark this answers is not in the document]
ab9rf: i'd avoid the pejorative "cheating" entirely and instead say something like "without spawning them in manually" or some other phrase that describes the action neutrally

don't embed judgments about how people choose to play the game into the documentation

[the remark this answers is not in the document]
ProtonNumber: How about "some crucial items may need to be spawned in"?

docs/embark-anyone.rst
```diff
@@ -0,0 +1,18 @@
+embark-anyone
+=============
+
+.. dfhack-tool::
+    :summary: Allows you to embark as any civilisation, including dead and non-dwarven ones.
+    :tags: embark gameplay armok
```
myk002, line 6: TODO: think about the implications of combining `gameplay` and `armok`

(If I submit this review without finishing my thoughts here, please poke me about this)

docs/embark-anyone.rst
```diff
@@ -0,0 +1,18 @@
...
+
```
myk002, line 7: Please add an introductory paragraph before the Usage section that at least goes over when to run the command

embark-anyone.lua
```diff
@@ -0,0 +1,108 @@
...
+function embarkAnyone()
+   local viewscreen = dfhack.gui.getDFViewscreen(true)
+
+
+   if viewscreen._type ~= df.viewscreen_choose_start_sitest then
+      qerror("This script can only be used on the embark screen!")
+   end
+
+   for i, civ in ipairs (df.global.world.entities.all) do
+      if civ.type == df.historical_entity_type.Civilization then
```
myk002, line 22: instead of placing the entire body of the for loop in an if statement, consider inverting the logic and skipping the for body (that is, a "continue" statement in other languages) using a Lua label:
```lua
for i, civ in ipairs(df.global.world.entities.all) do
    if civ.type ~= df.historical_entity_type.Civilization then goto continue end
    ...
    ::continue::
end
```
This reduces the overall indentation and can make the code easier to reason about and follow

embark-anyone.lua
```diff
@@ -0,0 +1,108 @@
...
+         local available = false
+
+         -- Check if civ is already available to embark
+         for _, item in ipairs (viewscreen.start_civ) do
+            if item == civ then
+               available = true
+               break
+            end
+         end
```
myk002, line 31: this would be more efficient if you built a set outside of this loop. e.g.:
```lua
local existing_civs = {}
for _,existing_civ in ipairs(viewscreen.start_civ) do
    existing_civs[existing_civ.id] = true
end
```
Then later you can test for membership by evaluating `existing_civs[civ.id]`:
```lua
if existing_civs[civ.id] then goto continue end
```

embark-anyone.lua
```diff
@@ -0,0 +1,108 @@
...
+                  -- DF stores population info as an array of groups of residents (?).
+                  -- Inspecting these further could give a more accurate count.
+                  for _, group in ipairs(site.populace.inhabitants) do
+                     pops = pops + group.count
+                  end
+               end
+
+               -- Count living nemeses
+               for _, nem in ipairs (civ.nemesis_ids) do
+                  if df.global.world.nemesis.all[nem].figure.died_year == -1 then
```
myk002, line 56: nemesis IDs are *usually* sequential, but not always. Use `df.nemesis.find(nem)` instead

embark-anyone.lua
```diff
@@ -0,0 +1,108 @@
...
+            local sites = 0
+            local pops = 0
+            local nemeses = 0
+            local histfigs = 0
+            local label = ''
+
+            -- Civs keep links to sites they no longer hold, so check owner
+            -- We also take the opportunity to count population
+            for j, link in ipairs(civ.site_links) do
+               local site = df.global.world.world_data.sites[link.target]
```
myk002, line 43: if `link.target` is the site id, you'd get the site with `df.site.find(link.target)`, which does a binary search through the sites vector. the ID is not necessarily the same as the offset.

embark-anyone.lua
```diff
@@ -0,0 +1,108 @@
...
+               for _, nem in ipairs (civ.nemesis_ids) do
+                  if df.global.world.nemesis.all[nem].figure.died_year == -1 then
+                     nemeses = nemeses + 1
+                  end
+               end
+
+               -- Count living histfigs
+               -- Used for death detection. May be redundant.
+               for _, fig in ipairs (civ.histfig_ids) do
+                  if df.global.world.history.figures[fig].died_year == -1 then
```
myk002, line 64: here too. use `df.historical_figure.find(fig)`. I didn't mention this above, but you also have to test the result of `find` for nil, since it's possible that the IDs in the vectors are stale, and the referent has been deleted.

embark-anyone.lua
```diff
@@ -0,0 +1,108 @@
...
+            label = label .. dfhack.units.getRaceNamePluralById(civ.race) .. "\n"
+
+            -- Add pop & site count or mark civ as dead.
+            if histfigs == 0 and pops == 0 then
+               label = label .. "Dead"
+            else
+               label = label .. "Pop: " .. (pops + nemeses) .. " Sites: " .. sites
+            end
+
+            table.insert(choices, {text = label, search_key = label:lower(),
```
myk002, line 89: you can let the search key be automatically generated if it's just going to be a version of the label (the match is case insensitive anyway)

embark-anyone.lua
```diff
@@ -0,0 +1,108 @@
...
+                                   info = {civ = civ, pops = pops, sites = sites,
+                                           nemeses = nemeses}})
+         end
+      end
+   end
+   dialogs.ListBox{
```
myk002, line 95: you should probably handle the case where there are no more civs to add to the list. A dialogs.MessageBox with the message should be sufficient.

docs/embark-anyone.rst
myk002: more words:

I've been treating "gameplay" and "armok" as mutually exclusive, with "gameplay" meaning "changes the gameplay in some consistent way that may not be achievable in vanilla but is not inherently "cheaty" or game-breaking" and armok meaning "no restrictions -- the world conforms to your will".

In other words, "gameplay" tools can have some "armoky" elements to them, but they are intended to be a balanced extension to the world.

This script straddles that line closer than most other tools.

Both `unretire-anyone` and `gui/embark-anywhere` have the `armok` tag and not the `gameplay` tag, so my suspicion is that we should go with `:tags: embark armok` here and remove the `gameplay` tag.

ProtonNumber: Given that some civs are borderline unplayable, I'll agree that armok alone is appropriate.

embark-anyone.lua
ProtonNumber: Ahhhh okay

Sorry, I'm still used to C, where using goto is _heavily_ frowned upon.

myk002: totally understood. I had to get used to this Lua-ism as well

docs/embark-anyone.rst
```diff
@@ -0,0 +1,23 @@
+embark-anyone
+=============
+
+.. dfhack-tool::
+    :summary: Allows you to embark as any civilisation, including dead and non-dwarven ones.
+    :tags: embark armok
+
+This script can be ran on the embark screen and presents a list of civilisations that
+are not normally playable to add to the "Select Origin Civ" menu. This includes dead
+civilisations, the normal non-dwarven civilisations, and subterranean animal people.
```
myk002, line 10:
```suggestion
This script must be run on the embark screen when choosing an origin civilization.
When run, you can add any civilization, including dead or non-dwarven civilizations,
to the list of choices.
```

docs/embark-anyone.rst
```diff
@@ -0,0 +1,23 @@
...
+-----
+
+::
+
+    embark-anyone
+
+Note
+-----
+Non-dwarven civs have their own mechanics which can render fortress mode difficult
+to unplayable. Preparing carefully is advised, and some crucial items may need to be spawned in.
```
myk002, line 23:
```suggestion
or unplayable. Preparing carefully is advised, and some crucial items may need to be
spawned in with other DFHack tools (e.g. `gui/create-item`).
```

embark-anyone.lua
```diff
@@ -0,0 +1,117 @@
...
+
+   viewscreen.start_civ:insert ('#', info.civ)
+   viewscreen.start_civ_nem_num:insert ('#', info.nemeses)
+   viewscreen.start_civ_entpop_num:insert ('#', info.pops)
+   viewscreen.start_civ_site_num:insert ('#', info.sites)
+end
+
+function embarkAnyone()
+   local viewscreen = dfhack.gui.getDFViewscreen(true)
+   local existing_civs = {}
```
myk002, line 16:
```suggestion
   local choices, existing_civs = {}, {}
```

embark-anyone.lua
```diff
@@ -0,0 +1,117 @@
+local dialogs = require 'gui.dialogs'
+local utils = require 'utils'
+local choices = {}
```
myk002, line 3:
```suggestion
local dialogs = require('gui.dialogs')
local utils = require('utils')
```

embark-anyone.lua
```diff
@@ -0,0 +1,117 @@
...
+      label = label .. dfhack.units.getRaceNamePluralById(civ.race) .. "\n"
+
+      -- Add pop & site count or mark civ as dead.
+      if histfigs == 0 and pops == 0 then
+         label = label .. "Dead"
+      else
+         label = label .. "Pop: " .. (pops + nemeses) .. " Sites: " .. sites
+      end
+
+      table.insert(choices, {text = label, search_key = label:lower(),
```
myk002, line 89:
```suggestion
      table.insert(choices, {text = label,
```

embark-anyone.lua
```diff
@@ -0,0 +1,117 @@
...
+                             info = {civ = civ, pops = pops, sites = sites,
+                                     nemeses = nemeses}})
+
+      ::continue::
+   end
+
+   if #choices then
```
myk002, line 96:
```suggestion
   if #choices > 0 then
```

myk002: in Lua, 0 is truthy

embark-anyone.lua
```diff
@@ -0,0 +1,117 @@
...
+      dialogs.ListBox{
+         frame_title = 'Embark Anyone',
+         text = 'Select a civilisation to add to the list of origin civs:',
```
myk002, line 99:
```suggestion
         text = 'Select a civilization to add to the list of origin civs:',
```

embark-anyone.lua
```diff
@@ -0,0 +1,117 @@
...
+         on_select = function(id, choice)
+            addCivToEmbarkList(choice.info)
+         end,
+         with_filter = true,
+         row_height = 4,
+      }:show()
+   else
+      dialogs.MessageBox{
+         frame_title = 'Embark Anyone',
+         text = 'No additional civilisations found.'
```
myk002, line 111:
```suggestion
         text = 'No additional civilizations found.'
```

myk002: using american spelling to conform to other uses in the game and on the wiki

embark-anyone.lua
myk002: use preferred formatting; move `choices` to block where it is used

embark-anyone.lua
myk002: not only is the search key correctly autogenerated in this case, the autogenerated version additionally handles special characters in names (e.g. umlauts and other diacritics)

myk002:
At your convenience, please also add your name to https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst

myk002 (commented):
looks great! there's one last thing I forgot to mention: could you add a line under the "New tools" header in the "Future" version in `changelog.txt`?

https://github.com/DFHack/scripts/blob/master/changelog.txt#L30

myk002 (approved):

PR #1261, opened 2024-08-08
[gui/journal] don't include trailing newlines when copying a line

myk002:
it causes artifacts when pasting that line in `gui/launcher` or other edit boxes


myk002:
we can do this more effectively in the DFSDL module; closing this PR

PR #1262, opened 2024-08-10
[timestream] ensure we don't skip unit birthdays

myk002:
Fixes: https://github.com/DFHack/dfhack/issues/4848

### DIFF
--- a/timestream.lua
+++ b/timestream.lua
@@ -2,6 +2,7 @@
 --@enable = true
 
 local argparse = require('argparse')
+local eventful = require('plugins.eventful')
 local repeatutil = require("repeat-util")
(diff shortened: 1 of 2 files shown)


myk002:
@Tjudge1 FYI

timestream.lua
```diff
@@ -68,18 +69,51 @@ local TICK_TRIGGERS = {
 }
 
 -- "owed" ticks we would like to skip at the next opportunity
-local timeskip_deficit = 0.0
+timeskip_deficit = timeskip_deficit or 0.0
```
myk002, line 72: state is made global so we can print them out as debug messages

PR #1267, opened 2024-08-13
new tool: gui/pregnancy

myk002:
productionization of #873

changes:
- UI now dynamically enables valid widgets, e.g. disabling UI elements that don't make sense in the current state
- if the same unit is named multiple times in the UI, the name will have the same color in all places
- business logic significantly simplified
- window appears off to the side so the player can more easily select units as the parents
- automatically load the spouse if no other partner has been chosen yet
- show whether the currently selected female is pregnant (and how long it will be before she comes to term)
- redo layout using dividers for a cleaner look


myk002:
demo: 
gui_pregnancy

PR #1268, opened 2024-08-13
new tool: json -- like the markdown tool

RobertHolstein:
Creates a json file with information about the select unit or item. Similar to the markdown tool but exports more data.

### DIFF
--- a/docs/json.rst
+++ b/docs/json.rst
@@ -1,0 +1,78 @@
+json
+====
+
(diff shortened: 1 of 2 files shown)


myk002:
yeah, we could combine this script and `markdown` (and `forum-dwarves`) into a generic unit and item export tool named.. what, `export`? `textify`?

PR #1284, opened 2024-09-01
Fixed mothers attempting to seek their units that were babies

ahlove3:
Fixed mothers attempting to seek their units that were babies by removing the unit id from the ANY_BABY table. Also made those grown babies act like adults by updating some flags.

Fixes: https://github.com/DFHack/dfhack/issues/4894

### DIFF
--- a/rejuvenate.lua
+++ b/rejuvenate.lua
@@ -24,6 +24,36 @@ function rejuvenate(unit, force, dry_run, age)
         unit.old_year = new_birth_year + 160
     end
     if unit.profession == df.profession.BABY or unit.profession == df.profession.CHILD then
+        if unit.profession == df.profession.BABY then
+            local leftoverUnits = {}
+            local shiftedLeftoverUnits = {}
+            -- create a copy
+            local babyUnits = df.global.world.units.other.ANY_BABY
+            -- create a new table with the units that aren't being removed in this iteration
+            for _, v in ipairs(babyUnits) do
+                if not v.id == unit.id then
+                    table.insert(leftoverUnits, v)
+                end
+            end
+            -- create a shifted table of the leftover units to make up for lua tables starting with index 1 and the game starting with index 0
+            for i = 0, #leftoverUnits - 1, 1 do
+                local x = i+1
+                shiftedLeftoverUnits[i] = leftoverUnits[x]
+            end
+            -- copy the leftover units back to the game table
+            df.global.world.units.other.ANY_BABY = shiftedLeftoverUnits
+            -- set extra flags to defaults
+            unit.flags1.rider = false
+            unit.relationship_ids.RiderMount = -1
+            unit.mount_type = 0
+            unit.profession2 = df.profession.STANDARD
+            unit.idle_area_type = 26
+            unit.mood = -1
+
+            -- let the mom know she isn't carrying anyone anymore
+            local motherUnitId = unit.relationship_ids.Mother
+            df.unit.find(motherUnitId).flags1.ridden = false
+        end
         unit.profession = df.profession.STANDARD
     end
     print(name .. ' is now ' .. age .. ' years old and will live to at least 160')


rejuvenate.lua
```diff
@@ -24,6 +24,36 @@ function rejuvenate(unit, force, dry_run, age)
...
+            local shiftedLeftoverUnits = {}
+            -- create a copy
+            local babyUnits = df.global.world.units.other.ANY_BABY
+            -- create a new table with the units that aren't being removed in this iteration
+            for _, v in ipairs(babyUnits) do
+                if not v.id == unit.id then
+                    table.insert(leftoverUnits, v)
+                end
+            end
+            -- create a shifted table of the leftover units to make up for lua tables starting with index 1 and the game starting with index 0
```
ab9rf, line 38: please use the `erase` method on the `world.units.all` vector for this. this method directly calls the appropriate C++ method on the underlying vector instead of rewriting the vector

rejuvenate.lua
```diff
@@ -24,6 +24,36 @@ function rejuvenate(unit, force, dry_run, age)
...
+            for i = 0, #leftoverUnits - 1, 1 do
+                local x = i+1
+                shiftedLeftoverUnits[i] = leftoverUnits[x]
+            end
+            -- copy the leftover units back to the game table
+            df.global.world.units.other.ANY_BABY = shiftedLeftoverUnits
+            -- set extra flags to defaults
+            unit.flags1.rider = false
+            unit.relationship_ids.RiderMount = -1
+            unit.mount_type = 0
```
myk002, line 48: should use `df.rider_positions_type.STANDARD` instead of `0`

rejuvenate.lua
```diff
@@ -24,6 +24,36 @@ function rejuvenate(unit, force, dry_run, age)
...
+            unit.profession2 = df.profession.STANDARD
+            unit.idle_area_type = 26
```
myk002, line 50: Should use `df.unit_station_type.MillBuilding` instead of `26`

rejuvenate.lua
```diff
@@ -24,6 +24,36 @@ function rejuvenate(unit, force, dry_run, age)
...
+            unit.mood = -1
+
+            -- let the mom know she isn't carrying anyone anymore
+            local motherUnitId = unit.relationship_ids.Mother
+            df.unit.find(motherUnitId).flags1.ridden = false
```
myk002, line 55: it is a good idea to always protect from `nil` when using any `find` method. The mother could have just been obliterated and the unit could be gone by the time this is run.

rejuvenate.lua
```diff
@@ -24,6 +24,36 @@ function rejuvenate(unit, force, dry_run, age)
...
+        end
         unit.profession = df.profession.STANDARD
```
myk002, line 57: Also check and update `unit.profession2` and the `profession` in the histfig (`df.historical_figure.find(unit.hist_figure_id)`)

rejuvenate.lua
myk002: you can also avoid finding the index entirely by using `utils.erase_sorted(df.global.world.units.other.ANY_BABY, unit, 'id')`

ab9rf: > you can also avoid finding the index entirely by using `utils.erase_sorted(df.global.world.units.other.ANY_BABY, unit, 'id')`

do note that classification subvectors (such as `ANY_BABY`) are not guaranteed to be sorted. the `categorize` methods simply insert the unit/item/building at the end of the classified subvector. using tools that expect these vectors to be sorted may result in unexpected behavior

ahlove3: Thanks for the heads up. I had trouble with lua doing a 1 based index vs C++ (as well as what I'm used to from other languages) doing a 0 based index and then that translating back into the game vector. I'll update with these fixes later this week/this weekend. Appreciate the suggestions

myk002:
Do you think you'll have time to finish this up this weekend? If so, we can get it into the upcoming release

myk002 (commented):
Please also add a line to `changelog.txt` with a description of the fix

myk002:
There are a few other necessary fixes to `rejuvenate`, so I can follow up on the PR comments. Thanks!

PR #1285, opened 2024-09-02
position.lua: additional info, option to copy cursor pos

Bumber64:
Adds an option (`-c`, `--copy`) to copy the current cursor position to the system clipboard.

Also implements display of the current historical era.

Fort site position isn't implemented yet. We might want to handle adventurer world position along with that (refer to #1194 for how that's done w/r/t travel map).

### DIFF
--- a/docs/position.rst
+++ b/docs/position.rst
@@ -5,13 +5,30 @@ position
     :summary: Report cursor and mouse position, along with other info.
     :tags: fort inspection map
 
-This tool reports the current date, clock time, month, and season. It also
-reports the cursor position (or just the z-level if no cursor), window size, and
-mouse location on the screen.
+This tool reports the current date, clock time, month, season, and historical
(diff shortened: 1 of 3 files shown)


docs/position.rst
```diff
@@ -5,13 +5,30 @@ position
...
     :tags: fort inspection map
 
-This tool reports the current date, clock time, month, and season. It also
-reports the cursor position (or just the z-level if no cursor), window size, and
-mouse location on the screen.
+This tool reports the current date, clock time, month, season, and historical
+era. It also reports the keyboard cursor position (or just the z-level if no
+active cursor), window size, and mouse location on the screen.
+
+Can also be used to copy the current cursor position for later use.
```
myk002, line 12:
```suggestion
Can also be used to copy the current keyboard cursor position for later use.
```

myk002 (approved):

PR #1287, opened 2024-09-03
`notes` script

wiktor-obrebski:
# Notes
This PR introduce notes feature that allows to add simple notes to the game map.

screen

Features:
 - `notes add` to add new note to the map (in place of keyboard cursor)
 - enable `notes.map_notes` overlay to render notes on the map as green pins
 - click on map notes to open note details and edit screen
 - click few times to toggle between notes if they are all in the same tile
 - the overlay has been written with performance in mind (and profiled)

TODO:
- [x] documentation
- [x] way to trigger note modal in ASCII mode
- [x] one-line-mode for text editor, for the note name
- [x] move `?` button from the text area component to the journal itself

### DIFF
--- a/docs/notes.rst
+++ b/docs/notes.rst
@@ -2,7 +2,7 @@ notes
 =====
 
 .. dfhack-tool::
-    :summary: Manage map-specific notes
+    :summary: Manage map-specific notes.
     :tags: fort interface map
 
 The `notes` tool enables players to annotate specific tiles
@@ -25,18 +25,18 @@ Usage
 Add new note in the current position of the keyboard cursor.
 
 Creating a Note
-~~~~~~~~~~~~~~~
+---------------
 1. Use the keyboard cursor to select the desired map tile where you want to place a note.
 2. Execute ``notes add`` via the DFHack console.
 3. In the pop-up dialog, fill in the note's title and detailed comment.
 4. Press :kbd:`Alt` + :kbd:`S` to create the note.
 
 Editing or Deleting a Note
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+--------------------------
 - Click on the green pin representing the note directly on the map.
 - A dialog will appear, offering options to edit the title or comment, or to delete the note entirely.
 
 Managing Notes Visibility
-~~~~~~~~~~~~~~~~~~~~~~~~~
-- Access the ``DFHack Control Panel`` / ``UI Overlays`` tab.
+-------------------------
+- Access the `gui/control-panel` / ``UI Overlays`` tab.
 - Toggle the ``notes.map-notes`` overlay to show or hide the notes on the map.
--- a/docs/notes.rst
+++ b/docs/notes.rst
@@ -1,0 +1,42 @@
+notes
+=====
+
+.. dfhack-tool::
+    :summary: Manage map-specific notes
+    :tags: fort interface map
+
+The `notes` tool enables players to annotate specific tiles
+on the Dwarf Fortress game map with customizable notes.
+
+Each note is displayed as a green pin on the map and includes a one-line title and a detailed comment.
+
+It can be used to e.g.:
+ - marking plans for future constructions
+ - explaining mechanisms or traps
+ - noting historical events
+
+Usage
+-----
+
+::
+
+    notes add
+
+Add new note in the current position of the keyboard cursor.
+
+Creating a Note
+~~~~~~~~~~~~~~~
+1. Use the keyboard cursor to select the desired map tile where you want to place a note.
+2. Execute ``notes add`` via the DFHack console.
+3. In the pop-up dialog, fill in the note's title and detailed comment.
+4. Press :kbd:`Alt` + :kbd:`S` to create the note.
+
+Editing or Deleting a Note
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+- Click on the green pin representing the note directly on the map.
+- A dialog will appear, offering options to edit the title or comment, or to delete the note entirely.
+
+Managing Notes Visibility
+~~~~~~~~~~~~~~~~~~~~~~~~~
+- Access the ``DFHack Control Panel`` / ``UI Overlays`` tab.
+- Toggle the ``notes.map-notes`` overlay to show or hide the notes on the map.
--- a/internal/journal/text_editor.lua
+++ b/internal/journal/text_editor.lua
@@ -234,7 +234,7 @@ function TextEditor:onInput(keys)
         return self.subviews.scrollbar:onInput(keys)
     end
 
-    if keys._MOUSE_L then
+    if keys._MOUSE_L and self:getMousePos() then
         self:setFocus(true)
     end
 
--- a/internal/journal/text_editor.lua
+++ b/internal/journal/text_editor.lua
@@ -94,6 +94,7 @@ TextEditor.ATTRS{
     select_pen = COLOR_CYAN,
     on_text_change = DEFAULT_NIL,
     on_cursor_change = DEFAULT_NIL,
+    one_line_mode = false,
     debug = false
 }
 
@@ -104,25 +105,27 @@ function TextEditor:init()
         TextEditorView{
             view_id='text_area',
             frame={l=0,r=3,t=0},
-            text = self.init_text,
+            text=self.init_text,
 
-            text_pen = self.text_pen,
-            ignore_keys = self.ignore_keys,
-            select_pen = self.select_pen,
-            debug = self.debug,
+            text_pen=self.text_pen,
+            ignore_keys=self.ignore_keys,
+            select_pen=self.select_pen,
+            debug=self.debug,
+            one_line_mode=self.one_line_mode,
 
-            on_text_change = function (val)
+            on_text_change=function (val)
                 self:updateLayout()
                 if self.on_text_change then
                     self.on_text_change(val)
                 end
             end,
-            on_cursor_change = self:callback('onCursorChange')
+            on_cursor_change=self:callback('onCursorChange')
         },
         widgets.Scrollbar{
             view_id='scrollbar',
             frame={r=0,t=1},
-            on_scroll=self:callback('onScrollbar')
+            on_scroll=self:callback('onScrollbar'),
+            visible=not self.one_line_mode
         }
     }
     self:setFocus(true)
@@ -251,6 +254,7 @@ TextEditorView.ATTRS{
     on_cursor_change = DEFAULT_NIL,
     enable_cursor_blink = true,
     debug = false,
+    one_line_mode = false,
     history_size = 10,
 }
 
@@ -273,12 +277,22 @@ function TextEditorView:init()
         bold=true
     })
 
+    self.text = self:normalizeText(self.text)
+
     self.wrapped_text = wrapped_text.WrappedText{
         text=self.text,
         wrap_width=256
     }
 
     self.history = TextEditorHistory{history_size=self.history_size}
+end
+
+function TextEditorView:normalizeText(text)
+    if self.one_line_mode then
+        return text:gsub("\r?\n", "")
+    end
+
+    return text
 end
 
 function TextEditorView:setRenderStartLineY(render_start_line_y)
@@ -410,7 +424,7 @@ end
 
 function TextEditorView:setText(text)
     local changed = self.text ~= text
-    self.text = text
+    self.text = self:normalizeText(text)
 
     self:recomputeLines()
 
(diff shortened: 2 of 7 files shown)


notes.lua
```diff
@@ -0,0 +1,322 @@
...
+                    focus_path='notes/comment',
+                    frame_style=gui.FRAME_INTERIOR,
+                    init_text=self.note and self.note.comment or ''
+                },
+                widgets.Panel{
+                    view_id='buttons',
+                    frame={b=0,h=3},
+                    autoarrange_subviews=true,
+                    subviews={
+                        edit_mode and widgets.TextButton{
```
myk002, line 185: For footer elements, we usually use `HotkeyLabel`. `TextButton` is intended for use in places where there are no other identifying markings that the widget is added by DFHack, such as in overlay panels that don't have room for a branded frame (e.g. the search widget on the Tasks panel).

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+                    }
+                }
+            },
+        },
+    }
+end
+
+function NoteManager:createNote()
+    local x, y, z = pos2xyz(df.global.cursor)
+    if x == nil then
```
myk002, line 216: use `guidm.getCursorPos()` instead of accessing `df.global.cursor` directly. That will get you `nil` if the cursor is invalid. `df.global.cursor.x` is *never* `nil`, so this check will not work as written.

https://github.com/DFHack/dfhack/blob/develop/library/lua/gui/dwarfmode.lua#L40

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+        print('Enable keyboard cursor to add a note.')
```
myk002, line 217: A mouse-capable way of adding a note is generally expected of current tools. See `gui/blueprint` for a possible example. You click on a button and then click on the map. while the location is being selected, a marker is rendered under the mouse cursor: https://github.com/DFHack/scripts/blob/master/gui/blueprint.lua#L427

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+        return
+    end
+
+    local name = self.subviews.name:getText()
+    local comment = self.subviews.comment:getText()
+
+    if #name == 0 then
+        print('Note need at least a name')
```
myk002, line 225: even though this is never going to be executed (since the widget is disabled when `#name == 0`), you should use `dfhack.printerr` for these messages instead of just `print`

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+    self:dismiss()
+end
+
+function NoteManager:deleteNote()
+    if self.note == nil then
+        return
+    end
+
+    for ind, note in pairs(df.global.plotinfo.waypoints.points) do
+        if note == self.note then
```
myk002, line 282: be careful about comparing pointers. it doesn't always behave the way you expect (e.g. when used as keys in maps). For tests like these, the common practice in DF/DFHack is to compare the IDs, not the pointers. `if note.id == self.note.id`

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+}
+
+local function main(args)
+    if #args == 0 then
+        return
+    end
+
+    if args[1] == 'add' then
+        local x = pos2xyz(df.global.cursor)
+        if x == nil then
```
myk002, line 311: same as above -- use `guidm.getCursorPos()` for the behavior you expect here

myk002 (commented):
I'm unsure whether this should be its own tool or whether it should be a tab in `gui/journal`. Maybe that could take some more discussion on Discord.

One thing that is missing from this interface is a way to search existing notes by name.

internal/journal/text_editor.lua
```diff
@@ -234,6 +234,10 @@ function TextEditor:onInput(keys)
         return self.subviews.scrollbar:onInput(keys)
     end
 
+    if keys._MOUSE_L then
```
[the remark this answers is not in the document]
wiktor-obrebski: nice, I wondered how to limit this

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+    self:reloadVisibleNotes()
+end
+
+function NotesOverlay:overlay_trigger(args)
+    return self:showNoteManager()
+end
+
+function NotesOverlay:onInput(keys)
+    if keys._MOUSE_L then
+        local pos = dfhack.gui.getMousePos()
```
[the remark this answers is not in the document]
wiktor-obrebski: nice

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+                },
+                text_editor.TextEditor{
+                    view_id='name',
+                    focus_path='notes/name',
+                    frame={t=1,h=3},
+                    frame_style=gui.FRAME_INTERIOR,
+                    frame_style_b=nil,
+                    init_text=self.note and self.note.name or ''
+                },
+                widgets.HotkeyLabel {
```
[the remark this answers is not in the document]
wiktor-obrebski: (same reasonic live above: https://github.com/DFHack/scripts/pull/1287#discussion_r1742403806)

notes.lua
wiktor-obrebski: it is `nil` thanks to `pos2xyz` - but your solution is better.

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+        and self.note_manager.note
+        or nil
+
+    self.last_click_pos = click_pos
+
+    local notes = df.global.plotinfo.waypoints.points
+
+    local last_note_on_pos = nil
+    local first_note_on_pos = nil
+    for _, note in pairs(notes) do
```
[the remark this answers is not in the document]
wiktor-obrebski: I optimized it now in different way - I scan only notes that are visible (and cached before). What do you think?

notes.lua
wiktor-obrebski: seems very good, but for now we do not have any `notes` interface where such button can be placed.
Do you think I should replace `notes add` command by mouse cursor selection?
Or this should stay as it is and the new way should appear after there will be a `gui/notes` for managing notes?

notes.lua
```diff
@@ -0,0 +1,322 @@
...
+function NoteManager:init()
+    local edit_mode = self.note ~= nil
+
+    self:addviews{
+        widgets.Window{
+            frame={w=35,h=20},
+            frame_inset={t=1},
+            autoarrange_subviews=false,
+            subviews={
+                widgets.HotkeyLabel {
```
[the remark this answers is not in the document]
myk002: it gets worse when the window is resized larger (I forgot to mention this -- the window should be resizable). It looks very strange when hovering over a tile 20 tiles away makes the button text highlighted.

notes.lua
myk002: ah, I forgot that pos2xyz had that logic in it

notes.lua
myk002: I think there should be a `gui/notes` entrypoint, but I also think that this dialog should have a way to create (and switch to editing) an additional note

notes.lua
```diff
@@ -0,0 +1,324 @@
...
+            local note = self:clickedNote(pos)
+            if note ~= nil then
+                self:showNoteManager(note)
+            end
+        end
+    end
+end
+
+function NotesOverlay:clickedNote(click_pos)
+    local pos_curr_note = same_xyz(self.last_click_pos, click_pos)
```
myk002, line 53: should protect against `click_pos` being `nil`, which I believe can happen if the player clicks on the screen, but outside of the map area

notes.lua
```diff
@@ -0,0 +1,324 @@
...
+                    frame_style=gui.FRAME_INTERIOR,
+                    init_text=self.note and self.note.point.comment or '',
+                    init_cursor=1
+                },
+                widgets.Panel{
+                    view_id='buttons',
+                    frame={b=0,h=2},
+                    autoarrange_subviews=true,
+                    subviews={
+                        widgets.HotkeyLabel{
```
myk002, line 187: the save/create and delete button would make better use of space if they were on the same line

notes.lua
```diff
@@ -0,0 +1,324 @@
...
+                }
+            },
+        },
+    }
+end
+
+function NoteManager:createNote()
+    local cursor_pos = guidm.getCursorPos()
+    if cursor_pos == nil then
+        print('Enable keyboard cursor to add a note.')
```
myk002, line 223: should use `dfhack.printerr` here too, I think

notes.lua
```diff
@@ -0,0 +1,324 @@
...
+    local pos_curr_note = same_xyz(self.last_click_pos, click_pos)
+        and self.note_manager
+        and self.note_manager.note
+        or nil
+
+    self.last_click_pos = click_pos
+
+    local last_note_on_pos = nil
+    local first_note_on_pos = nil
+    for _, note in ipairs(self.notes) do
```
myk002, line 62: this is now O(N) in the number of notes -- *probably* ok for likely use cases. I just get nervous when I see O(N) algorithms for (technically) unbounded N

notes.lua
wiktor-obrebski: This would be confusing. 
1. You clicked on a note to edit it and in a dialog you have option "New note"? Did you current changes to note saved or not? Do it switch to new note overlay mouse hover mode? etc.

2. You creating a new note, and got to moment when you see this dialog. Then you switch to editing. What this really means? You are forced to select a note on the screen now? Its sounds very confusing.
If you want to switch to editing - just select a note on the map. You can do this right now, do not need any additional wat to "switch" to it.
--
I never met such features in my web/desktop experience. Edit dialog is for editing, new dialog is for creating.


I believe the needs you speak about will be fully solved by `gui/notes` tool.
I though I can create it later, but maybe it should be done from the begining to avoid the confusion.

notes.lua
wiktor-obrebski: but `self.notes` is only notes currently visible on the screen, so would be really hard to make it slow IMO.
I will improve the var name to make it obvious.

wiktor-obrebski: we could consider optimizing `:reloadVisibleNotes` function if you have an idea how to do this.
there, I believe, is weak point which can make e.g. map navigation slow if there will really big # of notes.

notes.lua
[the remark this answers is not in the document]
wiktor-obrebski: done

notes.lua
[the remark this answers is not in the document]
wiktor-obrebski: `auto_width` done

notes.lua
myk002: we can stress test this and see if we need to make any further optimizations

myk002 (commented):
needs docs/notes.rst and New tools line in changelog.txt

myk002:
regarding this PR title: this is a script, not a plugin. DFHack "plugins" are C++

docs/notes.rst
```diff
@@ -0,0 +1,42 @@
+notes
+=====
+
+.. dfhack-tool::
+    :summary: Manage map-specific notes
```
myk002, line 5:
```suggestion
    :summary: Manage map-specific notes.
```

myk002: summaries must end in a period

docs/notes.rst
```diff
@@ -0,0 +1,42 @@
...
+-----
+
+::
+
+    notes add
+
+Add new note in the current position of the keyboard cursor.
+
+Creating a Note
+~~~~~~~~~~~~~~~
```
myk002, line 28: to keep in line with documentation for existing tools, these can be dashes -- making the sections peers of Usage

docs/notes.rst
```diff
@@ -0,0 +1,42 @@
...
+4. Press :kbd:`Alt` + :kbd:`S` to create the note.
+
+Editing or Deleting a Note
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+- Click on the green pin representing the note directly on the map.
+- A dialog will appear, offering options to edit the title or comment, or to delete the note entirely.
+
+Managing Notes Visibility
+~~~~~~~~~~~~~~~~~~~~~~~~~
+- Access the ``DFHack Control Panel`` / ``UI Overlays`` tab.
```
myk002, line 41: can link to
```
`gui/control-panel`
```
instead of
```
``DFHack Control Panel``
```

docs/notes.rst
wiktor-obrebski: In my understanding, this sections are "Usage" explanations. Or should/can be grouped on some other section, as they all describe how to use the too.

myk002 (approved):

docs/notes.rst
myk002: I think it's fine to leave them all as top-level sections. Nothing else has subheaders under Usage

PR #1301, opened 2024-09-12
new tool: immortal-cravings

chdoc:
allow immortals to satisfy their cravings for food and drink


myk002:
this is an awesome name : D

however, I am seeing a parallel with the concept of `idle-crafting` I think it would be better to combine those tools into one script, and as we add more kinds of self-service autonomy, they can become features of the unified script instead of creating a new script for each one.

Would it make sense to rename `idle-crafting` to `autonomy` or `i-want-it-i-do-it` or something that encompasses these related ideas?

chdoc:
I'm not entirely sure. There obviously are parallels, but `idle-crafting` is (essentially) an automation tool while this one is clearly in the "gameplay" category. From the implementation perspective, I would have to think about where the synergies really are - apart from a shared main loop where units with needs get picked up. I don't see a natural place to configure this tool using an overlay. 

From the communication perspective, `idle-crafting` has just been announced as a feature and received a lot of positive attention. I'm not sure immediately renaming it is a good idea.

From the development perspective, every need that we satisfy using a tool requires some amount of research and discussion. And other needs may require entirely different techniques. So maybe it is good to keep this thing separate for now.

Ultimately, I expect `idle-crafting` and related tools to be re-implemented as a C++ plugin at some point. `idle-crafting` currently takes about 0.7% of CPU time in my forts. Not an immediate problem, but still consuming about 7% of the total DFHack allowance. This tools time usage is hardly measurable in forts with one or two resident necromancers.

immortal-cravings.lua
```diff
@@ -0,0 +1,231 @@
...
+
+--- Load the saved state of the script
+local function load_state()
+    -- load persistent data
+    local persisted_data = dfhack.persistent.getSiteData(GLOBAL_KEY, {})
+    enabled = persisted_data.enabled or false
+end
+
+DrinkAlcohol = df.need_type['DrinkAlcohol']
+EatGoodMeal = df.need_type['EatGoodMeal']
```
myk002, line 127:
```suggestion
DrinkAlcohol = df.need_type.DrinkAlcohol
EatGoodMeal = df.need_type.EatGoodMeal
```

immortal-cravings.lua
```diff
@@ -0,0 +1,231 @@
...
+
+---@type integer[]
+watched = {}
+
+threshold = -9000
```
myk002, line 132: these should be local or "protected" globals (`foo = foo or default_foo`). otherwise, they'll get reset whenever the player runs (probably mistakenly) `immortal-cravings`

myk002: `threshold` should definitely be `local` since it's read only. `watched` could be either

immortal-cravings.lua
```diff
@@ -0,0 +1,231 @@
...
+
+---unit loop: check for idle watched units and create eat/drink jobs for them
+local function unit_loop()
+    -- print(('immortal-cravings: running unit loop (%d watched units)'):format(#watched))
+    ---@type integer[]
+    local kept = {}
+    for _, unit_id in ipairs(watched) do
+        local unit = df.unit.find(unit_id)
+        if unit and not (unit.flags1.caged or unit.flags1.chained) then
```
myk002, line 141: opportunity to reduce code nesting here
```lua
        if not unit or unit.flags1.caged or units.flags1.chained then
            goto next_unit
        end
```

myk002: also, unit should be skipped if dead or not active (e.g. left map on a raid)

immortal-cravings.lua
```diff
@@ -0,0 +1,231 @@
...
+            if not idle.unitIsAvailable(unit) then
+                table.insert(kept, unit.id)
+            else
+                --
+                for _, need in ipairs(unit.status.current_soul.personality.needs) do
+                    if need.id == DrinkAlcohol and need.focus_level < threshold then
+                        goDrink(unit)
+                        goto next_unit
```
myk002, line 149: these `goto`s could be replaced with `break`s

immortal-cravings.lua
```diff
@@ -0,0 +1,231 @@
...
+    watched = kept
+    if #watched == 0 then
+        -- print('immortal-cravings: no more watched units, cancelling unit loop')
+        repeatutil.cancel(GLOBAL_KEY .. '-unit')
+    end
+end
+
+---main loop: look for citizens with personality needs for food/drink but w/o physiological need
+local function main_loop()
+    print('immortal-cravings watching:')
```
myk002, line 170: this prints in this function are probably too spammy. the prints for when actions are actually taken is enough

immortal-cravings.lua
```diff
@@ -0,0 +1,231 @@
...
+end
+
+
+
+-- script action
+
+--- Handles automatic loading
+dfhack.onStateChange[GLOBAL_KEY] = function(sc)
+    if sc == SC_MAP_UNLOADED then
+        enabled = false
```
myk002, line 210: maybe add a comment here explaining that repeat-util will cancel the scheduled callbacks for you

chdoc:
It may take until the next weekend for me to get back to this. Busy at work and other plans for the upcoming weekend.

myk002:
no hurry! I have plenty of other PRs to keep me busy

immortal-cravings.lua
```diff
@@ -0,0 +1,231 @@
...
+    return closest
+end
+
+---find a drink
+---@param pos df.coord
+---@return df.item_drinkst|nil
+local function get_closest_drink(pos)
+    local is_good = function (drink)
+        local container = dfhack.items.getContainer(drink)
+        return container and df.item_barrelst:is_instance(container)
```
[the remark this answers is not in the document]
chdoc: Update: no, backpacks are not food storage. The change you suggested was for drinks, but the same check should also be done when looking for meals. ^^

myk002 (commented):
Could you add `immortal-cravings` to the control panel registry (`internal/control-panel/registry.lua`)?

~~Also, could you add a line to the New Tools section of the changelog?~~ didn't see that you already did

docs/immortal-cravings.rst
```diff
@@ -0,0 +1,17 @@
...
+When enabled, this script watches your fort for units that have no physiological
+need to eat or drink but still have personality needs that can only be satisfied
+by eating or drinking (e.g. necromancers). This enables those units to help
+themselves to a drink or a meal when they crave one and are not otherwise
+occupied.
+
+Usage
+-----
+
+``enable immortal-cravings``
```
myk002, line 17:
```suggestion
::

    enable immortal-cravings
```

myk002: looks better this way when rendered

PR #1306, opened 2024-09-14
[gui/journal] migrate TextEditor to use new text navigation keys

myk002:
Corresponds with: ~~https://github.com/DFHack/dfhack/pull/4933~~ https://github.com/DFHack/dfhack/pull/4935

@wiktor-obrebski FYI


myk002:
oh whoops -- this won't work until 50.14 is out. I'll put this on hold

PR #1309, opened 2024-09-24
Silence stuck-worship.lua spam

Bumber64:
Stop `fix/stuck-worship.lua` from spamming the console every time it shuffles prayer.

### DIFF
--- a/docs/fix/dry-buckets.rst
+++ b/docs/fix/dry-buckets.rst
@@ -19,6 +19,7 @@ manually.
 Usage
 -----
 
-::
-
-    fix/dry-buckets
+``fix/dry-buckets``
+    Empty water buckets not currently used in jobs.
+``fix/dry-buckets -q``, ``fix/dry-buckets --quiet``
+    Empty water buckets not currently used in jobs. Don't print to the console.
--- a/changelog.txt
+++ b/changelog.txt
@@ -33,6 +33,7 @@ Template for new versions:
 ## Fixes
 
 ## Misc Improvements
+- `fix/stuck-worship`: reduced console output by default. ``--verbose`` option to print all affected units.
 
 ## Removed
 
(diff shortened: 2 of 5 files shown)


myk002:
In general, I want the tools printing out when they do something so the player can see that enabling the tool is actually working. The problem here, of course, is that fixing prayer needs is *super frequent*. I don't want to remove the output outright, but I'd be open to reducing it somehow.

You could introduce `-q/--quiet` and `-v/--verbose` options. The verbose output could be the current full name output. regular output would be the number of units that were unstuck. quiet output would avoid printing the number of unstuck units if that number is 0. would that be better? the control panel registry could then be updated to pass `-q`.

Bumber64:
> quiet output would avoid printing the number of unstuck units if that number is 0.

Should that not just be the default? If you're running it manually, you can assume no output means 0 units. If it's running automatically, then we already want to suppress 0.

myk002:
The behavior I suggest (that most other tools conform to, but not `fix/dry-buckets`) is to print out how many things were affected/changed by default, even if it's 0. This is to give the player some feedback that it did what it was supposed to do and didn't just fail.

This is, of course, not required when it is run from `repeat`, so the control panel is configured to pass `-q` to many tools to not print out anything when zero things happened.

Bumber64:
pre-commit.ci autofix

docs/fix/dry-buckets.rst
```diff
@@ -19,6 +19,7 @@ manually.
 Usage
 -----
 
-::
-
-    fix/dry-buckets
+``fix/dry-buckets``
```
myk002, line 22:
```suggestion
::

    fix/dry-buckets [<options>]

Examples
--------

``fix/dry-buckets``
```

myk002: this would normally have an "Options" section as well, but I don't think it's strictly required here as there's only one

myk002 (approved):

PR #1311, opened 2024-09-28
Add realistic-melting tweak to control-panel registry

Birkow:
Add realistic-melting tweak from https://github.com/DFHack/dfhack/pull/4958 to control-panel registry

### DIFF
--- a/internal/control-panel/registry.lua
+++ b/internal/control-panel/registry.lua
@@ -125,6 +125,8 @@ COMMANDS_BY_IDX = {
     {command='partial-items', help_command='tweak', group='gameplay', mode='tweak', default=true,
         desc='Displays percentages on partially-consumed items like hospital cloth.'},
     {command='pop-control', group='gameplay', mode='enable'},
+    {command='realistic-melting', help_command='tweak', group='gameplay', mode='tweak', default=false,
+        desc='Adjust selected item types melt return for all metals to ~95% of forging cost. Reduce melt return by 10% per wear level. Affects weapons, shields, armor parts, tools, and trap components.'},
     {command='starvingdead', group='gameplay', mode='enable'},
     {command='timestream', group='gameplay', mode='enable'},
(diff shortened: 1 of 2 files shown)


internal/control-panel/registry.lua
```diff
@@ -125,6 +125,8 @@ COMMANDS_BY_IDX = {
     {command='partial-items', help_command='tweak', group='gameplay', mode='tweak', default=true,
         desc='Displays percentages on partially-consumed items like hospital cloth.'},
     {command='pop-control', group='gameplay', mode='enable'},
+    {command='realistic-melting', help_command='tweak', group='gameplay', mode='tweak', default=false,
```
myk002, line 128: you can leave off the `default=false` part -- we only add it if it's set to `true`

internal/control-panel/registry.lua
```diff
@@ -125,6 +125,8 @@ COMMANDS_BY_IDX = {
...
+        desc='Adjust selected item types melt return for all metals to ~95% of forging cost. Reduce melt return by 10% per wear level. Affects weapons, shields, armor parts, tools, and trap components.'},
```
myk002, line 129: this is pretty long -- could you open `gui/control-panel` and make sure this fits in the help area without scrolling?

Birkow: Does not fit without scrolling, Sentence with "Affects weapons, shields, armor parts, tools, and trap components." has to be scrolled to.
obraz
obraz

myk002: see if you can trim the description down so it fits in the two-line space at the default (minimum) `gui/control-panel` window size

otherwise, scroll keys start acting funny

Birkow: Ok, removed last sentence.

internal/control-panel/registry.lua
Birkow: ok, removed

myk002 (approved):

PR #1326, opened 2024-10-08
Add a display for melt remainders

gamma-delta:
Hello, I wrote this script that adds an overlay for how much metal a smelter has from "Melt an Item" jobs.

If there's 0 or 1 melt remainder types, it'll say what it is and how much; otherwise it will tell you the number of metals inside. You can click for a full view of all the metals and their remainders.

This is my first foray into DFHack-ing, so please give me feedback! (Especially, I can't figure out how to use the actual material name, so I'm just using its internal raws name. Also, I bet it doesn't handle divine metals properly, but I only have a few divine metal items in my fort and I'm using them to keep my monarch happy, so I can't test it :sweat_smile: )


gamma-delta:
pre-commit.ci autofix

gui/melt-remainder.lua
```diff
@@ -0,0 +1,119 @@
...
+]====]
+
+local overlay = require("plugins.overlay")
+local widgets = require("gui.widgets")
+-- the existence of this is not documented anywhere of course
+local dialogs = require("gui.dialogs")
+
+ENABLED = ENABLED or false
+
+local function i_hate_lua(tbl)
```
quietust, line 23: This isn't really an appropriate place to express your opinions on the scripting language DFHack uses.

gui/melt-remainder.lua
```diff
@@ -0,0 +1,119 @@
...
+    worst_language = worst_language + 1
+  end
+  return worst_language
+end
+
+local function get_melt_remainders(smelter)
+  if not smelter.melt_remainder then return nil end
+  local fractions = {}
+  local mat_count = #df.global.world.raws.inorganics
+  for i = 0, mat_count - 1 do
```
myk002, line 35: although `#smelter.melt_remainder` *should* be the same as `#df.global.world.raws.inorganics`, it would be more consistent to define `mat_count = #smelter.melt_remainder` since you are indexing into `smelter.melt_remainder` in the loop.

gui/melt-remainder.lua
```diff
@@ -0,0 +1,119 @@
...
+    local melt_frac = smelter.melt_remainder[i]
+    if melt_frac > 0 then
+      fractions[i] = melt_frac
+    end
+  end
+  return fractions
+end
+
+-- lua doesn't hoist functions nerd emoji
```
myk002, line 44: this is true, but the comment doesn't add value

gui/melt-remainder.lua
```diff
@@ -0,0 +1,119 @@
...
+local function popup_full_list()
+  local workshop = dfhack.gui.getSelectedBuilding(true)
+  if not workshop then return end
+  local rems = get_melt_remainders(workshop)
+  if not rems then return end
+
+  printall(rems)
+  local lines = {}
+  for mat_id, tenths in pairs(rems) do
+    local mat_name = df.global.world.raws.inorganics[mat_id].id
```
myk002, line 54: you should probably be using the material lookup functions for this: https://docs.dfhack.org/en/stable/docs/dev/Lua%20API.html#material-info-lookup

e.g. `dfhack.matinfo.decode(0, mat_id)`

then you can get the in-game name from the returned object.

gui/melt-remainder.lua
```diff
@@ -0,0 +1,119 @@
...
+    table.insert(lines, mat_name .. ": " .. (tenths * 10) .. "%\n")
+  end
+  if #lines == 0 then
+    table.insert(lines, "<There were no melt remainders>")
+  end
+  dialogs.DialogScreen{
```
myk002, line 60: `DialogScreen` is lower level than you need to be here. The `dialogs.showMessage()` function would be more appropriate. https://github.com/DFHack/dfhack/blob/develop/library/lua/gui/dialogs.lua#L121

gui/melt-remainder.lua
```diff
@@ -0,0 +1,119 @@
...
+  desc = "Displays the fractions of a complete bar 'stored' in the smelter by melting",
+  default_pos = { x = -39, y = 41 },
+  version = 1,
+  default_enabled = true,
+  viewscreens = {
+    'dwarfmode/ViewSheets/BUILDING/Furnace/Smelter/Tasks',
+    'dwarfmode/ViewSheets/BUILDING/Furnace/MagmaSmelter/Tasks',
+  },
+  frame = { w = 58, h = 1 },
+  visible = function() return ENABLED end,
```
myk002, line 77: you can take this out. overlay widget visibility is controlled by the framework

gui/melt-remainder.lua
```diff
@@ -0,0 +1,119 @@
...
+
+function isEnabled()
+  return ENABLED
+end
+if dfhack_flags.enable then
+  ENABLED = dfhack_flags.enable_state
+  return
+end
+
+print("gui/melt-remainder is " .. (ENABLED and "enabled" or "disabled") .. ".")
```
myk002, line 119: when this script is run by name, it should check to see if it is in an appropriate context (a smelter is selected). If not in an appropriate context, `qerror`. if a smelter is selected, it should show the dialog with the remainder info.

PR #1329, opened 2024-10-21
Add `devel/export-map` script

timothymtorres:
This adds a Lua tool that exports all the tile data in a fortress map to a JSON file. It's based on [export-map.lua](https://github.com/mikerenfro/df-map-export/blob/main/export-map.lua) by @mikerenfro that I rewrote. I used his [repo](https://github.com/mikerenfro/df-map-export) as my starting point while I was looking for DFHack scripts to export tile data.

With this script you can select tile flags and attributes to be copied and extracted:
- tiletype
- shape
- special
- variant
- hidden
- light
- subterranean
- outside
- aquifer
- material
- evilness
- liquid
- flow

Depending on the options selected, the material tables will be included as KEYS to reference the number IDs. Also big shoutout to @quietust for helping me find the aquifer data.


timothymtorres:
pre-commit.ci autofix

timothymtorres:
pre-commit.ci autofix

timothymtorres:
pre-commit.ci autofix

docs/export-map.rst
```diff
@@ -0,0 +1,163 @@
...
+
+::
+
+    export-map [include|exclude] [<options>]
+
+Examples
+--------
+
+``export-map``
+    Exports the fortress map to JSON with ALL data included
```
myk002, line 27: Here and below: `Exports` -> `Export`. These are descriptions of what the command represents, not a description of what happens. Also, please end these sentences in a period.

docs/export-map.rst
```diff
@@ -0,0 +1,163 @@
...
+Required
+--------
+
+When you are using options, you must include one of these settings.
+
+``include``
+    Include only the data listed from options to the JSON (whitelist)
+
+``exclude``
+    Exclude only the data listed from options to the JSON (blacklist)
```
myk002, line 46: with the update to the Usage section, you can remove this section

docs/export-map.rst
```diff
@@ -0,0 +1,163 @@
...
+
+Options
+-------
+
+``help``, ``--help``
+    Shows the help menu
+
```
myk002, line 53: we generally don't document "help" options since the canonical way to get help is `help devel/export-map`

docs/export-map.rst
```diff
@@ -0,0 +1,163 @@
...
+``-u``, ``--underworld``
+    Whether the underworld z-levels will be included
+
+``-e``, ``--evilness``
+    Whether the evilness value will be included in MAP_SIZE table. This only
+    checks the value of the center map tile at ground level and will ignore
+    biomes at the edges of the map.
+
+JSON DATA
```
myk002, line 97: What are the client requirements that led to this output format? The choices made here seem strange to me, and I could use some context.

docs/export-map.rst
```diff
@@ -0,0 +1,163 @@
+export-map
+==========
+
+.. dfhack-tool::
+    :summary: Export fortress map tile data to a JSON file
+    :tags: dev map
+
+WARNING - This command will cause the game to freeze for minutes depending on
+map size and options enabled.
```
myk002, line 9: if the runtime for large maps is untenable, it probably would not be too hard to move the property scanning logic a Lua-accessible function of `probe` and just use Lua for the output formatting (which is a pain to do in C++).

myk002 (commented):
please also add a line to the New Tools section of changelog.txt: https://github.com/DFHack/scripts/blob/master/changelog.txt#L29

docs/export-map.rst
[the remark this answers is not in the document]
timothymtorres: Applied with https://github.com/DFHack/scripts/pull/1329/commits/c1857eedb6fd0c43ee99c946f04bb583e1857712

docs/export-map.rst
```diff
@@ -0,0 +1,163 @@
+export-map
+==========
+
+.. dfhack-tool::
+    :summary: Export fortress map tile data to a JSON file
```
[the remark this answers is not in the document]
timothymtorres: Applied via https://github.com/DFHack/scripts/pull/1329/commits/041d415c9169157bc3e309721d9bb33150aea514

docs/export-map.rst
timothymtorres: Applied via https://github.com/DFHack/scripts/pull/1329/commits/035a515d3badd24d7e23b6aa5eafed5326de2d8e

docs/export-map.rst
timothymtorres: Removed via https://github.com/DFHack/scripts/pull/1329/commits/e77e63c1dd5685d0dc7a4c2d5ecedb61b75899f8

docs/export-map.rst
timothymtorres: Removed via https://github.com/DFHack/scripts/pull/1329/commits/0082af75ec6e3d69bf3f023941279f6e9c417746

docs/export-map.rst
timothymtorres: So right now I am taking the JSON data and importing it via another language called DM that is built around a game called Space Station 13. The conversion code is here:

https://github.com/timothymtorres/tgstation/blob/b19d164b370a7408b2ae0ec10c88553ba34a33a5/code/modules/admin/verbs/map_export.dm#L316-L578

The end goal here is to import DF maps directly without having to write my own terrain generation from scratch. I'm aiming to mod SS13 heavily enough so it can become a multiplayer version of DF.

This is what it looks like when you convert a DF map to SS13:

jA2y2MfDgz

docs/export-map.rst
timothymtorres: To export a DF map to JSON that is 192x192 with 66 z-levels and using all options took my computer roughly 7 minutes. A large part of the time cost is write operations since the JSON file is massive.

I don't think it's a big deal because this is a rare tool that will be used sparingly. People might want to export their maps to some 3D modeling software but it's a one and done type of deal.

docs/export-map.rst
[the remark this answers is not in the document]
timothymtorres: Removed fortress map desc via https://github.com/DFHack/scripts/pull/1329/commits/dd6b878e28fa710514b1db01bc13e592bc2eada3

timothymtorres:
pre-commit.ci autofix

docs/export-map.rst
myk002: I meant what led to this particular output format? I've never seen this style before. For example:
```
``ARGUMENT_OPTION_ORDER``
    The order of the selected options for how data is arranged at a map
    position.

    Example 1:
        ``{"material": 1, "shape": 2, "hidden": 3}``

        ``map[z][y][x] = {material_data, shape_data, hidden_data}``

    Example 2:
        ``{"variant": 3, "light": 1, "outside": 2, "aquifer": 4}``

        ``map[z][y][x] = {light_data, outside_data, variant_data, aquifer_data}``
```

Why not have string keys instead of a dynamic-but-consistent ordering of elements? Doesn't that require undue complexity in the client? I'd expect something like:

```
map[z][y][x] = {light=light_data, outside=outside_data, variant=variant_data, aquifer=aquifer_data}
```
and not have an "argument option order" at all.

docs/devel/export-map.rst
```diff
@@ -0,0 +1,158 @@
...
+
+    Note - when using the ``materials`` option, you need to pair the [number ID]
+    with the correct ``KEYS`` material table. Generally you use ``tiletype``
+    option as a helper to sort tiles into different material types. I would
+    recommend consulting ``tile-material.lua`` to see how materials are sorted.
+
+``map``
+    JSON map data is arranged as: ``map[z][y][x] = {tile_data}``
+
+    JSON maps start at index [1]. (starts at map[1][1][1])
```
myk002, line 133: I don't think this is true. It's **Lua** lists that start at [1]. JSON arrays are zero-indexed.

docs/export-map.rst
timothymtorres: > I've never seen this style before.

I sort of wrote it on the fly, but I'm sure other libraries have used a similar technique.

_Initially_ I considered using keys but the problem is time. I/O operations like writing to and reading from JSON are expensive in any language, especially with large amounts of data. Using dynamic positional arguments is a compression method for our data so we don't have to write as many bytes of data. I could have also used a library like [messagepack ](https://msgpack.org/) to really squeeze the data together even further, but then it would become even more unreadable. It's also why I use numbers for most of the data instead of strings but provide a matching key table to convert the numbers to strings.

Anyway math time!

Using the example: `map[z][y][x] = {light_data, outside_data, variant_data, aquifer_data}`
- JSON with positional args - `{true, true, 5, 2}` ~7 bytes of data
- JSON with string keys - `{light=true, outside=true, variant=5, aquifer=2}` ~38 bytes of data

##### _Assuming true/false is 1 byte._
In our example string keys results in 5 times slower performance. Considering how big maps are, this is a big deal. This only considers _exporting_ the data. Importing runs into the same performance cost, so really it's x2. If you export 7 bytes of data, you need to import 7 bytes of data as well. 

tldr; - It's a compression method so data processing is significantly faster.

docs/devel/export-map.rst
timothymtorres: Interesting. 

The JSON library I was using in DM to import the data counts the arrays starting at 1. They probably did that for compatibility reasons since DM arrays start at [1]. Thought it was universal.

Let me tweak the wording to reference that.

timothymtorres: Applied via https://github.com/DFHack/scripts/pull/1329/commits/9338977b7965c5a5d13020ddc1d4bd555affce95

docs/devel/export-map.rst
```diff
@@ -0,0 +1,159 @@
+devel/export-map
+==========
```
myk002, line 2:
```suggestion
devel/export-map
================
```

devel/export-map.lua
```diff
@@ -0,0 +1,332 @@
...
+    local center_x, center_y = math.floor(xmax/2), math.floor(ymax/2)
+    local rgnX, rgnY = dfhack.maps.getTileBiomeRgn(center_x, center_y, ground_z)
+    local biome = dfhack.maps.getRegionBiome(rgnX, rgnY)
+
+    return biome and biome.evilness or 0
+end
+
+local function classify_tile(options, x, y, z)
+    -- The last z-levels of hell shrink their x/y size unexpectedly! (ಠ_ಠ)
+    -- if your map is 190x190, the last hell z-levels are gonna be like 90x90
```
myk002, line 47: not all the time. this is only likely to happen if your map happens to cross a region boundary and different regions are different depths.

devel/export-map.lua
```diff
@@ -0,0 +1,332 @@
...
+        df.global.world.world_data.active_site[0].name
+    )
+    local export_path = "map-exports/" .. fortress_name
+    dfhack.filesystem.mkdir_recursive(export_path)
+    export_all_z_levels(fortress_name, export_path, options)
+end
+
+if dfhack_flags.module then
+    return
+end
```
myk002, line 246: this script is not declared as a module, so this does nothing

changelog.txt
```diff
@@ -27,6 +27,8 @@ Template for new versions:
 # Future
 
 ## New Tools
+- `fix/wildlife`: prevent wildlife from getting stuck when trying to exit the map. This fix needs to be enabled manually in `gui/control-panel` on the Bug Fixes tab since not all players want this bug to be fixed.
```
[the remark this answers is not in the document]
timothymtorres: Hmmm strange. Not sure how that got in there? Maybe I was fixing a merge conflict or accidentally c/p it in while I was using a changelog example.

devel/export-map.lua
timothymtorres: Removed via https://github.com/DFHack/scripts/pull/1329/commits/eac50825c0f59f457e725d8c17a528f882a67759

myk002:
When you get a chance, please add your name to the Authors.rst file! https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst

devel/export-map.lua
timothymtorres: Add those extra details via https://github.com/DFHack/scripts/pull/1329/commits/019527f8ee79cd1d5b71b1c91d40d0252aba3ec1

myk002 (approved):

PR #1330, opened 2024-10-27
Persist the last cycle tick in `emigration` for a more consistent gameplay experience

master-spike:
Fixes: https://github.com/DFHack/dfhack/issues/5011

### DIFF
--- a/emigration.lua
+++ b/emigration.lua
@@ -1,17 +1,26 @@
 --@module = true
 --@enable = true
 
+local utils = require('utils')
+
 local GLOBAL_KEY = 'emigration' -- used for state change hooks and persistence
 
-enabled = enabled or false
+local function get_default_state()
+    return {enabled=false, last_cycle_tick=0}
+end
+
+state = state or get_default_state()
 
 function isEnabled()
-    return enabled
+    return state.enabled
 end
 
 local function persist_state()
-    dfhack.persistent.saveSiteData(GLOBAL_KEY, {enabled=enabled})
+    dfhack.persistent.saveSiteData(GLOBAL_KEY, state)
 end
+
+local TICKS_PER_MONTH = 33600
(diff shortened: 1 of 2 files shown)


myk002:
I fixed the formatting of the "Fixes" line in the description so GitHub can understand it and link the issue properly

emigration.lua
```diff
@@ -210,8 +219,9 @@ dfhack.onStateChange[GLOBAL_KEY] = function(sc)
         return
     end
 
-    local persisted_data = dfhack.persistent.getSiteData(GLOBAL_KEY, {enabled=false})
+    local persisted_data = dfhack.persistent.getSiteData(GLOBAL_KEY, {enabled=false,last_cycle_tick=0})
```
myk002, line 222: this will leave `last_cycle_tick` as `nil` in current games with saved state. It would be better to normalize the value here so you can remove the check for `nil` in `event_loop`.

I suggest a pattern like this:

```lua
    state = get_default_state()
    utils.assign(state, dfhack.persistent.getSiteData(GLOBAL_KEY, state))
```

emigration.lua
```diff
@@ -4,13 +4,14 @@
 local GLOBAL_KEY = 'emigration' -- used for state change hooks and persistence
 
 enabled = enabled or false
+last_cycle_tick = last_cycle_tick or 0
```
myk002, line 7: now that we have more than just `enabled` here, I suggest making this clearer by putting the state in a table.

```lua
local function get_default_state()
    return {enabled=false, last_cycle_tick=0}
end

state = state or get_default_state()
```

master-spike:
Tested and works

myk002 (approved):
Just to mention, GitHub loses review state when you force push, so it's easier to review if you leave the commits separate. That is, if you force-push, I lose all review state and have to review from scratch. You can squash at the very end if you want to, but it's not necessary.

PR #1333, opened 2024-11-11
Fix confirm hover instructions

robob27:
The `df.main_hover_instruction` enum was updated [here](https://github.com/DFHack/df-structures/commit/04e2b185f2a2e09388079e7a1f37b7b016dca46f#diff-05529cd0f67c2aa615be427227c4060f584d82ff55924ed7ee5bd917a5849faf). Updating references to this enum in confirm to fix some confirm dialogs not working due to referencing the old enum values.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -38,7 +38,6 @@ Template for new versions:
 - `makeown`: halt any hostile jobs the unit may be engaged in, like kidnapping
 - `fix/loyaltycascade`: allow the fix to work on non-dwarven citizens
 - `control-panel`: fix setting numeric preferences from the commandline
-- `gui/confirm`: fix some confirm prompts not working
 
 ## Misc Improvements
 - `control-panel`: Add realistic-melting tweak to control-panel registry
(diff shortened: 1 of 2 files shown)


myk002:
hrm... another option would be to revert the name changes for the enum. @lethosor what do you think?

ab9rf:
I support staying with the canonical names going forward. Given that we've already made all the effort to move to canonical names, sticking with them going forward seems more reasonable than going back to the noncanonical ones we were using before. And Bay12 will no doubt make further changes, for which having our usage match theirs will make updating that much easier.

ab9rf (approved):

myk002:
fair enough. there is an issue that the names used in df-structures master are now different from those in df-structures adv-beta, but we can get around that with code like 

```lua
    predicate=function() return mi.current_hover == (df.main_hover_instruction.RouteRemove or df.main_hover_instruction.HAULING_REMOVE_ROUTE) end,
```

ab9rf:
We should probably just canonicalize df-structures for 50.14 as well. Shouldn't take long.

changelog.txt
```diff
@@ -38,6 +38,7 @@ Template for new versions:
 - `makeown`: halt any hostile jobs the unit may be engaged in, like kidnapping
 - `fix/loyaltycascade`: allow the fix to work on non-dwarven citizens
 - `control-panel`: fix setting numeric preferences from the commandline
+- `gui/confirm`: fix some confirm prompts not working
```
myk002, line 41: no need for a changelog since there hasn't been a stable release with this broken yet

myk002, line 41:
```suggestion
```

myk002 (approved):

PR #1339, opened 2024-11-26
Fix permanence of ungeld

NicksWorld:
Geld/ungeld now properly update `unit.body.components.body_part_status[?].gelded`.

Through testing I have discovered the following behavior guiding the implementation:
- On load the `body_part_status[?].gelded` flag will cause `unit.flags3.gelded` to be updated.
- Gelding a unit with a histfig naturally does not cause any wounds to be added to the histfig.

Ungelding will remove all gelding wounds and flags to avoid any conflicts caused by old versions marking all body parts from wounds gelded.

Fixes https://github.com/DFHack/dfhack/issues/4988

### DIFF
--- a/geld.lua
+++ b/geld.lua
@@ -5,11 +5,10 @@ local validArgs = utils.invert({
     'toggle',
     'ungeld',
     'help',
-    'find',
 })
 local args = utils.processArgs({...}, validArgs)
 
-unit=nil
+local unit = nil
 
 if args.help then
     print(dfhack.script_help())
@@ -21,7 +20,8 @@ if args.unit then
     if id then
         unit = df.unit.find(id)
     else
-        qerror("Invalid ID provided.")
+        qerror("Invalid unit ID provided.")
+        return
     end
 else
     unit = dfhack.gui.getSelectedUnit()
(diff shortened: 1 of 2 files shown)


geld.lua
```diff
@@ -21,66 +20,89 @@ if args.unit then
...
+        qerror("Invalid unit ID provided.")
+        return
     end
 else
     unit = dfhack.gui.getSelectedUnit()
 end
 
 if not unit then
     qerror("Invalid unit selection.")
+    return
```
myk002, line 32: `qerror` throws, so a `return` is not necessary. static analyzers don't understand this, though, so they sometimes give a warning.

geld.lua
```diff
@@ -5,11 +5,10 @@ local validArgs = utils.invert({
     'toggle',
```
myk002, line 5: above: `utils =` -> `local utils =`

myk002 (approved):

myk002:
At your convenience, please consider adding your name to our Authors file: https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst

PR #1340, opened 2024-11-27
Convert age argument to number in rejuvenate.lua

Eldresh:
due to the less than comparison, age must be converted to a number or else any attempt to use that argument causes the script to fail with "attempt to compare string with number".


Eldresh:
I'm still figuring this out and have no idea why it deleted, I'll try this again.

myk002 (commented):
Looks good! Could you also add a line to the Fixes section of `changelog.txt`?

Eldresh:
> Looks good! Could you also add a line to the Fixes section of `changelog.txt`?

I can do that. Thanks for your patience, this is my first PR and I'm still figuring stuff out.

myk002 (approved):

PR #1345, opened 2024-12-04
Add alternative rightclick behavior overlay to `gui/design`

nibirubingus:
This adds the overlay proposed in https://github.com/DFHack/dfhack/issues/5025. Name tentative.


myk002:
since this "script" does nothing when run and is controllable entirely via other interfaces, I think we should attach it to an existing tool and document it with that tool. This is thematically linked with the tooltip overlay, so I'd suggest tacking this code onto `gui/design.lua` after the `DimensionsOverlay` class.

Documentation can be added to `docs/gui/design.rst`. Here's another example of docs for a tool with multiple overlays: https://github.com/DFHack/dfhack/blob/develop/docs/plugins/dig.rst?plain=1#L194

fixrightclick.lua
```diff
@@ -0,0 +1,37 @@
...
+        },
+}
+
+local selection_rect = df.global.selection_rect
+local buildreq = df.global.buildreq
+
+function RightClickWidget:onInput(keys)
+    if keys._MOUSE_R or keys.LEAVESCREEN then
+        -- building mode, do not run if buildingplan.planner is enabled since it already provides this functionality
+        if buildreq.selection_pos.x >= 0 and not overlay.get_state().config['buildingplan.planner'].enabled then
```
myk002, line 24: does it work fine if you just don't check for `buildingplan`? I don't think it matters which tool resets `selection_pos`

myk002: that would be fine, yeah.

docs/gui/design.rst
```diff
@@ -19,7 +19,18 @@ Usage
...
-designating digging. The dimensions show up in a tooltip that follows the mouse
-cursor.
+This tool also provides two overlays that are managed by the `overlay` framework.
+
+dimensions
+~~~~~~~~~~
+
+The ``gui/design.dimensions`` overlay shows the selected dimensions when designating
+with vanilla tools, for example when painting a burrow or designating digging.
+The dimensions show up in a tooltip that follows the mousecursor.
```
myk002, line 29:
```suggestion
The dimensions show up in a tooltip that follows the mouse cursor.
```

gui/design.lua
```diff
@@ -120,8 +120,42 @@ function DimensionsOverlay:preUpdateLayout(parent_rect)
...
+
+RightClickOverlay = defclass(RightClickOverlay, overlay.OverlayWidget)
+RightClickOverlay.ATTRS{
+    desc='When drawing boxes, makes right click cancel selection instead of exiting.',
+    default_enabled=true,
+    viewscreens={
+        'dwarfmode/Designate',
+        'dwarfmode/Burrow/Paint',
+        'dwarfmode/Stockpile/Paint',
+        'dwarfmode/Zone/Paint',
```
myk002, line 135: Ha - I need to handle this case for DimensionsOverlay too. Thanks!

gui/design.lua
```diff
@@ -120,8 +120,42 @@ function DimensionsOverlay:preUpdateLayout(parent_rect)
...
+            selection_rect.start_y = -30000
+            selection_rect.start_z = -30000
+            return true
+        end
+    end
+end
+
 OVERLAY_WIDGETS = {
     dimensions=DimensionsOverlay,
+    rightclick=RightClickOverlay
```
myk002, line 158:
```suggestion
    rightclick=RightClickOverlay,
```

myk002 (approved):

myk002:
At your leisure, we would be honored if you could add your name to https://github.com/DFHack/dfhack/blob/develop/docs/about/Authors.rst

gui/design.lua
myk002: after you made the edit, did you run `:lua require('plugins.overlay').rescan()`? changes to overlay widgets need to be reloaded at the framework level.

PR #1346, opened 2024-12-05
Fix usage of stale state in legends racefilter

NicksWorld:
Loading multiple worlds or timelines in legends mode originally did not reset the state, allowing stale data to be used in filtering. This resulted in incorrect results being presented.

This patch resets `racefilter` state on world or map load to ensure the state remains up to date.

Map loading is checked to prevent issues with `open-legends` which is able to access legends mode without reloading the world.

### DIFF
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -1040,7 +1040,7 @@ dfhack.onStateChange[GLOBAL_KEY] = function(sc)
     end
 
     -- Reset state when a world or map is loaded to ensure data remains current
-    if sc == SC_WORLD_LOADED or sc == SC_MAP_LOADED then
+    if sc == SC_WORLD_LOADED then
         racefilter.reset_state()
     end
 end
--- a/exportlegends.lua
+++ b/exportlegends.lua
@@ -1038,6 +1038,11 @@ dfhack.onStateChange[GLOBAL_KEY] = function(sc)
     if sc == SC_VIEWSCREEN_CHANGED and df.viewscreen_choose_game_typest:is_instance(dfhack.gui.getDFViewscreen(true)) then
         asyncexport.reset_state()
     end
+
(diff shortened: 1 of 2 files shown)


exportlegends.lua
```diff
@@ -1038,6 +1038,11 @@ dfhack.onStateChange[GLOBAL_KEY] = function(sc)
     if sc == SC_VIEWSCREEN_CHANGED and df.viewscreen_choose_game_typest:is_instance(dfhack.gui.getDFViewscreen(true)) then
         asyncexport.reset_state()
     end
+
+    -- Reset state when a world or map is loaded to ensure data remains current
```
myk002, line 1042: why map and not just world? isn't race specific to world?

myk002 (commented):
nice fix -- could you add a line to the changelog?

exportlegends.lua
NicksWorld: Changes can occur between loading the map and opening legends thanks to the `open-legends` tool. As fort mode could easily modify historical data, I thought it safest to ensure that path starts with a reset state.

myk002: this change won't catch that specific case -- if a map is loaded, then the race of a histfig is changed (dangerous, btw!), then `open-legends` is run, this check won't catch it. I'm not certain that it's a case we *really* need to address, though. resetting on world load, though, is *very* important and should absolutely be done here.

myk002: the only reason I'm being picky about this is that maps are loaded and unloaded all the time in adventure mode, and I don't think we should run more than we really have to

NicksWorld: I was originally worried about werebeasts changing the race of histfigs, but after some research that doesn't appear to occur. As we aren't particularly worried about the case of `open-legends`, removing the reset on map load seems safe to me.

myk002 (approved):

PR #1363, opened 2025-01-01
Added new slider widget to `devel/hello-world`

realSquidCoder:
Created `devel/helloSlider.lua` (a prototype for a new single-slider widget)


devel/helloSlider.lua
```diff
@@ -0,0 +1,212 @@
...
+    dc:char(nil, SLIDER_TAB_RIGHT)
+    -- manage dragging
+    if self.is_dragging_target then
+        Slider_do_drag(self, width_per_idx)
+    end
+    if df.global.enabler.mouse_lbut_down == 0 then
+        self.is_dragging_target = nil
+        self.is_dragging_idx = nil
+    end
+end
```
myk002, line 130: This code should be added to https://github.com/DFHack/dfhack/blob/develop/library/lua/gui/widgets/range_slider.lua, with the class passed back in the final return statement, where it can be stored by https://github.com/DFHack/dfhack/blob/develop/library/lua/gui/widgets.lua#L29

Docs for the new widget should go near here: https://github.com/DFHack/dfhack/blob/develop/docs/dev/Lua%20API.rst?plain=1#L6419

devel/helloSlider.lua
```diff
@@ -0,0 +1,212 @@
...
+        },
+        Slider{
+            frame={l=1, t=3},
+            num_stops=#LEVEL_OPTIONS,
+            get_idx_fn=function()
+                return self.subviews.level:getOptionValue()
+            end,
+            on_change=function(idx) self.subviews.level:setOption(idx) end,
+        },
+    }
```
myk002, line 188: this code should be added to https://github.com/DFHack/scripts/blob/master/devel/hello-world.lua, with frame sizes adjusted to accommodate.

devel/hello-world.lua
```diff
@@ -32,6 +42,30 @@ function HelloWorldWindow:init()
             frame={w=10, h=5},
             frame_style=gui.INTERIOR_FRAME,
         },
+        widgets.Divider{
```
myk002, line 45: please set the properties to make it a flat divider instead of a connected divider:
```lua
                frame={h=1},
                frame_style_l=false,
                frame_style_r=false,
```

devel/hello-world.lua
```diff
@@ -32,6 +42,30 @@ function HelloWorldWindow:init()
...
+            frame={l=0,t=3}
+        },
+        widgets.CycleHotkeyLabel{
+            view_id='level',
+            frame={l=0, t=3, w=16},
```
myk002, line 50: setting `t` properties in frames is not useful for autoarranged widgets

devel/hello-world.lua
```diff
@@ -32,6 +42,30 @@ function HelloWorldWindow:init()
...
+            label='Level:',
+            label_below=true,
```
myk002, line 52: for a non-range slider, I think `label_below` would only waste space

devel/hello-world.lua
```diff
@@ -32,6 +42,30 @@ function HelloWorldWindow:init()
...
+            key_back='CUSTOM_SHIFT_C',
+            key='CUSTOM_SHIFT_V',
+            options=LEVEL_OPTIONS,
+            initial_option=LEVEL_OPTIONS[1].value,
+            on_change=function(val)
+                self.subviews.level:setOption(val)
+            end,
```
myk002, line 59: this widget should not be setting its own state in its own callback

devel/hello-world.lua
myk002: however, since 2 lines is way too many to put between the CycleHotkeyLabel and the Slider widget, maybe create a containing Panel that does not have autolayout properties.

myk002:
also needs a changelog entry

devel/hello-world.lua
```diff
@@ -32,6 +42,32 @@ function HelloWorldWindow:init()
...
+            frame={l=0, w=20},
+            label='Level:',
+            label_below=false,
+            key_back='CUSTOM_SHIFT_C',
+            key='CUSTOM_SHIFT_V',
+            options=LEVEL_OPTIONS,
+            initial_option=LEVEL_OPTIONS[1].value,
+            on_change=function(val)
+                self.callback{Slider.on_change(val)}
+            end,
```
myk002, line 61: this on_change handler isn't necessary. it also appears to have a syntax error

devel/hello-world.lua
```diff
@@ -32,6 +42,32 @@ function HelloWorldWindow:init()
...
+        widgets.Divider{
+            frame={h=1},
+            frame_style_l=false,
+            frame_style_r=false,
+        },
+        widgets.CycleHotkeyLabel{
+            view_id='level',
+            frame={l=0, w=20},
+            label='Level:',
+            label_below=false,
```
myk002, line 54: you can remove the `label_below` attribute

realSquidCoder:
@ab9rf this should be good to go if all checks pass. i tested it loacally and it worked without needing change

changelog.txt
```diff
@@ -27,13 +27,13 @@ Template for new versions:
 # Future
 
 ## New Tools
+- `devel/hello-world`: updated to show off the new Slider widget
```
ab9rf, line 30: `devel/hello-world` isn't really a tool, and in the past changes to it have been logged under "Misc Improvements", so please move this to that category

ab9rf (changes requested):

PR #1365, opened 2025-01-04
Create gui/spectate

TymurGubayev:
Configuration UI for the spectate plugin

image


myk002:
I am intrigued and delighted : )

gui/tooltips.lua
```diff
@@ -0,0 +1,281 @@
...
+            label="Follow mouse",
+            key='CUSTOM_ALT_M',
+            on_change=change_follow_mouse,
+        },
+    }
+end
+
+local function GetUnitJob(unit)
+    local job = unit.job
+    if job and job.current_job then
```
myk002, line 64: job is not a pointer and can never be `nil`. you can just test `job.current_job`

myk002:
Let me finish reviewing https://github.com/DFHack/dfhack/pull/4959 so getUnitsInBox no longer returns inactive units

image

docs/gui/tooltips.rst
```diff
@@ -0,0 +1,15 @@
+gui/tooltips
+============
+
+.. dfhack-tool::
+    :summary: Show tooltips with useful info.
+    :tags: fort inspection
+
+This script shows "tooltips" following units and/or mouse with job names.
+
```
myk002, line 9: needs disclaimer that tooltips will show over any vanilla UI elements:
image

The `dig` ascii overlays suffer from the same problem. I don't know of any good solution here.

docs/gui/tooltips.rst
```diff
@@ -0,0 +1,15 @@
+gui/tooltips
+============
+
+.. dfhack-tool::
+    :summary: Show tooltips with useful info.
```
myk002, line 5:
```suggestion
    :summary: Show name and job tooltips near units on map.
```

docs/gui/tooltips.rst
```diff
@@ -0,0 +1,15 @@
...
+    :tags: fort inspection
+
+This script shows "tooltips" following units and/or mouse with job names.
```
myk002, line 8: needs more description of the two options and their effects.

gui/tooltips.lua
```diff
@@ -0,0 +1,281 @@
...
+    if jobName then
+        sb[#sb+1] = ": "
+        sb[#sb+1] = jobName
+    end
+    return table.concat(sb)
+end
+
+local function GetTooltipText(x,y,z)
+    local txt = {}
+    local units = dfhack.units.getUnitsInBox(x,y,z,x,y,z) or {} -- todo: maybe (optionally) use filter parameter here?
```
myk002, line 83: can take a `pos` param to this function and pass `pos, pos` to `getUnitsInBox` -- it supports `pos` params as of https://github.com/DFHack/dfhack/pull/5146

gui/tooltips.lua
```diff
@@ -0,0 +1,281 @@
...
+
+    for _,unit in pairs(units) do
```
myk002, line 85:
```suggestion
    for _,unit in ipairs(units) do
```

gui/tooltips.lua
```diff
@@ -0,0 +1,281 @@
...
+    self:addviews{controls, tooltip}
+end
+
+-- map coordinates -> interface layer coordinates
+function GetScreenCoordinates(map_coord)
+    if not map_coord then return end
+
+    -- -> map viewport offset
+    local vp = df.global.world.viewport
+    local vp_Coord = vp.window_x -- is actually coord
```
myk002, line 165: this field appears to be misnamed -- let me correct that to just `window` (or maybe "corner" since that's how it's named in the DF headers)

https://github.com/DFHack/df-structures/pull/818

gui/tooltips.lua
```diff
@@ -0,0 +1,281 @@
...
+    local map_offset_by_vp = {
+        x = map_coord.x - vp_Coord.x,
+        y = map_coord.y - vp_Coord.y,
+        z = map_coord.z - vp_Coord.z,
+    }
+    -- -> pixel offset
+    local gps = df.global.gps
+    local map_tile_pixels = gps.viewport_zoom_factor // 4;
```
myk002, line 173: this is incorrect for ascii mode:
image


see Gui.cpp::getMousePos:
```cpp
// returns the map coordinates that the mouse cursor is over
df::coord Gui::getMousePos(bool allow_out_of_bounds)
{
    df::coord pos;
    if (gps && gps->precise_mouse_x > -1) {
        pos = getViewportPos();
        if (Screen::inGraphicsMode()) {
            int32_t map_tile_pixels = gps->viewport_zoom_factor / 4;
            pos.x += gps->precise_mouse_x / map_tile_pixels;
            pos.y += gps->precise_mouse_y / map_tile_pixels;
        } else {
            pos.x += gps->mouse_x;
            pos.y += gps->mouse_y;
        }
    }
    if (!allow_out_of_bounds && !Maps::isValidTilePos(pos.x, pos.y, pos.z))
        return df::coord();
    return pos;
}
```

gui/tooltips.lua
```diff
@@ -0,0 +1,281 @@
...
+        local pos = xyz2pos(dfhack.units.getPosition(unit))
+        if not pos then goto continue end
+
+        local scrPos = GetScreenCoordinates(pos)
+        local y = scrPos.y - 1 -- subtract 1 to move the text over the heads
+        local x = scrPos.x + oneTileOffset.x - 1 -- subtract 1 to move the text inside the map tile
+
+        -- to resolve overlaps, we'll mark every coordinate we write anything in,
+        -- and then check if the new tooltip will overwrite any used coordinate.
+        -- if it will, try the next row, to a maximum offset of 4.
```
myk002, line 225: love it. great idea

myk002:
Btw this PR has started some discussion on the Discord server, if you'd like to join: https://discord.com/channels/793331351645323264/807444515194798090/1325754489659195502

gui/tooltips.lua
myk002: the structure field rename PR has been merged

gui/tooltips.lua
TymurGubayev: unit-tooltip starts just above units in ASCII now

docs/gui/tooltips.rst
TymurGubayev: add a bit of clarification...
It's kind of hard to describe though

docs/gui/tooltips.rst
TymurGubayev: added an `**IMPORTANT NOTE**` at the beginning of the description text

gui/tooltips.lua
myk002: you can subscribe to the `testing` or `testing-adv-beta` branch on Steam to get the latest structures and APIs

myk002 (commented):
My feedback:

- this would be better as a fullscreen overlay that could be toggled via a keybinding bound to `dwarfmode/Default` and `dungeonmode/Default` ([file](https://github.com/DFHack/dfhack/blob/develop/data/init/dfhack.keybindings.init)). It could also be toggleable from `gui/tooltips` (which becomes just the configuration interface instead of the rendering mechanism). I don't want to promote leaving a windowed DFHack tool on the screen indefinitely because of https://github.com/DFHack/dfhack/issues/4183
- the enabled state should be managed by the overlay framework. this means that whether the tooltips are displayed is a per-player global setting. The tooltip behavior/contents configuration would likewise be global, stored in `dfhack-config/tooltips.json`

There were some reservations about the name `gui/tooltips` -- I'll get more feedback from people about that on Discord.

not required for this PR but maybe consider as a future enhancement:

- the configuration for which elements to display could be a matrix of checkboxes so players could choose any combination of information for the tooltip and the mouseover.

TymurGubayev:
> * the configuration for which elements to display could be a matrix of checkboxes so players could choose any combination of information for the tooltip and the mouseover.

is there a CheckBox control, or what should I use for this?

myk002:
There isn't a super convenient widget for that yet, largely because every place I've needed toggle checkboxes so far, they've been in a list (which doesn't take widgets). See `gui/control-panel` or maybe the prototype code in `gui/manipulator` for examples.

TymurGubayev:
Config UI with current defaults

image

image

gui/tooltips.lua
```diff
@@ -267,7 +269,7 @@ function MouseTooltip:render(dc)
 
     local pos = dfhack.gui.getMousePos()
     local text = GetTooltipText(pos)
-    if #text == 0 then return end
+    if not text or #text == 0 then return end
```
myk002, line 272: not just an ascii mode exception -- dfhack.gui.getMousePos() would return nil in graphics mode too if the mouse cursor is not over the DF window

myk002:
should probably ensure the mouseover panel renders *over* the other floating text:
image

gui/tooltips.lua
```diff
@@ -0,0 +1,446 @@
+-- Show tooltips on units and/or mouse
+
+--@ module = true
+
+local RELOAD = false -- set to true when actively working on this script
+
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local overlay = require('plugins.overlay')
+local ResizingPanel = require('gui.widgets.containers.resizing_panel')
```
myk002, line 10: this doesn't need to be individually imported -- it can already be addressed as `widgets.ResizingPanel`

gui/tooltips.lua
```diff
@@ -0,0 +1,446 @@
...
+
+--------------------------------------------------------------------------------
+
+-- pens are the same as gui/control-panel.lua
+local textures = require('gui.textures')
+local function get_icon_pens()
```
myk002, line 16: this is ugly duplication, but we don't have a good widget API for it yet, so I don't think there's a better option. We can change this later once we have a better way to address this fairly common use case. I see `orders.lua` does some similar duplication for the workshop labor restrictions overlay.

gui/tooltips.lua
TymurGubayev: I couldn't reproduce it in graphics though, probably because of the check just above for the `dfhack.screen.getMousePos()` return value.

myk002 (commented):
as discussed on Discord, could you rename `gui/tooltips` to `gui/spectate` and move the overlay code to `plugins/lua/spectate.lua`? `spectate` itself will need a bit of an overhaul. I can do that or you can if you'd like to.

TymurGubayev:
> should probably ensure the mouseover panel renders _over_ the other floating text
yeah, that might be a good idea :) But how?

myk002:
order of rendering in `render()`

TymurGubayev:
> as discussed on Discord, could you rename `gui/tooltips` to `gui/spectate` and move the overlay code to `plugins/lua/spectate.lua`? `spectate` itself will need a bit of an overhaul. I can do that or you can if you'd like to.

I'll try to do that, just can't promise when.

myk002:
I can overhaul `spectate` in the 51.01-r2 release timeframe. that would give us about a month to get this all finalized and merged

TymurGubayev:
btw, in case anyone is wondering: I haven't forgotten about merging this with `spectate`, I'm just trying to finish the new stuff first.

myk002:
no worries! I haven't done my part on `spectate` yet anyway, so it's not quite ready for you

TymurGubayev:
some things there might deserve to be moved to some more central location:

* `ToggleLabel` is like ToggleHotkeyLabel, but uses icon instead of "On"/"Off" text. Maybe make it an option?
* `pairsByKeys` is a trivial table iterator in key order (with an optional sorting function). I didn't find one; this one I had to write twice (here and in the `plugins/spectate.lua`)
* `rpad` is just `return string.format("%-"..i.."s", s)`... which may be too trivial
* `append` (which should maybe called `concat`) concats two sequences together (which is also trivial enough)

myk002:
I'll see what I can upstream, then I'll clean things up

first upstreamed change: https://github.com/DFHack/dfhack/pull/5296

myk002 (approved):

TymurGubayev:
image

that's a lot of red :)

PR #1374, opened 2025-01-16
new tool: autocheese

chdoc:


chdoc:
pre-commit.ci autofix

docs/autocheese.rst
```diff
@@ -0,0 +1,38 @@
+autocheese
+==========
+
+.. dfhack-tool::
+    :summary: Automatically make cheese using barrels that have accumulated sufficient milk.
+    :tags: fort auto
+
+Cheese making is difficult to automate using work orders, because a single job
+can consume anything from a bucket was a single unit of milk to barrel
+containing up to 100 units of milk.
```
myk002, line 10:
```suggestion
Cheese making is difficult to automate using work orders. A single job
can consume anything from a bucket with a single unit of milk to a barrel
with 100 units of milk. This makes it hard to predict how much cheese will
actually be produced by an automated order.
```

docs/autocheese.rst
```diff
@@ -0,0 +1,38 @@
+autocheese
+==========
+
+.. dfhack-tool::
+    :summary: Automatically make cheese using barrels that have accumulated sufficient milk.
```
myk002, line 5:
```suggestion
    :summary: Schedule cheese making jobs based on milk reserves.
```

myk002: These short descriptions have a target length of ~53 chars.

changelog.txt
```diff
@@ -29,6 +29,7 @@ Template for new versions:
 ## New Tools
 - `fix/stuck-squad`: allow squads and messengers returning from missions to rescue squads that have gotten stuck on the world map
 - `gui/rename`: (reinstated) give new in-game language-based names to anything that can be named (e.g. units, governments, fortresses, or the world)
+- `autocheese`: automatically make cheese using barrels that have accumulated sufficient milk
```
myk002, line 32: needs to be moved up to the current version

autocheese.lua
```diff
@@ -0,0 +1,155 @@
...
+    elseif unit.flags1.caged or unit.flags1.chained then
+        return false
+    elseif unit.military.squad_id ~= -1 then
+        local squad = df.squad.find(unit.military.squad_id)
+        -- this lookup should never fail
+        ---@diagnostic disable-next-line: need-check-nil
+        return #squad.orders == 0 and squad.activity == -1
+    end
+    return true
+end
```
myk002, line 48: shouldn't this function already be in ic? why not call that one to keep the logic unduplicated?

autocheese.lua
```diff
@@ -0,0 +1,155 @@
...
+            if unit_skill > max_skill then
+                max_unit = unit
+                max_skill = unit_skill
+            end
+        end
+    end
+    return max_unit, max_skill
+end
+
+local function findMilkBarrel(min_liquids)
```
myk002, line 75: I don't see any logic to ensure that the milk barrel is reachable from the target workshop

autocheese.lua
```diff
@@ -0,0 +1,155 @@
...
+    for _, container in ipairs(df.global.world.items.other.FOOD_STORAGE) do
+        if
+            not (container.flags.in_job or container.flags.forbid) and
+            container.flags.container and #container.general_refs >= min_liquids
```
myk002, line 79: srsly? they are each individual items and not a stack?

autocheese.lua
```diff
@@ -0,0 +1,155 @@
...
+
+if not reagent then
+    -- print('autocheese: no sufficiently full barrel found')
+    return
+end
+
+local workshop = findWorkshop()
+
+if not workshop then
+    print('autocheese: no workshop available')
```
myk002, line 133:
```suggestion
    print('autocheese: no Farmer's Workshop available')
```

autocheese.lua
```diff
@@ -0,0 +1,155 @@
...
+            end
+        end
+    end
+end
+
+function findWorkshop()
+    for _,workshop in ipairs(df.global.world.buildings.other.WORKSHOP_FARMER) do
+        if
+            not workshop.profile.blocked_labors[df.unit_labor.MAKE_CHEESE] and
+            #workshop.jobs == 0 and #workshop.profile.permitted_workers == 0
```
myk002, line 97: what if the permitted worker is free to take the cheese making job?

autocheese.lua
```diff
@@ -0,0 +1,155 @@
...
+
+local worker, skill = findAvailableLaborer(df.unit_labor.MAKE_CHEESE, df.job_skill.CHEESEMAKING, workshop)
+if not worker then
+    print('autocheese: no cheesemaker available')
+    return
+end
+local job = makeCheese(reagent, workshop)
+
+print(('autocheese: dispatching cheesemaking job for %s (%d milk) to %s'):format(
+    dfhack.items.getDescription(reagent, 0),
```
myk002, line 145:
```suggestion
    dfhack.df2console(dfhack.items.getReadableDescription(reagent)),
```

autocheese.lua
chdoc: The answer is that these functions do subtly different things:
1. The one for `idle-crafting` will not interrupt social activities
2. This one will also check for caged and chained units, which is handled elsewhere in `idle-crafting`. 

The good solution, after this is merged, is probably to have a function in the C++ `Units` module that will take a few flags to disable or enable optional checks.

autocheese.lua
chdoc: Good catch! I suppose the right fix is to only consider workshops that the selected (sufficiently full) barrel can be brought to. After all, the full barrel is the actual "trigger".

autocheese.lua
chdoc: Why would they be stacks? The game creates them as individual items during a milk job. And then they just get transferred from a bucket to a barrel. The game also keeps around thousands of blocks inside constructions for no reason at all -- other than to mess up the stocks screen, I suppose.

autocheese.lua
chdoc: That is a good point. While having a dedicated cheese maker sounds strange to me, the tool should probably not fail if somebody does this, or otherwise assigns dedicated workers to all farmer's workshops. However, unless the dedicated cheese maker is kept on standby, there will be cycles missed; the approach taken inherently requires the unit to be assigned at the same moment as the job is created. But then, I suppose there is no harm in missing a barrel at 50 units and only creating cheese once you have 60 or 100.

autocheese.lua
chdoc: You successfully "tricked" me into blindly accepting this one. ALWAYS check for easy errors.  :roll_eyes:

chdoc:
Didn't have a whole lot of time to actually play/test tonight, so only did some minimal testing before pushing.

chdoc:
I'm not sure the merge was necessary, but I merged locally anyway. I also quickly checked that it still works with v51.02

changelog.txt
```diff
@@ -49,6 +51,7 @@ Template for new versions:
 
 ## New Tools
 - `fix/stuck-squad`: allow squads and messengers returning from missions to rescue squads that have gotten stuck on the world map
+- `gui/rename`: (reinstated) give new in-game language-based names to anything that can be named (e.g. units, governments, fortresses, or the world)
```
myk002, line 54:
```suggestion
```

myk002 (approved):

PR #1380, opened 2025-01-26
Fix df.talk_choice_type def not being up to date for internal/convo.lua

Crystalwarrior:
This also fixes the extra options not appearing in the conversation menu in the "Ask directions" screen

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -27,12 +27,12 @@ Template for new versions:
 # Future
 
 ## New Tools
-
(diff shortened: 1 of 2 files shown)


myk002 (commented):
Looks good. Could you add a changelog entry?

Crystalwarrior:
@myk002 ah heck I forgot 😭
Can you make a pr template for this repository so any contributors get immediately reminded of that standard? In the meantime I'll only be able to make the change log in like 13 or so hours.

myk002 (approved):

PR #1383, opened 2025-01-29
Fix some EditField hotkey conflicts

ChrisJohnsen:
Other EditField conflicts will be posted in an issue, but these seemed small- and effective-enough to put into commits.

The encapsulation violation for the CivBox fix isn't exactly pleasant, but it would probably be much more invasive to get EditField to support post-init reconfiguration.

### DIFF
--- a/gui/autochop.lua
+++ b/gui/autochop.lua
@@ -239,6 +239,7 @@ function Autochop:init()
                 self:update_choices()
             end,
             visible=is_not_minimal,
+            disabled=function() return self.subviews.target.focus end,
         },
         widgets.Label{
             view_id='summary',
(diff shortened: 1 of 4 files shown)


ChrisJohnsen:
Should the changelog have separate entries for each of the fixes? Or maybe just one entry to cover everything once the other UIs mentioned in DFHack/dfhack#5226 are also addressed? "Restore (or show when unavailable) some UI hotkeys that were previously masked by new text-editing features."

ChrisJohnsen:
The consensus on Discord seems to be in favor of adopting new UI hotkeys to let the "conflicting keys" be used uniformly as editing keys. Closing this since that probably applies to these cases too. The CivBox change (f2987e2df700ea3049880a46e0a50a6956cfdd51) can always be cherry picked later.

PR #1392, opened 2025-02-03
add stub gui/spectate

myk002:
@TymurGubayev here's the stub for the config UI for https://github.com/DFHack/scripts/pull/1365


TymurGubayev:
Is this published in Steam's testing branch?

PR #1394, opened 2025-02-05
Update advfort.lua

velanos:
If you just remove that check advfort does actually start in steam version but I still cannot mine with it

### DIFF
--- a/gui/advfort.lua
+++ b/gui/advfort.lua
@@ -1837,4 +1837,7 @@ function usetool:onRenderBody(dc)
     self:shopMode(self:isOnBuilding())
     self:renderParent()
 end
+if not (dfhack.gui.getCurFocus()[1]=="dungeonmode/Look" or dfhack.gui.getCurFocus()[1]=="dungeonmode/Default") then
+    qerror("This script requires an adventurer mode with (l)ook or default mode.")
+end
 usetool():show()
--- a/gui/advfort.lua
+++ b/gui/advfort.lua
@@ -97,7 +97,7 @@ function reverseRaceLookup(id)
 end
 function deon_filter(name,type_id,subtype_id,custom_id, parent)
     --print(name)
-    local adv=df.global.world.units.active[0]
+    local adv=dfhack.world.getAdventurer()
     local race_filter=build_filter[reverseRaceLookup(adv.race)]
     if race_filter then
         if race_filter.forbid_all then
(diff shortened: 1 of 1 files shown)


Bumber64:
All instances of `df.global.world.units.active[0]` need to be replaced with `dfhack.world.getAdventurer()`.

velanos:
@Bumber64 I believe you I just got here in this project so anyhting you tell me I will do

gui/advfort.lua
```diff
@@ -1837,7 +1837,7 @@ function usetool:onRenderBody(dc)
     self:shopMode(self:isOnBuilding())
     self:renderParent()
 end
-if not (dfhack.gui.getCurFocus()=="dungeonmode/Look" or dfhack.gui.getCurFocus()=="dungeonmode/Default") then
+if not (dfhack.gui.getCurFocus()[1]=="dungeonmode/Look" or dfhack.gui.getCurFocus()[1]=="dungeonmode/Default") then
```
myk002, line 1840: This should be using `dfhack.gui.matchFocusString('dungeonmode/Look') or dfhack.gui.matchFocusString('dungeonmode/Default')`

there are no guarantees that the string you want will be the first one in the list

myk002:
this is a good start for updating `advfort`. It will still be marked as "unavailable" since it doesn't actually function properly, but first steps first.

velanos:
Another update I found out when comparing to advfort 0.47 and probably a major thing: addJobAction does not trigger in the new version. I will continue to try and find out why.

myk002 (approved):

myk002:
> usetool:oninput does not trigger in the new version

are there any errors shown on the console? even though advfort uses the older Screen interface (newer tools use ZScreen for focus handling and draggable, resizable windows), the onInput function should still be called

velanos:
@myk002 gamelog shows no errors or at least I am unable to find which console those errors should be found in it was far more primitive than that namely pasting print('anything') into the logic of this file on 0.47 version on my laptop and the steam version on my pc to see which triggered and oninit does trigger in the old version and not in the new one. Why that is I do not know.

myk002:
If you're on Windows, run the `show` command to get the console. If you're on Linux, run the `./dfhack` launcher script from a terminal. Error messages will be shown there

velanos:
@myk002 
image

PR #1399, opened 2025-02-10
finish updating scripts for the Great Reorg

myk002:
ref: https://github.com/DFHack/df-structures/pull/824


devel/export-dt-ini.lua
myk002, line 1: @cvuchener FYI

myk002: there may be more testing/updates needed

fix/population-cap.lua
```diff
@@ -1,27 +1,30 @@
...
-    }
-    civ_stats = civ.activity_stats
+local function upsize(civ_vec, tasks_vec)
+    local tasks_vec_size = #tasks_vec
+    if #civ_vec < tasks_vec_size then
+        civ_vec:resize(tasks_vec_size)
+    end
 end
 
+upsize(civ_stats.created_weapons, tasks.created_weapons)
```
myk002, line 20: I rewrote the logic here because the original didn't seem to make sense -- `civ.activity_stats` is not a pointer

PR #1409, opened 2025-02-26
Feat/adventurer journal

wiktor-obrebski:
Add adventure mode support for `gui/journal`.
The journal is stored per ID for the character the player created on the save.

It's not support `bodyswap` and other tools - using them u changing ur character, and new one will have new journal!

Also, add dedicated intro text for journal on adventure mode.

I do not see a way to test journal on CI in adventure mode right now.

### DIFF
--- a/docs/gui/journal.rst
+++ b/docs/gui/journal.rst
@@ -6,11 +6,11 @@ gui/journal
     :tags: fort interface
 
 The `gui/journal` interface makes it easy to take notes and document
-important details for the fortresses.
+important details for your fortresses and adventurers.
 
 With this multi-line text editor,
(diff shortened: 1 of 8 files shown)


docs/gui/journal.rst
```diff
@@ -6,11 +6,11 @@ gui/journal
     :tags: fort interface
```
myk002, line 6:
```suggestion
    :tags: adventure fort interface
```

myk002 (approved):
Looks great! I suspect players will want the ability to pull up journals of non-current adventurers, but we need to provide a Lua API for pulling data by key pattern like we have for C++ to support that.

PR #1411, opened 2025-03-01
New Feature: `gym`

realSquidCoder:
Code for dwarves to hit the gym when they yearn for the gains. Assigns dwarves to a military squad until they have fulfilled their need for Martial Training

### DIFF
--- a/gym.lua
+++ b/gym.lua
@@ -1,0 +1,261 @@
+-- Code for dwarves to hit the gym when they yearn for the gains.
+--[====[
+Gym
+=================
+
+Tags: Fort| Needs | BugFix | Units
+
+Assigns Dwarves to a military squad until they have fulfilled their need for Martial Training. Also Passively builds military skills and physical stats.
+
+CRITICAL SETUP:
+01-Minimum 1 squad with the name "Gym"
+02-An assigned squadleader in "Gym"
+03-An assigned Barracks for the squad "Gym"
+04-Active Training orders for the squad "Gym"
+
+This should be a new non military use squad. The uniform should be set to "No Uniform" and the squad should be set to "Constant Training" in the military screen.
+Set the squad's schedule to full time training with at least 8 or 9 training.
+The squad doesn't need months off. The members leave the squad once they have gotten their gains.
+
+NOTE-Dwarfs with the labor "Fish Dissection" enabled are ignored
+Make a Dwarven labour with only the Fish Dissection enabled, set to "Only selected do this" and assign it to a dwarf to ignore them.
+
+Usage
+--------
+
+    Gym [<options>]
+
+Examples
+--------
+
+Gym
+    Current status of script
+
+Gym -start
+    checks to see if you have fullfilled the creation of a training gym
+    searches your fort for dwarves with a need to go to the gym, and begins assigning them to said gym.
+    Once they have fulfilled their need they will be removed from the gym squad to be replaced by the next dwarf in the list.
+
+Gym -stop
+    Dwarves currently in the squad ,with the exception of the squadleader, will be unassigned and no new dwarves will be added to the squad.
+
+Options
+-------
+    -start
+        Starts the script
+        If there is no squad named GYM with a squadleader assigned it will not proceed.
+
+    -stop
+        Stops the script
+
+    -t
+        Use integer values. (Default 3000)
+        The negative need threshhold to trigger for each citizen
+        The greater the number the longer before a dwarf is added to the waiting list.
+]====]
+
+local repeatUtil = require 'repeat-util'
+local utils=require('utils')
+
+validArgs = utils.invert({
+    'start',
+    'stop',
+	't'
+})
+
+local args = utils.processArgs({...}, validArgs)
+local scriptname = "Gym"
+local ignore_flag = 43 -- Fish Dissection labor id
+local ignore_count = 0
+local need_id = 14
+local squadname ="Gym"
+
+
+--######
+--Functions
+--######
+function getAllCititzen()
+	local citizen = {}
+	local my_civ = df.global.world.world_data.active_site[0].entity_links[0].entity_id
+	for n, unit in ipairs(df.global.world.units.all) do
+        if unit.civ_id == my_civ and dfhack.units.isCitizen(unit) then
+            if unit.profession ~= df.profession.BABY and unit.profession ~= df.profession.CHILD then
+                if ( not unit.status.labors[ignore_flag] ) then
+					table.insert(citizen, unit)				
+				else	
+					ignore_count = ignore_count +1 
+				end
+            end
+        end
+	end
+	return citizen
+end
+
+local citizen = getAllCititzen()
+
+function findNeed(unit,need_id) 
+    local needs =  unit.status.current_soul.personality.needs
+    local need_index = -1
+    for k = #needs-1,0,-1 do
+        if needs[k].id == need_id then
+            need_index = k
+            break
+        end
+    end    if (need_index ~= -1 ) then 
+        return needs[need_index]
+    end
+    return nil
+end
+
+--######
+--Main
+--######
+
+function getByID(id) 
+	for n, unit in ipairs(citizen) do
+		if (unit.hist_figure_id == id) then
+			return unit
+		end
+	end
+
+	return nil
+end
+
+-- Find all training squads
+-- Abort if no squads found
+function checkSquads()
+	local squads = {}
+	local count = 0
+	for n, mil in ipairs(df.global.world.squads.all) do
+		if (mil.alias == squadname) then 
+			local leader = mil.positions[0].occupant 
+			if ( leader ~= -1) then
+				table.insert(squads,mil)
+				count = count +1
+			end
+		end
+	end
+
+	if (count == 0) then
+		dfhack.print(scriptname.." | ")
+        dfhack.printerr('ERROR: You need a squad with the name ' .. squadname)
+		dfhack.print(scriptname.." | ")
+		dfhack.printerr('That has an active Squad Leader')
+		dfhack.color(-1)
+		return nil
+	end
+
+	return squads
+end
+
+function addTraining(squads,unit)
+	for n, squad in ipairs(squads) do
+		for i=1,9,1   do
+			if (unit.hist_figure_id == squad.positions[i].occupant) then
+				return true
+			end
+
+			if (unit.military.squad_id ~= -1) then
+				return false
+			end
+
+			if ( squad.positions[i].occupant  == -1 ) then
+				squad.positions[i].occupant = unit.hist_figure_id 
+				return true
+			end
+		end
+	end
+
+	return false
+end
+
+function removeTraining(squads,unit) 
+	for n, squad in ipairs(squads) do
+		for i=1,9,1   do
+			if ( unit.hist_figure_id  == squad.positions[i].occupant ) then
+				unit.military.squad_id = -1
+				unit.military.squad_position = -1
+				squad.positions[i].occupant = -1
+				return true
+			end
+		end
+	end
+	return false
+end
+
+function removeAll(squads) 
+	if ( squads == nil) then return end
+	for n, squad in ipairs(squads) do
+		for i=1,9,1 do
+			local dwarf = getByID(squad.positions[i].occupant)
+			if (dwarf ~= nil) then
+				dwarf.military.squad_id = -1
+				dwarf.military.squad_position = -1
+				squad.positions[i].occupant = -1
+			end
+		end
+	end
+end
+
+
+function check()
+	local squads = checkSquads()
+	local intraining_count = 0
+	local inque_count = 0
+	if ( squads == nil)then return end
+	for n, unit in ipairs(citizen) do
+		local need = findNeed(unit,need_id)
+		if ( need  ~= nil ) then
+			if ( need.focus_level  < threshold ) then
+				local bol = addTraining(squads,unit)
+				if ( bol ) then 
+					intraining_count = intraining_count +1
+				else 
+					inque_count = inque_count +1
+				end
+			else
+				removeTraining(squads,unit)
+			end
+		end
+	end
+
+	dfhack.println(scriptname .. " | IGN: " .. ignore_count .. " TRAIN: " .. intraining_count .. " QUE: " ..inque_count )
+end
+
+
+function start() 
+	threshold = -5000
+	dfhack.println(scriptname ..  " | START")
+	
+	if (args.t) then 
+		threshold = 0-tonumber(args.t) 
+	end
+
+	running = true
+	repeatUtil.scheduleEvery(scriptname,1000,'ticks',check)
+end
+
+function stop()
+	repeatUtil.cancel(scriptname)
+	local squads = checkSquads()
+	removeAll(squads)
+	running = false
+	dfhack.println(scriptname .. " | STOP")
+end
+
+if (args.stop) then
+    if (running) then stop() end
+    return
+end 
+
+if (args.start) then	
+    if (running) then stop() end
+    start()
+    return
+end
+
+if ( running ) then
+    dfhack.println(scriptname .."    | Enabled")
+else
+    dfhack.println(scriptname .."   | Disabled")
+end


chdoc:
Here are some observations:

1. I'm not particularly fond of the name `gym`, because it makes me think of using screw pumps to increase strength. Maybe just call it `martial-training` or `martial-arts-class`
2. I think the squad should just be chosen from a list, instead of requiring the user to name a squad in a particular way. I think that squads are intended to use DF names, and I would not be surprised if the ability to just input strings would disappear. Maybe, if there is only a single squad that does not have a uniform set, default to that.
3. I would like an option to not automatically replace the squad leader. This way you could actually put a decently skilled teacher in charge of the squad. Also, squad leaders are nobles and this would avoid at least some of the historical figure noise.

chdoc:
4. It would be nice to have an option to automatically exclude holders of relevant positions (e.g. the manager ...)

realSquidCoder:
> 4. It would be nice to have an option to automatically exclude holders of relevant positions (e.g. the manager ...)

you can set up ppl to be ignored but auto ignoring ppl is not a bad idea either. who all should be ignored you think?

realSquidCoder:
actually, because i intend this to join the automation tab, perhaps `autotraining` would be a better name

gym.lua
```diff
@@ -0,0 +1,262 @@
...
+            if (unit.hist_figure_id == squad.positions[i].occupant) then
+                return true
+            end
+
+            if (unit.military.squad_id ~= -1) then
+                return false
+            end
+
+            if ( squad.positions[i].occupant  == -1 ) then
+                squad.positions[i].occupant = unit.hist_figure_id
```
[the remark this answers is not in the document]
realSquidCoder: What do you mean here?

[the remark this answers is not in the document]
chdoc: There should be a bidirectional link between the squad and the unit. This only sets the squad half of this link. The other half is in the `military` structure inside the unit. The code is actually reading that structure a few lines above.

gym.lua
```diff
@@ -0,0 +1,262 @@
...
+
+    return nil
+end
+
+-- Find all training squads
+-- Abort if no squads found
+function checkSquads()
+    local squads = {}
+    local count = 0
+    for n, mil in ipairs(df.global.world.squads.all) do
```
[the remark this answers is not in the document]
chdoc: @realSquidCoder did you forget about this issue? As thing stand, I'm afraid that if you retire a fort that makes use of `autotraining`, and start a new fort that also makes use of `autotraining`, the tool will mix up the squads in those two sites.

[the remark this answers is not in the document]
realSquidCoder: I was still figuring out the correct way to check it. Fixed now, afaik.

autotraining.lua
```diff
@@ -0,0 +1,258 @@
...
+    local squads = checkSquads()
+    local intraining_count = 0
+    local inque_count = 0
+    if ( squads == nil)then
+        repeatUtil.cancel(GLOBAL_KEY)
+        state.enabled = false
+        dfhack.println(GLOBAL_KEY  .. " | STOP")
+        return end
+    local citizen = getAllCititzen()
+    for n, unit in ipairs(citizen) do
```
chdoc, line 201:
```suggestion
    for n, unit in ipairs(getAllCititzen()) do
```

autotraining.lua
```diff
@@ -0,0 +1,258 @@
...
+        dfhack.print(GLOBAL_KEY .." was persisted with the following data:\nThreshold: ".. state.threshold .. ' | Squad name: '..state.squadname ..'\n')
+    end
+    persist_state()
+end
+
+
+--######
+--Functions
+--######
+function getAllCititzen()
```
chdoc, line 67: In order to avoid confusions with `getCitizens`, I would rename this function into something like `getTrainingCandidates`.

autotraining.lua
```diff
@@ -0,0 +1,258 @@
...
+    local ret = {}
+    local citizen = dfhack.units.getCitizens(true)
+    ignore_count = 0
+    for _, unit in ipairs(citizen) do
+        if unit.profession ~= df.profession.BABY and unit.profession ~= df.profession.CHILD then
```
chdoc, line 72:
```suggestion
        if dfhack.units.isAdult(unit) then
```

autotraining.lua
```diff
@@ -0,0 +1,258 @@
...
+end
+
+-- Find all training squads
+-- Abort if no squads found
+function checkSquads()
+    local squads = {}
+    local count = 0
+    for _, mil in ipairs(df.global.world.squads.all) do
+        for _, link in ipairs(dfhack.world.getCurrentSite().entity_links) do
+            if mil.entity_id == link.entity_id then
```
chdoc, line 115: The check for the current site can be simplified to:
```suggestion
        if squad.entity_id == df.global.plotinfo.group_id then
```

chdoc (commented):
Making another pass to see what would be needed for a GUI tool

chdoc:
So here is what I think would be needed for the UI tool:

1. extend the state with tables containing the ids of training squads and ignored units. The most convenient to work with is to map unit ids to booleans: `state.ignored[unit.id] = true`. This can be iterated over with `pairs` and used as a test without iteration. However, this requires some care when persisting. See how this is done in [`idle-crafting`](https://github.com/DFHack/scripts/blame/90431556c4c2924c206403232ea1c24491355d07/idle-crafting.lua#L238-L247). It is important to store IDs and not pointers, because both units and squads can disappear between runs.
2. Add a function `persist_state`, because persisting the state is now nontrivial
3. provide public functions for accessing and modifying this new state (calling `persist_state` after modifications)
4. (optional) extend the command-line interface to the new configuration options
5. Make the tool iterate only over configured squads and make it skip over ignored units, adopting the configuration if a configured squad disappears between runs.

Coming to think of it, I'm not entirely certain whether it is more convenient to store `unit.id` or `unit.hist_figure_id` for ignored units. I would leave that decision up to you.

autotraining.lua
```diff
@@ -0,0 +1,254 @@
...
+    return nil
+end
+
+-- Find all training squads
+-- Abort if no squads found
+function checkSquads()
+    local squads = {}
+    local count = 0
+    for _, mil in ipairs(df.global.world.squads.all) do
+        if squad.entity_id == df.global.plotinfo.group_id then
```
realSquidCoder, line 113:
```suggestion
        if mil.entity_id == df.global.plotinfo.group_id then
```

autotraining.lua
```diff
@@ -0,0 +1,254 @@
...
+
+function addTraining(squads,unit)
+    for n, squad in ipairs(squads) do
+        for i=1,9,1   do
+            if (unit.hist_figure_id == squad.positions[i].occupant) then
+                return true
+            end
+
+            if (unit.military.squad_id ~= -1) then
+                return (squad.alias == state.squadname)
```
realSquidCoder, line 144:
```suggestion
                return (df.squad.find(unit.military.squad_id).alias == state.squadname)
```

realSquidCoder:
This is nearly done. I just want to polish a few things before i feel its ready to merge

autotraining.lua
```diff
@@ -0,0 +1,309 @@
...
+        end
+        return false
+    end
+    for _, squad in ipairs(getTrainingSquads()) do
+        for i=1,9,1   do
+            if ( squad.positions[i].occupant  == -1 ) then
+                dfhack.military.addToSquad(unit.id,squad.id,i)
+                -- squad.positions[i].occupant = unit.hist_figure_id
+                -- unit.military.squad_id = squad.id
+                -- unit.military.squad_position = i
```
chdoc, line 213:
```suggestion
```

autotraining.lua
```diff
@@ -0,0 +1,309 @@
...
+end
+
+function removeTraining(unit)
+    for _, squad in ipairs(getTrainingSquads()) do
+        for i=1,9,1   do
+            if ( unit.hist_figure_id  == squad.positions[i].occupant ) then
+                dfhack.military.removeFromSquad(unit.id)
+                -- unit.military.squad_id = -1
+                -- unit.military.squad_position = -1
+                -- squad.positions[i].occupant = -1
```
chdoc, line 229:
```suggestion
```

autotraining.lua
```diff
@@ -0,0 +1,309 @@
...
+    dfhack.println(GLOBAL_KEY  .. " | IGN: " .. ignore_count .. " TRAIN: " .. intraining_count .. " QUE: " ..inque_count )
+end
+
+function start()
+    dfhack.println(GLOBAL_KEY  .. " | START")
+
+    if (args.t) then
+        state.threshold = 0-tonumber(args.t)
+    end
+    repeatUtil.scheduleEvery(GLOBAL_KEY, 1, 'days', check) -- 997 is the closest prime to 1000
```
chdoc, line 280:
```suggestion
    repeatUtil.scheduleEvery(GLOBAL_KEY, 1, 'days', check)
```

docs/autotraining.rst
```diff
@@ -0,0 +1,41 @@
...
+
+.. dfhack-tool::
+    :summary: Assigns citizens to a military squad until they have fulfilled their need for Martial Training
+    :tags: fort auto bugfix units
+
+Automation script for citizens to hit the gym when they yearn for the gains. Also passively builds military skills and physical stats.
+
+You need to have at least one squad that is set up for training. This should be a new non-military-use squad. The uniform should be
+set to "No Uniform" and the squad should be set to "Constant Training" in the military screen. Edit the squad's schedule to full time training with around 8 units training.
+The squad doesn't need months off. The members leave the squad once they have gotten their gains.
```
chdoc, line 12:
```suggestion
You need to have at least one squad that is set up for training. The squad should be set to "Constant Training" in the military screen. The squad doesn't need months off. The members leave the squad once they have satisfied their need for military training.

The configured uniform determines the skills that are acquired by the training dwarves. Providing "No Uniform" is a perfectly valid choice and will make your militarily inclined civilians become wrestlers over time. However, you can also provide weapons and armor to pre-train civilians for future drafts.
```

docs/autotraining.rst
```diff
@@ -0,0 +1,41 @@
...
+--------
+
+``autotraining``
+    Current status of script
+
+``enable autotraining``
+    Checks to see if you have fullfilled the creation of a training gym.
+    If there is no squad marked for training use, a clickable notification will appear letting you know to set one up/
+    Searches your fort for dwarves with a need to go to the gym, and begins assigning them to said gym.
+    Once they have fulfilled their need they will be removed from the gym squad to be replaced by the next dwarf in the list.
```
chdoc, line 31: gym ... gym ...

internal/control-panel/registry.lua
```diff
@@ -34,6 +34,8 @@ COMMANDS_BY_IDX = {
         desc='Automatically shear creatures that are ready for shearing.',
         params={'--time', '14', '--timeUnits', 'days', '--command', '[', 'workorder', 'ShearCreature', ']'}},
     {command='autoslab', group='automation', mode='enable'},
+    {command='autotraining', group='automation', mode='enable',
+        desc='Automation script for citizens to hit the gym when they yearn for the gains.'},
```
chdoc, line 38:
```suggestion
        desc='Automatically assign units with training needs to training squads. '},
```

chdoc (commented):
@myk002 I hope you agree with me on the choice of language. Otherwise, I would say that this has come a long way and is almost ready for merging.

docs/autotraining.rst
```diff
@@ -0,0 +1,41 @@
+autotraining
+============
+
+.. dfhack-tool::
+    :summary: Assigns citizens to a military squad until they have fulfilled their need for Martial Training
+    :tags: fort auto bugfix units
+
+Automation script for citizens to hit the gym when they yearn for the gains. Also passively builds military skills and physical stats.
```
chdoc, line 8: Oops, it looks like I accidentally removed the most important comment regarding language: I don't think that the whole "gym" "yearn" "gains" terminology fits with the way that documentation is usually phrased.

```suggestion
This script automatically assigns citizens with the need for military training to designated training squads.
```

chdoc:
@myk002 I have been away for a while, and I was hoping that this would have been merged by now. I have been trying to shepherd this a little, but I'm still not entirely satisfied with the code quality. Could you take a look and give an assessment as to what would need to be done to get this ready to be merged. I could invest sometime this week to do some further cleanup, but being neither author nor maintainer, I suppose I can't actually push any changes.

autotraining.lua
```diff
@@ -0,0 +1,290 @@
...
+                else
+                    inque_count = inque_count +1
+                end
+            else
+                removeTraining(unit)
+            end
+        end
+    end
+
+    dfhack.println(GLOBAL_KEY  .. " | IGN: " .. ignore_count .. " TRAIN: " .. intraining_count .. " QUE: " ..inque_count )
```
chdoc, line 252: This should probably use full words rather than abbreviations and be the only output during regular operation. Similarly, enabling and disabling should not generate output. `enable` follows the unix philosophy: no output means success.

autotraining.lua
```diff
@@ -0,0 +1,290 @@
...
+local repeatUtil = require('repeat-util')
+local utils=require('utils')
+
+validArgs = utils.invert({
+    't'
+})
+
+local args = utils.processArgs({...}, validArgs)
+local GLOBAL_KEY  = "autotraining"
+local need_id = df.need_type['MartialTraining']
```
chdoc, line 14: I would call this constant `MartialTraining`, the current name seems awfully generic

autotraining.lua
```diff
@@ -0,0 +1,290 @@
...
+            table.insert(squads, squad)
+        else
+            -- setting to nil during iteration is permitted by lua
+            state.training_squads[squad_id] = nil
+        end
+    end
+    return squads
+end
+
+function findNeed(unit)
```
chdoc, line 150: How about naming this `getTrainingNeed`?

autotraining.lua
```diff
@@ -0,0 +1,290 @@
...
+
+function removeTraining(unit)
+    return dfhack.military.removeFromSquad(unit.id)
+end
+
+function removeAll()
+    if ( state.training_squads == nil) then return end
+    for _, squad in ipairs(getTrainingSquads()) do
+        for i=1,9,1 do
+            local dwarf = getByID(squad.positions[i].occupant)
```
chdoc, line 222: To convert the historical figure id in the occupant field, it should suffice to use `df.historical_figure.find` and then look up the unit_id field, because `dfhack.military.removedFromSquad` takes an id instead of a unit pointer.

autotraining.lua
```diff
@@ -0,0 +1,292 @@
...
+    persist_state()
+end
+
+
+--######
+--Functions
+--######
+function getTrainingCandidates()
+    local ret = {}
+    local citizen = dfhack.units.getCitizens(true)
```
chdoc, line 104: I would inline this variable

autotraining.lua
```diff
@@ -0,0 +1,292 @@
...
+        end
+        if not dfhack.units.isAdult(unit) then
+            goto next_unit
+        end
+        local need = getTrainingNeed(unit)
+        if ( need  ~= nil ) then
+            if ( need.focus_level >= state.threshold ) then
+                goto next_unit
+            end
+        end
```
chdoc, line 119: Doesn't this actually include units that don't have any need for training? Also, this isn't C++, no parentheses in conditionals.
```suggestion
        if not need or need.focus_level >= state.threshold  then
            goto next_unit
        end
```

autotraining.lua
```diff
@@ -0,0 +1,292 @@
...
+    local squads = checkSquads()
+    local intraining_count = 0
+    local inque_count = 0
+    if ( squads == nil) then return end
+    for _,squad in ipairs(squads) do
+        for i=1,9,1   do
+            if ( squad.positions[i].occupant  ~= -1 ) then
+                local hf = df.historical_figure.find(squad.positions[i].occupant)
+                if hf ~= nil then
+                    local unit df.unit.find(hf.unit_id)
```
chdoc, line 232:
```suggestion
                    local unit = df.unit.find(hf.unit_id)
```

autotraining.lua
```diff
@@ -0,0 +1,280 @@
...
+    if squads == nil then return end
+    for _,squad in ipairs(squads) do
+        for i=1,9,1   do
+            if squad.positions[i].occupant  ~= -1 then
+                local hf = df.historical_figure.find(squad.positions[i].occupant)
+                if hf ~= nil then
+                    local unit = df.unit.find(hf.unit_id)
+                    local training_need = getTrainingNeed(unit)
+                    if not training_need or training_need.focus_level >= state.threshold then
+                        dfhack.military.removeFromSquad(unit)
```
chdoc, line 232:
```suggestion
                        dfhack.military.removeFromSquad(unit.id)
```
Please test your code, this is an outright argument error.

autotraining.lua
```diff
@@ -0,0 +1,292 @@
...
+--@ module = true
+
+local repeatUtil = require('repeat-util')
+local utils=require('utils')
+
+validArgs = utils.invert({
+    't'
+})
+
+local args = utils.processArgs({...}, validArgs)
```
SilasD, line 12: Argument processing should only be done when *not* running as a module.

I suggest moving this line down to the 'called on the command-line' section.

SilasD (commented):
I didn't see anything bad enough to justify delaying a merge.

I have some quibbles with command-line processing and with the enable/disable code.
`enable autotraining`
`disable autotraining`
It doesn't appear that these command-line commands will work.  (not verified.)

I somewhat question the need for `autotraining` as a separate script.  It seems to me that it could be rolled into `gui/autotraining`.

autotraining.lua
```diff
@@ -0,0 +1,292 @@
...
+    else
+        disable()
+    end
+else
+    -- called on the command-line
+    if args.t then
+        state.threshold = 0-tonumber(args.t)
+    end
+    print(("autotraining is %s"):format(state.enabled and "enabled" or "disabled"))
+end
```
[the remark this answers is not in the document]
chdoc: @SilasD I think you are just plain wrong here. While the plugin manager requires that the script can be loaded as a module for enable to work properly, the actual invocation as `[enable|disable] tool` does *not* set `dfhack_flags.module`. The way I understand it, the reason for the module is so that the plugin manager can register the `onStateChange` handlers and query the enable state. The important thing is to not generate any side effects, including output, when being loaded as a module. The code you highlighted does exactly what it should do: allow calling through the `enable` interface and perform argument parsing and output only when not sourced as a module and not being called through the `enable` interface.

ab9rf:
> I somewhat question the need for `autotraining` as a separate script. It seems to me that it could be rolled into `gui/autotraining`.
i'm sorry, `gui/autotraining`? i looked for this and could not find it. to what are you referring?

chdoc:
This pull-request adds both: `autotraining` as the tool that performs the actual work and can also be (partially) configured using the command-line and `gui/autotraining`. If we want to follow the convention that tools that open windows are invoked as `gui/something` and tools that perform automation are enabled as `enable something` we need two commands and therefore two files.

realSquidCoder:
@chdoc frack its still doing that? as far as i know that thats a miscount bug not a wrong effect bug, i thought i had fixed all of them tho...

autotraining.lua
```diff
@@ -0,0 +1,293 @@
...
+        if isIgnoredNoble(unit) then
+            ignore_count = ignore_count + 1
+            goto next_unit
+        end
+        table.insert(ret, { unit = unit, need = need.focus_level })
+        ::next_unit::
+    end
+    table.sort(ret, function (a, b) return a.need < b.need end)
+    return ret
+end
```
chdoc, line 135: This should not return units that are already in squads, in particular not units in non-training squads. Otherwise, `addTraining` will report these units as waiting units even though they will never actually be put into the training squads.

realSquidCoder: so i need to add a check if they are already in a squad here you say?

realSquidCoder: @chdoc so if they are in a squad at all, i know i should skip them but should i increment the ignore count or should i leave it (maybe increment it if they are in a non training squad? idk)

realSquidCoder: done and resolving this conversation

realSquidCoder:
@ab9rf everything seems to have passed so if the others have no issues im cool with a squash and merge. (and of course i'll be around to fix any bugs that are found in the next release)

PR #1412, opened 2025-03-02
spectate.lua: add "hold-to-show" option

TymurGubayev:


TymurGubayev:
btw, this is why I originally used the `create_row({t=t},...); t=t+1` approach: currently to move the bottom 4 widgets down, I had to modify 4 additional lines.

myk002 (approved):

myk002:
> btw, this is why I originally used the `create_row({t=t},...); t=t+1` approach: currently to move the bottom 4 widgets down, I had to modify 4 additional lines.

In the past, I've tried to find a balance here. I felt that t=t+1 made the code too messy and prevented the usual idiom of adding the widgets in a single call to `self:addviews{}`. setting `autoarrange_subviews=true` might be an option. A compromise that I've found works well is to group widgets into Panels, so only the Panel frame changes when things get rearranged, and the only inner widget frames that need to be modified are the ones in the modified Panel.

PR #1424, opened 2025-03-27
Create livingworld.lua

DroppedTheMantleEli:
This is my super rudimentary script for checking how many sites of interest there are in the world and printing the result and then increasing the sitecap to allow new sites in post-world generation. Humans build new hamlets, Elves build new Forest Retreats, Goblins build new trenches and Dwarfs build new hillocks.  Ideally the script would increase the sitecap only if the number of sites of interest are +1 -1 away from the sitecap number, but as it is now it works by increasing the sitecap after a set amount of time instead and achieving my goal I hope!

Don't believe it will change anything already exisiting

### DIFF
--- a/livingworld.lua
+++ b/livingworld.lua
@@ -1,0 +1,61 @@
+-- Import repeat-util for automation
+local repeatUtil = require('repeat-util')
+
+-- Define a function to count site types, print totals, and update sitecap
+local function count_and_update_sitecap()
+    -- Initialize counters for site types
+    local player_fortress_count = 0
+    local dark_fortress_count = 0
+    local mountain_halls_count = 0
+    local forest_retreat_count = 0
+    local town_count = 0
+    local fortress_count = 0
+
+    -- Loop through all sites in the world
+    for _, site in ipairs(df.global.world.world_data.sites) do
+        local site_type = site.type
+
+        -- Increment counters based on site type
+        if site_type == df.world_site_type.PlayerFortress then
+            player_fortress_count = player_fortress_count + 1
+        elseif site_type == df.world_site_type.DarkFortress then
+            dark_fortress_count = dark_fortress_count + 1
+        elseif site_type == df.world_site_type.MountainHalls then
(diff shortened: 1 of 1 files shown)


myk002:
Have you tested that this actually allows the game to produce more sites during gameplay? I don't think anyone has tried something like this before.

myk002:
Also, is the slow rate increase from this script better (gameplay-wise) than just setting the site cap to -1?

DroppedTheMantleEli:
https://discord.com/channels/793331351645323264/807444467140788254/1334653065642643589
Yep!

Den lör 12 apr. 2025 kl 17:33 skrev Myk ***@***.***>:

DroppedTheMantleEli:
I haven't tested too extensively to know which methodology would be best,
when testing to see if it would work at all or not I increased the sitecap
to 200+ above the current one and the Goblins built a lot of trenches very
quickly, which is why I opted for smaller incremental increases so that the
world doesn't fill up too fast with sites - More testing in concurrent
worlds would be required to see what's the best gameplay wise. I cobbled
this script together and it's not working as intended currently, as of
right now it keeps increases the sitecap maybe a tad bit too much and maybe
it should only increase the sitecap once every 2-3 years gameplay years.
Most players do not stay in a given world too long anyway, but this would
be something for those that play in their worlds longer.

Den lör 12 apr. 2025 kl 17:43 skrev Elias ***@***.***>:

ab9rf:
this looks like a useful addition and i do not see any reason to believe that altering the site cap on the fly is likely to destabilize the simulation

i would suggest in addition, as a command line option, adding the ability to set the site cap to specific values (since there is no vanilla UI for changing this post-worldgen of which I am aware) to facilitate this

i would also make the increment amount configurable, support a "query only" mode where it simply reports on the number of sites without changing the cap, add a "quiet" mode that doesn't print anything, and finally i'd remove the autorepeat functionality since that can be accomplished by combining a one-and-done tool with the existing `repeat` utility

PR #1425, opened 2025-03-28
Add toolbar button for `gui/design`

realSquidCoder:


ChrisJohnsen:
If you try a minimum-size DF window (912x552px, 114x46 UI tiles@8x12px), I think you will see that there are only 7 unused-by-DF columns between the left and "center" toolbars. So the sitemap button and this button will not both fit in a minimum size interface area.

Not sure what might be best
- squash them both in?
  - one would need to overwrite the center toolbar's "left border" (just to the right of the dig button)
  - maybe they could "spread out" when there is enough area, but this might get complicated to manage, especially if more buttons are to be added
- hide one when there isn't enough space?
- if one is disabled, should the other "take over" the whole area?
- move one to the other side of the window where there is more room (14 unused-by-DF columns between center and right toolbars)?

gui/design.lua
```diff
@@ -1565,6 +1567,94 @@ function DesignScreen:onDismiss()
...
+                    visible=function() return self.subviews.icon:getMousePos() end,
+                },
+            },
+        },
+    }
+end
+
+function DesignToolbarOverlay:onInput(keys)
+    return DesignToolbarOverlay.super.onInput(self, keys)
+end
```
myk002, line 1650: this function is a noop and can be removed

gui/design.lua
```diff
@@ -1565,6 +1567,94 @@ function DesignScreen:onDismiss()
...
 
+
+-- --------------------------------
+-- DesignToolbarOverlay
+--
+
+DesignToolbarOverlay = defclass(DesignToolbarOverlay, overlay.OverlayWidget)
+DesignToolbarOverlay.ATTRS{
+    desc='Adds a button to the toolbar at the bottom of the screen for launching gui/design.',
+    default_pos={x=50, y=-1},
```
myk002, line 1578: This position isn't going to work:
image

mayyybe here in the dig toolbar?
image

though that bar is already quite crowded..

PR #1431, opened 2025-04-07
`gui/mod-manager`: Add a window showing the active mods in a world

realSquidCoder:
image
Closes DFHack/dfhack#3985

Depends on: https://github.com/DFHack/dfhack/pull/5395

### DIFF
--- a/docs/gui/mod-manager.rst
+++ b/docs/gui/mod-manager.rst
@@ -5,7 +5,7 @@ gui/mod-manager
     :summary: Manange your active mods.
     :tags: dfhack interface
 
-In a loaded world, shows a list of active mods with the ability to copyto clipboard.
+In a loaded world, shows a list of active mods with the ability to copy to clipboard.
 Adds an optional overlay to the mod list screen that allows you to save and
 load mod list presets, as well as set a default mod list preset for new worlds.
 
--- a/docs/gui/mod-manager.rst
+++ b/docs/gui/mod-manager.rst
@@ -5,8 +5,8 @@ gui/mod-manager
     :summary: Manange your active mods.
     :tags: dfhack interface
 
-In a loaded world, shows a list of active mods with the ability to copy to clipboard.
-
+When run with a world loaded, shows a list of active mods. You can copy the
+list to the system clipboard for easy sharing or posting.
 
 Usage
 -----
@@ -24,8 +24,9 @@ framework.
 gui/mod-manager.button
 ~~~~~~~~~~~~~~~~~~~~~~
 
-Adds an optional overlay to the mod list screen that allows you to save and
-load mod list presets, as well as set a default mod list preset for new worlds.
+Adds a widget to the mod list screen that allows you to save and load mod list
+presets. You can also set a default mod list preset for new worlds so you don't
+have to manualy re-select the same mods every time you generate a world.
 
 gui/mod-manager.notification
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/gui/mod-manager.rst
+++ b/docs/gui/mod-manager.rst
@@ -6,8 +6,7 @@ gui/mod-manager
     :tags: dfhack interface
 
 In a loaded world, shows a list of active mods with the ability to copy to clipboard.
-Adds an optional overlay to the mod list screen that allows you to save and
-load mod list presets, as well as set a default mod list preset for new worlds.
+
 
 Usage
 -----
@@ -15,3 +14,20 @@ Usage
 ::
 
     gui/mod-manager
+
+Overlay
+-------
+
+This tool also provides one overlay that is managed by the `overlay`
+framework.
+
+gui/mod-manager.button
+~~~~~~~~~~~~~~~~~~~~~~
+
+Adds an optional overlay to the mod list screen that allows you to save and
+load mod list presets, as well as set a default mod list preset for new worlds.
+
+gui/mod-manager.notification
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Displays a message when a mod preset has been auto-applied.
--- a/docs/gui/mod-manager.rst
+++ b/docs/gui/mod-manager.rst
@@ -2,9 +2,10 @@ gui/mod-manager
 ===============
 
 .. dfhack-tool::
-    :summary: Save and restore lists of active mods.
+    :summary: Manange your active mods.
     :tags: dfhack interface
 
+In a loaded world, shows a list of active mods with the ability to copyto clipboard.
 Adds an optional overlay to the mod list screen that allows you to save and
 load mod list presets, as well as set a default mod list preset for new worlds.
 
--- a/gui/mod-manager.lua
+++ b/gui/mod-manager.lua
@@ -10,6 +10,7 @@ local utils = require('utils')
 
 local presets_file = json.open("dfhack-config/mod-manager.json")
 local GLOBAL_KEY = 'mod-manager'
+local INSTALLED_MODS_PATH = 'data/installed_mods/'
 
 -- get_newregion_viewscreen and get_modlist_fields are declared as global functions
 -- so external tools can call them to get the DF mod list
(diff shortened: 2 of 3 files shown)


gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
     }
 end
 
+ModlistMenu = defclass(ModlistMenu, widgets.Window)
+ModlistMenu.ATTRS {
+    view_id = "modlist_menu",
```
myk002, line 407:
```suggestion
```

myk002: you don't need to add `view_id` attributes unless you need to access them by name from another context

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+    frame_title = "Active Modlist",
+    frame_style = gui.WINDOW_FRAME,
```
myk002, line 409:
```suggestion
```

myk002: you don't need to specify the `frame_style` for a `widgets.Window` unless you're making a non-standard Window

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+
+    resize_min = { w = 30, h = 15 },
+    frame = { w = 40, t = 10, b = 15 },
```
myk002, line 412: I suggest either a constant size or a size based on the length of the list (with some maximum). specifying `t` and `b` will result in a huge window with lots of empty space on larger screens.

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+
+    resizable = true,
+    autoarrange_subviews=false,
```
myk002, line 415:
```suggestion
```

myk002: this is already the default

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+}
+
+local function get_mod_id_and_version(path)
```
myk002, line 418: This looks copied from `script-manager`. Please refactor instead of duplicating code that will get out of sync. Duplicated code is a maintenance burden and a source of future bugs.

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+            text_pen=COLOR_YELLOW,
+            auto_width=true,
+            on_activate=function()
+                local mods = ''
+                for _,mod in ipairs(getWorldModlist()) do
+                    mods = (mods == '' and mod) or (mods .. ', ' .. mod)
+                end
+                dfhack.internal.setClipboardTextCp437(mods)
+            end,
+            enabled=function() return #modlist:getChoices() > 0 end,
```
myk002, line 497: this is where the `view_id` actually comes in handy. You can declare the `widgets.List` inline instead of storing it in a local variable and change this line to be
```lua
enabled=function() return #self.subviews.modlist:getChoices() > 0 end,
```

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+            frame={t=1, r=1},
+            label='Copy modlist to clipboard',
+            text_pen=COLOR_YELLOW,
+            auto_width=true,
+            on_activate=function()
+                local mods = ''
+                for _,mod in ipairs(getWorldModlist()) do
+                    mods = (mods == '' and mod) or (mods .. ', ' .. mod)
+                end
+                dfhack.internal.setClipboardTextCp437(mods)
```
myk002, line 495:
```suggestion
                local mods = table.concat(getWorldModlist(), NEWLINE)
                dfhack.internal.setClipboardTextCp437Multiline(mods)
```

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+    }
+    self:addviews{
+        widgets.Label{
+            frame = { l=0, t=0 },
+            text = {'Active mods:'},
+        },
+        widgets.HotkeyLabel{
+            view_id='copy',
+            frame={t=1, r=1},
+            label='Copy modlist to clipboard',
```
myk002, line 487: In addition to the name, please also include the version, mod ID, and (if STEAM_FILE_ID is defined) a link to the webpage where the user can subscribe to the mod. It should look like `https://steamcommunity.com/sharedfiles/filedetails/?id=2924079137` (with the correct STEAM_FILE_ID instead of 2924079137)

so the format of each line should be:

```lua
('%s %s (%s): %s'):format(name, version, id, url)
```

unless STEAM_FILE_ID is not defined (i.e. it's not a steam workshop mod), in which case the last part should be left off.

gui/mod-manager.lua
```diff
@@ -401,6 +402,120 @@ function ModmanageScreen:init()
...
+        return modlist
+    end
+    qerror('No world is loaded')
+end
+
+function ModlistMenu:init()
+    local modlist = widgets.List{
+        view_id='modlist',
+        frame = {t=3},
+        choices = getWorldModlist()
```
myk002, line 477: The list should display the name, version, and the mod id. That is,
```lua
('%s %s (%s)'):format(name, version, id)
```

gui/mod-manager.lua
realSquidCoder: i veto the newline as you cant then copy paste into twitch. i see most ppl using it like that

docs/gui/mod-manager.rst
```diff
@@ -2,15 +2,32 @@ gui/mod-manager
...
 -----
 
 ::
 
     gui/mod-manager
+
+Overlay
+-------
+
+This tool also provides one overlay that is managed by the `overlay`
```
realSquidCoder, line 21:
```suggestion
This tool also provides two overlays that are managed by the `overlay`
```

gui/mod-manager.lua
```diff
@@ -401,6 +404,104 @@ function ModmanageScreen:init()
...
+            on_activate=function()
+                local mods = table.concat(getWorldModlist(true), NEWLINE)
+                dfhack.internal.setClipboardTextCp437Multiline(mods)
+            end,
+            enabled=function() return #self.subviews.modlist:getChoices() > 0 end,
+        },
+        widgets.List{
+            view_id='modlist',
+            frame = {t=4},
+            choices = getWorldModlist(true)
```
myk002, line 484: could you add a toggle to the bottom of the window (off by default) for "Show vanilla mods"? Some people need to know whether they have disabled any vanilla mods (e.g. for total conversions)

myk002: It would also be useful to have another toggle (on by default) for mods that are contributing to the DFHack script path. it might take some thought for how to make the different "classes" of mods clear.

- active and contributing to the DFHack script path
- active and not contributing to the DFHack script path
- not active and contributing to the DFHack script path

gui/mod-manager.lua
```diff
@@ -401,6 +404,104 @@ function ModmanageScreen:init()
...
+        for _,path in ipairs(df.global.world.object_loader.object_load_order_src_dir) do
+            path = tostring(path.value)
+            -- skip vanilla "mods"
+            if not path:startswith(INSTALLED_MODS_PATH) then goto continue end
+            local id, numerical_version, name, steam_id, display_version = scriptmanager.get_mod_info(path)
+            if not id or not numerical_version then goto continue end
+            mods[id]= {handled=true, name=name, version=display_version, steam_id=steam_id}
+            scriptmanager.add_mod_paths(mod_paths, id, path, '.')
+            ::continue::
+        end
```
myk002, line 433: this is breaking the abstraction of the script manager. this script should not have knowledge of INSTALLED_MODS_PATH

gui/mod-manager.lua
realSquidCoder: While i like the script path idea, id rather not include it in this pr. Im already getting a little overwhelmed from this and just want to get the basic idea out bc this is what a few ppl have been wanting, a way to see the mids they have to tell stream. I think it would be good to extend this after the basics work in another pr

realSquidCoder: I can do the vanilla mods tho

PR #1434, opened 2025-04-12
make `deathcause` an api for getting the cause of death

realSquidCoder:
Made this on my phone, might have errors so please dbl check


myk002 (commented):
could you remove the leftover parens from the `return` statements?

Also, needs a module declaration and standard side effects guard. See: https://docs.dfhack.org/en/latest/docs/dev/Lua%20API.html#importing-scripts

deathcause.lua
```diff
@@ -29,8 +30,7 @@ function displayDeathUnit(unit)
     str = str .. dfhack.units.getReadableName(unit)
 
     if not dfhack.units.isDead(unit) then
-        print(dfhack.df2console(str) .. " is not dead yet!")
-        return
+        return str .. " is not dead yet!"
```
myk002, line 33: you should update the names of these functions to reflect their new semantics

deathcause.lua
```diff
@@ -1,4 +1,5 @@
 -- show death cause of a creature
+--@ module = true
```
myk002, line 2: since you are exposing this as a module, please mark everything that *shouldn't* be exposed as `local`

deathcause.lua
```diff
@@ -12,25 +13,24 @@ function getItemAtPosition(pos)
...
+local function getDeathStringFromCause(cause)
     if cause == -1 then
         return "died"
     else
         return DEATH_TYPES[cause]:trim()
     end
 end
 
-function displayDeathUnit(unit)
+function getDeathUnit(unit)
```
myk002, line 28: This is now an API function, so it needs a better name. "printDeathUnit" may have made enough sense when this was just a script, but `reqscript('deathcause').getDeathUnit()` is not a name that makes the behavior of the function or the return value obvious.

Try to think of this from the caller's perspective? What do they need to know about what this function does, and how can you communicate that in the name of the function?

This comment goes for all the other non-`local` functions below.

deathcause.lua
```diff
@@ -50,7 +50,7 @@ function displayDeathUnit(unit)
         end
     end
 
-    print(dfhack.df2console(str) .. '.')
+    return str .. '.'
 end
 
 -- returns the item description if the item still exists; otherwise
```
myk002, line 56: is this function part of the external API? If not, it should be `local`

ab9rf:
i have tested this PR locally and it doesn't cause any malfunction. i haven't confirmed that the added functionality works as expected, but at least it shouldn't introduce a regression. CI testing isn't working correctly for some reason i can't explain

i will approve this once the API it introduces is documented

ab9rf (approved):

docs/deathcause.rst
```diff
@@ -14,3 +14,29 @@ Usage
...
+commandline interface with ``dfhack.run_script()`` or via the API functions
+defined in :source-scripts:`deathcause.lua`, available from the return value of
+``reqscript('deathcause')``:
+
+* ``getDeathCauseFromHistFig(histfig)``
+
+Returns a string with the historical figure's cause of death, sometimes with more information
+than with a unit.
+
+* ``getDeathCauseFromUnit(unit)``
```
chdoc, line 31: Other places in the Lua API take a unit or historical figure and then just act accordingly. For instance, there is:
```
dfhack.units.getNoblePositions(unit or historical_figure)
```
So I would argue for:
```
getDeathCause(unit or historical_figure)
```
as the public interface, using `df.unit:is_instance` or similar to determine the argument type.

realSquidCoder: agreed and resolved

deathcause.lua
```diff
@@ -87,11 +89,11 @@ function displayDeathEventHistFigUnit(histfig_unit, event)
...
     end
 
-    print(dfhack.df2console(str) .. '.')
+    return  str .. '.'
 end
 
 -- Returns the death event for the given histfig or nil if not found
-function getDeathEventForHistFig(histfig_id)
+local function getDeathEventForHistFig(histfig_id)
     for i = #df.global.world.history.events - 1, 0, -1 do
```
SilasD, line 97: This is a very big loop.  >750000 entries in my current fort's world, year 515.

Since a histfig has a `.died_year` field, you might be able to early-out with not-found ( `nil`, probably?) when you reach an event that occurred in the previous year.

You might even be able to get really clever and use a binary search to narrow down to the proper year's range.

This assumes that `world.history.events` is sorted by event date, which I believe but have not verified.

Speaking of that loop, this code implicitly returns `nil` when no matching event is found.  The caller doesn't check for that.

deathcause.lua
```diff
@@ -102,17 +104,18 @@ function getDeathEventForHistFig(histfig_id)
     end
 end
 
-function displayDeathHistFig(histfig)
+-- Returns the cause of death given a histfig
+local function getDeathCauseFromHistFig(histfig)
     local histfig_unit = df.unit.find(histfig.unit_id)
     if not histfig_unit then
         qerror("Cause of death not available")
```
SilasD, line 111: I don't like that an API function can deliberately crash.

Consider returning this string instead of calling `qerror()`.

deathcause.lua
```diff
@@ -102,17 +104,18 @@ function getDeathEventForHistFig(histfig_id)
...
     end
 
     if not dfhack.units.isDead(histfig_unit) then
```
SilasD, line 114: Actually I would suggest not finding the unit at all, and testing `histfig.died_year == -1` instead of `not dfhack.units.isDead()`.

You would want to fold in the string formatting that is in `getDeathEventHistFigUnit()`.

`dfhack.units.getReadableName()` accepts histfigs as well as units.

The only other use you have for a unit is to get the race, which is also in a histfig.

SilasD (commented):
I saw some minor things, nothing that *really* justifies delaying a merge.

One performance-related, one deliberately crashing if an event record wasn't found at all, one that requires a histfig's unit to be in world.units.all as opposed to being flushed to disk.

PR #1438, opened 2025-04-18
gui/design: add option to draw N-point stars, hollow or filled or inverted, and change the main axis to orient in any direction

ryanbennitt:
gui/design

Add new option to draw N-point stars.

* The default has 5 points, use 'B'/'b' to increase/decrease points.
* They can be hollow or filled, and inverted.
* The next-point offset can be increased/decreased using 'N'/'n' which particularly affects 7 point stars and above to make them spikier or smoother, but can also be used to decrease to 1 to make symmetrical polygons or increase to N which only paints the vertexes.
* The orientation can be changed by adding a main axis point using 'v' and moving this to point in the desired direction.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,6 +16,8 @@ Template for new versions:
 
 ## New Features
 
+- `gui/design`: add option to draw N-point stars, hollow or filled or inverted, and change the main axis to orient in any direction
+
 ## Fixes
 
 ## Misc Improvements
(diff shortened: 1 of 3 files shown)


myk002 (commented):
This is 100% awesome!

The only issue I found (from play testing) is that the drawn lines can now exceed the bounding box. This is especially visible when Invert is on:

image

When thickness is increased, could the lines be made to expand towards the center?

myk002:
I think I'll approve this for now, and we can discuss the bulging issue on Discord

myk002 (approved):

PR #1444, opened 2025-04-28
remove-stress also removes long-term stress, which also immediately removes stressed and haggard statuses

blake-sc:
While in its current implementation remove-stress will over time bring down long-term stress, it seems much more intuitive to just immediately remove the long-term stress.

### DIFF
--- a/remove-stress.lua
+++ b/remove-stress.lua
@@ -1,7 +1,6 @@
 -- Sets stress to negative one million
 --By Putnam; http://www.bay12forums.com/smf/index.php?topic=139553.msg5820486#msg5820486
 --edited by Bumber
---edited by BlakeMW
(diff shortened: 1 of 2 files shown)


myk002 (commented):
please remove the additional attribution line. otherwise, looks good!

remove-stress.lua
```diff
@@ -1,6 +1,7 @@
 -- Sets stress to negative one million
 --By Putnam; http://www.bay12forums.com/smf/index.php?topic=139553.msg5820486#msg5820486
 --edited by Bumber
+--edited by BlakeMW
```
myk002, line 4:
```suggestion
```

myk002 (approved):

PR #1445, opened 2025-04-28
Reduced diameter of star as thickness increases to fit within bounding box

ryanbennitt:
Reduced diameter of hollow star as the thickness increases so it does not exceed its bounding box.

Adjusted original changelog entry to contain more details about usage.


changelog.txt
```diff
@@ -16,10 +16,16 @@ Template for new versions:
 
 ## New Features
 
-- `gui/design`: add option to draw N-point stars, hollow or filled or inverted, and change the main axis to orient in any direction
+- `gui/design`: add new option to draw N-point stars.
+  - The default has 5 points, use 'B'/'b' to increase/decrease points.
+  - They can be hollow or filled, and inverted.
+  - The next-point offset can be increased/decreased using 'N'/'n' which particularly affects 7 point stars and above to make them spikier or smoother, but can also be used to decrease to 1 to make symmetrical polygons or increase to N which only paints the vertexes.
+  - The orientation can be changed by adding a main axis point using 'v' and moving this to point in the desired direction.
```
myk002, line 23: the changelog won't render multi-line messages correctly. could you add this info to the `docs/gui/design.rst` file instead?

changelog.txt
```diff
@@ -16,10 +16,16 @@ Template for new versions:
...
 
 ## Fixes
 
+- `gui/design`: reduced diameter of star as thickness increases to remain within bounding box
```
myk002, line 27: no need for a "fixes" line for the star shape since it hasn't been out in a release yet. However, if this fixes the bulge for the line shape, I'd say that deserves a "fixes" entry.

myk002:
I'm not sure if this fixes the issue. I was still able to draw a star that extended beyond the edge of the map:
image

changelog.txt
ryanbennitt: What's the bulge in the line shape issue?

ryanbennitt: Made the horizontal and vertical thickness consistent for curved lines as well, maybe this is the bulge you were referring to.

myk002:
> Am I missing something?

I'll test again, likely not properly synced on my part

changelog.txt
myk002: The bulge in the line shape was similar to what was happening for the star. When increasing the line thickness, the designated tiles can bulge beyond the selected endpoint, leading to potentially "selecting" tiles that are outside of the map:
image

This PR fixes the star shape, but the line still suffers from bulging.

I'll merge this PR as-is because the line bulge was a pre-existing issue, but if you have ideas on how to fix that, I'd love to hear them

myk002 (approved):

PR #1453, opened 2025-05-06
Improve `fix/stuck-squad`

Bumber64:
Attach new controller based on how DF handles cancelled missions, rather than requiring a returning squad.
Fix assumptions about one squad per army.
Some functions that might be useful to manipulate armies in general. (E.g., force any army to join a site.)
`verbose` and `quiet` options.

### DIFF
--- a/fix/stuck-squad.lua
+++ b/fix/stuck-squad.lua
@@ -1,126 +1,131 @@
+--Rescue stranded squads. (Also contains functions for messing with armies.)
 --@ module=true
 
-local utils = require('utils')
+local argparse = require('argparse')
+local plotinfo = df.global.plotinfo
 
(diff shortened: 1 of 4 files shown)


fix/stuck-squad.lua
```diff
@@ -1,126 +1,131 @@
...
-        ::continue::
     end
+    return armies
+end
 
-    if #stuck_armies == 0 then return stuck_armies, nil, nil end
+--From observing bugged saves, this condition appears to be unique to stuck armies
+local function is_army_stuck(army)
+    return army and not army.flags.dwarf_mode_preparing and --Let DF handle cancelled missions
+        army.controller_id ~= 0 and not army.controller
```
Bumber64, line 79: Why is a `controller_id` of `0` excluded, specifically?

PR #1454, opened 2025-05-15
Add research-info script

brmley:
The script to show more information about scholars activities, adapted for use with spectate-overlay or dfhack-console

### DIFF
--- a/internal/research-info/info.lua
+++ b/internal/research-info/info.lua
@@ -1,0 +1,126 @@
+-- Main module for research-info
+--@ module = true
+
+---@param unit_id df.unit.id
+function getHistoricalFigure(unit_id)
+    for _, historical_figure in ipairs(df.global.world.history.figures) do
+        if historical_figure.unit_id == unit_id then
+            return historical_figure
+        end
+    end
+    print("No historical figure found for unit ID: " .. unit_id)
+    return nil
+end
+
+function getKnowledgeGoalCategoryInfo(category_id)
+    -- alt: df.scholar_knowledgest
+    local topic_categories = {
+        [-1] = "None",  -- None
+        [0] = "Philosophy",  -- PHILOSOPHY_FLAG
+        [1] = "Philosophy Adv",  -- PHILOSOPHY_FLAG2
+        [2] = "Mathematics",  -- MATHEMATICS_FLAG
+        [3] = "Mathematics Adv",  -- MATHEMATICS_FLAG2
+        [4] = "History",  -- HISTORY_FLAG
+        [5] = "Astronomy",  -- ASTRONOMY_FLAG
+        [6] = "Naturalist",  -- NATURALIST_FLAG
+        [7] = "Chemistry",  -- CHEMISTRY_FLAG
+        [8] = "Geography",  -- GEOGRAPHY_FLAG
+        [9] = "Medicine 1",  -- MEDICINE_FLAG
+        [10] = "Medicine 2",  -- MEDICINE_FLAG2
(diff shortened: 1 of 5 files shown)


ab9rf:
I haven't moved forward to review this PR because the pre-commit check hasn't been resolved. Is the OP still interested in pursuing this PR?

research-info.lua
```diff
@@ -0,0 +1,69 @@
...
+
+if unit then
+    local research_data_list = {}
+    for i, unit in pairs(dfhack.units.getCitizens()) do
+        
+        if selected_unit and selected_unit.id ~= unit.id then
+            goto continue
+        end
+
+        local historical_figure = researchInfo.getHistoricalFigure(unit.id)
```
chdoc, line 52:
```suggestion
        local historical_figure = df.historical_figure.find(unit.hist_figure_id)
```
There are two observations here: units have a historical figure id field, and the idiomatic way to turn a numeric field into the corresponding structure is to use `df.<structure_name>.find(id)`. This makes functions like your `getHistoricalFigure` basically useless. Please revisit your code and see whether this same pattern can be applied elsewhere.

research-info.lua
```diff
@@ -0,0 +1,69 @@
...
+            print("\t" .. topic.name .. ": " .. topic.info)
+        end
+    end
+end
+
+--
+local unit  = dfhack.units.getCitizens()
+local selected_unit = dfhack.gui.getSelectedUnit()
+
+if unit then
```
chdoc, line 44: `dfhack.units.getCitizens` returns a table, and even the empty table (which this will never return in an active fort) is a "truthy" value in lua. So this condition can never evaluate to false.

internal/research-info/overlay.lua
```diff
@@ -0,0 +1,43 @@
...
+local researchInfo = reqscript("internal/research-info/info")
+
+local function formatKnowledgeString(str)
+    if not str or type(str) ~= "string" or str == "" then
+        return str or ""
+    end
+
+    local parts = {}
+    for part in str:gmatch("[^_]+") do
+        -- Первую букву делаем заглавной, остальные - строчными
```
chdoc, line 13: If you add comments, please add them in English.

internal/research-info/info.lua
```diff
@@ -0,0 +1,126 @@
...
+    end
+
+    local goal_info = ""
+    for i = 0, 31 do
+        if goal[i] then
+            local global_flag_index = category * 32 + i -- wizards move
+            goal_info = string.format("%s", df.dfhack_knowledge_scholar_flag[global_flag_index]) or string.format("Unknown Flag: %s", i)
+        end
+    end
+    return goal_info
```
chdoc, line 50: I don't understand what you're doing here. Please explain, so that we can see whether there is a better way.

chdoc (changes requested):
Since this tool can be invoked on its own from the command-line, it needs documentation. Without having a good understanding what this tool is supposed to do, it is hard to give good feedback. Please have a look at the .rst files in the docs folder. And then, of course, this also needs a changelog entry like every other PR.

internal/research-info/info.lua
brmley: This func returns name of researchable topic 

it takes data from https://github.com/DFHack/df-structures/blob/e4993ceef7907351fd2ac6f9bd3a9e5ff24699db/df.history_figure.xml#L308 and match value with name from https://github.com/DFHack/df-structures/blob/e4993ceef7907351fd2ac6f9bd3a9e5ff24699db/df.dfhack.xml#L221 there is a trouble in relations between these structures and these is doesnt match directrly. So, i use some magic.

PR #1463, opened 2025-06-01
Labor Control Overlay

unboundlopez:
Select Lock Overlay
===================

This is a DFHack overlay plugin for Dwarf Fortress that simulates the selection and locking of multiple units within the **Work Details** screen. It provides a simple UI interface to batch-toggle units' labor assignments, either by selecting, locking, or both.

Features
--------

- **Overlay Interface**: Integrated directly into the `LABOR/WORK_DETAILS` viewscreen.
- **Action Modes**: Choose between `Select only`, `Lock only`, or `Select + Lock`.
- **Batch Processing**: Specify how many entries to affect and apply actions with one click.
- **Non-Intrusive**: Uses DFHack GUI input simulation to trigger existing functionality.

Usage
-----

Once the plugin is loaded and you're in the `Work Details` screen (e.g., `u` -> `Work Details`), the overlay will automatically appear.

1. Use the **Mode** dropdown to select what action(s) to simulate:
   - `Select only`: Just toggles unit selection.
   - `Lock only`: Just toggles the lock status.
   - `Select + Lock`: Toggles both selection and lock.

2. Adjust the number of entries to apply actions to (default is 100).

3. Press the **RUN** button (or the hotkey defined for it) to execute the actions.

### DIFF
--- a/docs/zSelectLockLabors.rst
+++ b/docs/zSelectLockLabors.rst
@@ -24,25 +24,3 @@ Once the plugin is loaded and you're in the `Work Details` screen (e.g., `u` -> 
 2. Adjust the number of entries to apply actions to (default is 100).
 
 3. Press the **RUN** button (or the hotkey defined for it) to execute the actions.
-
-Install
--------
-
-1. Place this script in your DFHack `hack/scripts` directory.
-2. Ensure the plugin is listed in your `dfhack.init` or loaded manually via `:lua require('select_lock_overlay')`.
-
-Contributing
-------------
-
-Pull requests and issues are welcome! Please make sure your changes follow the existing code style.
-
-License
--------
-
-This plugin is distributed under the MIT License. See ``LICENSE`` for more information.
-
-Author
-------
-
-This plugin was written for DFHack by [Your Name Here].
-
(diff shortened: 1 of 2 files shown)


unboundlopez:
> I haven't tried it so it may be buggy or non-functional, but there's a bit of code in gui/manipulator that seems to deal with work details, does this help you avoid the UI approach?
> 
> https://github.com/DFHack/scripts/blob/master/gui/manipulator.lua#L657

Hello ryanbennitt

It’s definitely possible to make this run instantly using an interface. However, achieving that would require significantly more code and interactions with variables that I don’t fully understand yet.

Yesterday, while working on a different program, I experimented with enabling and disabling the checkbox for all labors. Visually, everything appeared to work correctly, but the dwarves didn’t actually perform the tasks. Because of that, I believe sticking with a true macro is the best option for simplicity.

Previously, before using this labor control overlay interface, I relied on a third-party macro tool called VoiceAttack, which I still use for other tasks.

Also, if someone runs this program extremely fast say, within 1 frame it does execute successfully, but the checkboxes may appear disabled even though they are actually enabled. I was surprised to see this behavior, but confirmed that if the user exits and re-enters the labor menu, all the labors are indeed selected and locked as expected.

This is the path to enable/disable the checboxes: gui/gm-editor df.global.plotinfo.labor_info.work_details[x].flags

unboundlopez:
pre-commit.ci autofix

PR #1465, opened 2025-06-12
Add zprospectanalyzer script and documentation

unboundlopez:
zprospectanalyzer
=================

A DFHack CLI utility and Lua module for scanning and reporting material occurrences
and elevation ranges using the built-in `prospect` command. It provides an easy-to-use
interface to list material counts and elevation ranges, with sorting and preset lists.

Features
--------

- **Output Parsing**: Runs `prospect all` or `prospect all --show <section>` and parses the text output.
- **Section Filtering**: Filter materials by specific sections like layers, ores, gems, or globally.
- **Presets**: Built-in "blocks" preset for common stone materials.
- **Missing Material Handling**: Lists missing materials at the end with `<not found>` marker.

Usage
-----

Load the script in DFHack and run:

.. code-block:: bash

  zprospectanalyzer [section] [material1] [material2] ...

Examples:

- **Default** (blocks preset):

  .. code-block:: bash

  zprospectanalyzer

- **Filter by section**:

  .. code-block:: bash

  zprospectanalyzer layer_materials chert granite

- **Global search**:

  .. code-block:: bash

  zprospectanalyzer jet ruby tetrahedrite

### DIFF
--- a/zprospectanalyzer.lua
+++ b/zprospectanalyzer.lua
@@ -1,0 +1,160 @@
+-- scripts/zprospectanalyzer.lua
+
+local zprospectanalyzer = {}
+
+--- Scans the world using the `prospect` command and returns a table of materials.
+-- @return table data[section][material_key] = { count=number, minElev=number?, maxElev=number? }
+function zprospectanalyzer.scanProspect(section)
+    local filterSection = section and section:lower():gsub("%s+","_") or "all"
+    local showMap = {
+        base_materials   = "base",
+        liquids          = "liquids",
+        layer_materials  = "layers",
+        features         = "features",
+        ores             = "ores",
+        gems             = "gems",
+        other_vein_stone = "veins",
+        shrubs           = "shrubs",
+        wood_in_trees    = "trees",
+    }
+    local cmd = { "prospect", "all" }
+    if filterSection ~= "all" and showMap[filterSection] then
+        table.insert(cmd, "--show")
+        table.insert(cmd, showMap[filterSection])
+    end
+    local output, status = dfhack.run_command_silent(table.unpack(cmd))
+    if status ~= CR_OK then
+        error(("prospect failed (code %d): %s"):format(status, tostring(output)))
+    end
+    local data = {}
+    local current
+    for line in output:gmatch("([^\n]+)") do
+        local header = line:match("^%s*([%a%s_]+)%s*:%s*$")
+        if header then
+            current = header:lower():gsub("%s+","_")
+            data[current] = {}
+        elseif current then
+            local name, count, elev = line:match(
+                "^%s*([%u_]+)%s*:%s*(%d+)%s+Elev:?%s*([-%d%.]+)"
+            )
+            if not name then
+                name, count = line:match("^%s*([%u_]+)%s*:%s*(%d+)")
+            end
+            if name and count then
+                local key = name:lower()
+                local entry = { count = tonumber(count) }
+                if elev then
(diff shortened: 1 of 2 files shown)


zprospectanalyzer.lua
```diff
@@ -0,0 +1,163 @@
...
+--- Main entry point for CLI.
+-- Supports "blocks" preset or custom section/material arguments.
+-- Not-found entries are printed last.
+-- blocks preset includes stones that are worth 3 pts.
+function zprospectanalyzer.main(...)
+    local args = { ... }
+    if #args == 0 then args = { "blocks" } end
+    local presets = {
+        blocks = {
+            "Alabaster", "Alunite", "Andesite", "Anhydrite", "Basalt",
```
quietust, line 97: This shouldn't be hardcoding - it should fetch the actual stone IDs from the raws.

ab9rf: There is no guarantee even that any of those stones exist in any given world, or if they do that they're worth 3 points in the embark menu. For that you need to pull the material definitions out of the world

DFHack tools should, to the extent practical, work even with radically modified raws

unboundlopez: Yesterday I tried looking for the skill point value of items and failed. I don’t like the way I’m doing it now but can’t seem to find way to achieve a preset of blocks worth 3 points.

zprospectanalyzer.lua
```diff
@@ -0,0 +1,163 @@
+-- scripts/zprospectanalyzer.lua
+
+local zprospectanalyzer = {}
+
+--- Scans the world using the `prospect` command and returns a table of materials.
```
ab9rf, line 5: instead of running the `prospect` command and grubbing through its output, you should add (or requested added) an API to the `prospect` plugin that will return the requested data

scraping is always a fragile approach and creates undocumented dependencies - if someone later changes the `prospect` tool's output, even in the slightest way, this tool will break

zprospectanalyzer.lua
ab9rf: you should (a) make the target value or values a script argument (what exactly is magical about 3 points?)
and (b) use the material definitions in the game to identify which minerals qualify

i'm restating this to add it here as a review blocker

ab9rf (changes requested):

zprospectanalyzer.lua
unboundlopez: I hadn't thought about prospect updating.

I haven't been able to figure out where the block point values are defined, and I haven’t had any luck locating them in gui/gm-editor either. What’s special about 3-point blocks? It’s helpful to know on the embark screen which blocks are worth 3 points, so I (the user) can make informed decisions and aim for a consistent material when making rock blocks. I usually bring around 180 stones to turn into blocks, so it's frustrating to start a large project with, say, rock salt, only to discover that chert is more abundant and at a higher tighter elevation.

I do want to switch to using the actual stone IDs in my presets, but without first uncovering where the point values are defined, I can't justify moving away from using the names.

unboundlopez: I hope you're able to test what I'm referring to. If you run prospect, it takes a while to make an informed decision about which stones to bring. But if you run my program, you can significantly cut down the time needed to decide.

unboundlopez: I don't know how to search for the point values of items anymore. I feel I've exhausted all avenues through gui/gm-editor and using lua.

PR #1475, opened 2025-07-04
[immortal-cravings] prioritize high-value meals 

chdoc:
... and don't go eating or drinking on a full stomach

fixes: https://github.com/DFHack/dfhack/issues/5401


immortal-cravings.lua
```diff
@@ -13,6 +13,27 @@ function distance(p1, p2)
     return math.max(math.abs(p1.x - p2.x), math.abs(p1.y - p2.y)) + math.abs(p1.z - p2.z)
 end
 
+---find best item in an item vector (according to some metric)
```
ab9rf, line 16: it seems to me that `findClosest` is `findBest` but with a specific metric (distance) and so `findClosest` could be written as a specialization of `findBest`

chdoc: You are of course right. I realized this at some point and then forgot about it. I just tried it, but the result is barely simpler and less efficient, because one now needs to compute the position (`xyz2pos(dfhack.items.getPosition(item))`) separately for the metric and the accessibility check. I can do it, but I don't think it is worth it.

ab9rf: no, i don't insist on this, it was an observation only

chdoc:
If there are no more required changes, can we get this merged?

ab9rf (approved):
lgtm

PR #1478, opened 2025-07-09
Fix for rename

Snowstar38:
Names are stored in 3 locations (the main unit, historical figure, and soul), not 2

I think this is fine but ***if anything in the vanilla game can cause current_soul to have a different name than the unit and historical figure***, then never mind - I have never had current_soul be changed to a separate individual with a different name from the unit itself, but I am new to the game.


chdoc:
Units can absolutely have multiple "souls" with multiple different names. Different "souls" are used by units that take on fake identities. I don't know whether `rename` handles those correctly, or what the correct behavior even should be.

PR #1480, opened 2025-07-09
Create setup-editor.lua

Snowstar38:
This adds a new GUI popup to edit game setup points, in both Fortress and Adventure mode.

"embark-skills" already does something very similar for Fortress mode, but this is a GUI vs a command script.

setup editor

I am quite new to programming, so for all I know, this could be absolutely horrific to an experienced programmer. But I hope that I'm providing something useful here!

### DIFF
--- a/gui/setup-editor.lua
+++ b/gui/setup-editor.lua
@@ -1,0 +1,282 @@
+-- gui/setup_editor.lua
+local gui     = require('gui')
+local widgets = require('gui.widgets')
+local dialog  = require('gui.dialogs')
+
+---------------------------
+-- Utility
+---------------------------
+
+local function getStringValue(val)
+    return tostring(val)
+end
+
+local function make_editable_list(rows, get_tab_index, flip_tabs)
+    local function build_choices()
+        local choices = {}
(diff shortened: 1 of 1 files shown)


chdoc:
I don't have time to do a full review right now, but here are a few things I can say right off the bat:

1. this pull-request needs both an entry in the `changelog.txt`  and a documentation file (i.e. `.rst` file) for the new tool. For the latter, please have a look at the `docs` folder. 
2. From the pictures, it looks like the user has to manually select the `Fortress` or `Adventure` tabs. Since those are different game modes, I don't think it is useful to put them in tabs. The tool should simply always display the information relevant for the active game mode. Of course, this requires detecting the game mode, which shouldn't be too hard.
3. I'm not a fan of wrapping all the accesses to game variables in `pcall`. There should be a way to determine the number of dwarves or the number of adventure characters from the game state. I don't like the idea of the tool silently failing in case the used data structures of the game are changed.
4. You should not assume that a fort always starts with 7 dwarves. There is also the `startdwarf` tool.
5. Apart from the liberal use of `pcall`, I think the quality of the code is pretty decent :+1:

PR #1481, opened 2025-07-12
Edits to the mod manager

ong-yinggao98:
Fixes DFHack/dfhack#4799 

I really do not like the idea of hardcoding all the vanilla modules, but for some reason (as documented erroneously), calling:

```lua
fields.src_dir[i].value:startswith('data/vanilla')
```

leads to a Lua error stating "cannot read field string.startswith: not found".

TODO:
* Resolve DFHack/dfhack#3746
* Resolve DFHack/dfhack#4402

### DIFF
--- a/gui/mod-manager.lua
+++ b/gui/mod-manager.lua
@@ -12,11 +12,45 @@ local widgets = require('gui.widgets')
 local presets_file = json.open("dfhack-config/mod-manager.json")
 local GLOBAL_KEY = 'mod-manager'
 
+local vanilla_modules = {
+    ['vanilla_text'] = true,
+    ['vanilla_languages'] = true,
+    ['vanilla_descriptors'] = true,
+    ['vanilla_materials'] = true,
+    ['vanilla_environment'] = true,
+    ['vanilla_plants'] = true,
+    ['vanilla_items'] = true,
+    ['vanilla_buildings'] = true,
+    ['vanilla_bodies'] = true,
+    ['vanilla_creatures'] = true,
+    ['vanilla_entities'] = true,
+    ['vanilla_reactions'] = true,
+    ['vanilla_interactions'] = true,
+    ['vanilla_descriptors_graphics'] = true,
+    ['vanilla_plants_graphics'] = true,
+    ['vanilla_items_graphics'] = true,
+    ['vanilla_buildings_graphics'] = true,
+    ['vanilla_creatures_graphics'] = true,
+    ['vanilla_interactions_graphics'] = true,
+    ['vanilla_world_map'] = true,
+    ['vanilla_interface'] = true,
+    ['vanilla_music'] = true,
+}
+
+function get_moddable_viewscreen(type)
+    local vs = nil
+    if type == 'region' then
+        vs = dfhack.gui.getViewscreenByType(df.viewscreen_new_regionst, 0)
+    elseif type == 'arena' then
+        vs = dfhack.gui.getViewscreenByType(df.viewscreen_new_arenast, 0)
+    end
+    return vs
+end
+
 -- get_newregion_viewscreen and get_modlist_fields are declared as global functions
 -- so external tools can call them to get the DF mod list
 function get_newregion_viewscreen()
-    local vs = dfhack.gui.getViewscreenByType(df.viewscreen_new_regionst, 0)
-    return vs
+    return get_moddable_viewscreen('region')
 end
 
 function get_modlist_fields(kind, viewscreen)
@@ -62,7 +96,9 @@ local function move_mod_entry(viewscreen, to, from, mod_id, mod_version)
     local mod_index = nil
     for i, v in ipairs(from_fields.id) do
         local version = from_fields.numeric_version[i]
-        if v.value == mod_id and version == mod_version then
+        local vanilla = vanilla_modules[mod_id]
+        -- assuming that vanilla mods will not have multiple possible indices
+        if v.value == mod_id and (vanilla or version == mod_version) then
             mod_index = i
             break
         end
--- a/gui/mod-manager.lua
+++ b/gui/mod-manager.lua
@@ -12,6 +12,9 @@ local widgets = require('gui.widgets')
 local presets_file = json.open("dfhack-config/mod-manager.json")
 local GLOBAL_KEY = 'mod-manager'
 
+-- hardly an elegant solution, but mysteriously,
+-- using from_fields.src_dir[i].startswith('data/vanilla') in move_mod_entry()
+-- leads to lua complaining that it 'cannot read field string.startswith: not found'
 local vanilla_modules = {
     ['vanilla_text'] = true,
     ['vanilla_languages'] = true,
@@ -97,7 +100,7 @@ local function move_mod_entry(viewscreen, to, from, mod_id, mod_version)
     for i, v in ipairs(from_fields.id) do
         local version = from_fields.numeric_version[i]
         local vanilla = vanilla_modules[mod_id]
-        -- assuming that vanilla mods will not have multiple possible indices
+        -- assumes that vanilla mods will not have multiple possible indices.
         if v.value == mod_id and (vanilla or version == mod_version) then
             mod_index = i
             break


ab9rf:
I suspect the reason why `fields.src_dir[i].value:startswith('data/vanilla')` does not work is because `fields.src_dir[i].value` is a C++ string (being accessed via proxy), not a Lua string, and so the Lua method `startswith` (which is defined for Lua strings) doesn't work.

ab9rf:
> Interesting, but in that case why does it work in `script-manager.lua::get_active_mods()`? It seems to call `startswith` just fine and the target seems to also be a C++ string.
> 
> ```lua
> -- local ol = df.global.world.object_loader
>         local path = ol.object_load_order_src_dir[idx].value
>         table.insert(mods, {
>             id=ol.object_load_order_id[idx].value,
>             name=ol.object_load_order_name[idx].value,
>             version=ol.object_load_order_displayed_version[idx].value,
>             numeric_version=ol.object_load_order_numeric_version[idx],
>             path=path,
>             vanilla=path:startswith('data/vanilla/'), -- windows also uses forward slashes
>         })
> ```

Because `path=path` results in it being converted from a C++ string to a Lua string, and so `path:startswith` is looking for that named method on a Lua string

This is an artifact of how our C++ object proxying into Lua works. When you directly "read" from a C++ string in order to use its value as a Lua value, the proxy handler automatically converts it into a Lua string. However, when you call a method directly on an object, this doesn't happen: instead, the metatable of the C++ proxy object is consulted, and as there is no `startswith` method in the C++ string proxy object metatable, the method call fails.

It might be an idea to "chain" the Lua string metatable onto the C++ string proxy metatable, but I'm not at all sure how to go about doing this. Will require considerable thought.

Copying the proxy object into a local variable before calling the method should provide a workaround.

ong-yinggao98:
I haven't tried SilasD's suggestion, but ab9rf's works like a charm.

gui/mod-manager.lua
```diff
@@ -12,34 +12,15 @@ local widgets = require('gui.widgets')
...
-    ['vanilla_creatures_graphics'] = true,
-    ['vanilla_interactions_graphics'] = true,
-    ['vanilla_world_map'] = true,
-    ['vanilla_interface'] = true,
-    ['vanilla_music'] = true,
-}
+-- Shamelessly taken from hack/library/lua/script-manager.lua
+function vanilla(dir)
+    dir = dir.value
+    dir = dir -- better safe than sorry i guess
```
ab9rf, line 18: this isn't required

ong-yinggao98: As requested (and tested)

ab9rf:
I just tested this locally with 52.02 and the arena mode functionality does not work. It's possible that the viewscreen name has changed in 52.xx, but that needs to be addressed before I can consider merging this.

SilasD:
As a minor note, Lua supports multiple return values.  
[Lua 5.3 Reference Manual, section 3.3.4](https://www.lua.org/manual/5.3/manual.html#3.3.4)
> Functions can return more than one value, (... specific syntax ...)

I'm going to go through and flag these, with suggested code.

gui/mod-manager.lua
```diff
@@ -84,13 +111,15 @@ local function move_mod_entry(viewscreen, to, from, mod_id, mod_version)
         v:erase(mod_index)
     end
 
-    return true
+    return { success= true, version= loaded_version }
 end
 
+--- @return { success: boolean, version: string }
```
SilasD, line 117: If you convert to multi-return, I believe the syntax here would be
```
---@return boolean  # returns true if the mod entry was moved; returns false if the mod or mod version was not found.
---@return version  # string - DISPLAYED_VERSION from the mod's info.txt 
```

gui/mod-manager.lua
```diff
@@ -84,13 +111,15 @@ local function move_mod_entry(viewscreen, to, from, mod_id, mod_version)
...
 local function enable_mod(viewscreen, mod_id, mod_version)
     return move_mod_entry(viewscreen, "object_load_order", "available", mod_id, mod_version)
 end
 
+--- @return { success: boolean, version: string }
```
SilasD, line 122: If you convert to multi-return, I believe the syntax here would be
```
---@return boolean  # returns true if the mod entry was moved; returns false if the mod or mod version was not found.
---@return version  # string - DISPLAYED_VERSION from the mod's info.txt 
```

ab9rf:
> This doesn't really make sense to me. C++ strings returned through the Lua wrapper _are_ Lua strings.

Only if they pass through `stl_string_identity::lua_write`. My best guess, based on the observed behavior, is that this isn't happening when the object is being pulled out of a "container" object; instead of ultimately pushing a Lua string onto the stack, it's pushing the Lua userdata object that proxies the C++ string _without_ using the `lua_write` specialization that converts C++ strings into Lua strings. It has to retain the proxy object so that if it gets assigned to the assignment uses the STL string assignment method.

ong-yinggao98:
> I just tested this locally with 52.02 and the arena mode functionality does not work. It's possible that the viewscreen name has changed in 52.xx, but that needs to be addressed before I can consider merging this.

Correct, please do not merge it yet. I mentioned that I need the the related PR merged but I'll move it to the top of the description in bold.
> Requires https://github.com/DFHack/dfhack/pull/5500

gui/mod-manager.lua
```diff
@@ -12,10 +12,29 @@ local widgets = require('gui.widgets')
 local presets_file = json.open("dfhack-config/mod-manager.json")
 local GLOBAL_KEY = 'mod-manager'
 
--- get_newregion_viewscreen and get_modlist_fields are declared as global functions
--- so external tools can call them to get the DF mod list
-function get_newregion_viewscreen()
+-- Shamelessly taken from hack/library/lua/script-manager.lua
```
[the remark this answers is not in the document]
ong-yinggao98: Ah, will take note in future.

ab9rf:
> > I just tested this locally with 52.02 and the arena mode functionality does not work. It's possible that the viewscreen name has changed in 52.xx, but that needs to be addressed before I can consider merging this.
> 
> Correct, please do not merge it yet. I mentioned that I need the the related PR merged but I'll move it to the top of the description in bold.
> 
> > Requires [DFHack/dfhack#5500](https://github.com/DFHack/dfhack/pull/5500)

Gotcha, I will go look at that PR now.

ong-yinggao98:
> I'm going to go through and flag these, with suggested code.

In hindsight unpacked return types should have been obvious to me, thanks a ton for the comments and suggestions.

ab9rf:
I just tested this locally and while the panel appears in arena mode and I can _save_ profiles I cannot _load_ a profile.

I suspect that this needs updates to reflect the type changes with 52.02; if you look at [the changes made for 52.02 compatibility](https://github.com/DFHack/dfhack/pull/5514/files#diff-e9af1675bcf3f2450f180a2662be150d39b322cee8cb1c6f2ecb11de186c18c4) you will note that I had to make adjustments to code in `script-manager.lua` and I suspect parallel changes need to be made here.

ong-yinggao98:
> I suspect that this needs updates to reflect the type changes with 52.02; if you look at [the changes made for 52.02 compatibility](https://github.com/DFHack/dfhack/pull/5514/files#diff-e9af1675bcf3f2450f180a2662be150d39b322cee8cb1c6f2ecb11de186c18c4) you will note that I had to make adjustments to code in `script-manager.lua` and I suspect parallel changes need to be made here.

Yep, we no longer need to pass `.value` anymore. I'll do a final pass tomorrow to make sure none of the changes have affected script behaviour.

ab9rf (approved):

ab9rf:
this looks ready to merge to me, but i'll wait if you want to give it one more pass first

i want to get a release out soon for the fix to mod manager but i'd like to slide this in with that release if i can

ong-yinggao98:
> this looks ready to merge to me, but i'll wait if you want to give it one more pass first
> 
> i want to get a release out soon for the fix to mod manager but i'd like to slide this in with that release if i can

I've just finished testing, thanks!

PR #1482, opened 2025-07-15
add new tool `entomb`

amade-w:
This is a new tool that will allow players to assign a tomb zone to any unit regardless of citizenship, pet status, etc. Once successfully assigned, haulers will place the corpse items into the assigned tomb. The tool also includes the option to ~instantly teleport~ task all burial items for simultaneous hauling into the tomb's coffin (thanks to SilasD) and the option to add arbitrary items for burial with the unit.

entomb-preview-1
entomb-preview-2
entomb-preview-3


entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+        end
+    else
+        qerror('No item selected or unit specified.')
+    end
+end
+
+-- Validate tomb zone assignment.
+local function CheckTombZone(building, id)
+    if df.building_civzonest:is_instance(building) then
+        if building.type == 97 then
```
SilasD, line 32: Consider changing this to explicitly reference the civzone type, for clarity. (untested code)
```if building.type == df.civzone_type.Tomb then```

File: entomb.lua [hunk 2]
```diff
@@ -0,0 +1,193 @@
...
+    if not tomb then
+        qerror('No unassigned tomb zones are available.')
+    end
+end
+
+-- Set corpse items to be valid for burial.
+local function FlagForBurial(corpseParts)
+    -- Undead units have empty corpse_parts vector.
+    if unit.enemy.undead then
+        for _, item in pairs(df.global.world.items.other.IN_PLAY) do
```
SilasD, line 87: Iterating over ```df.global.world.items.other.ANY_CORPSE``` would be substantially faster, and should give the same result.  You could even skip the two ```:is_instance()``` tests, but perhaps leave them in for type safety.

entomb.lua
```diff
@@ -0,0 +1,193 @@
+-- Entomb corpse items of any dead unit.
+--@module = true
+
+local utils = require('utils')
+
+local unit_id
```
SilasD, line 6: Consider making all of these internal to functions, and passing them in to subfunctions as necessary.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+local unit
+local building_id
+local tomb
+local forceBurial
+
+local args = {...}
+
+-- Get unit from selected corpse or corpse piece item.
+local function GetUnitFromCorpse()
```
SilasD, line 15: Consider allowing a caller to pass in `item`, and only using `getSelectedItem()` if the passed-in item was nil.  This would generalize the function.

Consider removing the `local` from the function definition; this function may be useful to other scripts.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+                buildingItem.use_mode = 0
+            end
+        end
+        print('Corpse items have been teleported into a coffin.')
+    else
+        print('No coffin in the assigned tomb zone.\nCorpse items will not be teleported into the tomb zone.')
+    end
+end
+
+local function AssignToTomb()
```
SilasD, line 138: Consider passing in `unit` for variable locality.

Consider removing the `local`; this function may be useful to other scripts.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+    local building
+    if #args > 0 then
+        for i, v in ipairs(args) do
+            if v == 'unit' then
+                unit_id = tonumber(args[i+1]) or nil
+                unit = unit_id and df.unit.find(unit_id)
+                if not unit then qerror('Invalid unit ID.') end
+            end
+            if v == 'tomb' then
+                building_id = tonumber(args[i+1]) or nil
```
SilasD, line 167: Consider making `building_id` local; this is the only function that uses it.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+            local isCoffin = coffin and df.building_coffinst:is_instance(coffin) or nil
+            -- Return TRUE if even one item is not interred.
+            if not isCoffin then
+                return true
+            end
+        end
+    end
+end
+
+local function GetEmptyTombZone()
```
SilasD, line 64: Consider passing in `unit` for variable locality, and getting a local `unit_id` from the `unit`.

Consider returning the unit's tomb zone for variable locality.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+    -- Check if unit is already assigned to a tomb zone.
+    local isAlreadyAssigned = IterateTombZone(unit_id)
+    if isAlreadyAssigned then
+        if isNotBuried() or forceBurial then
+            tomb = isAlreadyAssigned
+            print('Unit is already assigned to a tomb zone but may still have uninterred corpse or body part(s).')
+        else
+            qerror('Unit is already interred in a tomb zone.')
```
SilasD, line 72: I think this shouldn't be a crash-the-program error case.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+            local coffinBuilding_id = inCoffin and inCoffin.building_id or nil
+            local coffin = coffinBuilding_id and df.building.find(coffinBuilding_id) or nil
+            local isCoffin = coffin and df.building_coffinst:is_instance(coffin) or nil
+            -- Return TRUE if even one item is not interred.
+            if not isCoffin then
+                return true
+            end
+        end
+    end
+end
```
SilasD, line 62: Consider inserting `return false` before this line; currently this returns `nil`, which is not desirable for a boolean test function.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+
+-- Check if any of the unit's corpse items are still not in a coffin.
+local function isNotBuried()
+    for _, item_id in pairs(unit.corpse_parts) do
+        local item = df.item.find(item_id)
+        if item then
+            local inCoffin = dfhack.items.getGeneralRef(item, df.general_ref_type.BUILDING_HOLDER)
+            local coffinBuilding_id = inCoffin and inCoffin.building_id or nil
+            local coffin = coffinBuilding_id and df.building.find(coffinBuilding_id) or nil
+            local isCoffin = coffin and df.building_coffinst:is_instance(coffin) or nil
```
SilasD, line 55: Because this is a boolean variable, consider changing the `nil` to `false`.

Consider putting parentheses around the `coffin and df.building_coffinst:is_instance(coffin)`, for clarity.  At first I thought this was an instance of the `test and value1 or value2` idiom; it works the same but that's not the intent of the code.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+        -- Find an unassigned tomb zone.
+        tomb = IterateTombZone(-1)
+    end
+    if not tomb then
+        qerror('No unassigned tomb zones are available.')
+    end
+end
+
+-- Set corpse items to be valid for burial.
+local function FlagForBurial(corpseParts)
```
SilasD, line 84: Consider passing in `unit` for variable locality.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+                else
+                    qerror('Specified zone ID does not point to an unassigned tomb zone.')
+                end
+            end
+            if v == 'now' then forceBurial = true end
+        end
+    end
+end
+
+local function Main()
```
SilasD, line 182: Consider passing in `args`, for variable locality.

entomb.lua
```diff
@@ -0,0 +1,193 @@
+-- Entomb corpse items of any dead unit.
```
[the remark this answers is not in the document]
amade-w: Hey, no worries! I'm a novice coder so I welcome any opportunity to learn best practices and code better!
I'll get down to revising the script later today.

entomb.lua
```diff
@@ -0,0 +1,193 @@
...
+            if building.assigned_unit_id == id then
+                return true
+            end
+        end
+    end
+end
+
+-- Iterate through all available tomb zones.
+local function IterateTombZone(id)
+    for _, building in pairs(df.global.world.buildings.all) do
```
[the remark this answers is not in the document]
amade-w: I believe I've misunderstood a few things about vectors and pairs/ipairs here. Since the index starts at `0`, I thought ipairs would ignore the first element. But apparently ipairs work just fine.

I suppose I should change the same for other vector iterations currently using pairs.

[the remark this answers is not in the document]
SilasD: It's a DFHack-specific extension to `pairs()` and `ipairs()`, so it won't be in the standard [Lua documentation](https://www.lua.org/manual/5.3/manual.html).

It is mentioned in the [DFHack Lua API](https://docs.dfhack.org/en/stable/docs/dev/Lua%20API.html#id65), but you basically have to already know how it works to make sense of it:

> `pairs(ref)`, `ipairs(ref)`

> If the container has no associated enum, both behave identically, iterating over numerical indices in order. Otherwise, ipairs still uses numbers, while pairs tries to substitute enum keys whenever possible.

and

> `ref[index]`

> Accesses the container element, using either a ***0-based*** numerical index, or, if an enum is associated, a valid enum key string.

entomb.lua
```diff
@@ -1,171 +1,173 @@
...
-            if building.assigned_unit_id == id then
-                return true
-            end
+local function CheckTombZone(building, unit_id)
+    if building.type == df.civzone_type.Tomb then
+        if building.assigned_unit_id == unit_id then
+            return true
         end
     end
 end
```
SilasD, line 24: Probably should explicitly `return false` before this line.

entomb.lua
```diff
@@ -1,171 +1,173 @@
...
-        -- it to TEMP to turn it into an interred corpse item.
-        for _, item_id in pairs(corpseParts) do
-            local item = df.item.find(item_id)
-            if item then
-                dfhack.items.moveToBuilding(item, coffin, 2)
-            end
+function PutInCoffin(coffin, item)
+    if item then
+        -- Set df.building_item_role_type.PERM first before changing it to TEMP to turn the items
+        -- into interred burial items, otherwise the items will be hauled back to stockpiles.
```
SilasD, line 96: Aha!  I know why this happens!

It's the `item.flags.in_building` flag that I mentioned an hour ago.

I would bet a thousand dollars that 
```
if (dfhack.items.moveToBuilding(item, coffin, df.building_item_role_type.TEMP) ) then
    item.flags.in_building = true  -- if successfully moved, mark as interred
end
```

will work.

amade-w: You're right! Sadly I don't have a thousand dollars to my name :-)

Here's a test where I arbitrarily teleported a non-corpse item into an unassigned tomb:
entomb-preview-5
The item remained in the coffin without further interference from haulers.

I've also included a check for `flags.in_job`, it now removes any jobs associated with the item so it can be teleported properly.

entomb.lua
```diff
@@ -52,17 +52,24 @@ end
...
 local function IterateTombZones(unit_id)
     for _, building in ipairs(df.global.world.buildings.other.ZONE_TOMB) do
-        if CheckTombZone(building, unit_id) then return building end
+        if unit_id == -1 then
+            -- Use only active (unpaused) zones when assigning unassigned tomb zones.
+            if building.spec_sub_flag.active then
+                if CheckTombZone(building, unit_id) then return building end
+            end
+        else
+            if CheckTombZone(building, unit_id) then return building end
```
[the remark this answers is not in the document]
amade-w: Yeah, that was really awful. I've changed it.

PR #1485, opened 2025-07-23
embark-anyone.lua bugfix: Test the current viewscreen

SilasD:
Test the current viewscreen to ensure that it is the choose_start_site viewscreen, before trying to use it.

The bug: if the embark-anyone script is executed on any viewscreen **other than** the embark viewscreen, it aborts with a stack trace.

This bugfix makes it cleanly abort with a reasonably-descriptive error message.

This was found while diagnosing Issue #5509, but is not related.

Minimal changes to the script.

### DIFF
--- a/embark-anyone.lua
+++ b/embark-anyone.lua
@@ -1,19 +1,17 @@
 local dialogs = require('gui.dialogs')
 local utils = require('utils')
 
-function addCivToEmbarkList(info)
-   local viewscreen = dfhack.gui.getDFViewscreen(true)
-   if viewscreen._type ~= df.viewscreen_choose_start_sitest then
-      qerror("This script can only be used on the embark screen!")
+function embarkAnyone()
+
+   function addCivToEmbarkList(info)
+      local viewscreen = dfhack.gui.getDFViewscreen(true)
+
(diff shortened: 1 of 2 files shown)


embark-anyone.lua
```diff
@@ -3,6 +3,9 @@ local utils = require('utils')
 
 function addCivToEmbarkList(info)
    local viewscreen = dfhack.gui.getDFViewscreen(true)
+   if viewscreen._type ~= df.viewscreen_choose_start_sitest then
```
ab9rf, line 6: this code is duplicated, which means it should be pulled out as its own function that is called in both places that it appears

ab9rf (changes requested):
we might want to consider adding a utility function for asserting a specific viewscreen as I see that similar code is found in `unretire-anyone` and equality checks are found in four other scripts, but that's not required for this PR

embark-anyone.lua
SilasD: My intent was to keep the changes minimal.  I'm willing to refactor if you think it's important.

SilasD: Okay, found a minimal-change that doesn't duplicate code.

PR #1489, opened 2025-08-08
Update exterminate script and documentation

unboundlopez:
Please update exterminate to add a -c flag that targets only child units. I plan to build a timer-based script to run this so poults don’t wander off and put dwarven children at risk.

Argument parsing: adds -c alongside existing flags.
-c purpose: target children only.

Example:
exterminate -f -c bird_turkey -m knockout

### DIFF
--- a/exterminate.lua
+++ b/exterminate.lua
@@ -1,6 +1,8 @@
 --@module = true
 
 local argparse = require('argparse')
+-- for profession constants
+local profession = df.profession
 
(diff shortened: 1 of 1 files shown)


ab9rf:
needs changelog

unboundlopez:
I do not know exactly how to successfully add a changelog.txt so I'm going to add it here in the description of what I believe needs to bee added in this changelog.txt.

exterminate: now accepts a "-c" argument which targets only children.

PR #1491, opened 2025-08-16
new tool: Automatically milk and shear animals at nearby farmer's workshops

chdoc:
This is intended as a replacement/alternative for `automilk` and `autoshear`. It solves the problem that the workorder based approach loses the connection to the specific animals, possibly queuing jobs at workshops on the other side of your fort.

### DIFF
--- a/husbandry.lua
+++ b/husbandry.lua
@@ -1,0 +1,326 @@
+
+--@enable = true
+--@module = true
+
+local utils = require 'utils'
+local repeatutil = require("repeat-util")
+local ic = reqscript('idle-crafting')
+
+local verbose = true
+---conditional printing of debug messages
+---@param message string
+local function debug(message)
+    if verbose then
+        print(message)
+    end
+end
+
+-- From workorder.lua
+---------------------------8<-----------------------------
+
+local function isValidAnimal(unit)
+    -- this should also check for the absence of misc trait 55 (as of 50.09), but we don't
+    -- currently have an enum definition for that value yet
+    return  dfhack.units.isOwnCiv(unit)
+        and dfhack.units.isAlive(unit)
+        and dfhack.units.isAdult(unit)
+        and dfhack.units.isActive(unit)
+        and dfhack.units.isFortControlled(unit)
+        and dfhack.units.isTame(unit)
+        and not dfhack.units.isMarkedForSlaughter(unit)
+        and not dfhack.units.getMiscTrait(unit, df.misc_trait_type.Migrant, false)
+end
+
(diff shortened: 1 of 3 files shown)


husbandry.lua
```diff
@@ -0,0 +1,326 @@
...
+    dfhack.job.addGeneralRef(job, df.general_ref_type.UNIT_SHEAREE, unit.id)
+    ic.assignToWorkshop(job, workshop)
+end
+
+local function milkCreature(unit, workshop)
+    local job = ic.make_job()
+    job.job_type = df.job_type.MilkCreature
+    dfhack.job.addGeneralRef(job, df.general_ref_type.UNIT_MILKEE, unit.id)
+    ic.assignToWorkshop(job, workshop)
+end
```
ab9rf, line 134: this to me indicates that the functionality that's here being used from idle-crafting should at least be considered for being moved to a library

husbandry.lua
```diff
@@ -0,0 +1,326 @@
...
+                local milkee = dfhack.job.getGeneralRef(job, df.general_ref_type.UNIT_MILKEE)
+                if milkee then
+                    unit_milking[milkee.unit_id] = true
+                end
+            elseif state.shearing and job.job_type == df.job_type.ShearCreature then
+                local shearee  = dfhack.job.getGeneralRef(job, df.general_ref_type.UNIT_SHEAREE)
+                if shearee then
+                    unit_shearing[shearee.unit_id] = true
+                end
+            end
```
ab9rf, line 217: can't help but notice that this code is somewhat repetitive

also this pattern appears likely to recur, may be worth abstracting to a library routine

husbandry.lua
```diff
@@ -0,0 +1,326 @@
...
+                shearCreature(unit, workshop)
+            end
+        end
+
+        if state.milking and canMilkCreature(unit) and not unit_milking[unit.id] then
+            local workshop = getAppropriateWorkshop(unit, farmer_milking)
+            if workshop then
+                milkCreature(unit, workshop)
+            end
+        end
```
ab9rf, line 237: same with this - iterating over all workshops looking for a predicate or iterating over all units looking for a predicate both seem like things we should abstract

ab9rf (approved):

husbandry.lua
chdoc: Those are part of the functions that were recently merged into the library. I can fix this PR, but I can also do a follow up where I replace these functions in all the related tools.

ab9rf: A follow up is fine. This code is fine as it stands and I see no reason to delay merging it so we can include it in the next pre-release for UAT

PR #1492, opened 2025-08-17
Pedestal fixes3

SilasD:
Two minor fixes to `internal/common/pedestal.lua`.  No user-visible changes, except that removing items from a pedestal using the DFHack interface will work correctly.

Well-tested.  *Extensive* comments in each commit.


ab9rf:
This alters user-facing behavior and thus needs a changelog

SilasD:
Thanks!

So GitHub doesn't carry per-commit comments along when it transfers commits to a merge.

There's not even an indication that the comments exist.  I don't like that.

I would guess that's because the comments are not core Git.

PR #1497, opened 2025-08-24
Support for modded honey in ban-cooking

Droseran:
This patch allows ban-cooking to ban honey/brewable materials added by creatures other than honey bees. This includes vanilla bumblebees, though their honey is not currently obtainable in vanilla. However, there are mods that make bumblebees hiveable, as well as adding other types of hiveable bees.

### DIFF
--- a/ban-cooking.lua
+++ b/ban-cooking.lua
@@ -80,9 +80,18 @@ funcs.booze = function()
 end
 
 funcs.honey = function()
-    local mat = dfhack.matinfo.find("CREATURE:HONEY_BEE:HONEY")
-    if mat then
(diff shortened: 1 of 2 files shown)


SilasD (commented):
This works; there's nothing *wrong* with it.

But I would argue that the entire `funcs.honey` shouldn't *really* exist.  Instead, this code should be folded into `funcs.brew`.

`funcs.brew` searches for materials that have a reaction product that is alcohol.  And that's what this code does.  The only difference here is iterating over creatures instead of plants.

Separating the code just means that anybody who is trying to ban-cooking of brewables needs to ban two categories instead one.

(Combining the two would also make it easier to refactor out the copy-paste code.)

ab9rf (approved):

PR #1500, opened 2025-09-06
add new tool `store-owned`

amade-w:
This is a new tool that will allow any selected owned item to be tasked for storage in the owner's room furniture (cabinets, boxes, etc).

It will check to see if there's a container building of the appropriate type with sufficient space in the owner's bedroom, before looking in any owned office, dining hall or tomb, in that order.
If no container is found (or lacks sufficient space), it will attempt to find a tile unoccupied by any building in said rooms in the same order, to place the item on the floor of said tile.
If neither container nor tile can be found in any owned rooms, it will check if there's a dormitory zone available and check for container buildings in it, or an unoccupied floor if no container is available.
If it still can't find a container or tile, the job destination will default to the trade depot's center tile.
Finally, a "Store owned item" job will be created and assigned to the item's owner. 
The tool will also remove the item from the unit's uniform.

Players may also opt to task the item to be stored in the dorm or sent straight to the depot.
The tool also includes option to force the owner to relinquish ownership of the item.

Possible use cases:
- Unburden units that have acquired too many trinkets.
- Clean up scattered owned items selectively without removing ownership.

### DIFF
--- a/store-owned.lua
+++ b/store-owned.lua
@@ -1,0 +1,410 @@
+-- Task units to store their owned items.
+
+local utils = require('utils')
+
+local function GetUnit(item, options)
+    local unit = dfhack.items.getOwner(item)
+    if not unit then
+        print('Selected item is not owned by any unit.')
+    elseif not dfhack.units.isJobAvailable(unit) and not options.discard then
+        unit = nil
+        print('Owner of selected item is currently busy.')
+    end
+    return unit
+end
+
+local function isValidItem(item, unit)
+    local validItem = true
+    local itemPos = item and xyz2pos(dfhack.items.getPosition(item))
+    local building = dfhack.items.getHolderBuilding(item)
+    if item.flags.in_job or
+        item.flags.hostile or
+        item.flags.removed or
+        item.flags.dead_dwarf or
+        item.flags.rotten or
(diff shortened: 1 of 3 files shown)


store-owned.lua
```diff
@@ -0,0 +1,408 @@
...
+        df.item_toyst:is_instance(item) or
+        df.item_toolst:is_instance(item) -- Owned books and scrolls, which can be stored in boxes.
+    then
+        return true
+    end
+    return false
+end
+
+local function GetContainerType(item)
+    local containerType
```
ab9rf, line 114: explicit default value here maybe?

store-owned.lua
```diff
@@ -0,0 +1,408 @@
...
+    end
+    local unit = GetUnit(item, options)
+    local strCannotStore = 'Unable to task selected item for storage.'
+    if not unit and not options.discard then
+        qerror(strCannotStore)
+    elseif options.discard then
+        if not unit then
+            qerror('Cannot remove ownership of an ownerless item.')
+        else
+            goto removeItem
```
ab9rf, line 387: is this really the best way to do this?

amade-w: I ought to have been eaten by a raptor https://xkcd.com/292/

PR #1530, opened 2026-01-03
fix for modtools/create-unit

dvantwisk:
This fix changes the "arena spawn" paradigm that no longer works with the new version of DF to use `dfhack.units.create()`. It guts most arena-specific code, changes the method in which skills and equipment are applied, updates unit block info, and teleports the unit to the specified location. There's likely more to do, but for now this seems to revive the script's functionality as it was before the steam version of DF.

### DIFF
--- a/modtools/create-unit.lua
+++ b/modtools/create-unit.lua
@@ -129,21 +129,7 @@ function createUnit(raceStr, casteStr, pos, locationRange, locationType, age, do
 end
 
 function createUnitBase(...)
-  local old_gametype = df.global.gametype
-  local old_mode = df.global.plotinfo.main.mode
-  local old_popups = {} --as:df.popup_message[]
-  for _, popup in pairs(df.global.world.status.popups) do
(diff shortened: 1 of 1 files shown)


ab9rf (changes requested):
This looks good to me, but needs a changelog entry

modtools/create-unit.lua
chdoc, line 951: This also needs to be updated:
```suggestion
if dfhack_flags.module then
```

chdoc, line 955: Should be updated to use the `argparse` module.
```suggestion
local argparse = require('argparse')
local args = argparse.processArgs({...}, validArgs)
```

chdoc (commented):
Apologies for the long delay. There simply wasn't anyone around to do proper reviews for a long time.

I gave it a look, and I didn't find anything that really stood out, except for two (outside of the modified part) where the way that things are done now have changed a bit.

Is there still interest in getting this merged?

PR #1566, opened 2026-03-06
ignore language_name_component NONE element with value -1

Halavus:
added a break in the 2 concerned for loops


SilasD:
reference: PR #1595 which also fixed a problem in this code.  I do not think this PR caused the issue 1595 targetted, but I am still confuzzled why 1595 was needed; I still suspect a race.

PR #1568, opened 2026-03-16
Issue #5751 - gui/siegemanager - cleaner error logging if no siege engine selected

Halavus:

### DIFF
--- a/gui/siegemanager.lua
+++ b/gui/siegemanager.lua
@@ -419,7 +419,7 @@ function SiegeEngineList:set_selected_action(action)
     if not selected then
         qerror('No siege engine selected')
     end
-    
+
     local successful = set_siege_engine_action({selected.data}, action)
     if not successful then
         self:refresh_view(true)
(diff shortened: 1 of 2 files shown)


Halavus:
pre-commit.ci autofix

gui/siegemanager.lua
```diff
@@ -416,6 +416,10 @@ end
 function SiegeEngineList:set_selected_action(action)
     local _, selected = self.subviews.list:getSelected()
 
+    if not selected then
+        qerror('No siege engine selected')
+    end
+
```
chdoc, line 422: I think this is fundamentally the wrong way to fix the problem. All it does is change the error instead of avoiding it. The underlying issue is that the labels for setting the siege engines stance are enabled even if there is no siege engine selected. Thus, the right way to address this is to provide a function/callback that checks whether the labels should be enabled, as has been done here for example:

https://github.com/DFHack/scripts/blob/56c934eb5d4c957ca731705783a0a43397a9ba2c/idle-crafting.lua#L576-L580

chdoc (changes requested):

chdoc:
Superseded by #1602

PR #1571, opened 2026-03-24
Add gear assembly toggle widget

### LINKED ISSUE #5763: Idea: Tool that toggles gear assembly
Not sure of the feasibility of this one. The game seems to create a new machine whenever parts are separated. Toggling a flag on the gear isn't enough to get DF to do that, cause building collapse, etc.

Bumber64:
Add a toggle button on the gear assembly building that allows engaging/disengaging them.

Toggling a gear assembly at will is possibly Armok mode level control. However, I think it's already possible to disengage machines through spooky action by cancelling construction of unfinished machine parts to cause a collapse. I'd require dwarf interaction if I could.

`RollerOverlay:render` was using `TrackStopOverlay.super.render`. I fixed that, but I'm not sure there's a visible effect to that change.

It may be worth renaming the `trackstop` script to something fitting the increased scope at some point.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,6 +29,7 @@ Template for new versions:
 ## New Tools
 
 ## New Features
+- `trackstop`: can now toggle gear assemblies
 
 ## Fixes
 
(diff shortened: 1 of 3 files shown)


ab9rf:
This needs to have armok mode filtering enabled since it allows bypassing normal game mechanics

Bumber64:
> This needs to have armok mode filtering enabled since it allows bypassing normal game mechanics

How do we do that for this one specific widget?

Bumber64:
Replacing with #1590

PR #1582, opened 2026-04-25
Fix gm-unit: dynamically update size_cur when modifying body appearance

sizzlins:
Fixes #5793.

This pull request implements the math to dynamically update size_cur when modifying body appearance values (Height, Broadness, Length) in gm-unit. Previously, modifying these values left size_cur untouched (fulfilling the 3-year-old TODO in editor_body.lua). It uses standard DF scaling integer math to perfectly match the engine's volume calculations.


internal/gm-unit/editor_body.lua
```diff
@@ -45,6 +57,18 @@ function Editor_Body_Modifier:setBodyModifier(modifierIndex, value)
   -- Update the world texture
   self.target_unit.flags4.any_texture_must_be_refreshed = true
 
+  -- Manually recalculate body size
```
ab9rf: this code is an exact duplicate of the code above. write a function and call it in both places. Don't Repeat Yourself.

ab9rf (changes requested):

sizzlins:
superseded by #1594

PR #1583, opened 2026-04-25
Add manual-save script (closes DFHack/dfhack#5792)

sizzlins:
Adds a `manual-save` command that creates persistent, named save snapshots that won't be overwritten by autosaves. Works by triggering a native autosave and then duplicating the result into a timestamped folder. Includes a `--cleanup` flag to prune old snapshots. Compatible with the `repeat` command for rolling saves.


Bumber64:
It may be better to use `simulateInput` to press the buttons required to perform a manual save. (Hopefully there's a way to set the save name on the viewscreen rather than simulating the name inputs.) It's clunky, but better than messing with the file system.

sizzlins:
image

manual-save.lua
```diff
@@ -0,0 +1,203 @@
...
+
+    local date_str = os.date("%Y-%m-%d_%H-%M-%S")
+    local final_folder_name = positional_name or (fort_name .. "-Manual-" .. date_str)
+
+    print("Initiating manual save...")
+    print("Saving natively to: " .. final_folder_name)
+
+    -- Inject the text and trigger the save UI natively
+    local options = df.global.game.main_interface.options
+    options.open = true
```
Bumber64: It's my understanding that this isn't safe. Viewscreens need to be opened via simulated input to initialize properly.

sizzlins: @Bumber64 fixed, now uses simulated input to open options menu instead of options.open=true

manual-save.lua
```diff
@@ -0,0 +1,205 @@
...
+    return dfhack.filesystem.rmdir(path)
+end
+
+-- ---------------------------------------------------------------------------
+-- Save-directory resolution
+-- ---------------------------------------------------------------------------
+
+-- DF Premium stores saves under %APPDATA%/Bay 12 Games/Dwarf Fortress/save.
+-- Classic/portable installs keep them next to the executable.
+local function getTrueSaveDir()
```
ab9rf: I am not at all confident that this is safe across all platforms and it does not reflect portable mode behavior correctly. It also replicates code found in `Filesystem::getBaseDir` which should be used instead

manual-save.lua
```diff
@@ -0,0 +1,205 @@
...
+
+if not dfhack.world.isFortressMode() then
+    qerror('This script can only be used in fortress mode')
+end
+
+-- ---------------------------------------------------------------------------
+-- File helpers
+-- ---------------------------------------------------------------------------
+
+-- Recursively delete a directory tree. Returns true on success.
```
ab9rf: I don't like using `os.remove`, and furthermore `std::filesystem` already has a recursive directory delete which I'd prefer to proxy into Lua (via the `Filesystem` module, if needed) instead of replicating an already implemented mechanic in obscure Lua code

sizzlins: ive refactored this to use dfhack.filesystem.rmdir_recursive as requested, with a graceful qerror fallback for now since it hasnt been exposed to lua yet, let me know if you plan to proxy that api in the c++ core, or if youd like me to open a separate PR on the main dfhack repo to add it

sizzlins: actually I just did create dfhack.filesystem.rmdir_recursive, heres the link to the pull
https://github.com/DFHack/dfhack/pull/5834

PR #1585, opened 2026-05-15
`caravan`: fix special case for obsidian swords

### LINKED ISSUE #5660: Elf trading ethics warning
Wooden training spears, axes, and swords are not flagged with an ethics warning against elves in either the vanilla or dfhack trading windows.

chdoc:
fixes: https://github.com/DFHack/dfhack/issues/5660

It turns out that obsidian short swords are not the only wooden weapons...

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -32,6 +32,8 @@ Template for new versions:
 
 ## Fixes
 
+- `caravan`: fix ethics warning for wooden weapons
+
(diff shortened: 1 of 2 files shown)


quietust:
Looks reasonable to me - the weapon check should be in addition to the material checks (on the item and its improvements), not *instead* of them.

ab9rf (approved):

PR #1588, opened 2026-05-19
`fix/stuck-worship`: fix lookup of units current prayer target(s)

chdoc:
Issue: fix/stuck-worship considers the possible targets of a prayer to be all the deity links of the person organizing the worship. This is highly dubious, in particular for undedicated temples. 

The new function looks up the deity in the prayer sub-activity (to the worship activity in the respective temple) of the praying unit.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -32,6 +32,8 @@ Template for new versions:
 
 ## Fixes
 
+- `fix/stuck-worship`: fix lookup of units current prayer target(s).
+
 ## Misc Improvements
 
 ## Removed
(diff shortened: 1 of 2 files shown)


ab9rf:
I remember that you did some research on this after this PR was submitted, but don't remember what that research showed. Is this PR still an improvement in light of that? I'm inclined to merge but I wanted to check with you first.

chdoc:
Sorry for being so unresponsive lately. While I am convinced that the way the prayer target is currently determined is wrong, the way the game currently assigns happy thoughts after praying is also wrong (in the sense that neither is following what the game shows as the prayer target for the unit). I am no longer convinced that fixing this actually makes the situation better. Also, my testing has mostly been done in undedicated temples. 

I am downgrading this to draft status until I get around to do some more thorough testing including various dedicated temples.

PR #1591, opened 2026-06-11
scripts/add-recipe bugfixes

huesatlight:
first time using github, so feel free to correct any practices I got wrong.

add-recipe all was excluding recipes it shouldn't because of an adjective (eg: "long skirt") or because they had the same subtype number as picks. It was also including recipes the documentation says it wouldn't include (foreign instruments).


changelog.txt
```diff
@@ -36,6 +36,19 @@ Template for new versions:
...
 ## Removed
 
+# 53.14-alpha3
+
+## New Tools
+
+## New Features
+
+## Fixes
+- ``script\add-recipe``: fix to include incorrectly excluded recipes
```
ab9rf: this should be simply `add-recipe`, with _single_ backticks because it's a reference to a tool which has its own documentation page

ab9rf (approved):

PR #1596, opened 2026-06-24
Fix "combine" breaking dye mixes

Snowstar38:
There was a bug where using combine with mixed dyes would revert the mix to one of its component parts.

Ex: You have 2 stacks of 2 Dye A, 3 stacks of 1 Dye B, and 1 stack of 3 of the mix Dye AB

Expected behavior: use combine, you end up with 1 stack of 4 Dye A, 1 stack of 3 Dye B, and 1 stack of 3 Dye AB

Actual behavior: use combine, you end up with 1 stack of **7** Dye A, 1 stack of 3 Dye B, and no Dye AB (because it combined AB with A)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -47,6 +47,7 @@ Template for new versions:
 - `add-recipe`: fix to include incorrectly excluded recipes
 - `gui/control-panel`: fixed incorrect description of deteriorate commands
 - `gui/petitions`: fix deity display
+- `combine`: stopgap fix for incorrectly combining dyes, no longer affects dyes
 
 ## Misc Improvements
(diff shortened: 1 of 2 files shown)


ab9rf (changes requested):
This is a stopgap measure. the issue is that we _really_ need to check the fields added in the dye update, but `combine` was never updated to do this.

That said, worth doing. No behavior is better than buggy behavior.

Needs changelog, will approve once changelog is added.

Snowstar38:
Thanks! I wasn't sure I'm talented enough to dig through fixing it to work properly with dyes, but at least this way no one will have whole stacks of dye effectively disappearing on them!

ab9rf (approved):

PR #1599, opened 2026-07-10
update color-schemes for v5X

Moth-Tolias:
also, gui/color-schemes suggests registering from `/dfhack-config/colors` instead of `/raw/colors/`now.

note: doesn't use the `getConfigPath()` api, since that would be a pretty involved change and i'd rather get feedback first.
also doesn't update the gui style, that's out of my league

### DIFF
--- a/gui/color-schemes.lua
+++ b/gui/color-schemes.lua
@@ -611,7 +611,7 @@ function screen:onRegister()
         "Register",
         "Register color schemes by file or directory",
         COLOR_WHITE,
-        "/raw/colors/",
+        "/dfhack-config/colors/",
         function(path)
             ColorSchemes.register(path)
             self:populateList()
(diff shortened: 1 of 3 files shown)


gui/color-schemes.lua
```diff
@@ -611,7 +611,7 @@ function screen:onRegister()
         "Register",
         "Register color schemes by file or directory",
         COLOR_WHITE,
-        "/raw/colors/",
+        "/dfhack-config/colors/",
```
ab9rf, line 614: Please update this to use `dfhack.getConfigPath()` (see DFHack/DFHack#5823)

ab9rf (changes requested):

chdoc:
replaced by #1600

PR #1600, opened 2026-07-17
Overhaul color-schemes for v5X

Moth-Tolias:
Obsoletes #1599.
Cuts out an overwhelming majority of the old script in favour of dfhack-native implementations and a simpler interface. Uses Milo Christiansen's implementation of parsing the color files, because we think it's more elegant.

Removes the in-file help text in favour of the .rst documentation.
Removes registration as a concept.
Stores the default scheme in prefs/color.txt instead of a dedicated file.

Anything that uses this script expecting it to work the same as before will break. But then, v5x broke it anyway, so i think that's fine.
Only really works with the ASCII interface.
We haven't written a changelog entry yet. I wanted to get feedback on this first.

If this looks good to you [barring a changelog entry, and any other changes], we'll move on to season-palette.

### DIFF
--- a/color-schemes.lua
+++ b/color-schemes.lua
@@ -1,35 +1,7 @@
 -- Color schemes manager.
---[====[
-
-color-schemes
-=============
-A script to manage color schemes.
-
-Current features are :
-    * :Registration:
-    * :Loading:
-    * :Listing:
-    * :Default: Setting/Loading
-
-For detailed information and usage, type ``color-schemes`` in console.
-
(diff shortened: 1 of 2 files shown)


Moth-Tolias:
replaced tabs with spaces, and configured our editor so it won't happen again. whoops!

oh, and we forgot to mention, too. the format of the in-memory scheme is `local`, as are the functions that take said scheme. but if for some reason people want to edit schemes programmatically - that's something that could be changed, i suppose.

PR #1603, opened 2026-08-13
Add "civ styles" GUI

Snowstar38:
Adds gui/civ-styles, letting people edit the item styles a civilization knows how to make.

"add-recipe" can only add; there's no way to see what your civ already knows or remove any styles. This lists every style per category, marks which ones came from the civ's raws [helpful for modded civs/races], and lets you take styles away as well as grant them.

It also flags duplicate styles and can clear them (in case add-recipe was accidentally called 2x, lets users fix duplicate entries in menus). There are toggles to show known, unknown, etc.

Covers: armor, helms, gloves, shoes, pants, shields, weapons, diggers, training weapons, ammo, siege ammo, trap components, tools, toys, and instruments.

### DIFF
--- a/docs/gui/civ-styles.rst
+++ b/docs/gui/civ-styles.rst
@@ -1,0 +1,30 @@
+gui/civ-styles
+==============
+
+.. dfhack-tool::
+    :summary: View, add, and remove the item styles a civ can make.
+    :tags: adventure fort gameplay
+
+Civs pick randomly from a pool of item types at worldgen, so not every civ gets
+togas, whips, or high boots. This lists every weapon, armor, clothing, ammo,
+tool, toy, trap component, siege ammo, and instrument in the raws, marks the
+ones your civ knows how to make, and lets you toggle them.
+
+Unlike `add-recipe`, you can see what your civ already knows, and you can take a
+style away again. Removing a style only stops new orders; items already made are
+unaffected.
+
+Pick a category on the left, or search the ``All`` list, then hit :kbd:`Enter` to
+toggle a row. Changes are saved with the fortress.
+
+Rows are marked ``native`` if the style came from your civ's raws, ``exotic`` if
+it did not, ``gen`` for procedurally generated items, and ``xN`` if the style is
+listed more than once, which repeated `add-recipe` runs will do.
+:kbd:`Ctrl`:kbd:`D` removes duplicates.
+
+Usage
+-----
+
+::
+
+    gui/civ-styles
--- a/gui/civ-styles.lua
+++ b/gui/civ-styles.lua
@@ -1,0 +1,489 @@
+local gui = require('gui')
+local widgets = require('gui.widgets')
+local dialogs = require('gui.dialogs')
+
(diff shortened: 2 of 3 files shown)


chdoc:
First of, thank you for contributing, this is a significant piece of work. At first glance, this looks to be in reasonably good shape. But the devil is usually in the detail. I'm still expecting this to require multiple rounds of code review before it can get merged. Given our currently limited reviewing capacity, this may take weeks or even months.

Please indicate the level of AI-support that was used in the development of this tool (if any) and your willingness to iterate on this for several weeks? I would like to avoid spending hours on an in-depth code review, just for another pull request to get abandoned.

docs/gui/civ-styles.rst
```diff
@@ -0,0 +1,30 @@
+gui/civ-styles
```
chdoc, line 1: Why call is tool `civ-styles`? the concept is already known as `recipes`. I would think the tool would be better named `gui/recipes`.

docs/gui/civ-styles.rst
```diff
@@ -0,0 +1,30 @@
...
+style away again. Removing a style only stops new orders; items already made are
+unaffected.
+
+Pick a category on the left, or search the ``All`` list, then hit :kbd:`Enter` to
+toggle a row. Changes are saved with the fortress.
+
+Rows are marked ``native`` if the style came from your civ's raws, ``exotic`` if
+it did not, ``gen`` for procedurally generated items, and ``xN`` if the style is
+listed more than once, which repeated `add-recipe` runs will do.
+:kbd:`Ctrl`:kbd:`D` removes duplicates.
```
chdoc, line 23: Instead of fixing this behavior after the fact, please file a bug report against `add-recipe`.

gui/civ-styles.lua
```diff
@@ -0,0 +1,489 @@
...
+local function try_field(obj, field)
+    if not obj or not field then return nil end
+    local ok, val = pcall(function() return obj[field] end)
+    if ok then return val end
+    return nil
+end
+
+local function get_civ()
+    local civ_id = df.global.plotinfo.civ_id
+    if not civ_id or civ_id < 0 then return nil end
```
chdoc, line 47: `civ_id` is an integer that is always defined when `plotinfo` is defined (basically always). So it is fine to directly call `find` with it.

gui/civ-styles.lua
```diff
@@ -0,0 +1,489 @@
...
+
+--
+-- edits
+--
+
+function CivStyles:add(cat_key, subtype)
+    local cat = self.by_key[cat_key]
+    local vec = cat and self:get_vec(cat)
+    if not vec then return end
+    if count_in(vec, subtype) == 0 then
```
chdoc, line 377: If all you need is a membership test, please use `utils.linear_index(vector,key[,field])` and remove `count_in`.

chdoc (changes requested):
Here are some general remarks for the first round. I will do a more in-depth review once these are taken care of.

PR #1604, opened 2026-08-13
gui/tiletypes: remove COLOR_RESET use with Pen

ChrisJohnsen:
Ref DFHack/dfhack#5864

This Pen use of COLOR_RESET as a foreground color never actually caused any OOB reads (before DFHack/dfhack#5864) since the foreground value already had masking applied to it (related to bold handling).

The EMPTY_LABEL is used to fill (with space characters) the blank portions of the `gui/tiletypes` "more options" entries that are not text label nor checkbox. Since spaces usually don't have any foreground pixels, the foreground used here mostly doesn't matter.

To prevent a visual change for fonts that have foreground pixels in the space character, the foreground should be COLOR_WHITE since this is equivalent to what happens when a COLOR_RESET foreground is specified. Luckily, `UI_COLORS.HIGHLIGHTED` (used for the text portion of the entry while in the hover state) is also COLOR_WHITE.

Suggestion: Switch to specifying `hpen=UI_COLORS.HIGHLIGHTED`. This matches the text portion of the options and produces an effective Pen that is unchanged from when COLOR_RESET was used. Thus, there are no visible changes.

Pictured here:
The (top of the) "more options" window of `gui/tiletypes` with mouse hovering over the "Light" entry. This shows the COLOR_WHITE text under the mouse and the COLOR_GREY text for the non-hover entries. The bottom row shows what happens if the background of EMPTY_TOKEN is COLOR_WHITE instead of COLOR_BLACK.

gui/tiletypes more options

### DIFF
--- a/gui/tiletypes.lua
+++ b/gui/tiletypes.lua
@@ -203,7 +203,7 @@ end
 --#region UI Utilities
 
 ---@type widgets.LabelToken
-local EMPTY_TOKEN = { text=' ', hpen=dfhack.pen.make(COLOR_RESET), width=1 }
+local EMPTY_TOKEN = { text=' ', hpen=UI_COLORS.HIGHLIGHTED, width=1 }
 
 ---@class InlineButtonLabelSpec
 ---@field left_specs? widgets.ButtonLabelSpec


ab9rf:
Needs changelog

ChrisJohnsen:
To be clear, there is no user-visible change in behavior or appearance after the change in this PR. The code change here is just to get rid of a "bad example" of using COLOR_RESET as a Pen color.

Do you still want a changelog?

I just re-tested this by patching a copy of 53.16-r2rc1 downloaded from Steam (which includes DFHack/dfhack#5910). I also used a "visible space" font (modified from the default font to include foreground pixels in the space character) to make sure there was no behavior change even in that edge case.

Both before and after the change in this PR, the relevant UI looks like the top half of the image.
The bottom half of the image uses an separate EMPTY_TOKEN value (not the original value, and not the one from this PR) that was selected to make visible the usually invisible use of the `hpen` in EMPTY_TOKEN. In other words, it show *where* to look to verify that the code changes in this PR are not actually making any visible changes.

---

BTW, the the original COLOR_RESET usage in this code is not actually affected by the changes in 5910 since that only touched `decode_pen` (decoding a Lua table to a Pen instance). The "number only" usage here never uses `decode_pen`; it is handled in `CheckPen` where the number is simply masked off and used as the foreground color of a fresh Pen instance.

PR #1606, opened 2026-08-15
new tool: `gui/export-world-map`

chdoc:
Depends on https://github.com/DFHack/dfhack/pull/5376 to be merged first.

### DIFF
--- a/docs/gui/export-world-map.rst
+++ b/docs/gui/export-world-map.rst
@@ -5,10 +5,16 @@ gui/export-world-map
     :summary: Export world map data for GIS and other external tools.
     :tags: inspection embark map
 
-This tool provides a GUI for the ``export-world-map`` plugin and a number of
+This tool provides a GUI for the `export-world-map` plugin and a number of
 additional exports written in Lua. Moreover, this tool automates the process of
 scrolling around the embark map to generate the region tiles used for the
 exports.
+
+The additional exports provided are:
+
+* Roads and tunnels
+* Geological layers and the veins included in them (pure data layer)
+* Animal and plant populations of regions (pure data layer)
 
 Usage
 -----
--- a/docs/gui/export-world-map.rst
+++ b/docs/gui/export-world-map.rst
@@ -1,0 +1,18 @@
+gui/export-world-map
+====================
+
+.. dfhack-tool::
+    :summary: Export world map data for GIS and other external tools.
+    :tags: inspection embark map
+
+This tool provides a GUI for the ``export-world-map`` plugin and a number of
+additional exports written in Lua. Moreover, this tool automates the process of
+scrolling around the embark map to generate the region tiles used for the
+exports.
+
+Usage
+-----
+
+::
+
+    gui/export-world-map
--- a/gui/export-world-map.lua
+++ b/gui/export-world-map.lua
@@ -1,0 +1,224 @@
+---@diagnostic disable: missing-fields
+
+local gui = require('gui')
+local widgets = require('gui.widgets')
+
+local roads = reqscript('internal/export-world-map/export-roads')
+
+---@type df.viewscreen_choose_start_sitest
+local viewscreen = dfhack.gui.getDFViewscreen(true)
+
+if not df.viewscreen_choose_start_sitest:is_instance(viewscreen) then
+    qerror("Tool must be used while choosing an embark location.")
+    return
+end
+
+ExportMap = defclass(ExportMap, widgets.Window)
+ExportMap.ATTRS {
+    frame_title='Export World Map',
+    frame={w=58, h=25},
+    resizable=false,
+}
+
+local function getLoadedTileRatio()
+    return #df.global.world.world_data.midmap_data.region_details,
+            df.global.world.world_data.world_width * df.global.world.world_data.world_height
+end
+
+function ExportMap:updateRatio()
+    cur,max = getLoadedTileRatio()
+    self.subviews.midmap_ratio:setText(
+        ("%d out of %d region tiles loaded"):format(cur,max)
+    )
+end
+
+function ExportMap:initializeMapScan()
+    local pixel_x = df.global.gps.screen_pixel_x
+    local pixel_y = df.global.gps.screen_pixel_y
+
+    -- conservative approximation of screen dimensions in tiles
+    self.screen_w = (pixel_x // 16) - 2
+    self.screen_h = (pixel_y // 16) - 2
+
+    -- ensure that we are zoomed in
+    viewscreen.zoomed_in = true
+
+    viewscreen.zoom_cent_x = self.screen_w // 2
+    viewscreen.zoom_cent_y = self.screen_h // 2
+
+    self.done = false
+end
+
+
+
+-- keep scrolling around, until the entire world has been covered
+function ExportMap:onRenderBody(painter)
+    self:updateRatio()
+    if self.done then
+        return
+    end
+    if viewscreen.zoom_cent_x + self.screen_w // 2 < df.global.world.world_data.world_width * 16 then
+        viewscreen.zoom_cent_x = viewscreen.zoom_cent_x + self.screen_w
+    elseif viewscreen.zoom_cent_y + self.screen_h // 2 < df.global.world.world_data.world_height * 16 then
+        viewscreen.zoom_cent_x = self.screen_w // 2
+        viewscreen.zoom_cent_y = viewscreen.zoom_cent_y + self.screen_h
+    else
+        self.done = true
+    end
+end
+
+---@param by_world boolean
+---@param by_date boolean
+---@param fn fun(boolean,boolean):(string|fun():nil)
+local function invokeExport(by_world, by_date, fn)
+    local command = fn(by_world, by_date)
+    if type(command) == "string" then
+        dfhack.run_command(command)
+    else
+        command()
+    end
+end
+
+function ExportMap:startExports()
+    self.subviews.indicator_label:setText("working ...")
+
+    local by_date = self.subviews.by_date:getOptionValue()
+    local by_world = by_date or self.subviews.by_world:getOptionValue()
+
+    for _, export in ipairs(exports) do
+        if export.enabled then
+            invokeExport(by_world, by_date, export.command)
+        end
+    end
+
+    self.subviews.indicator_label:setText("")
+end
+
+---launch command from the C++ plugin
+---@param export string
+---@return fun(boolean,boolean):string
+local function pluginCommand(export)
+    return function(by_world, by_date)
+        -- grouping by date implies grouping by world
+        return ('export-world-map %s %s'):format(by_date and '--group-by-date' or (by_world and '--group-by-world' or ''), export)
+    end
+end
+
+exports = {
+    { id = 1, key = 'regions', desc = 'Region Map Export (regions.csv)' , enabled = true, command = pluginCommand("regions") },
+    { id = 2, key = 'rivers', desc = 'River Export (rivers.csv)' , enabled = true, command = pluginCommand("rivers") },
+    { id = 3, key = 'sites', desc = 'Site Export (sites.csv)' , enabled = true, command = pluginCommand("sites") },
+    { id = 4, key = 'elevation', desc = 'Elevation Grid Export (elevation.dat, elevation.vrt)' , enabled = true, command = pluginCommand("elevation") },
+    { id = 5, key = 'roads', desc = 'Road Export (roads.geojson)' , enabled = true, command = roads.export }
+}
+
+local SELECTED_ICON = dfhack.pen.parse{ch=string.char(251), fg=COLOR_LIGHTGREEN}
+local DISABLED_ICON = dfhack.pen.parse{ch='x', fg=COLOR_RED}
+
+function ExportMap:getChoices()
+    print("getChoices")
+    local choices = {}
+    for _, export in ipairs(exports) do
+        table.insert(choices, {
+            icon = function()
+                return export.enabled and SELECTED_ICON or DISABLED_ICON
+            end,
+            text = export.desc,
+            id = export.id
+        })
+    end
+    return choices
+end
+
+function ExportMap:toggleExport(_, choice)
+    exports[choice.id].enabled = not exports[choice.id].enabled
+    self:updateLayout()
+end
+
+function ExportMap:init()
+    self.done = true
+    self:addviews{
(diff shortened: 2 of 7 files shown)


chdoc:
gui-export-world-map

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+end
+
+function ExportMap:initializeMapScan()
+    local pixel_x = df.global.gps.screen_pixel_x
+    local pixel_y = df.global.gps.screen_pixel_y
+
+    -- conservative approximation of screen dimensions in tiles
+    self.screen_w = (pixel_x // 16) - 2
+    self.screen_h = (pixel_y // 16) - 2
+
```
SilasD: would it be worth implementing
> `function screen:onResize(w, h)`
> Called before `onRender` or `onIdle` when the window size has changed.

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+function ExportMap:onRenderBody(painter)
+    self:updateRatio()
+    if self.done then
+        return
+    end
+    if viewscreen.zoom_cent_x + self.screen_w // 2 < df.global.world.world_data.world_width * 16 then
+        viewscreen.zoom_cent_x = viewscreen.zoom_cent_x + self.screen_w
+    elseif viewscreen.zoom_cent_y + self.screen_h // 2 < df.global.world.world_data.world_height * 16 then
+        viewscreen.zoom_cent_x = self.screen_w // 2
+        viewscreen.zoom_cent_y = viewscreen.zoom_cent_y + self.screen_h
```
SilasD, line 66: what happens when the viewscreen is given coordinates outside the world?

if it crashes, I think there ought to be better guardrails here.  technically.

I note that `self.screen_w` and `self.screen_h` are cached at startup and not recalculated if the window size changes.

if there's no misfunction with invalid coordinates, then I don't care here.  if DF crashes, maybe this needs better guardrails.

you could catch
> `function screen:onResize(w, h)`
> Called before `onRender` or `onIdle` when the window size has changed.

or you could calculate then clamp the new coordinates.

*on the other hand*, this could be lumped under "Doctor, it hurts when I do this." "Well then, don't do that."

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+
+ExportMap = defclass(ExportMap, widgets.Window)
+ExportMap.ATTRS {
+    frame_title='Export World Map',
+    frame={w=58, h=25},
+    resizable=false,
+}
+
+local function getLoadedTileRatio()
+    return #df.global.world.world_data.midmap_data.region_details,
```
SilasD, line 26: so no region detail entries get knocked out of this cache during your map scan, even on the largest worlds?

that's a bit surprising to me, since DF is pretty aggressive about cleaning this cache when walking around in adventure mode.  but ok.

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+    )
+end
+
+function ExportMap:initializeMapScan()
+    local pixel_x = df.global.gps.screen_pixel_x
+    local pixel_y = df.global.gps.screen_pixel_y
+
+    -- conservative approximation of screen dimensions in tiles
+    self.screen_w = (pixel_x // 16) - 2
+    self.screen_h = (pixel_y // 16) - 2
```
SilasD, line 43: does this assume a 16x16 font?  if the 8x12 or 10x12 font is used, could a stride based on 16x16 skip over map blocks?  or is it the other way, where *actually* if the font size is lower, the stride is just moving less than it could?  probably that, yeah.

on checking, I see `df.global.gps.tile_pixel_{x,y}` -- ~~is that the text-mode font size, or is that graphics-mode tile size?~~  starting DF with the 8x12 font instead of my preferred 16x16 font showed these change to 8 and 12 instead of 16 and 16.  I think that's the font size *and* tile size.

still, this presumably works fine.  if it didn't, the region detail entries wouldn't be fully filled in.

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
+---@diagnostic disable: missing-fields
```
SilasD, line 1: I assume this is because of LuaLS not quite understanding the Widget system.  so annoying, need to figure out how to resolve that.  putting the field-is-optional flag `?` after every field name works, but would have to be done to every single field of every single Widget.  anyway, not relevant to this PR.

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+
+local gui = require('gui')
+local widgets = require('gui.widgets')
+
+local roads = reqscript('internal/export-world-map/export-roads')
+local layers = reqscript('internal/export-world-map/export-layers')
+local pops = reqscript('internal/export-world-map/export-pops')
+
+---@type df.viewscreen_choose_start_sitest
+local viewscreen = dfhack.gui.getDFViewscreen(true)
```
SilasD, line 11: can user input change the view while the scan is running?  or worse, change to a different viewscreen?

WASD, Escape to zoom out, left-click a button, middle-click-drag, right-click to zoom out?

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+    self.done = true
+    self:addviews{
+        widgets.Label{
+            frame = { t = 1 },
+            view_id = 'midmap_ratio',
+            text = "counting..."
+        },
+        widgets.TextButton{
+            frame = { w = 19, h = 1 , t = 3 },
+            label = "Generate Midmaps!",
```
SilasD, line 148: callback to yesterday's Discord discussion:
> rome of oxtrot — Yesterday at 1:55 PM
> it has the standard open source problem of the UI being mainly designed by programmers, which means it's heavily influenced by the internal data architecture, not by the cognitive model of people who actually use the tool

this should be named "**Load the entire world!**" or some such.

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+            frame = { t = 5, h = 1 },
+            frame_style_l = false,
+            frame_style_r = false
+        },
+        widgets.Label{
+            frame={ t = 7 , h = 1 },
+            text = "Select exports to place in dfhack-config/map-export:",
+        },
+        widgets.List{
+            frame={t=9, h = 7},
```
SilasD, line 161: could make the height data-driven:
`frame={t=9, h = #exports},`

gui/export-world-map.lua
```diff
@@ -0,0 +1,219 @@
...
+    local by_world = by_date or self.subviews.by_world:getOptionValue()
+
+    for _, export in ipairs(exports) do
+        if export.enabled then
+            invokeExport(by_world, by_date, export.command)
+        end
+    end
+end
+
+---launch command from the C++ plugin
```
SilasD, line 95: I spent some time trying to figure out "how is this launching something?" before realizing that it's actualy just formatting commands that will be executed later.

what I'm saying is, the action word in the comment threw me off.

docs/gui/export-world-map.rst
```diff
@@ -0,0 +1,24 @@
...
+* Roads and tunnels
+* Geological layers and the veins included in them (pure data layer)
+* Animal and plant populations of regions (pure data layer)
+
+Usage
+-----
+
+::
+
+    gui/export-world-map
```
SilasD, line 24: would it be useful to document / discuss the data formats of the output files?  for instance:
* The abc file is formatted as CSV data with fields of integer a, integer b, and string c.  String c can contain non-ASCII {choose one: CP437,UTF8} characters. The file does not have a header line.

SilasD (commented):
No red flags in the GUI script.  Some quibbles and some questions, but nothing to hold up accepting the PR.

I'll do the other scripts later, I have to go grocery shopping.

internal/export-world-map/export-layers.lua
```diff
@@ -0,0 +1,38 @@
+--@ module = true
+local util = reqscript('internal/export-world-map/util')
+
+function writeOutput(out)
+    local column_headers = "geo_index;type;material;top_height;bottom_height\n"
+    out:write(column_headers)
+
+    for _, geo_biome in ipairs(df.global.world.world_data.geo_biomes) do
+        for _, layer in ipairs(geo_biome.layers) do
+            out:write(('%s;%s;%s;%s;%s\n'):format(
```
SilasD, line 10: why semicolons when the output is named as a *CSV* file?  [Comma-separated values](https://en.wikipedia.org/wiki/Comma-separated_values)

(later) I see you're using this for most/all output files.  I'm not going to comment on each occurance.

as an aside, I've never understood why people use the idiom
`("a format string"):format(val, val)`
instead of 
`string.format("a format string", val, val)`
treating it as a `sprintf`-alike

if the format string is already defined, it makes sense, sure.  but that's so seldom the case with `string.format`.

internal/export-world-map/export-layers.lua
```diff
@@ -0,0 +1,38 @@
...
+                geo_biome.index,
+                "LAYER",
+                df.inorganic_raw.find(layer.mat_index).id,
+                layer.top_height,
+                layer.bottom_height
+            ))
+            for i = 0, #layer.vein_mat - 1 do
```
SilasD, line 17: nothing *wrong* here, ~~but why not `ipairs` ?~~  because you're walking more than one vector in parallel, okay.

internal/export-world-map/export-pops.lua
```diff
@@ -0,0 +1,59 @@
...
+
+    plants:write(column_headers)
+    creatures:write(column_headers)
+    vermin:write(column_headers)
+
+    for _, region in ipairs(df.global.world.world_data.regions) do
+        for _, pop in ipairs(region.population) do
+            local raw_id
+            local out
+            if pop.type >= 5 then
```
SilasD, line 22: is there a symbolic constant for this raw number?  from context, I think it is `df.world_population_type.Tree`.
I don't espect that the type is ever going to change, but I would prefer to have explicit tests for `Tree`, `Grass`, and `Bush` instead of a `>= Tree`.

internal/export-world-map/export-pops.lua
```diff
@@ -0,0 +1,59 @@
...
+                raw_id = df.plant_raw.find(pop.plant).id
+                out = plants
+            else
+                raw_id = df.creature_raw.find(pop.plant).creature_id
+                if
+                    pop.type == df.world_population_type.Vermin or
+                    pop.type == df.world_population_type.VerminInnumerable
```
SilasD, line 29: yes, like this.

SilasD (approved):
no red flags, it all looks reasonable.  another question and a quibble or two.

gui/export-world-map.lua
chdoc: I tried stealing the focus and then holding down `S` as well as changing the size of the window from large to small. Both resulted in not all tiles being generated, but there were no crashes. The same goes when pressing escape (twice) to actually go to the zoomed-out map. When there are no tiles to render, the "background" task terminates quickly.

gui/export-world-map.lua
chdoc: Indeed, it was assuming 16x16 graphical tiles. Problems could potentially arise if the computed screen width or height is too large and the tool skips over things. That being said, I couldn't even get it to skip tiles on the 14x21 font (on an large screen). Presumably this is because DF must generate all tiles that overlap the viewport. Still, doing this properly isn't that hard, so I will do it.

internal/export-world-map/export-layers.lua
chdoc: > why semicolons when the output is named as a _CSV_ file? [Comma-separated values](https://en.wikipedia.org/wiki/Comma-separated_values)
> 
> (later) I see you're using this for most/all output files. I'm not going to comment on each occurance.

This is a case of Europeans doing European things: https://discord.com/channels/793331351645323264/1340358673003712654/1540610890192134195

However, I agree that this should be documented for the benefit of others.

gui/export-world-map.lua
chdoc: If you leave the screen to the main menu with the tool open, DwarfFortress will destroy the DFHack viewscreen.

SilasD:
looks great.

internal/export-world-map/export-layers.lua
ab9rf: BTW, the rule for when you use a semicolon is "if the decimal separator is a comma, then the CSV separator is a semicolon, otherwise it's a comma". Microsoft's locale processing does this for you (they have a `ListSeparator` element in their locale library which is inferred using this rule) but in POSIX (and thus in the C++ standard, which is heavily POSIX-leaning) you have to do it yourself.

Most CSV libraries will accept either. Arguably, the best approach is to grab the current locale, identify its decimal separator, and from that infer whether a comma or a semicolon is most appropriate. 

```c++
std::locale current_locale = std::locale{""};    
auto const& numeric_facet = std::use_facet<std::numpunct<char>>(current_locale);
char decimal_sep = numeric_facet.decimal_point();
char list_sep = decimal_sep == ',' ? ';' : ',';
```

chdoc: I really would prefer if output of the world export tools would not depend on the user's locale. As far as I know, no other behavior of DFHack depends on it. If only you knew how much grief Microsoft's locale-dependent interpretation of CSV has caused on this side of the Atlantic...

Moreover, while semicolons do not naturally occur in our data, commas very much do
```
plant_pops.csv:736;Grass;BAMBOO, GOLDEN;10000001;10000001
plant_pops.csv:737;Grass;BAMBOO, ARROW;10000001;10000001
plant_pops.csv:746;Grass;BAMBOO, HEDGE;10000001;10000001
```

```
sites.csv:1434;-1;-10000;-1;lair;Emanomba Etrujulosm;Birdsrivers the Earthen Crab;NONE;NONE;NONE;NONE;NONE;POLYGON((84336 -52896,84384 -52896,84384 -52944,84336 -52944,84336 -52896))
```
The commas inside the WKT geometries obviously affects all layer files, including those generated by the C++ plugin.

Currently, all exports get away without doing any escaping. If you want commas, you also will get escaping.

SilasD: ＜meme＞ **Um, ackchyually,** ＜/meme＞

We've got a couple of locale-specific numbers.  The trade overlays format prices per locale, and the notifications floating overlay formats time-since-last-save per locale.

Which is just quibbling really.

Commas crop up in the exported data; I think that's a good enough reason to go with semicolons.  All good.

PR #1610, opened 2026-09-05
Remove `Really disband squad` and `Really change recenter location`

magnus-ISU:
DF asks 'Really disband?' when you disband a squad, and 'Really overwrite' when you reassign a recenter hotkey, so squad-disband and hotkey-reset put a second dialog in front of the game's own.

image

image

image

This is probably one of the only cases DFhack makes the game worse, likely because it used to be necessary and no one revisited this code.


chdoc (approved):
I can confirm that the game now asks before disbanding a squad or reassigning a hotkey. So these are indeed no longer necessary.

magnus-ISU:
Anything I need to do more to get this merged?

chdoc:
No, I was merely waiting whether someone else was going to chime in or not.

PR #1612, opened 2026-09-13
exportlegends: restore classic companion exports

inder-chakal:
Add the population, history, and world map companion exports that were originally present in Classic DF so that downstream applications like legends-viewer-next have more data to work with like site populations/history/map.


inder-chakal:
Alright - I'm done. Please merge. I tested the downstream changes in a Legends-Viewer-Next branch locally as well. The new map looks very nice.

PR #1617, opened 2026-09-15
combine: differentiate dyes by dye_profile instead of excluding them

Alistair-Afton:
## Summary

Implements the real fix for DFHack/dfhack#5849.

Since 52.03, dye powders carry an embedded `dye_profile` (color index plus the component material lists). Mixed dyes share `mat_type`/`mat_index` with their components, so combining stacks by material alone reverts a mixed dye to one of its components — the bug that #1596 avoided by excluding dyes from combining entirely.

This change fingerprints `dye_profile` into the comparison key for `POWDER_MISC` items, so dyes combine again — but only with an identical blend.

Dyes whose profile was lost to the vanilla merchant bug are never merged: the embedded `dye_profile` compound is always present on real items, so the unprofiled state is detected via `color_index == -1`, and each such item gets a unique comparison key. Non-dye powders (flour, etc.) are unaffected.

#### Test plan

- `test/combine.lua` covers same/different profiles, mixed-vs-component keys, unprofiled dyes (both embedded-empty and absent profile), and non-dye powders
- 6/6 tests pass in-game on 53.16


ab9rf:
There are unfixed vanilla game bugs regarding dyes with incorrect or missing dye profiles (specifically, dyes and dyed cloth purchased from merchants sometimes have incorrect or missing dye information). Until those bugs are fixed, I'm not comfortable with possibly combining bugged and unbugged dye stacks, at least not until the consequences of doing this are known and understood.

Alistair-Afton:
@ab9rf Good point. Dyes missing their dye_profile now get a unique comparison key each, so they can never be merged into another stack or each other. Unprofiled non-dye powders (flour and such) are unaffected, and properly profiled dyes still combine per profile.

Alistair-Afton:
Follow-up on the unprofiled-dye handling: the embedded `dye_profile` compound is never actually absent on real items — the missing-profile state shows up as `color_index == -1` with empty vectors. The detection now keys off that, so merchant-bugged dyes each get a unique key as described. Pushed as part of a rebase onto current master.

Regarding dyes with *incorrect* (rather than missing) profiles: they still merge only with items carrying an identical profile, so a wrong profile can never collapse a different blend — the residual risk is two items sharing the same incorrect profile, which are indistinguishable to begin with.

ab9rf (approved):

PR #1620, opened 2026-09-15
`caravan`: fix doubled pending value and lost marks across filter views

### LINKED ISSUE #5829: Minor visual bug in caravan movegoods interface
If you already have some items moved to trade depot and you open the movegoods dfhack interface and click "inside containers" and make it "yes", the "total value of items marked for trade" doubles and then stays double for that screen whichever flag you change. Its in [https://github.com/DFHack/scripts/blob/master/internal/caravan/movegoods.lua](url)
Video : https://github.com/user-attachments/assets/9fe18e53-93ac-487a-af76-2ddbe9e5ce5a

Alistair-Afton:
## Summary

Fixes DFHack/dfhack#5829

The movegoods overlay caches a separate copy of choice state for each filter combination (grouped x inside containers). Rebuilding a combination added all selected values to the pending total again, which doubled the displayed "total value of items marked for trade". Marks made in one view were also invisible in the others, and dismissing the modal only committed the currently active view, so marks were silently lost.

Marked state is now a shared set of explicit user decisions on the widget. Each cached view derives its pending flags from it, the pending total is recomputed once per scan, and dismiss commits every explicit mark regardless of the active view.

## Testing

In game with a fort and trade depot: instantiated the modal and exercised all four filter combinations. The pending total stays constant instead of doubling. Marks made in the ungrouped view remain visible in the grouped and inside containers views, and dismissing while viewing a different filter still creates the BringItemToDepot job. Unmarking and remarking adjusts the total correctly, and icons update on toggled items.


Alistair-Afton:
Added in-engine regression tests (`test/caravan/movegoods.lua`): the pending total must equal the canonical recomputation after every choices cache build, and marks made in one filter view must propagate to rebuilt views. Verified in game: 2/2 tests, 13/13 checks.

ab9rf (approved):

PR #1623, opened 2026-09-16
docs: rename the stockpiles tag to stockpile

### LINKED ISSUE #5843: `stockpiles` plugin shares name with DFHack tool tag
The `stockpiles` plugin shares a name with the `stockpiles` DFHack tool help tag. When clicking on the "?" button on the plugin's GUI, it shows the documentation for the plugin, but also sets the filter to tools with the `stockpiles` tag.

One of these should probably be changed. It seems easiest to just change the tag to `stockpile`.

Alistair-Afton:
## Summary

Fixes DFHack/dfhack#5843

Companion to DFHack/dfhack#5917. The ``stockpiles`` tag shares its name with the `stockpiles` plugin, so `gui/launcher` treats the tool name as a tag filter instead of showing the plugin's entry. Renames the tag to ``stockpile`` in the script docs that carry it.

## Testing

- Verified with the dfhack-side text docs build; the tag index now generates ``stockpile``.


Alistair-Afton:
The docs build here will fail until DFHack/dfhack#5917 merges. The tag indices are registered from dfhack's Tags.rst, which on develop still only knows the old tag name. Once that PR lands, rerunning the docs job should go green.

ab9rf (approved):

PR #1624, opened 2026-09-16
overlays: don't steal keys while renaming a building

### LINKED ISSUE #5764: Bug: `trackstop.pressureplate` overlay steals input from building renaming
The plate trigger hotkeys seem to take precedence over entering text in building renaming. `orders.skillrestrictions` doesn't have this problem, but I haven't had time to investigate.

This also seems to apply to other `trackstop` overlays, although those tend to use `Ctrl+` modifiers on their hotkeys.

Alistair-Afton:
## Summary

Fixes DFHack/dfhack#5764

The ``machine-toggle`` pressure plate / gear assembly overlays and the ``trackstop`` track stop / roller overlays consumed their hotkeys (x, shift+x, shift+r, shift+v, shift+t, shift+b) even while the building nickname field was open, so typing those characters changed settings instead of entering text.

Each overlay's ``onInput`` now skips event handling while ``view_sheets.building_entering_nickname`` is set, matching the guard already used by the orders skill restriction overlay.

## Testing

In a live fort, instantiated ``PlateOverlay`` and fed it a ``CUSTOM_X`` keypress with a stubbed super:

- ``building_entering_nickname = true``: key not consumed by the overlay (falls through to the rename field).
- ``building_entering_nickname = false``: normal handling resumes.


Bumber64:
I'd probably do something like `local vs = df.global.game.main_interface.view_sheets`, with how long that is.

Then it's easy enough to do `if not vs.building_entering_nickname then` instead of `isRenaming()`.

Alistair-Afton:
Done. `local vs = df.global.game.main_interface.view_sheets` at file scope (it's an inline compound in a static global, so the binding stays live) and `if not vs.building_entering_nickname then` in all four overlays. Re-verified live that the guard blocks input while renaming and passes through normally.

Alistair-Afton:
Superseded by DFHack/dfhack#5951, which fixes this at the framework level: overlay input dispatch is skipped entirely while a native text-entry flag is set, covering these overlays plus all others (e.g. the lever mechanism overlay's g/m keys, zone cage/chain overlays) without per-overlay guards.

PR #1629, opened 2026-09-16
source: validate cursor before source lookup

### LINKED ISSUE #5582: `source` throws an exception if keyboard cursor pos is invalid
`source` uses the keyboard cursor for its location determination but doesn't handle errors well

[per discord](https://discord.com/channels/793331351645323264/807444467140788254/1415782151592607776)

### LINKED ISSUE #5617: `source` throws a traceback instead of signalling a operator error if called with no tile selected
Instead of throwing a traceback the tool should signal a useful error with advice on how to correct the error

[discussion on steam](https://steamcommunity.com/app/2346660/discussions/0/667222425710250534/?tscn=1762425183)

Alistair-Afton:
## Summary

* validate the keyboard cursor before searching for an existing liquid source
* report the intended operator error instead of a nil position traceback
* add regression coverage for both add and delete

Fixes DFHack/dfhack#5582.
Fixes DFHack/dfhack#5617.

## Testing

* `test -t source`
* live `source delete` with no active cursor
* live `source add water` with no active cursor
* live `source list`


ab9rf:
No changelog.

PR #1634, opened 2026-09-17
gui/create-item: add 0-5 hotkeys to quality prompt

### LINKED ISSUE #5770: Idea: `gui/create-item` give hotkeys for item quality selection
The quality selection for `gui/create-item` doesn't seem to have keybinds like the other menus do, requiring use of arrow keys to select item quality when using the keyboard.

We could number the options 1-6 for convenience, or 0-5 to match other tools.

Alistair-Afton:
## Summary

Fixes DFHack/dfhack#5770

The quality prompt in `gui/create-item` now assigns digit hotkeys 0-5 to the six quality options, matching the internal `item_quality` values. `widgets.List` already supports per-choice keys, so this is just annotating the choice table.

## Testing

Verified the `widgets.List` choice-key path in game: feeding `STRING_A053` selects index 6 (Masterwork), `STRING_A048` selects index 1 (None).


Alistair-Afton:
Duplicate of #1616, which already implements this.

PR #1636, opened 2026-09-18
multihaul: let citizens haul multiple items to a stockpile in one trip

Alistair-Afton:
## Summary

New opt-in script `multihaul` (disabled by default; also registered in the control panel's gameplay section). When a citizen picks up an item for a stockpile, they also grab up to `max` additional loose items of the same type â€” detouring around the pickup site to collect them (`fetch` radius) and opportunistically picking up more along the way (`radius`) â€” then drop everything off in one trip.

    multihaul enable|disable
    multihaul status
    multihaul max <n>          (default 4)
    multihaul radius <n>       (default 2, en-route grab distance)
    multihaul fetch <n>        (default 8, detour distance around the
                               pickup site)
    multihaul weight <n>       (default 0 = unlimited; caps the total
                               carried load incl. the job's own item)
    multihaul weight auto      (derive the cap per dwarf from strength
                               scaled by body size â€” a typical dwarf can
                               manage ~900 mass units, about three boulders)
    multihaul types same|all|pile
                               (default same; "all" also grabs items that
                               other haul jobs are already taking to the
                               same destination, regardless of type; "pile"
                               additionally grabs unclaimed loose items of
                               any type the destination stockpile's filter
                               provably accepts, so a dwarf heading to a
                               mixed pile sweeps up everything nearby that
                               belongs there)
    multihaul targets piles|all
                               (default piles; "all" also piggybacks loads
                               into minecarts, barrels, and bins, inserting
                               extras into the container on delivery)

## Implementation notes

The job's own `job.items` vector is never touched â€” adding `job_item_ref`s to a live job reliably cancels it (verified in-game). Instead, extras ride as `Hauled` inventory items flagged `in_job` so other jobs don't claim them. Since the engine drops inventory items that aren't linked to the active job, a 3-frame poll re-attaches extras while the job is in flight. Wheelbarrow-assisted jobs are skipped since the wheelbarrow already multi-hauls.

Extras nearest to the origin are preferred. The pickup-site sweep uses `fetch`; en-route grabs (throttled to every few polls) use `radius` around the unit's current position.

On job end, extras are released onto the primary item's landing tile (if it landed on the destination pile), else the nearest pile tile when the unit finishes at/next to the pile (e.g. vehicle handoffs), else the unit's feet for mid-route cancellations. For container destinations, extras the unit still holds (or that were dropped at its feet during the handoff) are inserted via `dfhack.items.moveToContainer`, falling back to a ground drop if insertion fails (e.g. wrong material class). Extras dropped directly onto the destination pile or into the container mid-transit are released as delivered.

`types all` only ever takes items that provably share the job's destination: a claimed item's own `JOB` specific_ref must resolve to the same pile or container. Unrelated jobs are never touched, and an extra that gets legitimately re-claimed by a real job (flag vs. specific_ref check) is left alone rather than unflagged.

`types pile` re-implements DF's item-vs-stockpile-filter match in Lua (no such check is exposed). It covers every category with usable indices: stone, wood (via plant raws), bars/blocks (incl. coal/potash/ash/glass/soap `other_mats` mapping), gems, coins, weapons/trapcomps/ammo, armor, furniture (incl. tool buckets via the itemdef's `tool_use`), finished goods, food organics via `mat_table.organic_types/indexes` (incl. creature/caste-indexed fish and eggs), leather, cloth/thread by material class, paper/parchment sheets, corpses and corpse pieces/remains by race and part kind (fresh vs. rotten hide separately), and cages/traps by occupant race or the empty-container toggles. Quality (`quality_core`/`quality_total`, incl. improvements), dyed/undyed, dye color, usable/unusable, and links-only restrictions are honored; stockpile settings that were never configured (empty vectors, all-false arrays) correctly mean "all allowed". Anything that cannot be *proven* acceptable is rejected, so a wrong answer can only ever mean fewer extras, never a mis-stored item.

Candidate filter excludes items that are forbidden, owned, hostile, trader, rotten, dump/melt-designated, in containers, in buildings, spider webs, construction-designated, hidden, burning, storage-assigned (`getStockpile()`, pcall-wrapped since the field doesn't exist on all item classes), or already sitting on a building tile (so items are never pulled out of neighboring stockpiles).

Cleanup paths: `disable` releases all in-flight extras; a periodic sweep clears any `in_job` flag that has no `JOB` specific_ref behind it (heals leaks from script/env reloads); flags are also cleared on map/world unload so saves never persist bogus claims.

Additional audit fixes: the destination container itself can never be pocketed as an extra (a vehicle waiting on the ground for its own `StoreItemInVehicle` job matched the claim check against itself); items claimed by wheelbarrow-assisted haul jobs are not stolen (their jobs do not resolve through `get_job_dest`); dropped extras credit their weight back so the cap is not inflated; candidate scanning reads the overlapping map blocks' item lists instead of all of `IN_PLAY` (~26k items in the test fort); the repeat timer survives map unload and transient scan errors; tracked units that leave the citizen list (dead/off-map) release their extras immediately.

## Test plan

Verified live in a mature fort (DF 53.16):

- [x] Multiple `Hauled` items tolerated in one inventory
- [x] Extras attach at pickup (17 concurrent haulers observed carrying 1-4 extras each)
- [x] `fetch` collects extras 4-5 tiles beyond the pickup point and delivers them to the destination pile
- [x] Extras survive transit incl. z-level climbs (re-attach counters engine drops)
- [x] Extras land on destination pile tiles and stay stocked
- [x] `disable` with in-flight extras: all released cleanly, no orphaned `in_job` items (re-claimed stragglers verified via `specific_refs`)
- [x] `job.items` mutation confirmed unsafe (job cancel + orphan) â€” avoided entirely
- [x] Container jobs tracked (barrel/bin/vehicle); `moveToContainer` verified; primary items merging into container stacks detected as delivered
- [x] `weight auto`: per-unit caps verified (e.g. 703 for a weak hauler vs 1206 for a strong one; loads stop at the cap mid-pickup)
- [x] Orphan sweep clears leaked flags; no error spam (scan errors surface via `dfhack.printerr`)
- [x] `types pile`: cross-type extras attach at pickup and en route and are delivered into the destination pile (observed live: thread riding a corpse haul into a mixed refuse/leather/cloth pile)
- [x] Wheelbarrow-assisted jobs verified unstealable on live jobs: `claimed_for_dest` returns false for a wheelbarrow-claimed boulder, true for a normal same-pile claim
- [x] `types pile` discrimination on real piles: wood-only pile takes wood/rejects stone, bars piles distinguish metals vs. potash-only vs. soap-only, links-only piles reject everything, forbidden/owned/in-building candidates rejected
- [x] Mixed-type grab end-to-end: a REMAINS hauler to a refuse pile (temporarily widened to accept stone) swept a BOULDER extra, carried it down 5 z-levels, and dropped it at the destination; zero orphaned `in_job` flags after
- [x] `test/multihaul.lua`: 23 unit tests / 67 checks over the matcher (subtype, material, quality, dye/color, usability, freshness, race vectors, tool buckets, unconfigured-pile semantics) and the candidate filters `extra_ok`/`claimed_for_dest` (type modes, weight-cap boundary, destination-container exclusion, wheelbarrow-claim exclusion)


ab9rf:
we need to discuss whether this should be classified as an armok tool or if instead needs a new category for providing a material change in game mechanics that goes beyond mere automation

